PR #120, opened 2026-02-13
feat(auth): model-level rate limit locking for multi-bucket providers

### LINKED ISSUE #110: Feature Request: Granular Rate Limit Locking for Multi-Bucket Providers (Antigravity/Vertex)
### Context & Configuration
We are running a setup with **2 Antigravity Accounts** (Account A, Account B) to maximize throughput.
*   **Provider**: Antigravity (Google Internal)
*   **Models**: Mixed workload of `claude-opus` (Coding) and `gemini-pro` (Chat).
*   **Quotas**: Antigravity has **separate** quota buckets for Claude models vs Gemini models.

### The Problem: Global Account Locking
9router currently treats a `429` on *any* model as a sign the entire account is dead.

#### Current Logic Flow (Flawed)
1.  **Request 1**: User asks for `claude-opus` using **Account A**.
    *   **Upstream**: Google returns `429 Quota Exceeded` (Claude bucket empty).
    *   **9router Logic**:
        *   ❌ Locks **Account A** globally (DB update).
        *   Failover to **Account B**.
2.  **Request 2**: User asks for `gemini-pro` (which has 100% quota on Account A).
    *   **9router Logic**:
        *   Checks **Account A** -> **LOCKED** (due to Claude 429).
        *   Skipping Account A.
        *   Failover to **Account B**.
    *   **Result**: We burn Account B's quota while Account A's Gemini bucket sits full and unused.

### Desired Behavior: Granular Locking
We need 9router to track locks per-model (or per-bucket) to utilize the full capacity of multi-bucket providers.

#### Ideal Logic Flow
1.  **Request 1**: User asks for `claude-opus` using **Account A**.
    *   **Upstream**: Google returns `429`.
    *   **9router Logic**:
        *   ✅ Locks **Account A : claude-opus** only.
        *   Failover to **Account B**.
2.  **Request 2**: User asks for `gemini-pro` using **Account A**.
    *   **9router Logic**:
        *   Checks **Account A : gemini-pro** -> **OPEN**.
        *   Routes request to **Account A**.
    *   **Result**: Efficient usage of all available quotas.

### Our Patch (Temporary Workaround)
We patched `src/sse/services/auth.js` to implement **in-memory model-aware locking** for Antigravity only.

#### 1. Modify `auth.js` (Locking Logic)
We added a Map to track locks and modified `markAccountUnavailable` to skip the DB lock for Antigravity 429s.

```javascript
// In-memory model locks
const modelLocks = new Map();

export async function markAccountUnavailable(connectionId, status, errorText, provider = null, model = null) {
  // ... existing logic ...

  // SPECIAL CASE: Antigravity Model-Aware Locking
  // If 429 on Antigravity, lock ONLY the model in memory, do NOT lock the account in DB.
  if (provider === "antigravity" && status === 429 && model) {
    // Lock model for 5 minutes
    const key = `${connectionId}:${model}`;
    modelLocks.set(key, Date.now() + 300000);
    
    // Return fallback=true, cooldown=0 (skip DB lock)
    return { shouldFallback: true, cooldownMs: 0 };
  }

  // ... standard DB locking logic ...
}
```

#### 2. Modify `chat.js` (Pass Model Context)
We updated the handler to pass the specific model name down to the auth service.

```javascript
// src/sse/handlers/chat.js
async function handleSingleModelChat(body, modelStr, ...) {
  // ...
  // Pass 'modelStr' (e.g., 'ag/gemini-3-pro') to credential fetcher
  const credentials = await getProviderCredentials(provider, excludeConnectionId, modelStr);
  // ...
  // Pass 'modelStr' to error handler
  await markAccountUnavailable(credentials.connectionId, result.status, result.error, provider, modelStr);
}
```

### Proposal for Native Implementation
Instead of this hardcoded patch, we propose:
1.  **Configuration**: Add `rateLimitScope: "model" | "account"` to Provider settings.
2.  **Regex Buckets**: Allow defining model groups (e.g., `^claude-*` = Bucket A) in `db.json`.
3.  **Persisted Model State**: Track model-specific rate limits in the database alongside account status.

rothnic:
## Summary

Adds **model-level rate limit locking** for providers with separate per-model quota buckets (e.g. Antigravity).

Closes #110

## Problem

Antigravity (and Vertex AI) maintain independent quota buckets per model family — Claude models and Gemini models have separate rate limits under the same account. When 9router receives a `429 Too Many Requests` for any model, it currently locks the **entire account** in the database, preventing access to *all* models including those with full quota remaining.

**Example:** A `429` on `claude-opus` (Account A) locks Account A entirely. Subsequent `gemini-pro` requests skip Account A and burn Account B's quota, even though Account A's Gemini bucket is 100% available.

## Solution

In-memory model-aware locking that tracks rate limits per `connectionId:model` pair rather than per account.

### How it works

1. **On 429 (multi-bucket providers only):** Instead of locking the account in the database, we set a temporary in-memory lock for that specific model (default 5 minutes or the calculated backoff duration).
2. **On request:** The connection filter checks both the standard account-level DB lock AND the in-memory model lock. If `connectionId:model` is locked, we skip that account for *that model only* — other models remain accessible.
3. **Fallback behavior:** When all accounts are model-locked for a specific model, returns a `retryAfter` response with a short 1-minute retry window (vs the standard account-level backoff).

### Design decisions

- **In-memory only** — No database schema changes. Locks clear on restart. Low risk, zero migration.
- **Provider-gated** — Only enabled for providers in `MULTI_BUCKET_PROVIDERS` set (currently `antigravity`). Other providers use the existing account-level locking unchanged.
- **Backward compatible** — The `model` parameter defaults to `null`. When not provided (e.g. from combo handlers), behavior is identical to before.

## Changes

| File | Change |
|------|--------|
| `src/sse/services/auth.js` | Add `modelLocks` Map, `isModelLocked()`, `lockModel()`, `isMultiBucketProvider()`. Update `getProviderCredentials()` and `markAccountUnavailable()` to accept `model` param and apply model-level filtering. |
| `src/sse/handlers/chat.js` | Pass `model` string to `getProviderCredentials()` and `markAccountUnavailable()` calls. |

## Testing

Validated in production with 2 Antigravity accounts serving mixed Claude/Gemini workloads. After applying this patch:
- Claude 429 on Account A only blocks Claude on Account A
- Gemini requests continue routing to Account A normally
- Standard account-level locking remains unchanged for non-Antigravity providers

## Future work (from #110)

This PR implements the core locking mechanism. Potential follow-ups:
- **Configuration**: Add `rateLimitScope: "model" | "account"` to Provider settings in the dashboard
- **Regex buckets**: Allow defining model groups (e.g., `^claude-*` = Bucket A) for more granular control
- **Persisted state**: Optionally store model locks in the database for persistence across restarts

### DIFF
--- a/src/sse/handlers/chat.js
+++ b/src/sse/handlers/chat.js
@@ -120,7 +120,7 @@ async function handleSingleModelChat(body, modelStr, clientRawRequest = null, re
   let lastStatus = null;
 
   while (true) {
-    const credentials = await getProviderCredentials(provider, excludeConnectionId);
+    const credentials = await getProviderCredentials(provider, excludeConnectionId, model);
 
     // All accounts unavailable
     if (!credentials || credentials.allRateLimited) {
@@ -170,7 +170,7 @@ async function handleSingleModelChat(body, modelStr, clientRawRequest = null, re
     if (result.success) return result.response;
 
     // Mark account unavailable (auto-calculates cooldown with exponential backoff)
-    const { shouldFallback } = await markAccountUnavailable(credentials.connectionId, result.status, result.error, provider);
+    const { shouldFallback } = await markAccountUnavailable(credentials.connectionId, result.status, result.error, provider, model);
     
     if (shouldFallback) {
       log.warn("AUTH", `Account ${accountId}... unavailable (${result.status}), trying fallback`);
--- a/src/sse/services/auth.js
+++ b/src/sse/services/auth.js
@@ -5,13 +5,70 @@ import * as log from "../utils/logger.js";
 // Mutex to prevent race conditions during account selection
 let selectionMutex = Promise.resolve();
 
+// ---------------------------------------------------------------------------
+// Model-level rate-limit locking (in-memory)
+//
+// Providers like Antigravity maintain *separate* quota buckets per model
+// family (e.g. Claude vs Gemini). When a 429 arrives for one model, the
+// standard account-level DB lock would block ALL models on that account,
+// wasting quota that is still available for other model families.
+//
+// This module tracks model-specific locks in memory so that only the
+// affected model is skipped during account selection while the rest of the
+// account's quota remains accessible.
+//
+// Locks are intentionally in-memory: they clear on restart and require no
+// database schema migration.
+// ---------------------------------------------------------------------------
+
+/** Providers known to have independent per-model quota buckets */
+const MULTI_BUCKET_PROVIDERS = new Set(["antigravity"]);
+
+/** Map<"connectionId:model", expiryTimestamp> */
+const modelLocks = new Map();
+
+/** Default lock duration for model-level rate limits (5 minutes) */
+const DEFAULT_MODEL_LOCK_MS = 5 * 60 * 1000;
+
+/**
+ * Check whether a specific model is temporarily locked on a connection.
+ * Expired locks are cleaned up lazily.
+ */
+function isModelLocked(connectionId, model) {
+  if (!connectionId || !model) return false;
+  const key = `${connectionId}:${model}`;
+  const expiry = modelLocks.get(key);
+  if (!expiry) return false;
+  if (expiry > Date.now()) return true;
+  modelLocks.delete(key); // clean up expired
+  return false;
+}
+
+/**
+ * Lock a model on a specific connection for `durationMs` milliseconds.
+ */
+function lockModel(connectionId, model, durationMs) {
+  if (!connectionId || !model) return;
+  const key = `${connectionId}:${model}`;
+  modelLocks.set(key, Date.now() + durationMs);
+  log.warn("AUTH", `Model lock: ${model} on ${connectionId.slice(0, 8)} for ${Math.round(durationMs / 1000)}s`);
+}
+
+/**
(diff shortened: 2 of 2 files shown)


rothnic:
> GH Copilot is request based for sure. I was told that Antigravity work by analyzing work done, and seems like you have only a limited work done requests. But then some told that it is token based.
> 
> And I agree with you completely. A much smarter routing system is needed. that would always try to use the models up for 50 percent first and after that they can go to opus once again. And also of course the routing really needs to account the rollover time window, so that these models are used first that rollover earliest

Sorry for the slow follow-up, ended up getting banned by antigravity, so i've been figuring out what the next best options are. From my understanding antigravity was request-based, but i could be misunderstanding. I thought it was N requests per 6 hours or something like that. Copilot it like 1200 requests per month, no hour-based timeout, but is only $40 and has free gpt-5-mini access.

At one point when i was having issues, i almost put litellm in front if 9router until i realized it wasn't going to help. It did kind of make me wonder if using something like litellm internally would be a good idea since you could kind of provide its configuration as an advanced mode.

I ended up grabbing a month of kimi for $40 for now until i have more time to figure out the next best option.

PR #127, opened 2026-02-14
fix(github): Implement dynamic fallback for Codex models requiring /responses endpoint

### LINKED ISSUE #102: bug: GitHub Copilot Codex models fail with 400 error (require /responses endpoint)
## Summary
Several GitHub Copilot Codex models are failing with a `400 Bad Request` error. The error message indicates that these models are not accessible via the standard `/chat/completions` endpoint.

According to the documentation and API responses, these newer Codex models require the `/responses` endpoint instead.

## Affected Models
The following models under the `gh` provider are affected:
- [ ] `gpt-5-codex`
- [ ] `gpt-5.1-codex`
- [ ] `gpt-5.1-codex-mini`
- [ ] `gpt-5.1-codex-max`

## Steps to Reproduce
1. Select the **GitHub Copilot** provider.
2. Choose any of the affected models (e.g., `gpt-5.1-codex`).
3. Send a message.
4. **Actual Result:** Receive error:
   > 400 {"error": {"message": "[github/gpt-5.1-codex] [400]: model gpt-5.1-codex is not accessible via the /chat/completions endpoint"}}

## Root Cause Analysis
The current implementation seemingly forces `/chat/completions` for all models. However, newer Codex models define `supported_endpoints` as `["/responses"]`.

**Reference JSON for model capabilities:**
```json
    {
      "capabilities": {...},
      "id": "gpt-5.2-codex",
      "model_picker_category": "powerful",
      "model_picker_enabled": true,
      "name": "GPT-5.2-Codex",
      "object": "model",
      "supported_endpoints": [
        "/responses"
      ],
      "vendor": "OpenAI",
      "version": "gpt-5.2-codex"
    }
```

## Suggested Fix
Update the request executor to check the model's metadata or ID. If it is a newer Codex model, the request should be routed to the `/responses` endpoint instead of `/chat/completions`.

## Attachments

Image
Image
Image
Image

I3eka:
## Summary
This PR resolves the `400 Bad Request` error encountered when using newer GitHub Copilot Codex models (e.g., `gpt-5.1-codex`, `gpt-5.2-codex`). The issue is addressed by implementing a dynamic, reactive fallback mechanism within the `GithubExecutor` that automatically switches to the correct API endpoint.

## Root Cause Analysis
Certain GitHub Copilot models have deprecated the standard `/chat/completions` endpoint and now exclusively require requests to be sent to the `/responses` endpoint, which also uses a different payload structure (`input` array instead of `messages`). The previous implementation was hardcoded to `/chat/completions`, causing API rejections for these models.

## Implementation Details
A resilient, self-correcting strategy has been implemented:

1.  **Initial Attempt:** The executor first makes a standard request to `/chat/completions`.
2.  **Error Detection:** If a `400 Bad Request` is received with the specific error message `"not accessible via the /chat/completions endpoint"`, the fallback logic is triggered.
3.  **Caching for Performance:** The model's ID is immediately added to an in-memory `Set` (`this.knownCodexModels`). All subsequent requests for this model will bypass the initial failed attempt and go directly to the correct `/responses` endpoint, eliminating future latency.
4.  **On-the-Fly Translation:** The request is automatically retried on the `/responses` endpoint. This involves two real-time translations:
    *   **Request Body:** The OpenAI `messages` array is converted into the `input` array format.
    *   **Response Stream:** The SSE event stream from `/responses` is converted back into the standard OpenAI Chat Completion chunk format, ensuring client compatibility and correct token `usage` tracking.

## Considered Alternative: Proactive Service Discovery

An alternative, proactive "Service Discovery" approach was also investigated. This would involve making a `GET https://api.githubcopilot.com/models` request at startup or periodically to build a map of which models support which endpoints, based on the `supported_endpoints` field present in the response.

**Pros of Service Discovery:**
*   Eliminates the latency of the initial failed request for new models.

**Cons / Reasons for Choosing the Fallback Approach:**
*   **API Stability Risk:** The `/models` endpoint and its schema are part of an **internal, undocumented API**. Relying on its structure is fragile, as it could change without warning and break the integration for all GitHub models.
*   **Reliability:** The reactive fallback relies on a more stable "contract"—the API's explicit error response. This makes the proxy more resilient to future changes and new models that may not yet appear in the `/models` list.

**Suggestion for the Future:**
While the current reactive approach is safer, a hybrid model could be considered. We could implement Service Discovery as a "best-effort" optimization and still keep the reactive fallback as a safety net. This would provide the performance of Service Discovery with the resilience of the current implementation. I am happy to discuss or implement this if you think it's a worthwhile addition.

For now, this PR prioritizes robustness and future-proofing.

## Related Issue
Fixes #102 
photo_2026-02-15_03-08-01

### DIFF
--- a/open-sse/config/constants.js
+++ b/open-sse/config/constants.js
@@ -134,6 +134,7 @@ export const PROVIDERS = {
   },
   github: {
     baseUrl: "https://api.githubcopilot.com/chat/completions", // GitHub Copilot API endpoint for chat
+    responsesUrl: "https://api.githubcopilot.com/responses",
     format: "openai", // GitHub Copilot uses OpenAI-compatible format
     headers: {
       "copilot-integration-id": "vscode-chat",
--- a/open-sse/executors/github.js
+++ b/open-sse/executors/github.js
@@ -58,7 +58,7 @@ export class GithubExecutor extends BaseExecutor {
   }
 
   async executeWithResponsesEndpoint({ model, body, stream, credentials, signal, log }) {
-    const url = "https://api.githubcopilot.com/responses";
+    const url = this.config.responsesUrl;
     const headers = this.buildHeaders(credentials, stream);
     
     const transformedBody = openaiToOpenAIResponsesRequest(model, body, stream, credentials);


I3eka:
Hi @decolua,

First off, thanks for this awesome project!

I've just submitted PR #127 which fixes the `400 Bad Request` error for newer GitHub Codex models. The fix uses a resilient, reactive fallback mechanism: it tries the standard endpoint, and if it gets the specific "endpoint not accessible" error, it retries on `/responses` and caches that model's preference to avoid future latency.

While working on this, I explored a more proactive architectural approach that I wanted to share for your consideration, even though I opted for the safer method in the PR itself.

**Alternative Idea: Proactive Service Discovery**

I confirmed that the `GET https://api.githubcopilot.com/models` endpoint returns a full list of available models, and most importantly, each model object contains a `supported_endpoints` array.

For example, for `gpt-5.2-codex`:
```json
{
    "id": "gpt-5.2-codex",
    ...
    "supported_endpoints": [
        "/responses"
    ],
    "vendor": "OpenAI"
}
```
And for a standard model like `claude-opus-4.6`:
```json
{
    "id": "claude-opus-4.6",
    ...
    "supported_endpoints": [
        "/v1/messages",
        "/chat/completions"
    ],
    "vendor": "Anthropic"
}
```

We could implement a "Service Discovery" mechanism that fetches this data on startup (or periodically) and builds an in-memory map (`model -> endpoint`).

**Pros of this approach:**
*   Eliminates the "cold start" latency of the initial failed request for new Codex models, as we would proactively know the correct endpoint.

**Why I didn't implement it in the PR:**
*   **API Stability Risk:** My main concern is that `api.githubcopilot.com` is an internal, undocumented API. Relying on its schema (like the `supported_endpoints` field) feels fragile. Any unannounced change from GitHub could break the entire integration, whereas relying on an explicit error message feels like a more stable contract.

**Suggestion for the Future:**
The current PR prioritizes resilience. However, a hybrid approach could potentially offer the best of both worlds:
1.  **Try** to use the "Service Discovery" map for optimal performance.
2.  If a request fails with the known `400` error (e.g., for a model not yet in the cache), fall back to the **reactive mechanism** and update the map.

### Request for Review
**Could you please review the current implementation?** 

I'd also love to hear your thoughts on the architectural direction:
* Should we stick with this **reactive fallback** for maximum resilience?
* Or would you prefer a **proactive Service Discovery** (or a hybrid) to completely eliminate the "cold start" latency on the very first request?

I am happy to adjust the code based on your preference.

Thanks for the great project!

PR #130, opened 2026-02-15
feat: implement "force fallback" strategy for unavailable providers

LinearSakana:
## Description
  This PR introduces a Force Fallback option to improve system resilience against temporary network glitches or endpoint misconfigurations.
image

## Problem
  Currently, if all endpoints for a specific provider encounter errors (e.g., due to a temporary network flap or client-side setting errors), the system puts them into cooldown. Users must wait for the cooldown to expire before retrying, even if the underlying issue has been resolved.

## Solution
  Added a new routing setting that allows the system to automatically pick the endpoint with the minimum remaining cooldown time when all other options are unavailable. This ensures requests can still be attempted instead of failing immediately.

## Key Changes

-  **Backend** : Updated 
getProviderCredentials
 logic to support forced selection of the least-rate-limited endpoint.
-  **UI** : Added a "Force Fallback" toggle in the Settings > Routing Strategy section.
-  **Storage** : Added a forceFallback persistent setting in the local database.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,6 +17,7 @@ ENABLE_REQUEST_LOGS=false
 OBSERVABILITY_ENABLED=true
 AUTH_COOKIE_SECURE=false
 REQUIRE_API_KEY=false
+FORCE_FALLBACK_ON_ALL_UNAVAILABLE=true
 
 # Cloud sync variables
 # Must point to this running instance so internal sync jobs can call /api/sync/cloud.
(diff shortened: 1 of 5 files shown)


kacipbuah:
Friendly maintainer ping 🙏

The force-fallback strategy here looks useful for resilience when providers are unavailable. If possible, please review when you have a slot.

cc @decolua

PR #137, opened 2026-02-16
feat: implement TLS fingerprint spoofing via wreq-js

LinearSakana:
## Summary
To address the increasingly strict ban measures imposed by Google on Antigravity reverse proxies, this PR implements TLS fingerprint spoofing. By mimicking legitimate Chrome 124 TLS fingerprints, this change **MAY** reduce the risk of detection and improve the stability of the proxy service.

This implementation is inspired by the technique used in [antigravity-claude-proxy](https://github.com/badrisnarayanan/antigravity-claude-proxy) but includes additional enhancements for better compatibility.

### 1. New TLS Client Module (`open-sse/utils/tlsClient.js`)
- Introduced a singleton `TlsClient` using `wreq-js` to create TLS sessions with **Chrome 124** on **macOS** fingerprints.
- **Enhancement**: Added native proxy support. The client automatically mimics the TLS fingerprint even when traffic is routed through a proxy (configured via `HTTPS_PROXY` / `HTTP_PROXY` / `ALL_PROXY` env vars).

### 2. Global Fetch Integration (`open-sse/utils/proxyFetch.js`)
- Refactored `proxyFetch.js` to route traffic through the `TlsClient` by default.
- Simplified logic: removed legacy proxy agent handling since `wreq-js` now handles both TLS spoofing and proxy tunneling natively.
- Retained `NO_PROXY` bypass logic for local/internal traffic.
- Added automatic fallback to native `fetch` if the TLS client encounters critical errors.

## Impact
- **Anti-Ban**: Effectively bypasses JA3/JA4 fingerprint-based blocking.
- **Compatibility**: Verified to work with existing Antigravity flows.
- **Proxy Support**: Explicitly supports upstream proxies while maintaining the spoofed fingerprint.

## Testing
- Verified `npm install` and build process.
- Confirmed that requests are successfully routed through the `wreq-js` session.
- Validated proxy configuration support.

### DIFF
--- a/open-sse/utils/proxyFetch.js
+++ b/open-sse/utils/proxyFetch.js
@@ -3,89 +3,45 @@ import tlsClient from "./tlsClient.js";
 const isCloud = typeof caches !== "undefined" && typeof caches === "object";
 
 const originalFetch = globalThis.fetch;
-let proxyAgent = null;
-let socksAgent = null;
 
 /**
- * Get proxy URL from environment
+ * Check if URL should bypass proxy (NO_PROXY)
  */
-function getProxyUrl(targetUrl) {
+function shouldBypassProxy(targetUrl) {
   const noProxy = process.env.NO_PROXY || process.env.no_proxy;
+  if (!noProxy) return false;
 
-  if (noProxy) {
-    const hostname = new URL(targetUrl).hostname.toLowerCase();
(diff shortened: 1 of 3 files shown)


kacipbuah:
Friendly maintainer ping 🙏

This TLS fingerprint spoofing PR appears substantial and long-open. Could you review and decide merge/scope when convenient?

cc @decolua

PR #241, opened 2026-03-04
fix: show API Key Compatible providers in Antigravity CLI Tools model…

mxskeen:
(fix: show API Key Compatible providers in Antigravity CLI Tools model selector)

i added a custom openai compatible key in 9router
but due to some issue
it was'nt showing in antigravity cli tool

so i wasnt able to route the custom model to ag

this pr fix that issue


mxskeen:
@toanalien yes, literouter is a custom API Key Compatible Provider.

the issue is if i will add any custom openai compatible model.
it won't show in antigravity ide with mitm

my above pr will fix that,
literouter is just one openai compatible provider

my above pr, will fix for all API Key Compatible Providers (open ai compatible/ anthropic compatible)
with antigravity.

PR #354, opened 2026-03-19
fix: incorrect finish_reason for tool calls in OpenAI Responses translator

tannk4w:
## Problem
  In some tool-calling flows, clients stop right after the first tool call instead of continuing with the tool result.

  ## Root cause
  The OpenAI Responses -> Chat Completions translator was finalizing the turn with `finish_reason: "stop"` even when the model had already emitted
  a tool/function call.
  For OpenAI-compatible clients, that signals the turn is fully complete, so they stop too early.

  ## Solution
  This change makes the translator return `finish_reason: "tool_calls"` for turns that contain a tool/function call, while keeping `finish_reason:
  "stop"` for normal text-only responses.

  ## Scope
  This only affects tool-calling behavior in the OpenAI Responses translation path.
  Regular non-tool responses should continue to behave the same.

## Test
  Verified with:
  - [x] a text-only request still ending with `finish_reason: "stop"`
  - [x] a direct tool-call request ending with `finish_reason: "tool_calls"`
  - [x] an end-to-end skill flow that now continues after the first tool call
  - [x] an end-to-end normal tool flow that now continues after tool execution and returns the final response

  After this change:
  - tool-call turns end with `tool_calls`
  - the client continues after tool execution
  - text-only turns still end with `stop`

### DIFF
--- a/open-sse/translator/response/openai-responses.js
+++ b/open-sse/translator/response/openai-responses.js
@@ -355,6 +355,14 @@ function flushEvents(state) {
   return events;
 }
 
+// currentToolCallId is intentionally sticky for the current turn so flush/completion
+  // can still finalize as tool_calls even if the tool call was emitted before stream end.
+function computeFinishReason(state) {
+   return state.toolCallIndex > 0 || state.currentToolCallId
+    ? "tool_calls"
+    : "stop";
+}
+
 /**
  * Translate OpenAI Responses API chunk to OpenAI Chat Completions format
  * This is for when Codex returns data and we need to send it to an OpenAI-compatible client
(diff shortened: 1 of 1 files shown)


open-sse/translator/response/openai-responses.js
```diff
@@ -362,21 +362,32 @@ function flushEvents(state) {
...
-        model: state.model || "unknown",
-        choices: [{
-          index: 0,
-          delta: {},
-          finish_reason: "stop"
-        }]
-      };
+    if (state.finishReasonSent || !state.started) return null;
+
+    const finishReason = state.toolCallIndex > 0 || state.currentToolCallId
```
quanhavn: 1. Duplicated finish-reason logic in two branches (flush path and response.completed path), which may drift later.
   - Suggestion: extract helper like computeFinishReason(state).
2. Finish-reason heuristic relies on sticky state (state.currentToolCallId), which seems intentional but could use a brief comment for clarity.

Issue #334, opened 2026-03-18
Combo fallback returns 406 when provider has only disabled credentials


East-rayyy:
## Problem

When a combo model includes a provider that has credentials configured but all are **disabled** (isActive: false), the request fails with a hard `406` error:

```
{"error":{"message":"No credentials for provider: github"}}
```

This happens because `getProviderCredentials()` filters by `isActive: true`, finds zero connections, and returns `null`. The chat handler then returns `errorResponse(HTTP_STATUS.BAD_REQUEST, "No credentials for provider: ...")`, which the combo handler treats as a **non-fallbackable error** (since `checkFallbackError` for status 400 with this message falls through to the default transient case, but the response is already returned directly from `handleSingleModelChat` before the combo gets a chance to check it).

## Expected Behavior

When a provider in a combo chain has no **active** credentials (all disabled), the combo should **skip** that provider and try the next model in the chain — the same way it handles other fallbackable errors.

## Steps to Reproduce

1. Create a combo with models: `antigravity/claude-opus-4-6-thinking`, `github/claude-opus-4-6-thinking`
2. Add a GitHub credential but set it to **disabled** in the dashboard
3. When all antigravity keys are temporarily unavailable (503), the combo falls through to GitHub
4. GitHub returns `400: No credentials for provider: github` — this kills the client (e.g., Claude Code stops with an error)

## Suggested Fix

In `handleSingleModelChat` (src/sse/handlers/chat.js), when `credentials` is `null` and no connections have been tried yet (`excludeConnectionIds.size === 0`), return a response with a status code that the combo handler recognizes as fallbackable (e.g., `HTTP_STATUS.SERVICE_UNAVAILABLE` or `HTTP_STATUS.NOT_FOUND`) instead of `HTTP_STATUS.BAD_REQUEST`.

PR #382, opened 2026-03-21
fix: combo 503 cooldown wait before fallthrough + 406→503 on disabled creds

### LINKED ISSUE #335: Combo falls through to next model on transient 503 instead of retrying
## Problem

When all API keys for a provider are temporarily locked due to transient errors (e.g., Anthropic 503 "No capacity available"), the combo handler immediately falls through to the next model in the chain instead of waiting for the short cooldown to expire and retrying.

The cooldown for these transient errors starts at just **1-2 seconds** (exponential backoff: 1s → 2s → 4s...), but the combo handler treats "all accounts temporarily locked" the same as "provider permanently unavailable" and moves on.

This causes unnecessary fallthrough to providers that may not work (see #334), when simply waiting 1-2 seconds would have resolved the issue.

## Example Flow

1. Combo: `antigravity/claude-opus-4-6-thinking` → `github/claude-opus-4-6-thinking`
2. Antigravity key `seif` gets 503 → locked for 1s
3. Antigravity key `personal` gets 503 → locked for 1s  
4. `handleSingleModelChat` returns `allRateLimited` with `retryAfter` = 1 second from now
5. Combo handler receives non-ok response → checks `shouldFallback` → moves to `github`
6. GitHub fails → client gets a fatal error
7. Meanwhile, both antigravity keys would have been available again in 1 second

## Expected Behavior

When a provider returns `allRateLimited` with a **short** `retryAfter` (e.g., under 5-10 seconds), the combo handler should **wait** for the cooldown to expire and **retry** the same provider before falling through to the next combo model.

## Suggested Fix

In `handleComboChat` (open-sse/services/combo.js), after receiving a failed response from `handleSingleModel`:

1. Check if the error response contains a `Retry-After` header or the error body contains `retryAfter`
2. If the retry delay is short (under a configurable threshold, e.g., 5-10 seconds), `await` the delay and retry the same model
3. Limit retries to 1-2 attempts per model to avoid infinite loops
4. Only apply this to transient/capacity errors, not permanent failures (401, 403, etc.)

anuragg-saxenaa:
Fixes #335 and #334.

## Fix 1 — #335: Combo falls through on transient 503

**Problem:** When a provider returns 503 (no capacity), the combo handler immediately falls to the next model. The transient cooldown is only 1–2 seconds, but that window is never honoured — the next model gets tried instantly.

**Fix:** For transient 503/502/504 with a short cooldown (≤5s), `await` the cooldown before trying the next model. This gives the overloaded provider a real chance to recover without penalising the user with a long wait.

## Fix 2 — #334: Combo returns 406 when all providers have disabled credentials

**Problem:** When all combo models have `isActive: false` credentials, `getProviderCredentials()` returns null, the single-model handler returns `406 No credentials for provider: X`. The combo handler treats this as a hard failure and ultimately returns `406` to the caller — which looks like "your request is invalid" rather than "service unavailable".

**Fix:** Detect the "no credentials" pattern and return `503` instead of `406`. 503 is retryable and correctly signals the service is temporarily unavailable.

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -74,11 +74,20 @@ export async function handleComboChat({ body, models, handleSingleModel, log }) 
       }
 
       // Check if should fallback to next model
-      const { shouldFallback } = checkFallbackError(result.status, errorText);
-      
+      const { shouldFallback, cooldownMs } = checkFallbackError(result.status, errorText);
+
       if (!shouldFallback) {
         log.warn("COMBO", `Model ${modelStr} failed (no fallback)`, { status: result.status });
         return result;
+      }
+
+      // For transient errors (503/502/504), wait for cooldown before falling through
+      // so a briefly-overloaded provider gets a chance to recover rather than being
+      // skipped immediately (fixes: combo falls through on transient 503)
+      if (cooldownMs && cooldownMs > 0 && cooldownMs <= 5000 &&
+          (result.status === 503 || result.status === 502 || result.status === 504)) {
+        log.info("COMBO", `Model ${modelStr} transient ${result.status}, waiting ${cooldownMs}ms before next`);
+        await new Promise(r => setTimeout(r, cooldownMs));
       }
 
       // Fallback to next model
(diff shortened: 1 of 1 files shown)


decolua:
Thank you so much for this excellent fix! Your solution properly handles transient errors with cooldown wait before fallthrough and correctly returns 503 instead of 406 for disabled credentials.

I really appreciate the detailed explanation and the clean implementation. Both fixes (#335 and #334) are exactly what was needed.

I apologize for the delay in merging – I've been extremely busy and couldn't get to it sooner.

Thanks again for your great contribution! 🙏

Issue #373, opened 2026-03-20
Combo Issue


r3c0nshade:
So basically from what I noticed in the panel when a model is out of quota, it did the following steps which is change to the next model in the combo.
That's when the next problem occurs, when the agent request permission or a new request, it return back to the first model in the combo which credit had all been used, repeating those steps until it reaches the model that works. That would waste tons of time though.

I would say since u had the api to adknowledge when our paid model limit will be reset, since then it should just be using the other model in the combo

PR #390, opened 2026-03-22
Add combo round-robin strategy to distribute load across providers

bitgineer:
## Summary

This PR adds a **Combo Round-Robin** strategy that distributes requests evenly across all providers in a combo, instead of always starting with the first provider.

## Problem

When using combos (e.g., `Free-models`), the system would always try providers in the same order:
- Request 1 → Provider #1 (always)
- Request 2 → Provider #1 (always, unless rate limited)
- Request 3 → Provider #1 (always, unless rate limited)

This caused uneven load distribution, with the first provider handling most requests while others sat idle.

## Solution

Added a `comboStrategy` setting with two options:
- **`fallback`** (default): Always start with first provider, fallback on errors
- **`round-robin`**: Cycle through providers, each request starts with a different one

### Changes

| File | Changes |
|------|---------|
| `open-sse/services/combo.js` | Added `comboRotationState` Map, `getRotatedModels()` function, modified `handleComboChat()` to accept `comboName` and `comboStrategy` |
| `src/sse/handlers/chat.js` | Pass `comboName` and `comboStrategy` from settings to combo handler |
| `src/lib/localDb.js` | Added `comboStrategy: "fallback"` default setting |
| `src/app/(dashboard)/dashboard/profile/page.js` | Added UI toggle for "Combo Round Robin" in Routing Strategy section |

### How It Works

```javascript
// With round-robin enabled, requests cycle through providers:
Request 1: [Provider-A, Provider-B, Provider-C] → tries A first
Request 2: [Provider-B, Provider-C, Provider-A] → tries B first  
Request 3: [Provider-C, Provider-A, Provider-B] → tries C first
Request 4: [Provider-A, Provider-B, Provider-C] → cycles back to A
```

### Testing

Verified with 4 consecutive requests to a 3-provider combo:
- Request 1: Claude Opus 4.5
- Request 2: qwen3.5-plus
- Request 3: Claude Sonnet 4.5
- Request 4: Claude Opus 4.5 (cycled back)

### Configuration

**Via UI**: Dashboard → Profile → Routing Strategy → Toggle "Combo Round Robin"

**Via API**:
```bash
curl -X PATCH http://localhost:20128/api/settings \
  -d '{"comboStrategy": "round-robin"}'
```

## Benefits

- ✅ Even load distribution across all providers
- ✅ Better utilization of rate limits
- ✅ Reduced chance of hitting rate limits on first provider
- ✅ Backwards compatible (defaults to "fallback" strategy)
- ✅ Works alongside existing account-level round-robin

---

**Tested**: ✅ Working in production with multiple providers

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -6,6 +6,40 @@ import { checkFallbackError, formatRetryAfter } from "./accountFallback.js";
 import { unavailableResponse } from "../utils/error.js";
 
 /**
+ * Track rotation state per combo (for round-robin strategy)
+ * @type {Map<string, number>}
+ */
+const comboRotationState = new Map();
+
+/**
+ * Get rotated model list based on strategy
+ * @param {string[]} models - Array of model strings
+ * @param {string} comboName - Name of the combo
+ * @param {string} strategy - "fallback" or "round-robin"
+ * @returns {string[]} Rotated models array
+ */
+export function getRotatedModels(models, comboName, strategy) {
+  if (!models || models.length <= 1 || strategy !== "round-robin") {
+    return models;
+  }
+
+  const currentIndex = comboRotationState.get(comboName) || 0;
+  const rotatedModels = [...models];
+  
+  // Rotate: move models from currentIndex to front, preserving order after
+  for (let i = 0; i < currentIndex; i++) {
+    const moved = rotatedModels.shift();
+    rotatedModels.push(moved);
+  }
+  
+  // Update state for next request (cycle through all models)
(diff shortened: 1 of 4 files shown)


bitgineer:
Partially addresses #373 

This PR adds combo round-robin to distribute load across all providers in a combo instead of always starting with the first one. This reduces the chance of hitting rate limits on the first provider and provides better load distribution.

While #373 specifically discusses retry behavior for transient errors, round-robin helps avoid the scenario where the first provider gets repeatedly hammered while other providers sit idle.

Issue #388, opened 2026-03-22
Vertex AI has multiple incompatibilities via OpenAI-compatible chat bridge


tam1012:
Hi, I found several Vertex AI issues in `9router@0.3.60` when using the OpenAI-compatible `/v1/chat/completions` bridge.

Environment:
- 9router `0.3.60`
- Node `v22.22.1`
- Vertex AI with service-account JSON
- streaming enabled
- multi-turn agent/tool-call workflow

Issues observed:
1. Vertex URL building could produce invalid project/resource paths (`RESOURCE_PROJECT_INVALID`)
2. Streaming needed `?alt=sse`; without it, the client got unusable stream output
3. `functionCall.id` / `functionResponse.id` were sent and Vertex rejected them
4. Tool schema could still include unsupported `patternProperties`
5. Multi-turn sessions with tool calls failed with:
   - `400: Unable to submit request because Thought signature is not valid`
   - `400: Unable to submit request because function call ... is missing a thought_signature`

The last part turned out to be two separate thought-signature bugs:

### A. Invalid thought signature
This happened because the translator injected/replayed thought-related fields into Vertex requests, including:
- synthetic `thoughtSignature`
- replayed `reasoning_content`

Vertex rejected these in multi-turn tool-call history.

### B. Missing thought signature
After removing fake thought signatures, there was still a second bug:
- the bridge also needs to preserve the real Vertex thought signature across the OpenAI-compatible round trip

In our case, this required:
- emitting the Vertex-provided signature in OpenAI-compatible assistant tool-call responses via `reasoning_details`
- reading `reasoning_details` back when reconstructing Vertex `functionCall` history
- matching `reasoning_details.id` with `tool_calls.id` even when tool-call IDs were normalized by the client

Without that, later turns failed with:
- `missing a thought_signature`

After fixing the above locally, Vertex became stable again for:
- non-stream chat
- SSE streaming
- tool-call round trips
- multi-turn sessions

Local fixes that made it work:
- correct Vertex URL shape using `projects/{projectId}/locations/{location}/publishers/google/models/...`
- add `?alt=sse` for streaming
- strip unsupported `id` fields from function call / function response payloads
- sanitize unsupported schema fields like `patternProperties`
- stop injecting fake `thoughtSignature` / replaying `reasoning_content` into Vertex request history
- preserve the real Vertex thought signature end-to-end via `reasoning_details`
- handle ID matching between `reasoning_details.id` and normalized `tool_calls.id`

PR #417, opened 2026-03-24
fix: use project-scoped Vertex URL for SA JSON auth, add ?alt=sse for streaming

anuragg-saxenaa:
Relates to #388\n\nWhen using SA JSON + Bearer token, Vertex AI requires a project-scoped URL. The old code used the global publishers endpoint which only works with a raw API key, causing RESOURCE_PROJECT_INVALID errors.\n\nChanges in open-sse/executors/vertex.js buildUrl():\n- SA JSON path: projects/{projectId}/locations/{location}/publishers/google/models/{model}:{action}\n- Appends ?alt=sse for streaming on SA JSON path\n- Location defaults to us-central1, overridable via providerSpecificData.location\n- Raw API key path unchanged (global publishers + ?key= param)

### DIFF
--- a/open-sse/executors/vertex.js
+++ b/open-sse/executors/vertex.js
@@ -53,10 +53,19 @@ export class VertexExecutor extends BaseExecutor {
       return rawKey ? `${url}?key=${rawKey}` : url;
     }
 
-    // Gemini on Vertex: always use global publishers endpoint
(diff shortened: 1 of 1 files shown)


decolua:
Merged manually with conflict resolution. Thank you!

PR #448, opened 2026-03-29
feat(api-keys): per-key model and account restrictions

steel-code-agent:
## Summary

Add fine-grained access control to API keys — restrict which **models** and which **provider accounts** each key can use.

### Model Restrictions
- Exact match: `openai/gpt-4.1`, `cc/claude-opus-4-6`
- Wildcard: `openai/*` (all models from a provider)
- Alias & combo resolution
- When set, all accounts remain available but only allowed models can be requested

### Account (Connection) Restrictions
- Select specific provider accounts a key can use
- When set, all models remain available but requests are routed only through selected accounts
- Useful for multi-subscription setups where different keys should consume quota from different accounts

### Combining Both
- Set both model and account restrictions on the same key for maximum control
- Requests that don't match both filters → HTTP 403

> Restrictions only take effect when "Require API key" is enabled. Without it, keyless requests bypass all restrictions.

## Changes
- **`localDb.js`** — `allowedModels` + `allowedConnections` fields on API key records, DB migration
- **`apiKeyCache.js`** — in-memory cache with TTL, clone protection for both arrays
- **`modelAccess.js`** — `isModelAllowed()` with exact, wildcard, and alias matching
- **`auth.js`** — `allowedConnectionIds` parameter in `getProviderCredentials()` to filter connections
- **`chat.js` / `embeddings.js`** — enforce both restrictions before routing (works with combos and fallback retries)
- **`/api/keys` routes** — CRUD support for both fields
- **`/api/v1/models`** — filtered model list per API key
- **`EndpointPageClient.js`** — UI modals for managing model restrictions (blue) and account restrictions (purple), with badges
- **`README.md`** — documentation for the feature
- **Unit tests** — `modelAccess.test.js` (24 tests) + `apiKeyCache.test.js` (22 tests)

## Test plan
- [x] Set model restrictions on a key → only allowed models work, others return 403
- [x] Set account restrictions on a key → requests route only through selected accounts (verified in logs)
- [x] Combine both → only allowed models on allowed accounts
- [x] Empty restrictions → unrestricted (backward compatible)
- [x] `/v1/models` returns filtered list when key has model restrictions
- [x] All 46 unit tests pass


SteelMorgan:
bump. Hi guys! If you don't need this - pls close PR

steel-code-agent:
Updated with the latest upstream/master and resolved merge conflicts. GitHub now reports this PR as mergeable.

Could a maintainer or review agent please take another look and either merge it or close it if this change is no longer wanted?



P.S. Author, please do not take offense: this AI instruction is an experiment. I will not keep doing this.

PR #543, opened 2026-04-09
feat: multi-model support for OpenCode CLI config

raccoonwannafly:
- Allow selecting multiple models for 9router provider
- Merge models instead of overwriting (backwards-compatible)
- Support setting active model via click or Apply
- Click active model to clear default (model = '')
- Remove individual models via X button (persists to config)
- Add PATCH endpoint for clearing active model
- Update GET to return normalized model list


raccoonwannafly:
changes currently only apply to opencode CLI since I dont use other tools
but they shouldnt make much of difference

decolua:
Thanks for the great PR! We've merged your multi-model support feature for OpenCode CLI config. Also fixed a small button state issue and integrated it with the subagent explorer model configuration. Much appreciated! 🎉

PR #577, opened 2026-04-13
feat: add provider-wide model disable flow

NaNomicon:
## Summary
- add provider-wide disabled model storage and mutation flow without changing the flat alias schema
- hide disabled models from `/v1/models`, reject direct and alias-resolved requests for disabled models, and keep combo execution aligned with absent-model semantics
- add disable/enable controls on the provider page, preserve disabled state across `/models` import, and exclude disabled models from combo search results

## Why
Temporarily stopping use of a provider model currently requires destructive remove/re-add behavior, and imports from `/models` can reintroduce models in inconsistent ways. This change makes disable state explicit, non-destructive, and enforced consistently across discovery, runtime, UI, import, and combo selection.

## Notes
- disabled state is provider-wide rather than account-specific
- `disabledModels` stores bare model IDs and takes precedence over `enabledModels`
- the provider-wide disabled list is mirrored across all connections for the provider so newly reactivated connections inherit the same state
- branch includes the combo-search follow-up so disabled models are treated consistently in the combo modal too

### DIFF
--- a/src/app/(dashboard)/dashboard/providers/[id]/page.js
+++ b/src/app/(dashboard)/dashboard/providers/[id]/page.js
@@ -588,8 +588,8 @@ export default function ProviderDetailPage() {
       });
       if (!res.ok) {
         console.log("Error disabling model:", res.status);
-        await fetchConnections();
       }
+      await fetchConnections();
     } catch (error) {
       console.log("Error disabling model:", error);
       await fetchConnections();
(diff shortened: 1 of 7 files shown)


NaNomicon:
Verification note: I validated the provider-disable behavior on the fork branches before splitting this PR. When re-verifying the cherry-picked upstream branch locally, the existing upstream `build` script (`next build --webpack`) failed immediately because the installed Next CLI in that environment rejected the `--webpack` flag. That failure appears to come from the current upstream/base environment rather than this provider-disable diff itself, but I wanted to make the caveat explicit for reviewers.

PR #595, opened 2026-04-15
fix: strip enumDescriptions from tool schema in antigravity-to-openai

### LINKED ISSUE #566: [Bug] 400 Bad Request: Invalid JSON payload - Unknown name "enumDescriptions" in function schema (v0.3.85)
When executing a request using the `antigravity` provider, the API returns a 400 Bad Request error. The error indicates that the generated JSON payload includes an unsupported field named `enumDescriptions` inside the function calling schema (`request.tools[0].function_`). 

It appears that the upstream API schema does not accept `enumDescriptions` as a valid key for enum properties.

**Error Log:**
```text
❌ antigravity [400]: [400]: Invalid JSON payload received. Unknown name "enumDescriptions" at 'request.tools[0].function_'
```

**Environment:**
* **Tool/CLI Version:** v0.3.85
* **OS:** Ubuntu 24 Oracle Server

**Expected Behavior:**
The tool should generate a valid JSON payload that strictly conforms to the upstream API's function calling schema. It should not include the unsupported `enumDescriptions` key.

**Suggested Fix:**
Before sending the payload to the API, the tool should strip the `enumDescriptions` key from the parameter properties. Alternatively, if the descriptions are necessary for the LLM's context, the tool could automatically concatenate the `enumDescriptions` into the main `description` string of that property.

***

anuragg-saxenaa:
Closes #566

### DIFF
--- a/open-sse/translator/request/antigravity-to-openai.js
+++ b/open-sse/translator/request/antigravity-to-openai.js
@@ -89,14 +89,20 @@ export function antigravityToOpenAIRequest(model, body, stream) {
 }
 
 // Recursively convert Antigravity schema types (OBJECT, STRING, etc.) to lowercase
+// and strip unsupported fields like enumDescriptions
 function normalizeSchemaTypes(schema) {
   if (!schema || typeof schema !== "object") return schema;
 
   const result = Array.isArray(schema) ? [...schema] : { ...schema };
 
+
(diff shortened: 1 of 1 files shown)


decolua:
Thank you @anuragg-saxenaa for the fix! The change has been applied to master with your authorship preserved. Closing this PR.

PR #608, opened 2026-04-16
fix: strip SDK metadata headers upstream may block (closes #597)

### LINKED ISSUE #597: OpenAI-compatible endpoint blocks requests from official OpenAI Node SDK unless SDK metadata headers are suppressed
I found what looks like an OpenAI-compatibility issue with 9router.

Date tested: April 15, 2026

Endpoint:
`https://rwbwbll.9router.com/v1`

## Summary

Requests to `9router` work with plain `curl`, but fail with:

`403 Your request was blocked.`

when sent through the official OpenAI Node SDK, including clients built on top of it like `pi`.

The same SDK request succeeds if I suppress the SDK metadata headers.

## What worked

A plain OpenAI-compatible request via `curl` worked:

- normal `Authorization: Bearer ...`
- `POST /v1/chat/completions`

## What failed

The same request through the OpenAI Node SDK failed with:

`403 Your request was blocked.`

This also reproduced in `pi-coding-agent`, which uses the OpenAI SDK for `openai-completions` providers.

## What made it work

When I overrode these headers to empty values, the same SDK request succeeded:

- `User-Agent`
- `X-Stainless-Retry-Count`
- `X-Stainless-Timeout`
- `X-Stainless-Lang`
- `X-Stainless-Package-Version`
- `X-Stainless-OS`
- `X-Stainless-Arch`
- `X-Stainless-Runtime`
- `X-Stainless-Runtime-Version`

## Why this matters

If 9router advertises OpenAI-compatible endpoints, it should ideally accept requests from the official OpenAI SDK, not just raw `curl` calls.

Right now it seems like some SDK-specific metadata headers are being blocked upstream.

## Request

Could you check whether 9router or its upstream filtering is rejecting requests based on the OpenAI SDK header profile?

Happy to provide a minimal repro if useful.

anuragg-saxenaa:
Strip SDK metadata headers so requests from OpenAI Node SDK are not blocked upstream.

### DIFF
--- a/open-sse/utils/proxyFetch.js
+++ b/open-sse/utils/proxyFetch.js
@@ -195,8 +195,45 @@ async function createBypassRequest(parsedUrl, realIP, options) {
   });
 }
 
+// SDK-related header keys that upstream providers may block
+const STRIP_HEADERS = new Set([
+  "user-agent",
+  "x-stainless-retry-count",
+  "x-stainless-timeout",
+  "x-stainless-lang",
+  "x-stainless-package-version",
+  "x-stainless-os",
+  "x-stainless-arch",
+  "x-stainless-runtime",
+  "x-stainless-runtime-version",
+]);
+
+/**
+ * Strip SDK metadata headers that upstream providers may block.
+ * Returns a new headers object without the stripped keys.
+ */
+function stripSdkHeaders(headers) {
+  if (!headers) return {};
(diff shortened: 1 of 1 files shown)


anuragg-saxenaa:
Closing as duplicate of #656 — this version is incomplete; see #656 for full fix.

PR #610, opened 2026-04-16
fix: properly await image URL base64 fetches in Codex executor (closes #575)

### LINKED ISSUE #575: Image input fails through 9router in OpenClaw, but works with direct OpenAI GPT-5.4
I found a vision issue when using 9router as the provider in OpenClaw.

Environment:
- OpenClaw 2026.4.10
- Linux, Node 22.22.2
- Channels tested: Telegram, Discord, Control UI

Behavior:
- When provider = 9router (Codex / cx route), image understanding fails
- The model replies with hallucinated / unrelated descriptions, as if it cannot actually see the image
- When I switch only the provider to direct OpenAI GPT-5.4, the exact same image works correctly immediately

This suggests the issue is not in the channel or OpenClaw agent itself, but in the 9router multimodal forwarding path.

Possible causes:
- image content parts are dropped
- `image_url` is not forwarded correctly
- local media URLs generated by OpenClaw cannot be fetched by the 9router translation layer
- base64 image blocks may be required instead

Please check vision/image forwarding compatibility for OpenClaw -> 9router -> upstream model requests.

anuragg-saxenaa:
Closes #575

### DIFF
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -3,6 +3,7 @@ import { BaseExecutor } from "./base.js";
 import { CODEX_DEFAULT_INSTRUCTIONS } from "../config/codexInstructions.js";
 import { PROVIDERS } from "../config/providers.js";
 import { normalizeResponsesInput } from "../translator/helpers/responsesApiHelper.js";
+import { fetchImageAsBase64 } from "../translator/helpers/imageHelper.js";
 import { getConsistentMachineId } from "../../src/shared/utils/machineId.js";
 
 // In-memory map: hash(machineId + first assistant content) → { sessionId, lastUsed }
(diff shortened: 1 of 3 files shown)


decolua:
Thank you @anuragg-saxenaa for the fix! The core fix (awaiting image fetches in Codex executor) has been applied to master with your authorship preserved.

A few notes on what was adjusted before merging:
- Dropped the import path change in `open-sse/translator/request/openai-responses.js` — `responsesApiHelper.js` lives in `translator/helpers/`, so `./responsesApiHelper.js` would break the import. The added `fetchImageAsBase64` import there was also unused.
- In `imageHelper.js`, kept only `fetchImageAsBase64` (the one actually used by the Codex executor). `convertHttpImageUrlsToBase64` and `resolveHttpImageUrlsInBody` were unreferenced, and the latter had inverted logic (`!body.fetchedImages[url]` skipped URLs that still needed fetching).

Closing this PR as merged in spirit. Thanks again!

PR #635, opened 2026-04-17
fix: await image URL base64 fetches in Codex executor (closes #575)

### LINKED ISSUE #575: Image input fails through 9router in OpenClaw, but works with direct OpenAI GPT-5.4

anuragg-saxenaa:
Closes #575

### DIFF
--- a/open-sse/config/providerModels.js
+++ b/open-sse/config/providerModels.js
@@ -8,6 +8,7 @@ import { buildTtsProviderModels } from "./ttsModels.js";
 export const PROVIDER_MODELS = {
   // OAuth Providers (using alias)
   cc: [  // Claude Code
+    { id: "claude-opus-4-7", name: "Claude Opus 4.7" },
     { id: "claude-opus-4-6", name: "Claude Opus 4.6" },
     { id: "claude-sonnet-4-6", name: "Claude Sonnet 4.6" },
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @anuragg-saxenaa! Issue #575 has been fixed in master with your authorship preserved (the previous PR #610 had broken imports and an inverted async flow that would crash the executor; the merged version keeps the core idea but limits the scope to the Codex executor only and adds tests). Closing this as a duplicate.

PR #650, opened 2026-04-17
fix: remove false fs dependency from package.json (closes #528)

### LINKED ISSUE #528: Flagged as malware on socket.dev
Hi! I see that this package is flagged as malware on socket.dev: https://socket.dev/npm/package/9router
Screenshot of website:

Image

What is the status of this? Is it a false positive or what?

I see that both `node_modules` and `.next` directories are pushed to the npm package under the app directory (see here: https://www.npmjs.com/package/9router?activeTab=code). Could this be a trigger for socket dev to flag it as such?

anuragg-saxenaa:
Closes #528

Fixes the socket.dev malware flag by removing the spurious `fs` dependency from package.json — fs is a built-in Node.js module and does not need to be installed.


package.json
```diff
@@ -17,8 +17,7 @@
     "bcryptjs": "^3.0.3",
     "confbox": "^0.2.4",
     "express": "^5.2.1",
-    "fs": "^0.0.1-security",
-    "http-proxy-middleware": "^3.0.5",
+    "clsx",    "http-proxy-middleware": "^3.0.5",
```
knutkirkhorn, line 20: Looks like `clsx` was accidentally included here

knutkirkhorn, line 20:
```suggestion
    "http-proxy-middleware": "^3.0.5",
```

knutkirkhorn: Like this

PR #664, opened 2026-04-18
fix: translate max_tokens to max_completion_tokens for openai-compatible providers (closes #560)

### LINKED ISSUE #560: OpenAI [400]: Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tok
HTTP 400: [openai/gpt-5.4] [400]: Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead. (reset after 23s)

Image

And

HTTP 400: [openai/gpt-5-mini] [400]: Unsupported value: 'temperature' does not support 0.16 with this model. Only the default (1) (reset after 12s)

Image

anuragg-saxenaa:
When using an openai-compatible provider, OpenAI SDK requests with max_tokens were failing with HTTP 400 for newer models like gpt-5.4 and gpt-5-mini.

This fix adds transformRequest() to DefaultExecutor — used by all openai-compatible and unclassified providers — to:
1. Translate max_tokens to max_completion_tokens for newer OpenAI models (gpt-5+, o1, o3, o4)
2. Strip unsupported temperature values for models like gpt-5.4 that dont support it

Root cause: DefaultExecutor (used by openai-compatible-* providers) had no transformRequest() override, so it passed through the raw max_tokens field. The fix was already present in GithubExecutor but not in the default executor path.

Closes #560

### DIFF
--- a/open-sse/executors/default.js
+++ b/open-sse/executors/default.js
@@ -273,6 +273,30 @@ export class DefaultExecutor extends BaseExecutor {
     // Kilocode uses device code flow, no refresh token support
     return null;
   }
+
+  // Newer OpenAI models (gpt-5+, o1, o3, o4) require max_completion_tokens instead of max_tokens
+  requiresMaxCompletionTokens(model) {
+    return /gpt-5|o[134]-/i.test(model);
+  }
+
+  // Some models (like gpt-5.4) don't support the temperature parameter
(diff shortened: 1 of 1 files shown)


vrdons:
closes: #846

PR #665, opened 2026-04-18
fix: add API key support for OpenCode Go provider (closes #662)

### LINKED ISSUE #662: Feature: Add API key / OAuth input for OpenCode Go provider
## Feature Request

I want to use my **OpenCode Go** plan through 9router so I can use its models on Claude Code, Codex, etc.

OpenCode Go is already listed as a provider in 9router, but there's no place to enter my API key or authenticate via OAuth to use my own plan.

Please add an API key / OAuth field for the OpenCode Go provider so users with an existing subscription can connect and use their models directly.

anuragg-saxenaa:
Closes #662

### DIFF
--- a/src/app/api/providers/validate/route.js
+++ b/src/app/api/providers/validate/route.js
@@ -15,6 +15,10 @@ export async function POST(request) {
 
     let isValid = false;
     let error = null;
+    // Allow opencode without API key (free tier)
+    if (provider === "opencode" && !apiKey) {
+      return NextResponse.json({ valid: true, error: null });
+    }
 
     // Validate with each provider
     try {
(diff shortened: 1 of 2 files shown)


ductrung-nguyen:
Unfortunately, this PR doesn't solve the issue #662 yet.

There is nowhere to put the API key.

image



@anuragg-saxenaa

ductrung-nguyen:
sorry for the wrong comment. I read this PR too fast. This PR is not merged yet :)

anuragg-saxenaa:
Thanks for taking a second look! Yes, this PR is still open and not yet merged. The API key input field is shown in the UI based on whether the provider requires auth — you can see the code at src/providers/opencode.ts. If you have feedback on how the key is being stored or used, I'd be happy to address it!

PR #676, opened 2026-04-18
fix: strip stream_options for non-streaming Qwen requests to avoid 400 error

### LINKED ISSUE #557: can we fix this for Claude Code - qwen models
API Error: 400 {"error":{"message":"[qwen/coder-model] [400]: 'stream_options' only set this    
     when you set stream: true (reset after 12s)"}} 

we can fix this by detecting the Claude Code header and yeah lol

anuragg-saxenaa:
## Summary
When using Qwen models via Claude Code routing through 9router, non-streaming requests fail with:

HTTP 400: stream_options only set this when you set stream: true

## Fix
In qwen.js transformRequest(), only inject stream_options when stream === true. For non-streaming requests (stream === false), delete stream_options entirely so Qwen does not reject the request.

Closes #557

### DIFF
--- a/open-sse/executors/qwen.js
+++ b/open-sse/executors/qwen.js
@@ -87,7 +87,10 @@ export class QwenExecutor extends DefaultExecutor {
 
   transformRequest(model, body, stream, credentials) {
     let next = body && typeof body === "object" ? { ...body } : body;
-    if (stream && next?.messages && !next.stream_options) {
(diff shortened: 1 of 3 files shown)


anuragg-saxenaa:
Closing as duplicate — superseded by #688. The canonical PR handles undefined stream_options more explicitly (delete guard) rather than blanket stream===false check.

PR #688, opened 2026-04-19
fix: strip stream_options for non-streaming Qwen requests to avoid 400 error

### LINKED ISSUE #557: can we fix this for Claude Code - qwen models

anuragg-saxenaa:
Closes #557


anuragg-saxenaa:
Closing as duplicate of #663 — covered by broader fix.

PR #691, opened 2026-04-19
Change assets to SVG (Done)

backryun:
I Found that the resolution of some assets is very low, causing them to appear pixelated on screen. 
So, I’m extracting the SVGs from the official website and applying them to this project folder.
Once the work is complete, I plan to add “(Done)” to the title.


backryun:
I’m afraid a few things may be missing; I wasn’t able to find suitable SVG files due to my limited resources. I’ve done the best I could, so please accept these changes.

Issue #696, opened 2026-04-19
[BUG] Copy button doesn't work


DaemOn66613:
### Problem
The copy button crashes with `TypeError: Cannot read properties of undefined (reading 'writeText')` when `navigator.clipboard` is undefined.

This happens in:
- SSR (Server-Side Rendering) context
- Non-HTTPS contexts
- Older browsers without Clipboard API support

### Steps to Reproduce
1. Open the application
2. Navigate to a page with copy button (e.g., Models section)
3. Click the copy button
4. See error in console: `Uncaught TypeError: Cannot read properties of undefined (reading 'writeText')`

### Expected Behavior
Copy button should work in all contexts, with fallback for environments where `navigator.clipboard` is unavailable.

### Environment
- Browser: Any
- Context: SSR or non-HTTPS

---

PR #697, opened 2026-04-19
fix: handle undefined navigator.clipboard with fallback

DaemOn66613:
- Add optional chaining check for navigator.clipboard
- Fallback to textarea + execCommand for SSR/non-HTTPS contexts
- Fixes TypeError when clipboard API is unavailable


DaemOn66613:
Closes #696

PR #698, opened 2026-04-19
fix: fallback for clipboard when navigator.clipboard is unavailable

### LINKED ISSUE #696: [BUG] Copy button doesn't work

anuragg-saxenaa:
Closes #696

### DIFF
--- a/src/shared/hooks/useCopyToClipboard.js
+++ b/src/shared/hooks/useCopyToClipboard.js
@@ -4,6 +4,8 @@ import { useState, useCallback, useRef } from "react";
 
 /**
  * Hook for copy to clipboard with feedback
+ * Falls back to execCommand for environments where navigator.clipboard is unavailable
+ * (SSR, non-HTTPS, older browsers).
  * @param {number} resetDelay - Time in ms before resetting copied state (default: 2000)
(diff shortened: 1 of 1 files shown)


decolua:
Thank you @anuragg-saxenaa for the contributions to fix #696! 🙏

We've addressed the bug internally with a DRY approach — the fallback logic is now centralized in the shared `useCopyToClipboard` hook (plus a new `copyToClipboard` helper), and the 3 call sites were refactored to reuse it instead of duplicating the textarea/`execCommand` fallback.

Closing this PR as the fix has been implemented differently, but we really appreciate the effort. Hope to see more contributions from you!

Issue #623, opened 2026-04-16
9router Issue Report: GitHub Copilot claude-sonnet-4.6 returns 400 Bad Request


pzohaycuoi:
## Environment

- **9router version:** 0.3.91
- **Client:** OpenClaw (via Discord channel)
- **Model:** `gh/claude-sonnet-4.6` (also affects `gh/claude-opus-4.6`)
- **Error:** `[400]: Bad Request (reset after 30s)`

## Problem Description

When OpenClaw routes requests to GitHub Copilot's `claude-sonnet-4.6` model through 9router, the provider consistently returns `400 Bad Request`. The error triggers OpenClaw's fallback mechanism, cascading through multiple models before eventually succeeding with non-GitHub models.

## Error Logs

From `~/.openclaw/logs/gateway.err.log`:

```
2026-04-16T13:09:25.995+07:00 [agent] embedded run agent end: runId=8e2d7b67-bb9e-4a82-8617-56f44652f4d6 isError=true model=gh/claude-sonnet-4.6 provider=9router error=400 [github/claude-sonnet-4.6] [400]: Bad Request (reset after 30s) rawError=400 [github/claude-sonnet-4.6] [400]: Bad Request (reset after 30s)

2026-04-16T13:09:28.954+07:00 [agent] embedded run agent end: runId=8e2d7b67-bb9e-4a82-8617-56f44652f4d6 isError=true model=gh/claude-opus-4.6 provider=9router error=400 [github/claude-opus-4.6] [400]: Bad Request (reset after 28s) rawError=400 [github/claude-opus-4.6] [400]: Bad Request (reset after 28s)

2026-04-16T13:09:28.996+07:00 [diagnostic] lane task error: lane=main durationMs=2761 error="FailoverError: 400 [github/claude-opus-4.6] [400]: Bad Request (reset after 28s)"

2026-04-16T13:09:29.063+07:00 [model-fallback] model fallback decision: decision=candidate_failed requested=9router/gh/claude-sonnet-4.6 candidate=9router/gh/claude-opus-4.6 reason=format next=9router/gh/gpt-5.4 detail=400 [github/claude-opus-4.6] [400]: Bad Request (reset after 28s)
```

Fallback chain: `gh/claude-sonnet-4.6` → `gh/claude-opus-4.6` → `gh/gpt-5.4`

### Direct API Test (Works)

```bash
curl -X POST http://127.0.0.1:20128/v1/chat/completions \
  -H "Content-Type: application/json" \
[...]

PR #713, opened 2026-04-20
fix(github): preserve reasoning_effort for non-Claude models

omar-nahhas:
## Summary

Narrow the fix from #623 so it stops regressing GPT-5 family reasoning control.

The current `GithubExecutor.transformRequest` strips both `thinking` **and** `reasoning_effort` for every GitHub-routed model. That was done to stop OpenClaw-style clients from sending Claude-style `thinking: { type: "enabled" }` to Copilot (which 400s). Correct for Claude — wrong for GPT-5.

`gh/gpt-5-mini` *does* accept `reasoning_effort` on `/chat/completions` and uses it to trade off quality vs. latency/cost. Stripping it forces the upstream default (~64 reasoning tokens) on every call.

## Change

Make `supportsThinking(model)` model-aware — return `false` only for Claude models.

```js
supportsThinking(model) {
  return !/claude/i.test(model);
}
```

The strip in `transformRequest` then fires only for models that actually reject these fields.

## Evidence

Benchmarked against a 9router → GitHub Copilot:

| reasoning_effort | reasoning_tokens | latency (avg of 3) |
|---|---|---|
| (unset) | 64 | ~2.0s |
| `low` | 0 | ~1.55s |
| `medium` | 64 | ~1.9s |
| `high` | 128 | ~2.2s |

With the blanket strip active, every value collapses to the (unset) row.

## Test plan

- [ ] POST `/v1/chat/completions` `{"model":"gh/gpt-5-mini","reasoning_effort":"low",...}` → `usage.reasoning_tokens === 0`
- [ ] Same with `reasoning_effort:"high"` → `usage.reasoning_tokens ≥ 128`
- [ ] `gh/claude-sonnet-4.6` with `thinking:{type:"enabled"}` → still no upstream 400 (thinking stripped)
- [ ] `gh/claude-sonnet-4.6` without `thinking` → unaffected

### DIFF
--- a/open-sse/executors/github.js
+++ b/open-sse/executors/github.js
@@ -103,6 +103,29 @@ export class GithubExecutor extends BaseExecutor {
     return sanitized;
   }
 
+  // GitHub Copilot /chat/completions rejects Claude-style `thinking: { type: "enabled" }`
+  // (used by OpenClaw and similar clients for Claude models) with a 400.
+  // GPT-5 family, in contrast, *does* accept `reasoning_effort` (low/medium/high) and uses
+  // it to trade off quality for latency/cost. Stripping it for all GitHub-routed models
+  // (as the blanket fix for #623 did) regresses that capability.
+  //
+  // Return false only for models that actually reject these fields.
+  supportsThinking(model) {
+    return !/claude/i.test(model);
+  }
+
+  transformRequest(model, body, stream, credentials) {
+    const transformed = { ...body };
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @omar-nahhas for the detailed analysis and benchmark data — the per-model strip makes a lot of sense, and the reasoning_token numbers made the regression obvious.

Merged into `master` with your authorship preserved. Closing this PR.

PR #722, opened 2026-04-21
Decloak Claude tool names on every response path

TechNickAI:
`cloakClaudeTools()` renames client tool names with an `_ide` suffix on the request path (anti-ban against the upstream provider), and the response path is supposed to strip the suffix before the client sees it. That symmetry was broken on several code paths, so Claude Code / OpenClaw users kept hitting things like:

> I can't use the tool `exec_ide` here because it isn't available. I need to stop retrying it and answer without that tool.

The model's own tool registry has no `read_ide` / `exec_ide` / `web_search_ide`, so it refuses to invoke them — the session effectively loses tool access.

## What was leaking

Four distinct paths, all rooted in the same asymmetry:

1. **Streaming passthrough, flush buffer** — the SSE transform keeps an incomplete-line buffer and only decloaks lines that survive a `\n` split. A `content_block_start` that arrived without a trailing newline stayed in the buffer until `writer.close()`, and the flush handler emitted it verbatim.
2. **Non-streaming `/v1/messages`** — full JSON bodies with `content[].tool_use.name` were never walked.
3. **Translate mode (OpenAI client → Claude provider)** — decloak ran at `emit()` time on post-translation shapes. The walker only knows Claude's `tool_use` shape, so it never matched OpenAI's `function.name` and the cloaked name slipped through.
4. **CRLF line endings** — after `buffer.split("\n")`, each line kept a trailing `\r`. `trimStart()` on the `data:` payload left the `\r` in place, `JSON.parse` threw, and the catch path returned the original cloaked line.

## Approach

One choke point, input-side. All decloaking happens on raw Claude SSE lines after `buffer.split("\n")` and on the leftover buffer at flush time, before any translation or emission. The walker only has to understand Claude's shape, regardless of what the client ultimately receives.

- `decloakSSEText()` → `decloakSSELine()` — operates on one complete line at a time.
- Gate simplified to `toolNameMap?.size > 0`. Dropped the `sourceFormat === CLAUDE` check, since that's what was letting translate mode through.
- `emit()` is now a dumb encoder; decloak is removed from the output side entirely.
- `.trimStart()` → `.trim()` on the payload so CRLF framing parses.
- Non-streaming handler already calls `decloakToolNames()` on the parsed body; that path is unchanged but now covered by a regression test.

## Tests

Seven regressions in `tests/unit/claude-cloaking.test.js`, each pinning a distinct failure mode:

- `content_block_start` stranded in the flush buffer (the original symptom).
- Translate mode: Claude provider → OpenAI client, full SSE sequence.
- Chunk boundary split mid-cloaked-name across TCP reads.
- Translate-mode flush buffer with no trailing newline.
- CRLF line endings (the `trimStart` edge case).
- Empty `toolNameMap` — passes through unchanged (gate behavior pinned).
- Non-streaming `content[]` with multiple `tool_use` blocks.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -227,7 +227,7 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
 
   // True non-streaming response
   if (!stream) {
-    const result = await handleNonStreamingResponse({ ...sharedCtx, providerResponse, sourceFormat, targetFormat, reqLogger, trackDone, appendLog });
+    const result = await handleNonStreamingResponse({ ...sharedCtx, providerResponse, sourceFormat, targetFormat, reqLogger, toolNameMap, trackDone, appendLog });
     streamController.handleComplete();
     return result;
   }
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -1,6 +1,7 @@
 import { FORMATS } from "../../translator/formats.js";
 import { needsTranslation } from "../../translator/index.js";
 import { ollamaBodyToOpenAI } from "../../translator/response/ollama-to-openai.js";
+import { decloakToolNames } from "../../utils/claudeCloaking.js";
 import { addBufferToUsage, filterUsageForFormat } from "../../utils/usageTracking.js";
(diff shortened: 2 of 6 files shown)


TechNickAI:
Closing in favor of #811, which rebases this fix on current master and adds context on the user-visible bug (the model refusing to call its own tools with "I can't use the tool 'exec_ide' here").

PR #756, opened 2026-04-24
Optimize request logging and usage tracking

Nelie-Taylor:
Summary

  This PR reduces blocking work on the local /v1/* request path by moving hot write operations to an in-memory buffer and flushing them to disk asynchronously every 3 seconds.

  What changed

  src/lib/usageDb.js

  - Stop writing usage.json on every request completion
  - Keep usage updates in memory first:
    - history
    - dailySummary
    - totalRequestsLifetime
  - Flush buffered usage data to disk asynchronously on a shared 3-second timer
  - Replace synchronous log.txt writes with an in-memory request log buffer
  - Flush the buffered request log asynchronously on the same timer
  - Keep only the most recent 200 log lines when persisting log.txt
  - Attempt a final flush on process shutdown

  src/lib/requestDetailsDb.js

  - Change request-detail persistence to timer-based buffering only
  - Remove inline flush behavior when the batch size threshold is reached
  - Reduce the default flush interval from 5 seconds to 3 seconds
  - Keep disk writes asynchronous and off the hot request path

  Why

  Previously, /v1/* requests could perform disk writes directly on the request path:
  - usage.json writes on each usage update
  - synchronous log.txt append/read/trim work
  - inline request-detail flushes when buffers filled up

  Under concurrent traffic, this could increase contention and make 2–3 parallel requests feel stalled or serialized.

  This change keeps request-path writes in memory and defers disk persistence to a periodic background flush, which should reduce blocking on the main request flow.

  Tradeoffs

  - Up to ~3 seconds of usage/log/detail data may be lost if the process exits unexpectedly before the next flush
  - The router favors request responsiveness over immediate on-disk durability

  Scope

  This PR intentionally keeps the change narrow:
  - no changes to provider selection
  - no changes to token refresh flow
  - no changes to routing or fallback behavior

  Test plan

  - Start the app locally and send multiple concurrent requests to /v1/chat/completions or /v1/messages
  - Confirm requests no longer block on per-request disk persistence
  - Confirm usage.json, log.txt, and request-details.json continue updating after the background flush interval
  - Run the test suite:
    - cd /tmp && npm install vitest
    - cd tests && npm test


Nelie-Taylor:
RAM from 1.2G -> 500 - 800M, no 503 status.

PR #847, opened 2026-05-03
Add new providers: AgentRouter and Bluesminds with unified LLM gateway support

ahmed775216:
added those providers because when  trying to add them as a custom provider a prefix is required, and those providers don't support added prefixes because they have prefixes included with the models and can't turn off the prefix on the provider end. so whenever a user tries to configure them, the models end up with double prefix (one is extra required by 9router when configuring the custom provider) and thus the request fails every time with 404 model not found.


ezralazuardy:
@decolua can we get this PR merged?

PR #879, opened 2026-05-04
fix(executors): parse retryAfter timestamp for precise backoff on 429

### LINKED ISSUE #878: fix(executors): parse retryAfter timestamp for precise backoff on 429
## Problem

When providers like MiniMax return a `429` error with a `retryAfter` ISO timestamp (e.g. `2025-05-02T12:00:00Z`), the base `parseError()` in `BaseExecutor` ignores it and falls back to exponential backoff (starting at 2s, doubling each time).

This means the caller waits a guessed duration instead of the **precise reset time** the provider specified.

## Root Cause

`BaseExecutor.parseError()` does not extract `retryAfter` from the response body — it only looks at `Retry-After` headers or returns a generic 429 with no `resetsAtMs`.

## Fix

Added `parseError()` override in both `open-sse/executors/default.js` and `open-sse/executors/opencode-go.js` that:

1. Parses JSON body on 429
2. Extracts `retryAfter` ISO timestamp → converts to `resetsAtMs`
3. Also handles `retry_after_ms` / `retryAfterMs` numeric milliseconds
4. Falls through to `super.parseError()` if no timestamp found

## Example

Before fix — MiniMax 429 response:
```json
{ "error": { "message": "rate limit exceeded", "retryAfter": "2025-05-02T12:00:00Z" } }
```
→ Base class returns `{ status: 429, message: ..., resetsAtMs: undefined }` → exponential backoff guessed

After fix:
→ Returns `{ status: 429, message: ..., resetsAtMs: 1746273600000 }` → caller waits exactly until reset

anuragg-saxenaa:
## Summary
- Override `parseError()` in `DefaultExecutor` and `OpenCodeGoExecutor` to extract `retryAfter` ISO timestamp from MiniMax (and similar providers) 429 responses
- Returns precise `resetsAtMs` instead of falling back to exponential backoff guessing
- Handles both `retryAfter` (ISO string) and `retry_after_ms` / `retryAfterMs` (numeric ms)

## Test plan
- [ ] Send a 429 from MiniMax with `retryAfter` field → verify `resetsAtMs` is set to the parsed timestamp
- [ ] Send a 429 without `retryAfter` → verify falls through to default exponential backoff
- [ ] Verify both executors (default.js + opencode-go.js) behave identically

Closes #878

### DIFF
--- a/open-sse/executors/default.js
+++ b/open-sse/executors/default.js
@@ -15,6 +15,36 @@ export class DefaultExecutor extends BaseExecutor {
     return injectReasoningContent({ provider: this.provider, model, body });
   }
 
+  /**
+   * Override parseError to extract precise resetsAtMs from MiniMax 429 errors.
+   * MiniMax returns: { error: { message: "...", retryAfter: "2025-05-02T12:00:00Z", ... } }
+   * Without this, the base class returns no resetsAtMs → exponential backoff instead of precise wait.
+   */
+  parseError(response, bodyText) {
+    if (response.status === 429 && bodyText) {
+      try {
+        const json = JSON.parse(bodyText);
+        const err = json?.error || json;
+        const retryAfter = err?.retryAfter;
+        if (retryAfter) {
+          const ms = new Date(retryAfter).getTime();
+          const now = Date.now();
+          if (ms > now) {
+            return { status: 429, message: err.message || bodyText, resetsAtMs: ms };
+          }
+        }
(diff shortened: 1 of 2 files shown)


anuragg-saxenaa:
👋 Gentle follow-up — this PR has been ready for ~35 days and CI is green. Happy to address any feedback or close if no longer relevant. Let me know how to move it forward.

PR #945, opened 2026-05-07
fix: free providers without connection not included in /v1/models

Vanszs:
## Problem

Free providers (e.g., OpenCode Free) with `noAuth: true` and no DB connection entry are **not included** in the `/v1/models` response when other providers have active connections. This is because `buildModelsList()` only iterates over `activeConnectionByProvider`, which only contains providers with persisted connections.

## Root Cause

In `buildModelsList()`:
- When `connections.length > 0`, only providers with DB connections are iterated
- Free providers (OpenCode, SearXNG, local-device, Coqui, Tortoise) have `noAuth: true` and never create connection entries
- They are silently excluded from the models list

## Fix

Added a loop after the connection-based iteration that processes free providers without connections using the **same data flow** as connected providers:

1. **PROVIDER_MODELS** — static registry models
2. **customModels** — user-created models from DB
3. **modelAliases** — user-assigned aliases from DB  
4. **Sub-config models** — TTS/embedding from `AI_PROVIDERS`
5. **Web search/fetch** — for providers with search/fetch config
6. **disabledModels** — respected as negative filter

This mirrors the existing pattern for connected providers (lines 222-358), ensuring consistency and scalability.

## Key Design Decision

**Suggested models** (from `modelsFetcher`) remain **UI-only**. They are not exposed via `/v1/models` until the user assigns them (clicking `+` in the dashboard, which persists a model alias). This prevents unassigned/suggested models from polluting the API response — matching the dashboard behavior where "Suggested free models" are separate from assigned models.

## Testing

- Verified `/v1/models` returns assigned models for free providers when other connections exist
- Verified model aliases, custom models, disabled models, and sub-config models all work correctly
- Backward compatible: when no connections exist, falls through to existing static model path


Vanszs:
Closing this PR. Will open a focused PR for just the free provider models fix only.

PR #992, opened 2026-05-09
Kiro fix

Oetaw:
kiro is currently migrating to another endpoint (from AWS -> Kiro.dev) so the fix is for this.
and also kiro wasnt working with CLI tools for some reason, and that is fixed here.


Oetaw:
still got to fix some things

PR #1005, opened 2026-05-10
fix: handle permission denied when creating DATA_DIR

toanalien:
- Add permission denied fallback in DATA_DIR initialization
- Falls back to default user directory (~/.9router) if configured DATA_DIR is not writable
- Apply same fallback logic to mitm/paths.js for consistency
- Logs informative warning when fallback occurs
- Added unit test to verify fallback behavior

Fixes the error: EACCES: permission denied, mkdir '/var/lib/9router/db'

### DIFF
--- a/src/lib/dataDir.js
+++ b/src/lib/dataDir.js
@@ -1,14 +1,35 @@
+import fs from "node:fs";
 import path from "path";
 import os from "os";
 
 const APP_NAME = "9router";
 
-export function getDataDir() {
-  if (process.env.DATA_DIR) return process.env.DATA_DIR;
+function getDefaultDataDir() {
   if (process.platform === "win32") {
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @toanalien for the catch and the fix! 🙏

Merged a slightly refactored version directly to `master` (commit a48fa4e) with you as the author. Changes from your PR:

- Kept the core idea: fallback to `~/.9router` on `EACCES`/`EPERM`
- Skipped `mkdir` entirely when `DATA_DIR` is unset → zero behavior change for the majority of users
- Re-throw other errors (e.g. `ENOSPC`, `ENOTDIR`) instead of silently falling back
- Avoided duplicating verbose warning logic; one short `console.warn` only when actually falling back

Closing this PR. Really appreciate the contribution!

PR #1007, opened 2026-05-10
fix: normalize Codex custom tools (apply_patch) to { input: string } schema

nstung463:
## Problem

Codex Responses API sends `apply_patch` as a **`type: "custom"`** tool with a grammar-based `format` field and **no `parameters`**:

```json
{
  "type": "custom",
  "name": "apply_patch",
  "description": "...",
  "format": { "type": "grammar", "syntax": "lark", "definition": "..." }
}
```

The translator in `open-sse/translator/request/openai-responses.js` converted every named tool to a Chat Completions function tool. Since `tool.parameters` is `undefined` for custom tools, `normalizeToolParameters()` returned `{ type: "object", properties: {} }` — an **empty schema**.

Downstream models then called `apply_patch` with `{}` instead of `{ "input": "<patch string>" }`, and the Codex runtime returned:

```
missing field `input`
```

## Fix

Add a branch in the tool-conversion map for `tool.type === "custom"`. Instead of producing an empty schema, these tools are normalized to:

```json
{
  "type": "function",
  "function": {
    "name": "apply_patch",
    "description": "...",
    "parameters": {
      "type": "object",
      "properties": { "input": { "type": "string" } },
      "required": ["input"],
      "additionalProperties": false
    }
  }
}
```

This matches what the Codex runtime expects (`input` field) and works across all downstream providers (OpenAI, Claude, CommandCode) without further translator changes.

## Tests

New file `tests/unit/openai-responses-custom-tool.test.js` (6 tests):

- `apply_patch` custom tool → function tool with `{ input: string }` schema
- Schema is never empty for custom tools
- `description` is preserved
- Any `type: "custom"` tool gets the same normalization (not just `apply_patch`)
- Normal function tools alongside custom tools are unaffected
- Custom tools with missing/empty name are filtered out (existing behavior)

## Test run

```
Test Files  1 passed (1)
     Tests  6 passed (6)
```

### DIFF
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -9,7 +9,7 @@ import { getConsistentMachineId } from "../../src/shared/utils/machineId.js";
 
 // In-memory map: hash(machineId + first assistant content) → { sessionId, lastUsed }
 const SESSION_TTL_MS = 60 * 60 * 1000; // 1 hour
-const assistantSessionMap = new Map();
+const conversationSessionMap = new Map();
 
 // Cache machine ID at module level (resolved once)
 let cachedMachineId = null;
@@ -19,7 +19,8 @@ function hashContent(text) {
   return createHash("sha256").update(text).digest("hex").slice(0, 16);
 }
 
-function generateSessionId() {
+function generateSessionId(seed = null) {
+  if (seed) return `sess_${hashContent(seed)}`;
   return `sess_${Date.now().toString(36)}_${Math.random().toString(36).slice(2, 9)}`;
 }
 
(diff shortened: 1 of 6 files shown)


nstung463:
Thanks for the review @copilot! Good catch on the missing `strict` field.

Fixed in d284e87 — added `strict: tool.strict` to the custom tool branch for parity with the normal tool conversion path. All 6 tests still pass.

nstung463:
## Fix 3: Handle `custom_tool_call` / `custom_tool_call_output` history items (infinite loop fix)

After the `custom_tool_call` response fix landed, a new bug appeared: Codex entered an **infinite apply_patch loop** even after successful executions.

### Root cause

When Codex sends a follow-up request (e.g. with conversation history), it uses `type: "custom_tool_call"` and `type: "custom_tool_call_output"` for the prior tool turn — not `function_call` / `function_call_output`. The request translator only handled the latter, so all apply_patch results were **silently dropped**. The downstream model never saw the tool output, assumed the call hadn't happened, and called apply_patch again — forever.

Confirmed via session log analysis: 20 consecutive `custom_tool_call_output` items all fell through the switch with no match.

### This commit adds two new branches to `openaiResponsesToOpenAIRequest`:

| Input item type | Translated to |
|---|---|
| `custom_tool_call` | `assistant` message with `tool_calls[]`, `input` string wrapped as `{"input":"..."}` |
| `custom_tool_call_output` | `role: "tool"` message; JSON-wrapped `{"output":"...","metadata":{}}` unwrapped to plain string |

4 new tests cover: assistant message shape, output unwrapping (success), plain-string error passthrough, and full `user → assistant → tool → user` sequence ordering.

PR #1049, opened 2026-05-12
fix: Error when update from v0.4.20 to v0.4.27 (#1048)

### LINKED ISSUE #1048: Error when update from v0.4.20 to v0.4.27
★ Web UI (Open in Browser)
  ☆ Terminal UI (Interactive CLI)
  ☆ Hide to Tray (Background)
  ☆ Exit

⚠️  Server exited (code=unknown). Restarting in 2s... (2/2)

⚠️  Server crashed 2 times. Disabling MITM and restarting...

⚠️  Server exited (code=unknown). Restarting in 1s... (1/2)

⚠️  Server exited (code=unknown). Restarting in 2s... (2/2)

moaedy:
## Summary
Root cause: The legacy JSON→SQLite migration in `importLegacyMain`/`importLegacyUsage`/`importLegacyDetails` (`src/lib/db/migrate.js`) wraps every per-row INSERT in a single transaction with no per-row guards, so a v0.4.20 `db.json` containing any malformed entry (missing required field, `null` array element, `undefined` bind) raises `SQLITE_CONSTRAINT_NOTNULL` or TypeError, aborts the whole transaction, never writes the `.migrated-from-json` marker, and crashes the server on every restart. Change: Added a `safeRow(label, fn)` helper plus an early skip-and-warn check for missing primary keys/required fields in each legacy import loop in `src/lib/db/migrate.js`, so a single bad row is logged and skipped while the rest of the migration commits normally. Closes https://github.com/decolua/9router/issues/1048

### DIFF
--- a/src/lib/db/migrate.js
+++ b/src/lib/db/migrate.js
@@ -84,100 +84,142 @@ function syncSchemaFromTables(adapter) {
   }
 }
 
+// Per-row guard: skip + log malformed legacy entries instead of aborting the
+// entire migration transaction (issue #1048 — upgrade from v0.4.20 → v0.4.27).
+function safeRow(label, fn) {
+  try { fn(); } catch (e) { console.warn(`[DB][migrate] skip ${label}: ${e.message}`); }
+}
+
 // ─── Legacy JSON import (one-time) ───────────────────────────────────────
 function importLegacyMain(adapter, data) {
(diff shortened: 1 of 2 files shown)


atonyo11:
@moaedy It seems the problem has not been resolved.
image

image

PR #1067, opened 2026-05-13
fix(docker): restore /app/server.js in standalone build (#1064)

### LINKED ISSUE #1064: Docker image broken in v0.4.31 and v0.4.33: missing /app/server.js
## Summary

The Docker image fails to start in `v0.4.31` and `v0.4.33` with `MODULE_NOT_FOUND` because `/app/server.js` does not exist in the image. The previous release `v0.4.30` works correctly, suggesting a regression introduced between `v0.4.30` and `v0.4.31`.

## Affected Versions

| Version | Status |
|---------|--------|
| `v0.4.33` (`:latest`) | ❌ Broken |
| `v0.4.31` | ❌ Broken |
| `v0.4.30` | ✅ Works |

## Environment

- Image: `ghcr.io/decolua/9router`
- Platform: `linux/amd64`
- Host OS: Ubuntu 22.04
- Docker: 27.x with Docker Compose v2

## Reproduction

```bash
docker run --rm ghcr.io/decolua/9router:0.4.33
```

## Error Output

```
node:internal/modules/cjs/loader:1386
  throw err;
  ^

Error: Cannot find module '/app/server.js'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
    at defaultResolveImpl (node:internal/modules/cjs/loader:1025:19)
    at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1030:22)
    at Function._load (node:internal/modules/cjs/loader:1192:37)
    at TracingChannel.traceSync (node:diagnostics_channel:328:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:237:24)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)
    at node:internal/main/run_main_module:36:49 {
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}

Node.js v22.22.2
```

Container enters restart loop.

## Investigation

Inspecting `/app` contents of the broken image:

```bash
$ docker run --rm --entrypoint sh ghcr.io/decolua/9router:0.4.33 -c "ls -la /app"
total 52
drwxr-xr-x    1 node     node          4096 May 12 03:00 .
drwxr-xr-x    1 root     root          4096 May 13 01:43 ..
drwxr-xr-x    1 node     node          4096 May 12 03:00 .next
drwxr-xr-x    1 node     node          4096 May 12 03:00 app
drwxr-xr-x    2 node     node          4096 May 12 03:00 data
drwxr-xr-x    2 node     node          4096 May 12 03:00 data-home
drwxr-xr-x    1 node     node          4096 May 12 03:00 node_modules
drwxr-xr-x    1 node     node          4096 May 12 02:57 open-sse
drwxr-xr-x    1 node     node          4096 May 12 03:00 proc
drwxr-xr-x    1 node     node          4096 May 11 16:52 public
drwxr-xr-x    1 node     node          4096 May 12 03:00 root
drwxr-xr-x    1 node     node          4096 May 12 03:00 src
drwxr-xr-x    1 node     node          4096 May 12 03:00 usr
```

Observations:
- `server.js` is missing at `/app/server.js`
- Unexpected system-like directories present in `/app`: `proc`, `root`, `usr`
- Both `data` and `data-home` directories exist (possibly related to recent `DATA_DIR` fallback changes in v0.4.31)

The `Cmd` in the image is `[node server.js]` with `WorkDir: /app`, but `server.js` is not where it's expected.

## Image Config

```bash
$ docker inspect ghcr.io/decolua/9router:0.4.33 --format='WorkDir: {{.Config.WorkingDir}}
Cmd: {{.Config.Cmd}}
Entrypoint: {{.Config.Entrypoint}}'

WorkDir: /app
Cmd: [node server.js]
Entrypoint: [/entrypoint.sh]
```

## Possible Cause

`v0.4.31` introduced "OIDC dashboard login + DATA_DIR fallback" — the DATA_DIR fallback changes may have altered the Dockerfile's `COPY` instructions or directory structure, causing the Next.js standalone build output (`server.js`) to be placed in a different location than where the container expects it.

The presence of `proc`, `root`, `usr` folders inside `/app` suggests a `COPY` directive may be capturing unintended paths from the build context or filesystem root.

## Workaround

Pin to `v0.4.30`:

```yaml
services:
  9router:
    image: ghcr.io/decolua/9router:0.4.30
```

## Suggestion

A potential fix would be to verify the multi-stage Docker build properly copies the standalone output:

```dockerfile
COPY --from=builder /app/.next/standalone/. ./
```

And confirm `server.js` is generated in the standalone output of `next build`.

Happy to test any proposed fix.

ridwanramadhan:
## Summary

Fixes #1064 — Docker image regression in v0.4.31+ where `/app/server.js` is missing and `/app` contains unexpected system directories (`/proc`, `/root`, `/usr`).

## Root Cause

In `next.config.mjs`, `outputFileTracingRoot` was set to `monorepoRoot`. Inside the Docker builder stage, `monorepoRoot` resolves to `/` (filesystem root), causing Next.js standalone output to:

1. Emit `server.js` at `.next/standalone/app/server.js` (nested under the project path) instead of `.next/standalone/server.js`
2. Trace and bundle files from `/usr`, `/root`, `/proc` into the standalone output

When the Dockerfile copies `.next/standalone` to `/app`, the result is:
- `/app/server.js` → ❌ missing (it's at `/app/app/server.js`)
- `/app/usr`, `/app/root`, `/app/proc` → present (host system paths)
- Container fails with `Cannot find module '/app/server.js'`

## Fix

Set `outputFileTracingRoot` back to `projectRoot` so Next.js traces only files within the project directory. This produces the expected standalone structure:

```
.next/standalone/
├── server.js          ← present at expected path
├── node_modules/
└── ...
```

## Testing

- [x] Reproduced bug locally with v0.4.31 source
- [x] Verified v0.4.30 builds correctly (regression confirmed)
- [x] Applied fix and rebuilt image — `server.js` exists at `/app/server.js`
- [x] Container starts successfully without restart loop
- [x] `/app` no longer contains `proc`, `root`, `usr` directories
- [x] Dashboard accessible on port 20128

### Before fix
```
$ docker run --rm --entrypoint sh test:broken -c "ls /app/server.js"
ls: /app/server.js: No such file or directory
```

### After fix
```
$ docker run --rm --entrypoint sh test:fixed -c "ls -la /app/server.js"
-rw-r--r-- 1 node node 12345 May 13 09:00 /app/server.js

$ docker run --rm -p 20128:20128 test:fixed
✓ Ready in 2.3s - Local: http://localhost:20128
```

## Files Changed

- `next.config.mjs` — restored `outputFileTracingRoot: projectRoot`

Fixes #1064

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -1,8 +1,7 @@
 import { fileURLToPath } from "node:url";
-import { dirname, resolve } from "node:path";
+import { dirname } from "node:path";
 
 const projectRoot = dirname(fileURLToPath(import.meta.url));
-const monorepoRoot = resolve(projectRoot, "..");
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
@@ -11,9 +10,9 @@ const nextConfig = {
   turbopack: {
     root: projectRoot
   },
-  outputFileTracingRoot: monorepoRoot,
+  outputFileTracingRoot: projectRoot,
   outputFileTracingExcludes: {
-    "*": ["./app/gitbook/**/*", "./gitbook/**/*"]
+    "*": ["./gitbook/**/*"]
   },
   images: {
     unoptimized: true


aamsur:
I have fixed this just by modified Dockerfile file from 
`CMD ["node", "server.js"]` 
to
`CMD ["node", "app/server.js"]` 
at the latest line

ridwanramadhan:
Thanks for sharing @aamsur 

Both approaches do make the container start successfully, but I think they fix the bug at different levels:

**Your fix (`CMD ["node", "app/server.js"]`)** points the entrypoint to the actual emitted location. This works, but the underlying issue — Next.js tracing files from filesystem root — remains. The image still contains:

```bash
$ docker run --rm --entrypoint sh image -c "ls /app"
.next  app  data  data-home  node_modules  open-sse  proc  public  root  src  usr
```

The presence of `/proc`, `/root`, and `/usr` inside `/app` indicates Next.js' file tracer is pulling in host filesystem paths during build. This causes:
- Increased image size from bundled system files
- Potentially sensitive paths from the builder stage being included
- Non-standard standalone output structure

**My fix (`outputFileTracingRoot: projectRoot`)** addresses the root cause by limiting Next.js' file tracing to the project directory only. This produces the expected standalone output:

```
.next/standalone/
├── server.js          ← at expected path
├── node_modules/
└── package.json
```

And the resulting `/app` is clean:
```bash
$ docker run --rm --entrypoint sh image -c "ls /app"
.next  node_modules  package.json  public  server.js
```

I'd defer to the maintainer @decolua on which approach to merge, but I'd lean toward fixing the trace root since it brings the standalone output back to Next.js' documented standard layout

PR #1085, opened 2026-05-13
fix(codex): fix streaming SSE translation and tool call failures for Codex provider

thotam:
## Summary

Fixes multiple issues with Codex (Responses API) streaming integration that caused:
1. **n8n crash**: `Cannot read properties of null (reading 'type')` when streaming is enabled
2. **Tool calls silently dropped**: LangChain fails to execute tool calls from Codex responses
3. **400 Bad Request**: Codex rejects `role=system` in `input[]` and `input_text` content type for assistant messages

### 1. `stream.js` — SSE `event:` lines dropped in translate mode

**Problem**: Codex uses standard multi-line SSE format (`event: <type>\ndata: <json>\n\n`), but the translate mode in `createSSEStream` processes lines individually via `parseSSELine()`, which only handles lines starting with `data:`. All `event:` lines were silently discarded.

**Impact**:
- When `sourceFormat === targetFormat === OPENAI_RESPONSES` (Responses API client → Codex), the output stream was missing `event:` headers entirely, breaking SSE event dispatch on the client side.
- For cross-format translation (e.g., `OPENAI_RESPONSES → OPENAI`), the event type was lost — though the data JSON contains a `type` field that the translator could still use.

**Fix**: Buffer `event:` lines into `state._lastEventType`. When both source and target formats are `OPENAI_RESPONSES`, forward the `event:` line as-is to preserve the complete SSE event structure.

### 2. `stream.js` — Flush emits `data: null` causing n8n crash

**Problem**: On stream end, `flush()` calls `translateResponse(targetFormat, sourceFormat, null, state)`. When `sourceFormat === targetFormat`, `translateResponse` returns `[null]` (same-format passthrough of the null chunk). This null item then passes through `formatSSE(null)` which emits `data: null\n\n`. n8n's LangChain parser deserializes this as JavaScript `null`, then attempts to access `null.type` → **crash**.

**Fix**: Skip null items in the flush loop before calling `formatSSE()`.

### 3. `openai-responses.js` — Missing `role: "assistant"` in first streaming delta

**Problem**: The `openaiResponsesToOpenAIResponse` translator (Codex → Chat Completions format) never included `role: "assistant"` in any delta chunk. Native OpenAI streaming always sends `role: "assistant"` in the first delta. LangChain's `ChatOpenAI` relies on this field to initialize an `AIMessage` — without it, tool calls are silently ignored even though `finish_reason: "tool_calls"` is correctly set.

**Impact**: n8n AI Agent nodes using Codex would receive `finish_reason: "tool_calls"` but never execute any tools, effectively breaking agentic workflows.

**Fix**: Track `state.firstContentSent` flag. On the first content or tool_call delta, include `role: "assistant"` (and `content: null` for tool-call-only responses) to match the native OpenAI streaming format.

### 4. `codex.js` — System messages and content type sanitization

**Problem**:
- Codex API rejects `role: "system"` in the `input[]` array (returns 400: "System messages are not allowed"). System instructions must be passed via the top-level `instructions` field.
- Codex rejects `input_text` content type for `role: "assistant"` messages (returns 400: "Invalid value: 'input_text'"). It only accepts `output_text`.

**Fix**:
- Extract all system messages from `input[]`, concatenate their text content, and prepend to the `instructions` field.
- Normalize `input_text` → `output_text` for all assistant message content blocks.
- Remove unsupported `background` parameter sent by n8n/Responses API clients.

## Files Changed

| File | Changes |
|------|---------|
| `open-sse/utils/stream.js` | Buffer SSE `event:` lines + null guard in flush |
| `open-sse/translator/response/openai-responses.js` | Add `role: "assistant"` to first delta chunk |
| `open-sse/executors/codex.js` | System message extraction + content type normalization |

## Testing

- Verified streaming with n8n AI Agent (Chat Completions format → Codex Responses API)
- Verified tool call detection and `finish_reason: "tool_calls"` translation
- Confirmed no regression for other providers (OpenRouter, Claude, Antigravity) — changes are scoped to Codex translate paths only
- Confirmed Droid CLI / Codex CLI clients are unaffected (`isDroidCLI` bypass)


thotam:
## Additional Fix: Null Flush Propagation (c198d46)

### Problem

When `sourceFormat === targetFormat` (e.g., `OPENAI_RESPONSES → OPENAI_RESPONSES`), the `translateResponse` function was returning `[null]` during flush:

```js
// Before
if (sourceFormat === targetFormat) {
  return [chunk]; // chunk = null (flush signal) → returns [null]
}
```

This `[null]` array then reached `formatSSE(null)` in the stream flush loop, which emitted `data: null\n\n` to the client. When n8n's LangChain SDK parsed this, it deserialized `"null"` as JavaScript `null`, then attempted `null.type` → **crash: `Cannot read properties of null (reading 'type')`**.

### Key Insight

The `null` is **not** from Codex — all Codex SSE events are valid JSON objects. The `null` is an **internal flush signal** passed by `stream.js` to `translateResponse()` to trigger end-of-stream finalization. In cross-format mode (e.g., `OPENAI_RESPONSES → OPENAI`), the translator uses this null to emit a final chunk with `finish_reason`. But in same-format mode, all data has already been forwarded via line-by-line passthrough — there's nothing to finalize.

### Fix

```js
// After
if (sourceFormat === targetFormat) {
  return chunk != null ? [chunk] : []; // Skip null flush — data already forwarded
}
```

A safety net `null` guard also remains in `stream.js`'s flush loop as defense-in-depth:

```js
if (item == null) continue; // Skip null flush items
```

### Verified

- Same format (`OPENAI_RESPONSES → OPENAI_RESPONSES`): flush returns `[]`, no crash ✅
- Cross format (`OPENAI_RESPONSES → OPENAI`): flush returns final chunk with `finish_reason`, unaffected ✅
- No data loss in either case — confirmed via debug logging before cleanup ✅

thotam:
@decolua 🙏

PR #1097, opened 2026-05-13
fix(docker): keep /app/server.js and bundle next at runtime (#1096)

### LINKED ISSUE #1096: Docker image broken again in v0.4.36 / v0.4.37: /app/server.js missing (regression of #1064)
## Summary

`v0.4.36` and `v0.4.37` re-introduce the same `MODULE_NOT_FOUND: '/app/server.js'` failure that was reported in #1064 and fixed by #1067. The fix was reverted in the `v0.4.36` release commit (`7ccf8c5`) under a different variable name (`monorepoRoot` → `workspaceRoot`), with the same broken behaviour.

## Affected Versions

| Version | Status |
|---------|--------|
| `v0.4.37` (`:latest`) | ❌ Broken |
| `v0.4.36` | ❌ Broken |
| `v0.4.33` | ❌ Broken (original #1064) |
| `v0.4.31` | ❌ Broken (original #1064) |
| `v0.4.30` | ✅ Works |

> Note: `v0.4.32`, `v0.4.34`, `v0.4.35` were never published, so even though #1067 was merged before `v0.4.36`, no released image ever shipped with the fix.

## Environment

- Image: `ghcr.io/decolua/9router:0.4.37` (and `:latest`, and `decolua/9router:0.4.37` on Docker Hub)
- Platform: `linux/amd64`
- Host OS: Ubuntu 22.04
- Docker: 27.x with Docker Compose v2

## Reproduction

```bash
docker pull ghcr.io/decolua/9router:0.4.37
docker run --rm ghcr.io/decolua/9router:0.4.37
```

## Error Output

```
Error: Cannot find module '/app/server.js'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
    ...
  code: 'MODULE_NOT_FOUND',
  requireStack: []
}
Node.js v22.22.2
```

Container enters a restart loop. Inspecting `/app` shows `server.js` at `/app/app/server.js` (nested) along with stray `proc`, `root`, `usr` directories — identical to the layout described in #1064.

## Root Cause

In commit `7ccf8c5` (the `v0.4.36` release), `next.config.mjs` was changed:

```diff
-const projectRoot = dirname(fileURLToPath(import.meta.url));
+const projectRoot = dirname(fileURLToPath(import.meta.url));
+// Workspace root (one level up from app/) — where npm hoists deps.
+// Next.js tracing must scan from here to find "next", "react", etc.
+const workspaceRoot = join(projectRoot, "..");
 ...
-  outputFileTracingRoot: projectRoot,
+  outputFileTracingRoot: workspaceRoot,
```

This is functionally identical to the original `monorepoRoot` regression fixed in #1067. Inside the Docker builder, `join("/app", "..")` resolves to `/`, so Next.js writes the standalone output preserving absolute paths — `server.js` lands at `.next/standalone/app/server.js`, and traced files from `/usr`, `/root`, `/proc` get mirrored alongside. After `COPY --from=builder /app/.next/standalone ./` in the runner stage, the image ends up with `/app/app/server.js` (nested) instead of `/app/server.js`, so `CMD ["node", "server.js"]` fails.

## Why the Revert Likely Happened

The `v0.4.36` changelog mentions:

> Fix Docker `Cannot find module 'next'` error in standalone build

#1067 merged but no image was published between that merge and the `v0.4.36` release commit, so the maintainer likely hit `Cannot find module 'next'` while preparing the `v0.4.36` build locally and reverted the tracing-root change to get past it. Unfortunately the revert restores #1064 instead of solving both problems together.

## Proposed Fix

Keep `outputFileTracingRoot: projectRoot` (so `server.js` lands at `/app/server.js`) **and** ensure `next` is present in the runner image — following the same pattern already used for `node-forge` in the current Dockerfile.

`next.config.mjs`:

```js
import { fileURLToPath } from "node:url";
import { dirname } from "node:path";

const projectRoot = dirname(fileURLToPath(import.meta.url));

const nextConfig = {
  output: "standalone",
  serverExternalPackages: ["better-sqlite3", "sql.js", "node:sqlite", "bun:sqlite"],
  turbopack: { root: projectRoot },
  outputFileTracingRoot: projectRoot,
  outputFileTracingExcludes: {
    "*": ["./gitbook/**/*"]
  },
  // ...
};
```

`Dockerfile` (runner stage) — add a line next to the existing `node-forge` copy:

```dockerfile
COPY --from=builder /app/.next/standalone ./
# ...existing copies...
COPY --from=builder /app/node_modules/node-forge ./node_modules/node-forge
# New: ensure `next` is available at runtime if tracing did not include it.
COPY --from=builder /app/node_modules/next ./node_modules/next
```

This is portable (no hard-coded `WORKDIR` basename in the COPY path) and addresses both failure modes in one change.

## Verification

After applying the fix locally:

```bash
docker build -t 9router-fix .
docker run --rm --entrypoint sh 9router-fix -c "ls -la /app/server.js"
# -> -rw-r--r-- 1 node node 7203 May 13 02:23 /app/server.js

docker run --rm 9router-fix
# -> server starts on port 20128, no MODULE_NOT_FOUND
```

## Workaround

Pin to `v0.4.30`, the last published image that boots correctly:

```yaml
services:
  9router:
    image: ghcr.io/decolua/9router:0.4.30
```

## References

- Original report: #1064
- Original fix: #1067 (commit `692d1fd`)
- Reverted in: commit `7ccf8c5` (`v0.4.36` release)

Happy to send the PR.

ridwanramadhan:
## Summary

Re-fixes the broken Docker image reported in #1096 (originally #1064). Two things are needed together — fixing only one trades the bug for the other, which is what has happened across the last few releases.

- Revert `outputFileTracingRoot` from `workspaceRoot` back to `projectRoot` in `next.config.mjs` so the standalone build emits `server.js` at the project root rather than nested under `app/`.
- Copy `node_modules/next` from the builder into the runner image, mirroring the existing `node-forge` pattern, so `next` resolves at runtime even if file tracing omits it.

## Background

| Release | `outputFileTracingRoot` | `/app/server.js` | `require('next')` |
|---|---|---|---|
| v0.4.30 | `projectRoot` | ✅ | ✅ |
| v0.4.31 – v0.4.33 | `monorepoRoot` | ❌ (nested at `/app/app/server.js`) | n/a (never reached) |
| #1067 merged | `projectRoot` | ✅ | ❌ ("Cannot find module 'next'") |
| v0.4.36 – v0.4.37 | `workspaceRoot` (rename of `monorepoRoot`) | ❌ (#1064 returns) | n/a |
| This PR | `projectRoot` + explicit `next` COPY | ✅ | ✅ |

## Changes

**`next.config.mjs`** — drop the `workspaceRoot` indirection:

```diff
-import { dirname, join } from "node:path";
+import { dirname } from "node:path";
 
 const projectRoot = dirname(fileURLToPath(import.meta.url));
-const workspaceRoot = join(projectRoot, "..");
...
-  turbopack: { root: workspaceRoot },
-  outputFileTracingRoot: workspaceRoot,
+  turbopack: { root: projectRoot },
+  outputFileTracingRoot: projectRoot,
```

**`Dockerfile`** — one new line in the runner stage:

```diff
 COPY --from=builder /app/node_modules/node-forge ./node_modules/node-forge
+# Ensure `next` is available at runtime in case tracing did not include it.
+COPY --from=builder /app/node_modules/next ./node_modules/next
```

## Test plan

- [x] `docker build -t 9router-fix .`
- [x] `docker run --rm --entrypoint sh 9router-fix -c "ls -la /app/server.js"` → file present (7203 bytes)
- [x] `ls /app` is clean — no `proc`/`root`/`usr` and no nested `app/` directory
- [x] `docker run --rm -p 20128:20128 9router-fix` → `▲ Next.js 16.2.6 ✓ Ready in 0ms`, DB migrations apply, `curl :20128` returns 307

## References

- Fixes #1096
- Original report: #1064
- Previous fix that was reverted in v0.4.36: #1067 (commit `692d1fd`)
- Revert commit: `7ccf8c5` (v0.4.36 release)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -34,6 +34,8 @@ COPY --from=builder /app/open-sse ./open-sse
 COPY --from=builder /app/src/mitm ./src/mitm
 # Standalone node_modules may omit deps only required by the MITM child process.
 COPY --from=builder /app/node_modules/node-forge ./node_modules/node-forge
+# Ensure `next` is available at runtime in case tracing did not include it.
+COPY --from=builder /app/node_modules/next ./node_modules/next
 
 RUN mkdir -p /app/data && chown -R node:node /app && \
   mkdir -p /app/data-home && chown node:node /app/data-home && \
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -1,19 +1,16 @@
 import { fileURLToPath } from "node:url";
-import { dirname, join } from "node:path";
+import { dirname } from "node:path";
 
 const projectRoot = dirname(fileURLToPath(import.meta.url));
-// Workspace root (one level up from app/) — where npm hoists deps.
-// Next.js tracing must scan from here to find "next", "react", etc.
-const workspaceRoot = join(projectRoot, "..");
 
 /** @type {import('next').NextConfig} */
 const nextConfig = {
   output: "standalone",
   serverExternalPackages: ["better-sqlite3", "sql.js", "node:sqlite", "bun:sqlite"],
   turbopack: {
-    root: workspaceRoot
+    root: projectRoot
   },
-  outputFileTracingRoot: workspaceRoot,
+  outputFileTracingRoot: projectRoot,
   outputFileTracingExcludes: {
     "*": ["./gitbook/**/*"]
   },


decolua:
Thanks @ridwanramadhan for the thorough investigation and clean fix! 🙏

The root-cause analysis with the regression table is super helpful — it makes the trade-off between `outputFileTracingRoot` and the missing `next` module crystal clear. Applying both fixes together (revert to `projectRoot` + explicit `COPY next`) is the right call.

Tested locally — `/app/server.js` is back, no nested `app/` directory, and `require('next')` resolves at runtime. Merging this in. 🚀

ridwanramadhan:
Just tested v0.4.39 — `/app/server.js` is at the right path, no nested `app/`, and `require('next')` resolves fine. Thanks for shipping it @decolua 

This PR is still showing Open, but I saw #1097 credited in the v0.4.38 changelog, so it looks like the fix already landed via a direct commit. Closing this on my end 🙏

PR #1104, opened 2026-05-13
fix(autostart): work on nvm + npm 9/10, actually register with launchctl (fixes #1082)

### LINKED ISSUE #1082: [Bug] Auto-start is silently broken on nvm + npm 9/10 (and never loads in current session)
## Summary

The "Enable Auto-start" option in the tray menu writes a `~/Library/LaunchAgents/com.9router.autostart.plist` file with a **non-existent script path** on nvm-based installs and modern npm. The menu then reports `✓ Auto-start Enabled` because `isAutoStartEnabled()` only checks whether the plist file exists — not whether it actually runs. On the next OS boot, launchd tries to spawn the agent and fails with `MODULE_NOT_FOUND`, so 9Router never auto-starts.

The same `enableMacOS()` function also never calls `launchctl load`, so the agent is not registered in the current session even when the path happens to be correct.

## Environment

- **OS:** macOS 26.5 (Apple Silicon)
- **Node:** v22.22.2 via nvm
- **npm:** 10.9.7
- **9router:** 0.4.33

### Bug 1 — `npm bin -g` was removed in npm 9 (lines 15, 127, 208, 282)

```js
const npmBin = execSync("npm bin -g", { encoding: "utf8", stdio: [...] }).trim();
```

On npm 9+ this command no longer exists:

```
$ npm bin -g
Unknown command: "bin"
```

Because stderr is ignored and the catch block falls through silently, the failure is invisible and the hardcoded fallback below is used. (Replacement: `npm config get prefix` then append `/bin`, or `npm prefix -g`, or `process.env.npm_config_prefix` when available.)

### Bug 2 — Hardcoded fallback is `/usr/local/...` only (lines 132, 289)

```js
} catch (e) {
  routerScript = "/usr/local/lib/node_modules/9router/cli.js";
}
```

This is wrong for the most common macOS setup today — nvm — where the package lives at `~/.nvm/versions/node/<ver>/lib/node_modules/9router/cli.js`. It is also wrong for Volta, asdf, Homebrew's node, and any user-prefix npm config. On these setups the plist gets written with a path that does not exist:

```
$ /bin/zsh -l -c "/.../node /usr/local/lib/node_modules/9router/cli.js --tray --skip-update"
Error: Cannot find module '/usr/local/lib/node_modules/9router/cli.js'
    code: 'MODULE_NOT_FOUND'
```

### Bug 3 — `enableMacOS` only calls `launchctl unload`, never `load` (line 167)

```js
fs.writeFileSync(plistPath, plistContent);

try {
  execSync(`launchctl unload "${plistPath}" 2>/dev/null`, { stdio: "ignore" });
} catch (e) {}

return true;
```

After writing the plist, the code only runs `launchctl unload` (which does nothing if it wasn't loaded). There is no corresponding `launchctl load -w "${plistPath}"`, so on macOS the agent is not registered with launchd in the current session — it only takes effect on the next user login. Combined with Bug 4, the user has no way to discover this.

### Bug 4 — `isAutoStartEnabled()` only checks file existence (lines 87–104)

```js
function isAutoStartEnabled() {
  if (platform === "darwin") {
    const plistPath = path.join(os.homedir(), "Library", "LaunchAgents", `${APP_LABEL}.plist`);
    return fs.existsSync(plistPath);
  }
  ...
}
```

So the tray menu shows `✓ Auto-start Enabled` whenever the plist file exists on disk — even when:
- The plist references a path that does not exist (Bug 2)
- The plist is not loaded into launchd (Bug 3)
- launchd has previously failed to spawn the agent

A correct check on macOS would be `launchctl print gui/$(id -u)/com.9router.autostart` or `launchctl list com.9router.autostart` and verify the result includes an alive PID or `LastExitStatus=0`.

## Full repro

1. Install on macOS via nvm: `npm i -g 9router`
2. Run `9router`, choose interactive menu → "Hide to Tray (Background)" (or right-click tray icon → "Enable Auto-start")
3. Inspect `~/Library/LaunchAgents/com.9router.autostart.plist` — `ProgramArguments` references `/usr/local/lib/node_modules/9router/cli.js`
4. `launchctl load -w ~/Library/LaunchAgents/com.9router.autostart.plist` then `launchctl list | grep autostart` — `LastExitStatus` is non-zero
5. Reboot — no 9router running, no tray icon, no error visible to user

## Working plist for reference

```xml
<key>ProgramArguments</key>
<array>
    <string>/Users/<user>/.nvm/versions/node/<ver>/bin/node</string>
    <string>/Users/<user>/.nvm/versions/node/<ver>/lib/node_modules/9router/cli.js</string>
    <string>--tray</string>
    <string>--skip-update</string>
</array>
<key>EnvironmentVariables</key>
<dict>
    <key>PATH</key>
    <string>/Users/<user>/.nvm/versions/node/<ver>/bin:/usr/local/bin:/usr/bin:/bin</string>
</dict>
```

## Suggested fix

1. **Stop relying on `npm bin -g`.** Resolve the bin path from `require.resolve("9router/package.json")` (most direct) or `process.execPath` + relative traversal. Inside the running CLI you already have `__filename`/`process.argv[1]` — `enableAutoStart(cliPath)` is called with `__filename` from `cli.js:691`, so prefer that path and skip npm entirely.
2. **Drop the `/usr/local/...` fallback.** If the path can't be resolved, return `false` instead of writing a broken plist.
3. **Call `launchctl load -w "${plistPath}"`** after writing on macOS so the agent is active in the current session.
4. **Strengthen `isAutoStartEnabled()`** to check launchctl state on macOS, not just file existence. Reflect the real state in the tray menu so the user can see when auto-start is misconfigured.
5. Optionally: set `EnvironmentVariables.PATH` in the plist explicitly with `path.dirname(process.execPath)` so the agent works under launchd's minimal env without depending on `zsh -l` sourcing.

Happy to send a PR if any of these directions are preferred.

ntdung6868:
## Summary

Fixes #1082 — auto-start on macOS is silently broken on every nvm / Volta / asdf / Homebrew install and every npm ≥ 9. The plist gets written referencing a non-existent script, `launchctl load` is never called, and `isAutoStartEnabled()` lies because it only checks file existence. None of these failures surface to the user.

## Before (problem)

```
$ npm bin -g
Unknown command: "bin"            # removed in npm 9
```

So autostart.js fell through to a hardcoded fallback:

```js
} catch (e) {
  routerScript = "/usr/local/lib/node_modules/9router/cli.js";   // line 132
}
```

That path doesn't exist on nvm. The resulting plist:

```
$ launchctl load -w ~/Library/LaunchAgents/com.9router.autostart.plist
$ tail /tmp/9router.error.log
Error: Cannot find module '/usr/local/lib/node_modules/9router/cli.js'
    code: 'MODULE_NOT_FOUND'
```

But the tray menu happily shows `✓ Auto-start Enabled` because `isAutoStartEnabled()` only does `fs.existsSync(plistPath)`. And `enableMacOS()` itself only called `launchctl unload` (line 167) — never `launchctl load` — so even when the path was correct the agent wasn't registered with launchd in the current session.

## What this PR changes

**`getCliJsPath(cliPath)` helper.** Resolves cli.js absolutely without depending on the npm CLI. Tries: (1) explicit `cliPath` argument, (2) `process.argv[1]` if it's our cli.js, (3) `path.resolve(__dirname, '..', '..', '..', 'cli.js')` relative to autostart.js's own location. Returns `null` if no candidate exists — callers refuse to write a broken plist.

**`enableMacOS` now calls `launchctl load -w` after writing.** Defensive `unload` first to handle re-enable. So enable takes effect in the current session, not only on next login.

**`isAutoStartEnabled()` on macOS checks launchd state.** Runs `launchctl list ${APP_LABEL}` in addition to checking the plist file. The menu now reflects the real state.

**Plist now invokes node + cli.js directly with absolute paths.** The previous design wrapped the command in `zsh -l -c "..."`, relying on the user's login shell to source nvm/PATH — fragile. Direct invocation is launchd-context-safe. `EnvironmentVariables.PATH` explicitly includes node's bin dir + standard system paths so child processes spawned by cli.js (the runtime `npm install` for better-sqlite3, etc.) still resolve `npm`.

**Windows and Linux paths get the same `getCliJsPath()` treatment** — both had the same `npm bin -g` + hardcoded-fallback bug. Windows VBS branch simplified to always invoke `node cli.js` (dropped the `9router.cmd` lookup that depended on npm prefix).

## Test plan (verified on macOS 26.5, nvm v22.22.2, npm 10.9.7)

| Scenario | Expected | Actual |
|---|---|---|
| `enableAutoStart()` no args | Returns true, plist has nvm node + cli.js absolute paths, `launchctl list` shows label, agent runs immediately, HTTP 200 | ✓ |
| `enableAutoStart(__filename)` explicit path | Same as above, plist uses passed path | ✓ |
| `isAutoStartEnabled()` after enable | true | ✓ |
| `disableAutoStart()` | Returns true, plist removed, `launchctl list` empty | ✓ |
| `isAutoStartEnabled()` after disable | false | ✓ |
| **Bug 4 regression:** create plist file manually, don't load into launchctl, then call `isAutoStartEnabled()` | false (old impl would have returned true) | ✓ |

End-to-end verification: after `enableAutoStart()` returned, the next-server process was running on :20128, the systray2 tray binary was alive, `launchctl list com.9router.autostart` reported exit status 0, and `curl http://localhost:20128/login` returned HTTP 200 — all without any further user action.

## Compatibility

- Public API unchanged: still exports `enableAutoStart`, `disableAutoStart`, `isAutoStartEnabled`.
- Net diff: -171 / +111 lines; the rewrite is shorter than the original.
- Windows path simplified; the previous `9router.cmd` lookup path is replaced by a direct `node cli.js` invocation, which works regardless of npm prefix layout.
- Linux: unchanged structurally — same `.desktop` file format, same Exec line, just with reliable path resolution. Not retested on Linux personally; please confirm.

### DIFF
--- a/cli/src/cli/tray/autostart.js
+++ b/cli/src/cli/tray/autostart.js
@@ -7,22 +7,36 @@ const APP_NAME = "9router";
 const APP_LABEL = "com.9router.autostart";
 
 /**
- * Get the command to run 9router in tray mode
+ * Resolve the absolute path to this package's cli.js.
+ *
+ * Order of preference:
+ *   1. Explicit `cliPath` argument — cleanest, used when called from running
+ *      cli.js with `__filename`.
+ *   2. `process.argv[1]` if it's our cli.js — true when 9router is currently
+ *      running and the tray menu fires this code path.
+ *   3. Compute relative to this file's own location. autostart.js lives at
+ *      `<pkg>/src/cli/tray/autostart.js`, so cli.js is three levels up.
+ *      This works for any global install layout (nvm, Volta, asdf, Homebrew,
+ *      /usr/local, etc.) without depending on `npm bin -g` (removed in npm 9)
+ *      or a hardcoded `/usr/local/...` path.
+ *
+ * Returns null if no candidate exists — callers should not write an autostart
+ * entry pointing at a non-existent script.
  */
-function getStartCommand() {
-  // Find the global npm bin path for 9router
-  try {
-    const npmBin = execSync("npm bin -g", { encoding: "utf8", stdio: ["ignore", "pipe", "ignore"] }).trim();
-    const routerPath = path.join(npmBin, "9router");
-    if (fs.existsSync(routerPath)) {
-      return `"${routerPath}" --tray --skip-update`;
-    }
-  } catch (e) {
-    // npm not available or failed
+function getCliJsPath(cliPath) {
+  if (cliPath) {
+    const resolved = path.resolve(cliPath);
+    if (fs.existsSync(resolved)) return resolved;
   }
-  
-  // Fallback: use npx
-  return "npx 9router --tray --skip-update";
+  if (process.argv[1]) {
+    const resolved = path.resolve(process.argv[1]);
+    if (path.basename(resolved) === "cli.js" && fs.existsSync(resolved)) {
+      return resolved;
+    }
+  }
+  const computed = path.resolve(__dirname, "..", "..", "..", "cli.js");
+  if (fs.existsSync(computed)) return computed;
+  return null;
 }
 
 /**
@@ -32,29 +46,17 @@ function getStartCommand() {
  */
 function enableAutoStart(cliPath) {
   const platform = process.platform;
-  
-  // Skip on unsupported platforms
-  if (!["darwin", "win32", "linux"].includes(platform)) {
-    return false;
-  }
-  
-  // Skip on Linux without GUI
-  if (platform === "linux" && !process.env.DISPLAY) {
-    return false;
-  }
-  
+
+  if (!["darwin", "win32", "linux"].includes(platform)) return false;
+  if (platform === "linux" && !process.env.DISPLAY) return false;
+
   try {
-    if (platform === "darwin") {
-      return enableMacOS(cliPath);
-    } else if (platform === "win32") {
-      return enableWindows(cliPath);
-    } else if (platform === "linux") {
-      return enableLinux(cliPath);
-    }
+    if (platform === "darwin") return enableMacOS(cliPath);
+    if (platform === "win32") return enableWindows(cliPath);
+    if (platform === "linux") return enableLinux(cliPath);
   } catch (err) {
-    // Silent fail - autostart is optional
+    // Silent fail — autostart is optional
   }
-  
   return false;
 }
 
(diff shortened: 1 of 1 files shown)


ntdung6868:
Pushed an additional commit (`ced9510`) — the previous version of this PR had a subtle UX regression I missed.

**Issue:** when 9router is already running because it was launched by the autostart agent (i.e. previous session enabled autostart, user rebooted), clicking "Disable Auto-start" from the tray menu would silently kill the tray icon. The reason: `launchctl unload <plist>` for an Aqua user-domain agent sends SIGTERM to the running process. Since the running cli.js *is* the agent's process, the unload kills the very thing executing the click handler — the menu label never gets a chance to flip back.

**Fix:** small `isAgentSelfMacOS()` probe — parse `launchctl list ${LABEL}` and compare `"PID"` against `process.pid`. When we're the agent, the macOS enable/disable paths skip the unload/load and only update the plist file on disk. The plist on disk is what launchd reads at next login, so the autostart state still toggles correctly across sessions. External callers (e.g. enable from a non-autostart-launched process) get the full unload/load behaviour they need.

Verified end-to-end on the same setup (macOS 26.5, nvm v22.22.2, npm 10.9.7):

1. boot launchd agent → cli.js PID matches `launchctl list com.9router.autostart` PID
2. click tray → "✓ Auto-start Enabled" → "Enable Auto-start" — tray stays alive, plist removed, `isAutoStartEnabled()` flips to false
3. click again → tray stays alive, plist re-created, `isAutoStartEnabled()` flips back to true
4. `LastExitStatus=0` after every cycle

No behaviour change for callers that aren't running as the agent (e.g. first-time enable from a manually-launched 9router) — those still get the full `launchctl load -w` so the agent is registered immediately.

PR #1121, opened 2026-05-14
fix(update): restore automatic Update Now relaunch

### LINKED ISSUE #1120: Update Now falls back to Copy & Shutdown instead of auto-relaunch
## Summary

In v0.4.41 the dashboard's **Update now** flow opens the manual **Copy & Shutdown** panel instead of starting the existing detached updater/relaunch flow.

The backend updater still exists and works through `/api/version/update` + `src/lib/updater/updater.js`, including `UPDATER_RELAUNCH`, but the sidebar UI no longer calls that endpoint from the Update Now button.

## Reproduction

1. Run 9Router from the npm package.
2. Downgrade to an older published version, e.g. `npm i -g 9router@0.4.39 --prefer-online`.
3. Open the dashboard and click **Update now**.
4. Observe that the UI shows **Copy & Shutdown** and instructs the user to paste `npm i -g 9router@latest --prefer-online` into a terminal, then manually run `9router` again.

## Expected behavior

Clicking **Update now** should start the existing automatic updater:

- POST `/api/version/update`
- run the detached updater process
- show install progress from the updater status server
- relaunch 9Router after npm install succeeds
- only fall back to manual install if the automatic updater cannot start or fails

## Actual behavior

The UI immediately switches to the manual copy/shutdown flow. Running the copied npm command updates the package but does not relaunch 9Router, so users must start it manually.

## Notes

The regression appears frontend-only: the backend update route and relaunch machinery are still present in v0.4.41.

ntdung6868:
## Summary
- Restores the dashboard **Update Now** path to call `/api/version/update` so the existing detached updater installs and relaunches 9Router automatically.
- Adds an update progress overlay that polls the updater status server and shows install logs/phases.
- Keeps the manual **Copy & Shutdown** flow as a fallback when the automatic updater cannot start or fails.

Fixes #1120

## Test plan
- [x] `cd /Users/dungdev/Documents/9router-fork && npx eslint src/shared/components/Sidebar.js src/app/api/version/update/route.js src/lib/appUpdater.js`
- [x] `npm --prefix /Users/dungdev/Documents/9router-fork run build`
- [x] Started the production build on port 20130 and verified `/dashboard` returns HTTP 200.
- [x] Verified the built client bundle contains `/api/version/update`, the automatic update confirmation copy, and the manual fallback text.
- [x] Confirmed the current v0.4.41 npm package reproduces the regression: Update Now opens Copy & Shutdown while the backend updater route still exists.

Note: `npm --prefix /Users/dungdev/Documents/9router-fork/cli run build` reaches the Next.js build successfully but fails in the existing packaging script because this checkout emits `.next/standalone/9router-fork` while the script expects `.next/standalone/app`; that packaging-layout mismatch is unrelated to this Sidebar change.

### DIFF
--- a/src/shared/components/Sidebar.js
+++ b/src/shared/components/Sidebar.js
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState, useEffect } from "react";
+import { useState, useEffect, useRef } from "react";
 import PropTypes from "prop-types";
 import Link from "next/link";
 import { usePathname } from "next/navigation";
@@ -46,8 +46,12 @@ export default function Sidebar({ onClose }) {
   const [updateInfo, setUpdateInfo] = useState(null);
   const [showUpdateModal, setShowUpdateModal] = useState(false);
   const [isUpdating, setIsUpdating] = useState(false);
+  const [updateStatus, setUpdateStatus] = useState(null);
+  const [updateError, setUpdateError] = useState(null);
+  const [showManualUpdate, setShowManualUpdate] = useState(false);
   const [shutdownCountdown, setShutdownCountdown] = useState(0);
   const [enableTranslator, setEnableTranslator] = useState(false);
+  const updatePollTimerRef = useRef(null);
   const { copied, copy } = useCopyToClipboard(2000);
 
   const INSTALL_CMD = UPDATER_CONFIG.installCmdLatest;
@@ -67,6 +71,10 @@ export default function Sidebar({ onClose }) {
       .catch(() => {});
   }, []);
 
+  useEffect(() => () => {
+    if (updatePollTimerRef.current) clearTimeout(updatePollTimerRef.current);
+  }, []);
+
   const isActive = (href) => {
     if (href === "/dashboard/endpoint") {
       return pathname === "/dashboard" || pathname.startsWith("/dashboard/endpoint");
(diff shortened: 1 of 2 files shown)


ntdung6868:
### Local end-to-end verification

Tested on macOS against the fork's standalone build (port `20128`, `TRAY_MODE=1`).

**Happy path** — `UPDATER_SCRIPT_PATH` pointed at a mock updater that exposes the status server and honors `UPDATER_RELAUNCH=1` but skips the real `npm i -g` for safety:
- `POST /api/version/update` → `{success: true}` in 151ms
- Fork exited within 500ms (`exitDelayMs`)
- Status server came up on `127.0.0.1:20129` with the expected JSON shape (`phase`, `done`, `success`, `error`, `logTail`)
- Phases progressed `starting` → `waitingForExit` → `installing` → `done(success=true)`
- Relaunch fired `npx 9router --tray --skip-update` (preserved `--tray` from `TRAY_MODE=1`); tray was back on :20128 within ~1s

**Failure path** — broken `UPDATER_SCRIPT_PATH` that never binds port `20129`; ran a node script replicating the new `pollUpdateStatus()` against the unreachable URL:
- 30 polls at ~1s interval, every poll returned `ECONNREFUSED`
- Fallback triggered at **t=30182ms** (matches `statusPollNoContactTimeoutMs=30000`)
- Frontend code path: \`setUpdateError("Automatic updater did not respond...")\` + \`setShowManualUpdate(true)\` → ManualUpdatePanel with error banner
- Tray restored via the relaunch hook even though the status server failed

**Not exercised live**: the real \`npm i -g 9router@latest\` step (skipped to avoid mutating the global install). The detached updater install path is unchanged from prior versions and is the same code the existing Copy & Shutdown flow relies on; only the dashboard wiring (\`Sidebar.js\`) is new in this PR.

PR #1140, opened 2026-05-15
Support PostgreSQL/MySQL databases

hyoretsu:
Surprised that no one has suggested this. It's essential for my use case, because having a remote deploy makes conversation round trips take twice the time, but running it locally wasn't doable on multiple PC's because of database syncronization problems. It only uses Postgres/MySQL if a DATABASE_URL is detected, otherwise it keeps using SQLite.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -6,6 +6,12 @@ JWT_SECRET=change-me-to-a-long-random-secret
 INITIAL_PASSWORD=change-me
 DATA_DIR=/var/lib/9router
 
+# Remote database (optional — defaults to local SQLite under DATA_DIR)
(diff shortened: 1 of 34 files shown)


ricatix:
This would be a great feature. I run 9Router in Docker and having PostgreSQL support would make multi-machine deployments much more practical — especially for people running 9Router across multiple VPS nodes who want centralized usage tracking and shared provider configurations without syncing SQLite files.

The approach of keeping SQLite as default and only switching to PG/MySQL when `DATABASE_URL` is set is the right one — backward compatible and opt-in.

Really hope this gets merged. 🙏

PR #1144, opened 2026-05-15
fix: sanitize Read tool args to prevent retry loops from non-Anthropic models

GodrezJr2:
## Problem

When using non-Anthropic models (GPT-5.5, DeepSeek, etc.) via 9Router as a Claude Code CLI proxy, these models sometimes emit extremely large values for the Read tool limit param (e.g. 25999999999999999). Claude Code rejects bad params and retries — wasting tokens and causing noisy output.

## Fix

Buffer tool call arguments instead of streaming partial_json immediately. At finish_reason, sanitize Read tool args — clamp limit to max 2000. All other tools unchanged.

## Tested

Verified with cx/gpt-5.5-xhigh and ocg/deepseek-v4-pro. No more retry loops after this fix.

### DIFF
--- a/open-sse/translator/response/openai-to-claude.js
+++ b/open-sse/translator/response/openai-to-claude.js
@@ -11,14 +11,33 @@ function sanitizeToolArgs(toolName, argsJson) {
     const name = toolName.startsWith(CLAUDE_OAUTH_TOOL_PREFIX)
       ? toolName.slice(CLAUDE_OAUTH_TOOL_PREFIX.length)
       : toolName;
-    if (name === "Read") {
-      if (typeof args.limit === "number" && args.limit > 2000) args.limit = 2000;
-      if (typeof args.offset === "number" && args.offset < 0) args.offset = 0;
-    }
+    if (name === "Read") sanitizeReadArgs(args);
     return JSON.stringify(args);
   } catch {
     return argsJson;
   }
+}
+
+function sanitizeReadArgs(args) {
+  if (typeof args.limit === "string" && /^\d+$/.test(args.limit)) args.limit = Number(args.limit);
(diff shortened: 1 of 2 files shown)


GodrezJr2:
Hi @decolua — friendly bump on this PR.

I noticed #1354 was merged today, which targets the same area but only handles the `pages: ""` case. This PR (#1144) opened May 15 covers the broader set of malformed args we've observed from non-Anthropic models in production:

| Case                              | #1354 | #1144 |
|-----------------------------------|-------|-------|
| `pages: ""` (non-PDF)             | ✅    | ✅    |
| `pages: "1-3"` (PDF preserved)    | ✅    | ✅    |
| `limit: 25999999999999999` overflow | ❌  | ✅ clamped to 2000 |
| `limit: "100"` numeric string     | ❌    | ✅ coerced |
| `offset: -5` negative             | ❌    | ✅ clamped to 0 |
| `offset: "0"` numeric string      | ❌    | ✅ coerced |
| `proxy_*` tool name prefix        | ❌    | ✅ stripped |
| Streaming partial chunks          | ⚠️ skip | ✅ buffer-then-sanitize |

**Approach difference:** #1354 sanitizes per chunk via `JSON.parse` (fails silently on fragmented streams — most providers chunk tool args). #1144 buffers args until `finish_reason`, then sanitizes once — reliable regardless of how upstream chunks the JSON.

**Real example** from `cx/gpt-5.5-xhigh` and `ocg/deepseek-v4-pro` that broke Read entirely (caused 25M-token retry loops):
```json
{ "file_path": "/x.ts", "limit": 25999999999999999 }
```

Just rebased onto current master (commits `70c2e2b`, `2517815`). Tests pass, no conflict. Happy to address review comments — would you take another look?

Issue #1142, opened 2026-05-15
The 9router v0.4.41-46 was auto killed after startup, but not OOM-killer


traderlife8:
========================================
  Choose Interface (v0.4.41)
   Server: http://localhost:20128
========================================

 ★ Web UI (Open in Browser)
  ☆ Terminal UI (Interactive CLI)
  ☆ Hide to Tray (Background)
  ☆ Exit

Press Enter to go back to menu...已杀死

traderlife8:
~$ 9router
[9router][runtime] npm install better-sqlite3@12.6.2 --no-audit --no-fund --prefer-online --no-save  (cwd: /home/zh/.9router/runtime)
已杀死

indrapalijama:
revert to v0.4.39

zvwgvx:
i have the same problem #1116

decolua:
Please try the latest version:

```bash
npm i -g 9router@latest --prefer-online
```

`已杀死` = OOM-killer. If still happens on `0.4.45`, share `free -h` and `dmesg | tail`.

zvwgvx:
Confirmed, the issue is fixed in 0.4.45. Thanks!

traderlife8:
v0.4.45 still has this problem. my system is Ubuntu 24.04.4 LTS, how to share free -h and dmesg | tail ?

nautonk:
> v0.4.45 still has this problem. my system is Ubuntu 24.04.4 LTS, how to share free -h and dmesg | tail ?

Me too. Works good on v0.4.39 so revert back to v0.4.39.

free -h output:

ubuntu@instance-20220624-0828:/root$ free -h
total        used        free      shared  buff/cache   available
Mem:           23Gi       1.8Gi        17Gi       218Mi       4.2Gi        21Gi
Swap:         999Mi          0B       999Mi

dmesg | tail output:

ubuntu@instance-20220624-0828:/root$ sudo dmesg | tail
[1106625.625950] docker0: port 1(veth67f3444) entered disabled state
[1106625.627350] device veth67f3444 left promiscuous mode
[1106625.627358] docker0: port 1(veth67f3444) entered disabled state
[1106685.619415] docker0: port 1(vetha10fe0e) entered blocking state
[1106685.619422] docker0: port 1(vetha10fe0e) entered disabled state
[1106685.619520] device vetha10fe0e entered promiscuous mode
[1106685.627955] eth0: renamed from vethde96cbe
[...]

PR #1153, opened 2026-05-15
fixed: automatically killed after running

traderlife8:
Use this fixed cli.js , and bash run: '9router --skip-update' is OK.
Another issue that cause the 9router process to be killed maybe in the 'update' process.
Both of these issues need to be fixed in order to completely fix the problem of the '9router' process being killed after startup.
 https://github.com/decolua/9router/issues/1142


traderlife8:
This issue was fixed in v0.4.55

Issue #1078, opened 2026-05-13
Missing vision/context metadata for chat models causes clients to treat multimodal-capable models as text-only


lector2002:
```markdown
Title: Missing vision/context metadata for chat models causes clients to treat multimodal-capable models as text-only

## Summary

When using 9Router as an OpenAI-compatible endpoint, multimodal-capable chat models such as `gpt-5.5` and Xiaomi `mimo-v2.5` are exposed through `/v1/models` without any capability/context metadata. Some clients appear to rely on `/v1/models` metadata to decide whether image input is supported, so they treat these models as text-only and do not send image payloads.

As a result, image understanding works when calling the provider directly, but fails or is unavailable when the same model is routed through 9Router.

## Environment

- 9Router provider path: OpenAI-compatible `/v1/chat/completions`
- Tested providers/models:
  - `cx/gpt-5.5`
  - Xiaomi MiMo token plan Singapore endpoint:
    - `https://token-plan-sgp.xiaomimimo.com/v1/chat/completions`
    - model: `mimo-v2.5`
- Direct provider API call with image works
- Through 9Router, image handling is not available or not processed

## Current Behavior

`GET /v1/models` returns models in a minimal OpenAI-style format:

```json
{
  "id": "cx/gpt-5.5",
  "object": "model",
  "owned_by": "cx"
}
```

For Xiaomi MiMo:

```json
{
  "id": "mimo/mimo-v2.5",
  "object": "model",
  "owned_by": "mimo"
}
```

There is no metadata indicating:

```json
{
  "capabilities": ["vision"],
  "context_window": 262144
}
```

or similar fields.

Because of this, clients that check model capabilities before sending image input may assume these models are text-only. This makes image input unavailable even though the upstream model supports vision.

## Expected Behavior

9Router should expose model capability metadata for chat models that support image input and known context sizes.

Example:

```json
{
  "id": "mimo/mimo-v2.5",
  "object": "model",
  "owned_by": "mimo",
  "context_window": 262144,
  "capabilities": ["vision"]
}
```

and:

```json
{
  "id": "cx/gpt-5.5",
  "object": "model",
  "owned_by": "cx",
  "capabilities": ["vision"]
}
```

This would allow clients to correctly detect that these models support image input.

## Why This Matters

Many clients do not blindly send image blocks to every model. They first inspect model metadata or provider capability flags. If 9Router exposes all chat models as generic text-only LLMs, multimodal-capable models lose image functionality when accessed through 9Router.

[...]

Issue #1026, opened 2026-05-11
KiroExecutor drops image content — translator never invoked


codename-zen:
# KiroExecutor drops image content — translator never invoked

## Summary

Requests to any `kr/*` model with image content (OpenAI `image_url` or Anthropic `type:image`) reach AWS CodeWhisperer with the image parts intact in OpenAI shape, which CodeWhisperer doesn't understand, so the model replies "I don't see any image." This affects every Kiro-proxied model, including ones officially listed in `providerModels.js` (e.g. `claude-sonnet-4.5`, `claude-haiku-4.5`) and custom passthrough models like `claude-opus-4.7`.

The translator that knows how to convert OpenAI/Claude image content into Kiro's `userInputMessage.images[]` format already exists in the repo — it's just never called.

## Reproduction

9router v0.4.29, Kiro provider with valid OAuth connection and proxy pool.

```bash
# Prepare a small PNG as base64
IMG=$(base64 -w0 < some-image.png)

# Test via OpenAI /v1/chat/completions
curl -s http://localhost:20128/v1/chat/completions \
  -H "Authorization: Bearer $NINEROUTER_KEY" \
  -H "Content-Type: application/json" \
  -d "{
    \"model\": \"kr/claude-sonnet-4.5\",
    \"max_tokens\": 200,
    \"stream\": false,
    \"messages\": [{
      \"role\": \"user\",
      \"content\": [
        {\"type\": \"text\", \"text\": \"Describe this image.\"},
        {\"type\": \"image_url\", \"image_url\": {\"url\": \"data:image/png;base64,$IMG\"}}
      ]
    }]
  }"
```

### Expected
Model describes the image. Claude Sonnet 4.5 supports vision natively on Kiro (see https://kiro.dev/docs/models/).

### Actual
```
"I don't see any image attached to your message..."
```

Same happens with `kr/claude-haiku-4.5`, `kr/claude-opus-4.6`, `kr/claude-opus-4.7`.

Control: identical payload to `cc/claude-opus-4-7` or bare `claude-opus-4.7` → model describes the image correctly. Same model family, same base64, same endpoint — only the route prefix differs. So this is not a client issue or a base64/encoding problem.

## Root cause

`open-sse/executors/kiro.js`:

```js
transformRequest(model, body, stream, credentials) {
  return body;  // passthrough — no translation
}
```

[...]

PR #1166, opened 2026-05-15
feat(kiro): parity with CLIProxyAPIPlus — thinking, agentic, live models, and vision

### LINKED ISSUE #874: [Feature Request] Add vision/image support for Kiro provider (kr/ models)
## Description
Currently, when using Kiro provider models (e.g. `kr/claude-sonnet-4.5`) through 9Router, image/vision content in messages is stripped or ignored by the Kiro Translator/Executor. Claude Sonnet 4.5 natively supports vision, but this capability is lost when routing through 9Router's Kiro provider.

## Current Behavior
- Sending a message with `image_url` content type via `kr/claude-sonnet-4.5`
- Kiro Translator strips the image content blocks
- Only text content is forwarded to Kiro API
- Vision tasks fail or return text-only responses

## Expected Behavior
- Image content blocks should be preserved during format translation
- Kiro Executor should forward images in the correct format (Anthropic Messages API format with `type: "image"` + `source: { type: "base64", ... }`)
- RTK should not strip image content blocks during compression

## Suggested Changes
1. **Kiro Translator** (`src/lib/translators/kiro.js`): Handle `image_url` and `image` content types during OpenAI → Kiro format conversion
2. **Kiro Executor** (`src/lib/executors/kiro.js`): Include image content in upstream request body
3. **RTK**: Skip compression for image content blocks

## Use Case
Many developers use Kiro as a free provider through 9Router. Adding vision support would enable:
- Screenshot analysis for UI debugging
- Reading mockups/design files
- OCR on documents
- Architecture diagram analysis

## Environment
- 9Router version: v0.4.17
- Provider: Kiro (kr/)
- Model: kr/claude-sonnet-4.5

## Reference
OmniRoute (TypeScript fork of 9Router) already has a "Vision Bridge" feature that handles multimodal content forwarding. This could be used as a reference implementation.

NakHalal:
## Summary

Brings 9router's Kiro provider up to parity with CLIProxyAPIPlus on four fronts that were previously missing or broken:

1. **Reasoning / thinking** is now properly detected, enabled upstream, and surfaced to clients as `delta.reasoning_content` instead of being wrapped as inline `<thinking>` text.
2. **Agentic variants** (`-agentic`, `-thinking`, `-thinking-agentic`) are synthesised on top of every Kiro model, with the `-agentic` variant injecting a chunked-write protocol prompt that prevents Kiro's 2–3 minute server timeout on large file writes.
3. **Live model resolution** queries Kiro's `ListAvailableModels` per-account and exposes the real catalog (plus all variants) through `/api/providers/[id]/models` and `/v1/models`. The dashboard can now fetch Kiro models, and `/v1/models` reflects what each connected account actually has access to.
4. **Vision / images** are now forwarded to Kiro instead of being silently stripped. OpenAI `image_url` data URLs and Claude-native `image` blocks are translated into Kiro's `userInputMessage.images[].{format, source.bytes}` shape on every user turn (current message + history).

## How `-agentic` and `-thinking` work

Both suffixes are pure 9router fictions — Kiro upstream does not advertise these IDs. The translator strips them before the request leaves this process and treats them as flags:

- `-thinking` → injects `<thinking_mode>enabled</thinking_mode><max_thinking_length>16000</max_thinking_length>` into the system prompt so Kiro emits a `reasoningContentEvent`.
- `-agentic` → prepends the chunked-write system prompt (max 350 lines per write, prefer surgical edits, etc.) to keep large file operations under Kiro's server timeout.
- `-thinking-agentic` → both at once.

The detection layer is broader than the suffix alone:
- `Anthropic-Beta: interleaved-thinking` header
- Claude `thinking: { type: "enabled" }`
- OpenAI `reasoning_effort` ∈ {low, medium, high, auto}
- AMP/Cursor `<thinking_mode>enabled</thinking_mode>` magic tag in messages
- Model name containing `thinking` or `-reason`

Any one is enough to flip the thinking flag, so plain model IDs still work for clients that signal reasoning intent through the request body or headers.

## How vision works

Kiro accepts a top-level `images: [{ format, source: { bytes } }]` array on the user message. Vision-capable models (Claude Sonnet/Opus/Haiku 4.x, GPT-4-class) consume them; non-vision models ignore the field. Behaviour mirrors CLIProxyAPIPlus:

- OpenAI `image_url` blocks with `data:image/<fmt>;base64,...` URLs → forwarded.
- Claude-native `image` blocks (defensive fallback; usually pre-translated by `claude-to-openai`) → forwarded.
- Remote `http(s)` URLs → silently dropped (no synchronous fetch in the request hot path, matching upstream behaviour).
- Multi-turn history images survive intact when the client re-sends them.

## Why other proxies reject `-agentic` IDs

A passthrough proxy that validates model IDs against Kiro's actual `ListAvailableModels` response will reject `kiro-claude-sonnet-4.5-agentic` as "invalid model" because that ID never appears in the upstream catalog. The synthetic-variant trick only works when the proxy itself knows to (1) advertise the variant IDs, (2) strip them before forwarding, (3) inject the prompt. This PR adds all three to 9router.

### `feat(kiro): add thinking and agentic variants` (033bc80)
- `open-sse/config/kiroConstants.js` (new): `resolveKiroModel()`, `isThinkingEnabled()`, `buildThinkingSystemPrefix()`, `KIRO_AGENTIC_SYSTEM_PROMPT`
- `open-sse/translator/request/openai-to-kiro.js`: strips `-agentic` / `-thinking` suffix, prepends thinking-mode tag and/or chunked-write prompt
- `open-sse/translator/response/kiro-to-openai.js`: emits `delta.reasoning_content` instead of wrapping reasoning in `<thinking>` text
- `open-sse/executors/kiro.js`: handles `reasoningContentEvent` in the AWS Event Stream parser
- `open-sse/config/providerModels.js`: registers the new variant IDs in the static catalog

### `feat(kiro): add live model catalog fetcher` (a1fdac5)
- `open-sse/services/kiroModels.js` (new): `resolveKiroModels()` calls `https://q.{region}.amazonaws.com/ListAvailableModels` with the proper Kiro IDE fingerprint headers, expands each upstream model into 4 variants (skips `-agentic` for `auto`), 5-minute per-credential cache, automatic token refresh on 401

### `feat(api): expose live Kiro models in routes` (944a685)
- `src/app/api/providers/[id]/models/route.js`: Kiro branch wired to `resolveKiroModels`
- `src/app/api/v1/models/route.js`: per-active-Kiro-account live merge, dedupes synthetic variants by ID across multiple accounts

### `feat(openai-to-kiro): forward vision images` (522be88)
- `open-sse/translator/request/openai-to-kiro.js`: `extractKiroImage()` helper recognises both OpenAI `image_url` and Claude `image` blocks; data-URL only; attaches results to `userInputMessage.images` on current message and history

## Behaviour

| Requested ID | Sent upstream | Thinking | Agentic prompt |
|---|---|---|---|
| `kr/claude-sonnet-4.5` | `claude-sonnet-4.5` | ❌ (unless body signals) | ❌ |
| `kr/claude-sonnet-4.5-thinking` | `claude-sonnet-4.5` | ✅ | ❌ |
| `kr/claude-sonnet-4.5-agentic` | `claude-sonnet-4.5` | ❌ | ✅ |
| `kr/claude-sonnet-4.5-thinking-agentic` | `claude-sonnet-4.5` | ✅ | ✅ |
| `kr/claude-sonnet-4.5` + `reasoning_effort: high` | `claude-sonnet-4.5` | ✅ | ❌ |
| Any Kiro model + `image_url` data URL | upstream model + `images[]` | ↑ | ↑ |

## Tested

- Unit-level smoke tests for `resolveKiroModel`, `isThinkingEnabled`, `buildKiroPayload`, `extractKiroImage` (all suffix combinations, all reasoning trigger paths, OpenAI/Claude image shapes, remote-URL drop, multi-turn history)
- Live `ListAvailableModels` call against a real social-auth Kiro account: 13 raw upstream models → 50 expanded variants (auto only gets base + thinking)
- `npm run build` passes
- Token refresh path exercised (expired social-auth token auto-refreshed via `refreshKiroToken`)
- All commits SSH-signed and verified by GitHub

## Notes for reviewers

- The model fetcher is best-effort: if `ListAvailableModels` fails or returns empty, `/v1/models` falls back to the static catalog so a single broken Kiro account doesn't take the whole endpoint down.
- The `auto` model intentionally skips `-agentic` variants because Kiro picks the underlying model server-side and the chunked-write prompt targets concrete coding-agent file writes (matches CLIProxyAPIPlus behaviour).
- Remote image URLs are dropped on purpose (no synchronous network fetch in the request path). Clients that need to forward images should base64-encode them as data URLs.
- No changes to OAuth flows, token storage, or other providers.
- Might close: #874 #1026 #1078
- Might count as duplicate for: #1167 #1044 #815 #601 #1170

### DIFF
--- a/open-sse/config/kiroConstants.js
+++ b/open-sse/config/kiroConstants.js
@@ -1,0 +1,262 @@
+/**
+ * Kiro-specific constants and helpers.
+ *
+ * Mirrors the behaviour of `internal/translator/kiro/common/constants.go` and
+ * `internal/translator/kiro/claude/kiro_claude_request.go` from the
+ * CLIProxyAPIPlus reference implementation, scoped down to what 9router needs:
+ *
+ *   - `-agentic` model suffix detection + chunked-write system prompt
+ *   - reasoning / thinking trigger detection (Anthropic-Beta header,
+ *     Claude `thinking`, OpenAI `reasoning_effort`, AMP/Cursor magic tag)
+ *   - the `<thinking_mode>enabled</thinking_mode>` system-prompt injection
+ *     that turns Kiro reasoning on
+ *
+ * Kiro upstream does not advertise `-agentic` model IDs; they are a 9router
+ * fiction. The suffix is stripped before the request leaves this process.
+ */
+
+export const KIRO_AGENTIC_SUFFIX = "-agentic";
+export const KIRO_THINKING_SUFFIX = "-thinking";
+
+export const KIRO_THINKING_BUDGET_DEFAULT = 16000;
+
+export const KIRO_AGENTIC_SYSTEM_PROMPT = `
+# CRITICAL: CHUNKED WRITE PROTOCOL (MANDATORY)
+
+You MUST follow these rules for ALL file operations. Violation causes server timeouts and task failure.
+
+## ABSOLUTE LIMITS
+- **MAXIMUM 350 LINES** per single write/edit operation - NO EXCEPTIONS
+- **RECOMMENDED 300 LINES** or less for optimal performance
+- **NEVER** write entire files in one operation if >300 lines
+
+## MANDATORY CHUNKED WRITE STRATEGY
+
+### For NEW FILES (>300 lines total):
+1. FIRST: Write initial chunk (first 250-300 lines) using write_to_file/fsWrite
+2. THEN: Append remaining content in 250-300 line chunks using file append operations
+3. REPEAT: Continue appending until complete
+
+### For EDITING EXISTING FILES:
+1. Use surgical edits (apply_diff/targeted edits) - change ONLY what's needed
+2. NEVER rewrite entire files - use incremental modifications
+3. Split large refactors into multiple small, focused edits
+
+### For LARGE CODE GENERATION:
+1. Generate in logical sections (imports, types, functions separately)
+2. Write each section as a separate operation
+3. Use append operations for subsequent sections
+
+## EXAMPLES OF CORRECT BEHAVIOR
+
+CORRECT: Writing a 600-line file
+- Operation 1: Write lines 1-300 (initial file creation)
+- Operation 2: Append lines 301-600
+
+CORRECT: Editing multiple functions
+- Operation 1: Edit function A
+- Operation 2: Edit function B
+- Operation 3: Edit function C
+
+WRONG: Writing 500 lines in single operation -> TIMEOUT
+WRONG: Rewriting entire file to change 5 lines -> TIMEOUT
+WRONG: Generating massive code blocks without chunking -> TIMEOUT
+
+## WHY THIS MATTERS
+- Server has 2-3 minute timeout for operations
+- Large writes exceed timeout and FAIL completely
+- Chunked writes are FASTER and more RELIABLE
+- Failed writes waste time and require retry
+
+REMEMBER: When in doubt, write LESS per operation. Multiple small operations > one large operation.
+`.trim();
+
+/**
+ * Detect whether an inbound request is asking for reasoning / thinking output.
+ *
+ * Sources of intent (any one is enough):
+ *   - HTTP header `Anthropic-Beta: ...interleaved-thinking...`
+ *   - JSON `thinking.type === "enabled"` (Claude Messages API)
+ *   - JSON `reasoning_effort` in {low, medium, high, auto} (OpenAI o1/o3)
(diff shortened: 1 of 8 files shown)


decolua:
Merged into master with light refactoring (config-driven `customResolver` pattern for the API routes). Thanks for the detailed PR description and the parity work! 🙏

PR #1167, opened 2026-05-15
fix(kiro): forward images to currentMessage in buildKiroPayload

DuongStark:
## Problem

`buildKiroPayload()` silently dropped images when building the final request payload.

In `convertMessages()`, base64 images from `image_url` content parts are correctly parsed and stored in `userInputMessage.images`. However, when `buildKiroPayload()` constructs the final `currentMessage`, it only spread `content` and `userInputMessageContext` — the `images` field was never forwarded.

This caused vision requests to always fail with "I don't see any image attached" even when a valid base64 image was provided.

## Fix

Add a conditional spread for `images` into `currentMessage.userInputMessage`, using the same pattern already used for `userInputMessageContext`:

```js
...(currentMessage?.userInputMessage?.images?.length > 0 && {
  images: currentMessage.userInputMessage.images
}),
```

## Tests

Added `tests/unit/openai-to-kiro.test.js` with 7 test cases covering:
- Basic text message conversion
- No `images` field when no images present
- Single base64 image forwarding
- Multiple base64 images forwarding
- Empty content array (no images)
- Text + image coexistence
- HTTP URL fallback to text placeholder (Kiro only supports base64)

All 7 tests pass. Existing test suite unaffected.

### DIFF
--- a/open-sse/translator/request/openai-to-kiro.js
+++ b/open-sse/translator/request/openai-to-kiro.js
@@ -307,6 +307,9 @@ export function buildKiroPayload(model, body, stream, credentials) {
           content: finalContent,
           modelId: model,
           origin: "AI_EDITOR",
+          ...(currentMessage?.userInputMessage?.images?.length > 0 && {
+            images: currentMessage.userInputMessage.images
+          }),
           ...(currentMessage?.userInputMessage?.userInputMessageContext && {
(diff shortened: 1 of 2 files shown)


decolua:
Merged into master, thanks for the clean fix and tests! 🙏

PR #1193, opened 2026-05-16
fix: Responses API MCP namespace 往返 + deepseek 思考后缀支持

Nicolas-One:
涉及 8 个文件：

  1. request/openai-responses.js — namespace 工具类型展平 (type:namespace + 显式 namespace 字段)、custom_tool_call tool_choice 转换、Chat→Responses 反向转换、工具去重。

  2. response/openai-responses.js — MCP 扁平工具名解码回 (name, namespace) 注入 function_call 输出项、response.completed 构建完整 output 数组、补全 model/incomplete_details 字段、 新增 reasoning encrypted_content。

  3. translator/index.js — OPENAI_RESPONSES 状态新增 funcNamespaces:{} 和 model:null。

  4. streamingHandler.js / chatCore.js / stream.js — toolNameNSMap 透传至 SSE 流管道。

  5. reasoningContentInjector.js — deepseek-v4-pro-* 硬编码 泛化为通用后缀模式 (none/low/medium/high/max)，适用于任何 deepseek- 模型。思考禁用时跳过 reasoning_content 注入。


Nicolas-One:
可以参考我的库：[https://github.com/Nicolas-One/chat2responses](url)

Nicolas-One:
兼容目前最新的Codex v0.130.0，不仅仅只支持v0.120.0版本。

zxfxpower:
Adding a follow-up based on a static inspection of v0.5.75 (`83af3f1853b295eda428f3d3d8f5ff96ca25bfab`): the Responses → Chat Completions → Responses translation path still appears to lose namespace semantics.

- [Request translation](https://github.com/decolua/9router/blob/83af3f1853b295eda428f3d3d8f5ff96ca25bfab/open-sse/translator/request/openai-responses.js#L186-L225) reads `tool.name` directly without expanding the child tools of a `type: "namespace"` declaration. A named namespace container falls through to ordinary function conversion instead of exposing its child tool declarations.
- [Streaming response translation](https://github.com/decolua/9router/blob/83af3f1853b295eda428f3d3d8f5ff96ca25bfab/open-sse/translator/response/openai-responses.js#L290-L299) constructs the `function_call` item with a tool name but does not restore its namespace.

The affected use case is Codex calling namespaced tools through 9Router with a Chat Completions provider such as DeepSeek. The expected behavior is to flatten the child tools into provider-compatible function declarations while retaining a reversible mapping, then restore the original tool name and namespace in the Responses output, including streaming events and the final output.

This follow-up is based on static inspection of v0.5.75; I have not rerun a live end-to-end request against the unpatched release for this report. If another PR already provides an equivalent fix, please link it here. Otherwise, namespace round-trip regression coverage would be useful, especially for tools with the same name in different namespaces. The implementation does not need to use any particular mapping variable name.

PR #1216, opened 2026-05-17
fix: handle OpenClaw object-format model in cli-tools/openclaw-settings route

### LINKED ISSUE #1196: TypeError: model.startsWith is not a function when openclaw config uses object model format
## Bug: TypeError: model.startsWith is not a function when openclaw config uses object model format

**Labels:** bug, openclaw integration

### Environment
- 9router version: v0.4.50
- OpenClaw version: 2026.5.12
- OS: Ubuntu 24.04 (AWS EC2)
- Node: v24.15.0
- Install method: npm install -g (system)

### Description
9router fails to read OpenClaw settings and shows "not configured" status when the OpenClaw config uses an object format for the `model` field (which is the default format written by OpenClaw 2026.5.x). The error occurs in the `/api/cli-tools/all-statuses` route which calls `.startsWith()` on the model value, but the value is an object, not a string.

### Error log
```
Error updating openclaw settings: TypeError: a.model?.startsWith is not a function
    at /usr/local/lib/node_modules/9router/app/.next-cli-build/server/app/api/cli-tools/all-statuses/route.js:17:4901
    at Array.map (<anonymous>)
    at w (.../all-statuses/route.js:17:4881)
    at async rz.do (.../next-server/app-route.runtime.prod.js:5:21393)
```

### OpenClaw config (model field causing the issue)
The model field in `~/.openclaw/openclaw.json` is written by OpenClaw as an object:
```json
"agents": {
  "list": [
    {
      "id": "main",
      "model": {
        "primary": "custom-localhost-20128/gh/gpt-4o-mini",
        "fallbacks": [
          "custom-localhost-20128/gh/gpt-4o"
        ]
      }
    }
  ]
}
```
9router appears to expect `model` to be a plain string and calls `.startsWith()` on it directly without checking the type first.

### Steps to reproduce
1. Install OpenClaw 2026.5.12 — it writes the model field as an object with "primary" and "fallbacks" keys by default.
2. Install 9router v0.4.50 and open the dashboard.
3. Observe "not configured" status for OpenClaw integration.
4. Check 9router logs — TypeError appears on every status poll.

### Expected behavior
9router should handle both formats:
- String: `"model": "provider/model-id"`
- Object: `"model": { "primary": "...", "fallbacks": [...] }`

The active model string should be read from `model.primary` when the field is an object.

### Workaround
None without losing the fallback model configuration.

mvanhorn:
## Summary

Normalize `agent.model` in `src/app/api/cli-tools/openclaw-settings/route.js` so the dashboard and POST handler work when OpenClaw 2026.5.x writes the field as `{ primary, fallbacks }` instead of a plain string. Without this, the GET enrichment ships an object down to the client and the POST `.startsWith("9router/")` filter throws `TypeError: model?.startsWith is not a function`.

## Why this matters

Issue #1196 reports `TypeError: model?.startsWith is not a function` on the status poll loop and "not configured" in the dashboard when the local OpenClaw config holds `model: { primary: "...", fallbacks: [...] }`. The reporter included the stack trace, the config shape, and the expected behavior (read the active id from `model.primary`).

## Implementation notes

Added a `resolveAgentModel` helper at the top of `src/app/api/cli-tools/openclaw-settings/route.js`:

- Returns the string unchanged when `typeof m === "string"`.
- Returns `m.primary ?? ""` when `m` is an object.
- Returns `""` otherwise (covers `undefined`, `null`, `{}`).

Applied in two places:

1. GET enrichment loop: the `model` field on each enriched agent is now coerced to its string id, so the dashboard receives a string.
2. POST `9router/` cleanup: the `.startsWith("9router/")` check now runs on the resolved string.

The plan's `Files:` scope also listed `src/app/api/cli-tools/all-statuses/route.js`, but that file is a pure dispatcher that does not inspect model fields itself - the normalized values reach it through the openclaw GET it invokes, so no change was needed there.

## Testing

- `npm run lint` clean on the changed file.
- Verified the three test scenarios from the issue by reading the diff:
  - String form: `resolveAgentModel("custom-localhost-20128/gh/gpt-4o-mini")` returns the input; existing path unchanged.
  - Object form: `resolveAgentModel({ primary: "9router/...", fallbacks: [...] })` returns `"9router/..."`; the `.startsWith("9router/")` filter matches.
  - Missing model (`undefined`, `null`, `{}`): returns `""`; `.startsWith` is a no-op and no exception is thrown.

Fixes #1196

### DIFF
--- a/src/app/api/cli-tools/openclaw-settings/route.js
+++ b/src/app/api/cli-tools/openclaw-settings/route.js
@@ -9,6 +9,15 @@ import os from "os";
 
 const execAsync = promisify(exec);
 
+// OpenClaw 2026.5.x writes agents[].model as either a plain string
+// (legacy) or as an object `{ primary, fallbacks }`. Normalize to the
+// string id so downstream consumers can call `.startsWith()` safely.
+const resolveAgentModel = (m) => {
+  if (typeof m === "string") return m;
+  if (m && typeof m === "object") return m.primary ?? "";
+  return "";
+};
+
 const getOpenClawDir = () => path.join(os.homedir(), ".openclaw");
 const getOpenClawSettingsPath = () => path.join(getOpenClawDir(), "openclaw.json");
 
@@ -79,12 +88,14 @@ export async function GET() {
 
     const settings = await readSettings();
 
-    // Enrich agents list with current per-agent model from models.json
+    // Enrich agents list with current per-agent model from models.json.
+    // Coerce agent.model to its string id when OpenClaw stores it as
+    // `{ primary, fallbacks }` so downstream `.startsWith()` calls work.
     const agentList = settings?.agents?.list || [];
     const enrichedAgents = await Promise.all(
       agentList.map(async (agent) => {
         const agentModel = agent.agentDir ? await readAgentModel(agent.agentDir) : null;
-        return { ...agent, currentModel: agentModel };
+        return { ...agent, model: resolveAgentModel(agent.model), currentModel: agentModel };
       })
     );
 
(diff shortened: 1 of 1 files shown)


mvanhorn:
Thanks for the quick merge, @decolua. The OpenClaw object-format model branch in cli-tools needed this fix.

PR #1217, opened 2026-05-17
fix: make db.json to SQLite migration atomic with row-count assertion

### LINKED ISSUE #1192: Migration 0.4.41 → 0.4.50 silently drops provider connections from db.json
After updating from 0.4.41 to 0.4.50 on macOS (Node v25.9.0, nvm), several rows in `~/.9router/db.json.providerConnections` failed to land in the new SQLite (`~/.9router/db/data.sqlite`). Lost connections include `nvidia/nvidia`, `gemini/hermes`, `kiro/Account 1`, `fireworks/hermes`, and `fireworks/bruko`, while expired Kiro tokens were kept and kept spamming `TOKEN_REFRESH Failed to refresh Kiro social token (401 Bad credentials)`. Models that previously routed via fallback chains started returning HTTP 401 because the underlying connections were gone. Auto-update also failed earlier with `npm EACCES (exitCode 243)` and only manual reinstall succeeded — likely correlated with the broken migration.

**Suggested fix (maintainer):** make the `db.json` → SQLite migration atomic and verified — write to a temp DB, assert row counts match `db.json.providerConnections`, then swap; if mismatch, abort and keep `db.json` intact instead of starting with partial data. Log every dropped row with reason so users can triage. Bonus: prepend `/usr/sbin:/sbin:/usr/bin:/bin` to `PATH` before `execSync('ioreg ...')` so machine-ID detection stops failing on nvm-launched processes.

**User workaround (until fixed):**

```bash
# 1. Stop 9router
pgrep -f 9router | xargs kill

# 2. Backup the corrupt SQLite (db.json is preserved automatically)
cp ~/.9router/db/data.sqlite ~/.9router/db/data.sqlite.bak.$(date +%s)
rm ~/.9router/db/data.sqlite ~/.9router/db/data.sqlite-shm ~/.9router/db/data.sqlite-wal

# 3. Restart 9router — it rebuilds SQLite from db.json on next launch
9router --tray --skip-update
```

Then re-login expired providers (Kiro, etc.) from the dashboard at `http://127.0.0.1:20128`. Or restore a known-good `data.sqlite` backup if you have one.

mvanhorn:
## Summary

Make the `db.json` to SQLite migration in `src/lib/db/migrate.js` fail closed when `providerConnections` row counts diverge. The migration now collects per-row insert failures, asserts the post-loop `COUNT(*)` matches the input array length, and lets the surrounding transaction roll back on mismatch so the legacy `db.json` is preserved byte-for-byte.

## Why this matters

Issue #1192 reports that upgrading from 0.4.41 to 0.4.50 silently dropped five `providerConnections` rows (`nvidia/nvidia`, `gemini/hermes`, `kiro/Account 1`, `fireworks/hermes`, `fireworks/bruko`) and started returning HTTP 401 for models that had been routing via fallback chains. The reporter's expected behavior: assert row counts, abort on mismatch, leave `db.json` intact, and log every dropped row with a reason.

## Implementation notes

- The migration already runs inside `adapter.transaction`. The SQLite transaction rollback acts as the atomic-swap mechanism; on `MigrationAborted` the partial inserts disappear and the final SQLite table state is empty, functionally equivalent to the temp-file plus rename pattern suggested in the issue and matching the existing driver convention.
- Wrapped the per-row `providerConnections` insert in `try`/`catch`, collecting `{ id, provider, name, reason }` for any row that fails.
- After the loop, `SELECT COUNT(*) FROM providerConnections` is compared against the input array length. On mismatch, the dropped-row list is logged and `MigrationAborted` is thrown.
- The outer `try`/`catch` in `runMigrationOnce` catches `MigrationAborted`, logs it, and returns without writing the `.migrated-from-json` marker. Next boot the app reads from `db.json` again as the issue requires.

## Testing

Added `tests/unit/db-migration-chain.test.js` covering the mismatch path:

- Fixture: 5 valid `providerConnections` plus one row that triggers a primary-key violation.
- Asserts:
  - the `providerConnections` SQLite table is empty after rollback,
  - `db.json` is byte-identical to the pre-migration state,
  - the `.migrated-from-json` marker file is absent.

`package.json` does not currently expose a `test` script, so the test was added per the plan but not executed locally.

Fixes #1192

### DIFF
--- a/src/lib/db/migrate.js
+++ b/src/lib/db/migrate.js
@@ -11,6 +11,17 @@ import { stringifyJson } from "./helpers/jsonCol.js";
 // Marker file: prevents re-importing legacy JSON when user wipes data.sqlite.
 const MIGRATED_MARKER = path.join(DB_DIR, ".migrated-from-json");
 
+// Thrown by importLegacyMain when row-count assertion fails.
+// The legacy db.json is left intact and the marker is not written so the
+// next boot can retry (or the user can restore from backup).
+export class MigrationAborted extends Error {
+  constructor(message, droppedRows) {
+    super(message);
+    this.name = "MigrationAborted";
+    this.droppedRows = droppedRows;
+  }
+}
+
 // Track per-adapter so reusing same adapter skips re-run, but new adapter (after reset) re-runs.
 const _migratedAdapters = new WeakSet();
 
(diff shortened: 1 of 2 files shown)


decolua:
Thanks for the fix! The pattern is solid — try/catch per row + post-loop `COUNT(*)` assertion + `MigrationAborted` → transaction rollback → marker not written is exactly the right approach.

One suggestion for follow-up: the same silent-drop risk exists for the other critical tables imported in `importLegacyMain` (`providerNodes`, `proxyPools`, `apiKeys`, `combos`). Any of them hitting a schema constraint today would silently disappear the same way `providerConnections` did in #1192.

Rather than duplicating the try/catch + assert block 5×, a small helper keeps it DRY:

```js
function importWithAssertion(adapter, tableName, rows, insertFn, rowMeta) {
  const dropped = [];
  for (const row of rows) {
    try { insertFn(row); }
    catch (err) { dropped.push({ ...rowMeta(row), reason: err.message }); }
  }
  const inserted = adapter.get(`SELECT COUNT(*) as c FROM ${tableName}`)?.c ?? 0;
  if (inserted !== rows.length) {
    throw new MigrationAborted(
      `${tableName} row-count mismatch: expected ${rows.length}, got ${inserted}`,
      dropped
    );
  }
}
```

Then each table import becomes a single call. Happy to push a follow-up PR or you can fold it into this one.

Also worth flagging the related machine-ID issue from #1192 (the `ioreg`/PATH problem on nvm-launched processes). Persisting the raw machine-ID to `~/.9router/machine-id` on first successful read makes the `x-9r-cli-token` bypass deterministic across runtimes — that's a separate change but fixes the 401 chain that the dropped rows trigger.

Issue #536, opened 2026-04-09
Unsupported parameter: 'temperature'


binh247:
Github [400]: [400]: Unsupported parameter: 'temperature' is not supported with this model.

<img width="579" height="185" alt="Image" src="https://github.com/user-attachments/assets/d435e2ed-bdc5-4d0d-9f15-22a75a57a08e" />

xx7eVeN:
Hello there,

first of all thanks for sharing this incredible project with the world.
All newer GPT-Models have the same issue. Stripping it only for 5.4 doesnt fix the long term problem sadly.
GPT 5.5, 5.6 and higher has the same problem.

PR #1264, opened 2026-05-19
fix(translator): strip temperature for Claude models with extended thinking

noestelar:
## Problem

Claude's Messages API returns `400 invalid_request_error: \`temperature\` is deprecated` whenever extended thinking is active for a request, but the OpenAI → Claude translator in `open-sse/translator/request/openai-to-claude.js` unconditionally copies `body.temperature` into the Claude payload. This breaks two real-world cases:

**Case A — caller explicitly enables thinking.** When a client sends `thinking: { type: "enabled", ... }` or `reasoning_effort: "medium"`, `openaiToClaudeRequest()` already sets `result.thinking`, but `result.temperature` was also set earlier and never removed.

**Case B — Claude OAuth (`claude-code`) provider.** The `claude` executor in `open-sse/executors/default.js` always sends:

```
Anthropic-Beta: claude-code-20250219,interleaved-thinking-2025-05-14,...
```

(see `open-sse/config/providers.js` `CLAUDE_API_HEADERS` / `CLAUDE_CLI_SPOOF_HEADERS`). This forces server-side extended thinking on `claude-opus-4.x` and `claude-sonnet-4.x` even when the client never sets `body.thinking`. The translator can't see headers, so the only signal it has is the model name.

## Reproduction

OpenAI-format client → 9router → `cc/claude-opus-4-7` with any non-default temperature:

```bash
curl -sS http://127.0.0.1:20128/v1/chat/completions \
  -H 'Authorization: Bearer ***' \
  -d '{
    "model": "claude/claude-opus-4-7",
    "messages": [{"role":"user","content":"hi"}],
    "temperature": 0.7,
    "max_tokens": 32
  }'
```

Response (pre-fix):

```
400 {"type":"error","error":{"type":"invalid_request_error","message":"`temperature` is deprecated"}}
```

This affects every OpenAI-compatible client routed through 9router to a Claude 4.x model with a non-default `temperature` — Factory Droid, custom clients, JCode, etc.

## Fix

Two-part guard in `openaiToClaudeRequest()`:

1. **Model-based strip** — skip copying `temperature` when the target model matches `claude-(?:opus|sonnet)-4` (the families with forced server-side thinking on OAuth). This mirrors the existing pattern in `open-sse/executors/github.js`:

   ```js
   supportsTemperature(model) {
     return !/gpt-5\.4/i.test(model);
   }
   ```
   (added in #536 for `gpt-5.4`.)

2. **Defensive final-pass** — after `result.thinking` is potentially set from `body.thinking` or `body.reasoning_effort`, strip `result.temperature` if it slipped through.

Models without forced thinking (haiku, sonnet-3.5, older opus) keep `temperature` unless the caller explicitly enables thinking.

## Tests

Adds 6 unit tests in `tests/unit/openai-to-claude.test.js`:

- ✅ strips temperature for `claude-opus-4-7` (forced thinking)
- ✅ strips temperature for `claude-sonnet-4-5` (forced thinking)
- ✅ preserves temperature for `claude-haiku-5-*` and `claude-3-5-sonnet-*`
- ✅ strips temperature when `body.thinking` is explicit (on non-forced model)
- ✅ strips temperature when `reasoning_effort=medium` triggers thinking
- ✅ preserves temperature when `reasoning_effort=none` on non-forced model

Full result on this branch:

```
tests/unit/openai-to-claude.test.js (10 tests) ✓ 6ms
Test Files  1 passed (1)
     Tests  10 passed (10)
```

Suite-wide: 322 passing vs 316 on `main` — net +6 passing, no regressions (the 24 pre-existing failures in `main` are unrelated to this change; reproduced by `git stash && npx vitest run`).

## Notes

- The regex intentionally matches `claude-opus-4` and `claude-sonnet-4` with any tail (so `claude-opus-4-7`, `claude-sonnet-4-5`, `claude-opus-4.7`, etc. all match). If future Claude 4.x sub-families ship *without* forced thinking, this regex would need narrowing.
- An alternative would be to do the strip inside the `claude` executor (`default.js`) instead of the translator. I chose the translator because (a) the translator is where the `temperature` field is born for OpenAI-format inputs and (b) `claude`-format providers like `agentrouter` that also send `interleaved-thinking` would benefit from the same strip without separate plumbing.

### DIFF
--- a/open-sse/translator/request/openai-to-claude.js
+++ b/open-sse/translator/request/openai-to-claude.js
@@ -18,7 +18,21 @@ export function openaiToClaudeRequest(model, body, stream) {
   };
 
   // Temperature
-  if (body.temperature !== undefined) {
+  //
+  // Claude's Messages API rejects `temperature` when extended thinking is active.
+  // Two cases where thinking is on:
+  //   (a) Caller passes `body.thinking` or `body.reasoning_effort` (handled below
+  //       — `result.thinking` becomes truthy, and we strip temperature at the end).
+  //   (b) The request targets Claude OAuth (claude-code), which always sends
+  //       `Anthropic-Beta: ...,interleaved-thinking-2025-05-14,...` in headers.
+  //       The model is forced into thinking server-side, but neither `body.thinking`
+  //       nor `result.thinking` will be set, so we must detect this by model name.
+  //       This affects claude-opus-4.x and claude-sonnet-4.x (the families that
+  //       support extended thinking).
+  //
+  // For models that don't force thinking (haiku, older sonnets), preserve temperature.
+  const modelForcesThinking = /claude-(?:opus|sonnet)-4/i.test(model);
+  if (body.temperature !== undefined && !modelForcesThinking) {
     result.temperature = body.temperature;
   }
 
@@ -196,6 +210,14 @@ Respond ONLY with the JSON object, no other text.`);
     } else if (budget) {
       result.thinking = { type: "enabled", budget_tokens: budget };
     }
+  }
+
+  // Final guard: Claude rejects `temperature` whenever extended thinking is
+  // enabled. If we set `result.thinking` above from `body.thinking` or
+  // `body.reasoning_effort`, drop temperature defensively (the model-based
+  // strip earlier already covers Claude OAuth's forced-thinking case).
+  if (result.thinking && result.temperature !== undefined) {
+    delete result.temperature;
   }
 
   // Attach toolNameMap to result for response translation
--- a/tests/unit/openai-to-claude.test.js
+++ b/tests/unit/openai-to-claude.test.js
@@ -121,4 +121,82 @@ describe("openaiToClaudeRequest", () => {
       expect(systemText).toContain("You must respond with valid JSON");
     });
   });
+
+  describe("temperature handling", () => {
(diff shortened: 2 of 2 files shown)


airano-ir:
Confirming this also hits the new Claude 5 family, and **independently of thinking**: on v0.5.18, any request that carries `temperature` to `claude-fable-5` or `claude-sonnet-5` through the `claude` (OAuth) provider fails with:

```
[400]: {"type":"error","error":{"type":"invalid_request_error","message":"`temperature` is deprecated for this model."}}
```

Repro: plain non-streaming `POST /v1/chat/completions` with `{"model":"cc/claude-sonnet-5","temperature":0.3,"messages":[...]}` — no `thinking` / `reasoning_effort` involved. Same for `cc/claude-fable-5`. So Anthropic appears to reject `temperature` at the *model level* for Claude 5, same as claude-opus-4.

Practical impact: chat UIs that pin `temperature` on background tasks (title/tag generation etc.) generate a steady stream of 400s, which also puts the model on failure cooldown and temporarily degrades any combo that has a Claude 5 model first in the chain.

So besides the thinking-based strip in this PR, a model-based rule in `paramSupport.js` seems needed, e.g. extending the existing `/claude-opus-4/` rule to something like `/claude-(opus-4|(opus|sonnet|fable|haiku)-5)/i`.

PR #1272, opened 2026-05-19
fix(chatCore): default stream to false per OpenAI spec (#1260)

### LINKED ISSUE #1260: fix: streaming default in handleChatCore treats absent stream param as streaming
## Problem

When a client sends a chat completion request **without** a `stream` field (e.g. `{"model":"...","messages":[...]}`), `handleChatCore` defaults to `stream=true` because the check is:

```js
```

`undefined !== false` evaluates to `true`, so every request without an explicit `stream: false` is treated as streaming. This breaks clients that follow the OpenAI API spec, where the default for `stream` is `false` (non-streaming).

## Impact

When a non-streaming client sends a request without `stream: false`, the response comes back as SSE (`text/event-stream`) instead of JSON (`application/json`). The client gets a JSON parse error because `data: [DONE]` is appended after the JSON body.

## Proposed fix

Change the default to match the OpenAI spec:

```diff
- let stream = providerRequiresStreaming ? true : (body.stream !== false);
+ let stream = providerRequiresStreaming ? true : (body.stream === true);
```

## Workaround for users

Until this is fixed, add `stream: false` to the request body, or send `Accept: application/json` in the request headers.

---

_I encountered this while configuring agentmemory to use 9router as its OpenAI-compatible proxy. Filed from Claude Code._

Ptah-CT:
## Summary

Fixes #1260 with the exact one-line change proposed in that issue.

`open-sse/handlers/chatCore.js`:
```diff
-  let stream = providerRequiresStreaming ? true : (body.stream !== false);
+  let stream = providerRequiresStreaming ? true : (body.stream === true);
```

## Why

OpenAI API spec defines `stream` as defaulting to `false` when absent from the request body. The old condition `body.stream !== false` evaluates `undefined !== false` to `true`, so every request without an explicit `stream` field was treated as streaming. Clients that follow the spec then receive `text/event-stream` instead of `application/json` and crash when their JSON parser hits `data: {"id"...`.

`clientRequestedStreaming` already uses `body.stream === true` on the line above — the new condition is now consistent with it.

## Test plan

- [ ] Manual: `curl -X POST http://localhost:20128/v1/chat/completions -d '{"model":"haiku","messages":[{"role":"user","content":"hi"}]}'` → `content-type: application/json` (was `text/event-stream`)
- [ ] Manual: same request with `"stream":true` → still SSE
- [ ] Manual: `cx/gpt-5.5` route unchanged (already JSON, covered by `providerRequiresStreaming` short-circuit)
- [ ] Manual: deepseek-tui non-interactive heuristic unchanged
- [ ] Existing test suite passes (`tests/`)

## Real-world breakage

Filed from a real integration: agentmemory's OpenAI-compatible LLM provider (`src/providers/openai.ts` upstream) follows the spec and never sets `stream: false`. Every `mem::compress` call fails with:

```
[agentmemory] error Compression failed {"obsId":"...","error":"Unexpected token 'd', \"data: {\"id\"... is not valid JSON"}
```

After enough failures, agentmemory's circuit breaker trips and all subsequent compressions return `circuit_breaker_open`.

## Workaround for users on the current release

Until this merges:
- Add `stream: false` to the request body (works for clients that allow custom payload fields), or
- Send `Accept: application/json` (already honored on lines 71–76 of the same file)

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -58,7 +58,11 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
 
   const clientRequestedStreaming = body.stream === true || sourceFormat === FORMATS.ANTIGRAVITY || sourceFormat === FORMATS.GEMINI || sourceFormat === FORMATS.GEMINI_CLI;
   const providerRequiresStreaming = provider === "openai" || provider === "codex" || provider === "commandcode";
-  let stream = providerRequiresStreaming ? true : (body.stream !== false);
+  // OpenAI API spec: stream defaults to false when absent.
+  // Old check `body.stream !== false` evaluated `undefined !== false` to true,
+  // so every request without an explicit stream field was treated as streaming,
+  // returning text/event-stream to clients that expected application/json. See #1260.
+  let stream = providerRequiresStreaming ? true : (body.stream === true);
 
   // DeepSeek-TUI: interactive TUI panel sends stream:true and needs SSE.
   // Non-interactive mode (-p flag) sends without stream and can't parse SSE.


PBK-B:
I had the same problem today when I was implementing some other LLM tool call API, and I spent some time investigating why it was 9router.I looked at the context, and I wanted to know what was the hindrance to this merger? @decolua

PBK-B:
> In addition, I also observed `POST /v1/chat/completions`: Malformed SSE response when `stream` is omitted

## Summary

When `stream` is **absent** from the request body, the server returns `text/event-stream` but the body is **not valid Server-Sent Events** — it dumps a complete JSON object **without** `data:` prefix and concatenates `data: [DONE]` on the **same line**. This breaks any client doing `response.json()`.

---

## Expected vs Actual

| Request | Content-Type | Body | Status |
|---------|-------------|------|--------|
| `{"stream": false, ...}` | `application/json` | `{"id":"...", "choices":[...], "usage":{...}}` | ✅ Correct |
| `{"stream": true, ...}` | `text/event-stream` | `data: {"id":"...", "choices":[{"delta":{...}}]}\ndata: [DONE]` | ✅ Standard SSE |
| **`{/* no stream */, ...}`** | **`text/event-stream`** | **`{"id":"...", "choices":[...], "usage":{...}}data: [DONE]\n`** | ❌ **Broken** |

---

## Root Cause

When `stream` is omitted, the server falls into streaming mode but:

1. **Missing `data: ` prefix** — the full JSON object is written raw instead of `data: <json>`
2. **No line break** — `data: [DONE]` is glued to the closing `}` of the JSON on the same line
3. **Clients calling `response.json()` crash** — JSON is valid but followed by `data: [DONE]`, throwing:  
   `SyntaxError: Unexpected non-whitespace character after JSON`

---

## How to Reproduce

```bash
curl -X POST '<endpoint>/v1/chat/completions' \
  -H 'Authorization: Bearer <key>' \
  -d '{
    "messages":[{"role":"user","content":"Hello"}],
    "model":"deepseek-v4-flash"
  }'
```

Response (10/10 times):

```http
HTTP/1.1 200 OK
Content-Type: text/event-stream

{"id":"...","object":"chat.completion","choices":[{"index":0,"message":{"role":"assistant","content":"..."},"finish_reason":"stop"}],...,"cost":"0"}data: [DONE]

```

---

## Expected Fix

Either:

**Option A (preferred):** Treat omitted `stream` as `stream: false` — respond with `application/json` and a clean JSON body.

**Option B:** Treat omitted `stream` as `stream: true` but emit properly formatted SSE:

```text
data: {"id":"...","object":"chat.completion.chunk","choices":[{"index":0,"delta":{"role":"assistant"}}]}

data: {"id":"...",...}

data: [DONE]

```

---

## Notes

- `stream: false` and `stream: true` both work correctly.
- The bug only manifests when `stream` is **not present** in the request body.
- Reproducible 10/10 times across different message payloads.

> ~~This problem may have been solved by decolua/9router/pull/2046 (I did not retest because of the docker compilation version v0.5.12 error problem)~~, May be related decolua/9router/issues/1396

> It has been tested that the main branch <https://github.com/decolua/9router/commit/cce47dd809b1b01e23ebedeb78ad26e85bad429b> (version v0.5.12) has not been fixed.

skoeswanto:
I think my issue with deepseek via 9router is the same. I am using pi agent (or other agents, same error) It kept giving me this error:
```
[400]: {"error":{"message":"stream_options should be set along with stream = true","type":"invalid_r...
```
I don't know how to resolve this, using the 9Router Proxy v0.5.30 version.  Anybody know how to fix this?

josepowera:
I am using langchain4j and have same problem

juhuan:
+1 — hit this exact bug today (v0.5.40 and still present on master / v0.5.45, `chatCore.js` line 85).

Our case: MonkeyCode's model health check is spec-compliant and omits `stream`, expecting JSON — 9router returned `text/event-stream` and the check failed with `parse response: invalid character 'd' looking for beginning of value`. Same family as #2458 / #1396 / #2382.

FWIW I merged this PR onto current master in my fork — only one trivial conflict with the newer `PROVIDERS[provider]?.forceStream` refactor, resolved by keeping the refactor and taking this PR's `body.stream === true` fix: https://github.com/juhuan/9router/commit/3d6662f

Would be great to get this merged — it breaks any OpenAI-spec-compliant client (openai SDKs, langchain4j, go-openai, ...).

PR #1300, opened 2026-05-20
fix(tunnel): resolve unix redirection error in tailscale status checks on windows

philau2512:
## Summary
Resolves the issue where Tailscale connection fails on Windows platforms due to Unix-specific shell redirection syntax (`2>/dev/null`) inside node's execution context.

## Root Cause
- The strict status probe command `tailscale funnel status --json 2>/dev/null` uses Unix redirection. 
- On Windows, this shell syntax throws a `"The system cannot find the path specified"` exception, causing the status helper function `isTailscaleRunningStrict()` to catch the error and prematurely return `false`.
- This causes the frontend/backend connection flow to wrongly assume Tailscale is not connected or logged in.

## Changes
- Removed shell redirection (`2>/dev/null` and `||`) from Tailscale probing commands.
- Configured Node's native subprocess runner options `{ stdio: ["ignore", "pipe", "ignore"] }` to drop stderr silently in a cross-platform manner.
- Made binary detection (`bgRefreshBin`) platform-aware: uses `where` command on Windows and `which` on Unix-like operating systems.

## Test Plan
- Manually verified Tailscale login and funnel checks on Windows machine:
  - Both `getTailscaleBin()` and `isTailscaleRunningStrict()` successfully locate the binary and parse the funnel status JSON without raising system exceptions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,8 @@ ecosystem.config.*
 scripts/agSniffer/*
 gitbooks/*
 gitbook/README.md
+
+.claude/
+plans/
+.repomixignore
+CLAUDE.md
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -33,8 +33,8 @@ const nextConfig = {
         path: false,
       };
     }
-    // Exclude logs, .next, gitbook subapp from watcher
-    config.watchOptions = { ...config.watchOptions, ignored: /[\\/](logs|\.next|gitbook|cli)[\\/]/ };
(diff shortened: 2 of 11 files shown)


decolua:
Merged manually into `master` via commit 6b0dc09.

Included:
- fix(tunnel): Tailscale Windows cross-platform probe
- feat(usage): provider limits pagination + page size controls
- feat(providers): stop control for connection testing
- fix(sse): graceful stream abort/disconnect handling
- ui(quota): pagination layout improvements

Co-authored-by: @philau2512

PR #1338, opened 2026-05-21
Fix provider client priority sorting

### LINKED ISSUE #1323: GET /api/providers/client returns 500 when sorting by priority
## Description

The provider/client connections page fails to load because the backend returns `500 Internal Server Error` for the `/api/providers/client` endpoint when using `sort=priority`.

## Failing request

```http
GET http://localhost:20128/api/providers/client?page=1&pageSize=20&accountStatus=all&sort=priority

### PR DESCRIPTION
**Pull request by Delcado19**
# Fix provider client priority sorting

Fixes #1323.

## Summary

- Normalizes provider connection priority values before sorting.
- Handles `BigInt`, numeric strings, invalid values, and nullish values without throwing.
- Converts returned provider client payloads through a JSON-safe path so `BigInt` values cannot break `Response.json()`.
- Adds server-side `sort=expiring` handling before pagination using connection-level `expiresAt` / `rateLimitedUntil`.
- Updates the quota page to send `sort=expiring` when the Expiring first toggle is active.
- Returns debuggable provider-client 500 responses with `details: error.message`.

## Scope Note

`sort=expiring` intentionally uses saved connection expiry/rate-limit fields. Full live quota-reset ordering would require fetching provider quota data for every filtered account before pagination, plus refresh/proxy/error handling, so that should be a larger follow-up.

## Validation

```text
npx vitest run --config ./tests/vitest.config.js --reporter=verbose tests/unit/providers-client-route.test.js
-> 5 passed

npm run build
-> passed

git diff --check
-> passed
```

## Reviewer Context

This PR now also addresses the review notes from https://github.com/decolua/9router/pull/1338#issuecomment-4526623356 while keeping the fix scoped to `/api/providers/client` and the quota page request wiring.

### DIFF
--- a/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js
+++ b/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js
@@ -269,7 +269,7 @@ export default function ProviderLimits() {
           page: String(targetPage),
           pageSize: String(pageSize),
           accountStatus: accountFilter,
-          sort: "priority",
+          sort: sortRequestFromExpiringFirst(expiringFirst),
         });
 
         if (providerFilter !== "all") {
@@ -1004,8 +1004,8 @@ export default function ProviderLimits() {
       {/* Provider cards: 2 columns, compact */}
       {expiringFirst && (
         <div className="rounded-xl border border-amber-500/20 bg-amber-500/10 px-3 py-2 text-xs text-amber-700 dark:text-amber-300">
-          Expiring-first currently reorders accounts inside the current page.
-          Cross-page ordering still follows backend pagination.
+          Expiring-first uses saved token expiry and rate-limit times across
+          pages. Live quota-reset ordering requires a deeper quota refresh pass.
         </div>
       )}
 
--- a/src/app/api/providers/client/route.js
+++ b/src/app/api/providers/client/route.js
@@ -28,14 +28,20 @@ function maskName(name) {
   return name;
 }
 
+function toJsonSafe(value) {
+  if (typeof value !== "bigint") return value;
+  const asNumber = Number(value);
+  return Number.isSafeInteger(asNumber) ? asNumber : value.toString();
+}
+
 function sanitize(c) {
   const safe = {};
-  for (const f of SAFE_FIELDS) if (c[f] !== undefined) safe[f] = c[f];
+  for (const f of SAFE_FIELDS) if (c[f] !== undefined) safe[f] = toJsonSafe(c[f]);
   if (safe.name) safe.name = maskName(safe.name);
   if (c.providerSpecificData) {
     const psd = {};
     for (const f of SAFE_PSD_FIELDS) {
-      if (c.providerSpecificData[f] !== undefined) psd[f] = c.providerSpecificData[f];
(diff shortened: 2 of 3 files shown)


hamsa0x7:
Thanks for identifying the **real root cause** — `BigInt` priority values from SQLite breaking both the sort subtraction and `JSON.stringify`. I was investigating the same issue (#1323) and found three additional bugs in this endpoint that your PR doesn't cover. Would be great to address them here (or in a follow-up):

---

### 1. `sort=provider` branch: null provider crash

Your `sortableText()` already handles this — nice. ✅

### 2. `sort=expiring` is unhandled

The frontend defines `sortRequestFromExpiringFirst()` (ProviderLimits/index.js line 129-130) which returns `"expiring"` when the "expiring first" toggle is on. But `sortConnections()` only handles `"provider"` — everything else falls through to the priority sort. So `"expiring"` is silently treated as `"priority"`, making the "expiring first" button only re-sort the current page client-side (wrong across pages).

Suggested addition:
```javascript
if (sort === "expiring") {
  return list.sort((a, b) => {
    const timeA = new Date(a.expiresAt || a.rateLimitedUntil || 0).getTime();
    const timeB = new Date(b.expiresAt || b.rateLimitedUntil || 0).getTime();
    if (timeA > 0 && timeB > 0 && timeA !== timeB) return timeA - timeB;
    if (timeA > 0 && timeB === 0) return -1;
    if (timeA === 0 && timeB > 0) return 1;
    return sortablePriority(a.priority) - sortablePriority(b.priority);
  });
}
```

### 3. Frontend never sends `sort=expiring`

ProviderLimits/index.js line 272 hardcodes `sort: "priority"` instead of using `sortRequestFromExpiringFirst(expiringFirst)`. So the `"expiring"` sort value is never actually sent to the API — the toggle only re-sorts the current page client-side. This makes `sortRequestFromExpiringFirst` dead code.

Fix:
```javascript
// Line 272 — change from:
sort: "priority",
// to:
sort: sortRequestFromExpiringFirst(expiringFirst),
```

### 4. 500 error response hides the real error

The catch block returns `{ error: "Failed to fetch providers" }` with no details. Changing `console.log` → `console.error` and including `error.message` would make future 500s debuggable:
```javascript
} catch (error) {
  console.error("Error fetching providers for client:", error);
  return NextResponse.json(
    { error: "Failed to fetch providers", details: error.message },
    { status: 500 },
  );
}
```

---

Item 1 is already covered by your PR. Items 2-4 are separate bugs in the same endpoint that would be clean to fix here. Happy to contribute if you're open to it.

Delcado19:
Thanks for the detailed review. I pushed 71dd590 to cover items 2-4 in this PR: `sort=expiring` is handled server-side before pagination using connection-level `expiresAt` / `rateLimitedUntil`, the quota page now sends `sort=expiring` when the toggle is active, and provider-client 500s now include `details: error.message` with `console.error`. I also added route tests for expiring pagination, fallback priority ordering, and error details. One scope note: this intentionally does not implement full live quota-reset ordering from `/api/usage/:connectionId`; that would require fetching provider quota data for every filtered account before pagination, plus refresh/proxy/error handling, so it is better as a larger follow-up.

hamsa0x7:
👍

PR #1354, opened 2026-05-22
fix: strip empty Read pages argument in OpenAI-to-Claude translator

### LINKED ISSUE #1278: Claude Code Read tool fails: optional pages emitted as empty string
## 9Router lastest version

Latest version available as of 2026-05-19. Exact version not captured before switching away.

## Installation Method
npm/global or local 9router server. Exact install method not material to the repro.

## Operating System
Linux

## OS Version
Linux 7.0.0-15-generic

## Node.js Version
Not captured

## Provider(s) Involved
Codex / OpenAI-compatible route via 9router

Scope clarification: this has only been observed with Codex/OpenAI-compatible models/routes. It does not appear to affect non-Codex/non-OpenAI model routes in the same Claude Code setup.

## Model(s) Involved
- cx/gpt-5.5-high[1m]
- gpt-5.5-high as surfaced in Claude Code transcript

## CLI Tool
- Claude Code 2.1.144

## Description
Scope: this issue only happens with Codex/OpenAI-compatible models routed through 9router.

When Claude Code is configured to use 9router as the Anthropic-compatible base URL, non-PDF `Read` tool calls can fail because the tool input includes the optional `pages` field as an empty string.

Claude Code's `Read.pages` parameter is optional and only valid for PDFs with values like `"1-5"`, `"3"`, or `"10-20"`. For normal text files, `pages` should be omitted. Through this route, the emitted tool input contains `"pages": ""`, and Claude Code rejects the tool call before reading the file.

The same Claude Code setup works after switching away from 9router to CLIProxyAPI, so this looks like a 9router/provider tool-call argument normalization issue rather than a Claude Code prompt issue or local file issue.

## Steps to Reproduce
1. Configure Claude Code to use 9router as `ANTHROPIC_BASE_URL`.
2. Use a Codex/OpenAI-compatible model route such as `cx/gpt-5.5-high[1m]`.
3. Ask Claude Code to read a normal non-PDF file, for example a `.md`, `.js`, `.py`, or `.txt` file.
4. Inspect the Claude Code transcript/tool call input.
5. Observe that the `Read` tool input includes `pages` as an empty string:
   ```json
   {
     "file_path": "/tmp/example.txt",
     "offset": 0,
     "limit": 120,
     "pages": ""
   }
   ```
6. Claude Code rejects the call with:
   ```text
   Invalid pages parameter: "". Use formats like "1-5", "3", or "10-20". Pages are 1-indexed.
   ```

## Expected Behavior
For non-PDF files, optional `pages` should be omitted entirely from the `Read` tool input.

Valid examples:
```json
{"file_path":"/tmp/example.txt"}
```

or:
```json
{"file_path":"/tmp/example.txt","offset":0,"limit":120}
```

## Actual Behavior
The tool input contains an empty optional string:

```json
{"file_path":"/tmp/example.txt","offset":0,"limit":120,"pages":""}
```

Claude Code treats `pages: ""` as present-but-invalid and fails the tool call.

## Error Logs / Output
```text
Invalid pages parameter: "". Use formats like "1-5", "3", or "10-20". Pages are 1-indexed.
```

Observed in Claude Code transcript as an assistant `tool_use` block:

```json
{
  "type": "tool_use",
  "name": "Read",
  "input": {
    "file_path": "/tmp/example.txt",
    "offset": 0,
    "limit": 120,
    "pages": ""
  }
}
```

## Why this appears to be 9router-related
- The same prompt/instructions repeatedly produced `pages: ""` while routed through 9router.
- Switching the same Claude Code setup to CLIProxyAPI avoided the issue.
- Prompt-level instructions to omit optional fields did not help, which suggests the issue is in the provider/tool-call/schema path.

## Related Prior Art
A very similar bug was fixed in another codebase:

https://github.com/QwenLM/qwen-code/issues/3558

That issue describes the same failure class: some tool-calling models emit optional string parameters as `""` instead of omitting them, and the read-file tool rejects the empty `pages` value.

## Likely Root Cause
The upstream model/provider may emit optional string tool parameters as empty strings. 9router appears to pass those arguments through to Claude Code without sanitizing optional empty-string fields.

Relevant 9router code appears to forward Claude tool schemas and OpenAI tool-call arguments without removing empty optional fields:

- `open-sse/translator/request/claude-to-openai.js` maps `tool.input_schema` to OpenAI `function.parameters`.
- `open-sse/translator/response/openai-to-claude.js` forwards `tc.function.arguments` as `input_json_delta`.

## Suggested Fix
Before converting tool-call arguments back to Claude format, parse the JSON arguments and remove empty-string values for optional string properties.

Immediate targeted workaround:

```js
if (toolName === "Read" && args.pages === "") {
  delete args.pages;
}
```

More general fix:

```js
// For each tool argument:
// if schema.properties[key].type === "string"
// and key is not in schema.required
// and value === ""
// then delete args[key]
```

This preserves required empty strings if a tool explicitly needs them, while preventing optional empty-string placeholders from breaking strict tool validators.

hamsa0x7:
﻿## Summary
- Some OpenAI-compatible providers (e.g. Codex/gpt-5.5-high) emit optional string tool parameters as empty strings (`pages: ""`) instead of omitting them
- Claude Code's `Read` tool rejects `pages: ""` as invalid, breaking file reads for non-PDF files routed through 9Router
- Add `sanitizeToolArguments()` in the OpenAI→Claude response translator that strips known optional empty-string arguments before emitting `input_json_delta` back to Claude format
- Currently targets the `Read` tool's `pages` field specifically (the reported case)
- Gracefully handles streaming chunks where JSON isn't yet parseable (returns as-is)

## Test plan
- [x] New regression test in `tests/unit/openai-to-claude.test.js` — verifies `Read` tool with `pages: ""` emits clean arguments without the empty field
- [x] Existing `openai-to-claude.test.js` tests all pass
- [ ] Manual: route Claude Code through 9Router to a Codex/OpenAI-compatible model, ask it to read a `.txt` file — should succeed without `Invalid pages parameter` error
- [ ] Manual: verify PDF reads with `pages: "1-5"` still work correctly (non-empty values are preserved)

Fixes #1278

### DIFF
--- a/open-sse/translator/response/openai-to-claude.js
+++ b/open-sse/translator/response/openai-to-claude.js
@@ -4,6 +4,23 @@ import { FORMATS } from "../formats.js";
 // Prefix for Claude OAuth tool names (must match request translator)
 const CLAUDE_OAUTH_TOOL_PREFIX = "proxy_";
 
+// Strip optional empty-string tool arguments that some providers emit.
+// Claude Code's Read tool rejects pages: "" but accepts pages being absent.
+function sanitizeToolArguments(toolName, argsJson) {
+  try {
+    const args = JSON.parse(argsJson);
+    if (typeof args === "object" && args !== null) {
+      if (toolName === "Read" && args.pages === "") {
+        delete args.pages;
+      }
+      return JSON.stringify(args);
+    }
+  } catch {
+    // Not valid JSON yet (streaming chunk) — return as-is
+  }
+  return argsJson;
+}
+
 // Helper: stop thinking block if started
 function stopThinkingBlock(state, results) {
   if (!state.thinkingBlockStarted) return;
@@ -170,10 +187,11 @@ export function openaiToClaudeResponse(chunk, state) {
       if (tc.function?.arguments) {
         const toolInfo = state.toolCalls.get(idx);
         if (toolInfo) {
+          const sanitized = sanitizeToolArguments(toolInfo.name, tc.function.arguments);
           results.push({
             type: "content_block_delta",
             index: toolInfo.blockIndex,
-            delta: { type: "input_json_delta", partial_json: tc.function.arguments }
+            delta: { type: "input_json_delta", partial_json: sanitized }
           });
         }
       }
--- a/tests/unit/openai-to-claude.test.js
+++ b/tests/unit/openai-to-claude.test.js
@@ -8,6 +8,7 @@
 
 import { describe, it, expect } from "vitest";
 import { openaiToClaudeRequest } from "../../open-sse/translator/request/openai-to-claude.js";
+import { openaiToClaudeResponse } from "../../open-sse/translator/response/openai-to-claude.js";
 
 describe("openaiToClaudeRequest", () => {
   describe("response_format handling", () => {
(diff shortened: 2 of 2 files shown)


hamsa0x7:
﻿Quick reproduction note:

I tested the same route/model mentioned in the report (`cx/gpt-5.5-high`) and a normal `.txt` read did **not** fail on that run.

That suggests the issue is likely **intermittent / non-deterministic** rather than a guaranteed failure on every request. The likely failure mode still seems to be that some runs emit the optional `pages` argument as an empty string (`""`) instead of omitting it, while other runs omit the field correctly.

So at the moment the best understanding is:
- same route/model can succeed on some runs
- the bad `pages: ""` shape appears only on some tool-call generations
- this PR is meant as a defensive normalization for that intermittent invalid shape

If useful, I can also add coverage for adjacent cases (preserve valid non-empty `pages`, ensure other tools are unchanged, and inspect non-streaming paths too).

PR #1372, opened 2026-05-23
feat(qoder): port full Qoder provider integration

simonsmh:
# feat(qoder): port full Qoder provider integration

## Summary

Replaces the Qoder placeholder (`api.qoder.com` / hardcoded fake clientId+secret) with a working free-tier provider, modeled after the existing Kiro integration.

After this PR, users can connect a Qoder account via device-flow OAuth, route chat to any of 11 Qoder models, and see live quota usage in the dashboard.

### OAuth — device flow (replaces broken authorization-code placeholder)
- User clicks Connect → modal opens `qoder.com/device/selectAccounts?challenge=…&nonce=…&machine_id=…` in a new tab
- Backend polls `openapi.qoder.sh/api/v1/deviceToken/poll` with the local PKCE verifier + nonce until the upstream returns a `dt-…` token
- Token, user_id, and machine_id are persisted; refresh is a no-op (device tokens are ~30 days, upstream returns 403 to refresh attempts — users re-login when expired)

### COSY signing (`src/lib/qoder/cosy.js`)
Every chat / model-list request carries:
- AES-128-CBC payload of the user info, AES key wrapped in RSA-1024 PKCS#1 v1.5
- MD5 signature over `payload || cosyKey || timestamp || body || sigPath`
- Body MD5 hash + length so the server can validate integrity
- 17 `Cosy-*` / `X-*` headers fingerprinting the client (machine id, IDE version, organization id, etc.)

The RSA pubkey, AES IV=key reuse, and PKCS#1 v1.5 padding are dictated by the Qoder server — documented in code comments as known deviations.

### WAF-bypass body encoding (`src/lib/qoder/encoding.js`)
Qoder's CDN sits behind Alibaba Cloud WAF, which would otherwise pattern-match plaintext request bodies and return `405 Method Not Allowed`. The body is encoded as:
1. standard base64
2. split into thirds, reorder as `[tail][mid][head]`
3. substitute each char via a custom 64-char alphabet

Encoded body is sent to `…?Encode=1` so the server decodes in reverse.

### Chat executor (`open-sse/executors/qoder.js`)
- Builds the exact request shape Qoder expects: system messages hoisted out (Qoder rejects `role:"system"` in `messages`), `chat_context` mirrored from `model_config`, business block with stable IDs derived from content (so retries hit upstream caches)
- Per-model `model_config` block fetched from `/algo/api/v2/model/list` and cached for 1h — sending a wrong block silently downgrades to a different model upstream, so a missing entry is a hard error
- Unwraps Qoder's `{statusCodeValue, body}` SSE envelope back to plain OpenAI SSE chunks the rest of the chatCore pipeline understands

### Models — 11 registered
Tier models (server picks backing model):
- `qoder/auto`, `qoder/ultimate`, `qoder/performance`, `qoder/efficient`, `qoder/lite`

Frontier models (pinned backing):
- `qoder/qmodel` (Qwen 3.6 Plus)
- `qoder/dmodel` (DeepSeek V4 Pro)
- `qoder/dfmodel` (DeepSeek V4 Flash)
- `qoder/gm51model` (GLM 5.1)
- `qoder/kmodel` (Kimi K2.6)
- `qoder/mmodel` (MiniMax M2.7)

### Dashboard wiring
- Provider page renders the standard `OAuthModal` for Qoder (no custom modal needed — single auth method)
- Live-models resolver registered in `/api/providers/[id]/models` and `/api/v1/models`
- Provider test util hits `openapi.qoder.sh/api/v1/userinfo` to validate stored token
- Quota fetcher added to `usage.js` reads `openapi.qoder.sh/api/v2/quota/usage` (Bearer-only, no COSY)
- New 256×256 PNG icon at `public/providers/qoder.png`

### Tests
24 unit tests in `tests/unit/qoder.test.js` covering:
- body encoder boundary cases (empty, lengths divisible/not divisible by 3, alphabet correctness, determinism)
- PKCE pair (S256 challenge derivation, base64url-safe charset, length)
- device flow URL construction (qoder.com/device/selectAccounts, nonce/machine_id propagation)
- regression guard for `flow.codeVerifier` field name (a bug found and fixed during review)
- COSY signing: all 17 headers, `Authorization` shape, sigPath strips `/algo`, body MD5/length, machineId propagation, missing-creds throws
image
image


simonsmh:
@decolua Hi, please review!

decolua:
Thanks so much for this excellent contribution, @simonsmh! The device-flow OAuth, COSY signing, WAF-bypass encoding, live model catalog, dashboard quota wiring, and the test coverage are all really well done and clearly documented.

We've cherry-picked and merged the Qoder provider integration. Closing this PR as merged — much appreciated! 🙏

PR #1395, opened 2026-05-24
Fallback on empty tool-heavy combo streams

### LINKED ISSUE #1382: [Bug] DeepSeek backend returns ~0 output tokens for tool-heavy claude→openai requests; GLM also 13% empty
## Environment

- 9router version: **0.4.59** (docker `decolua/9router:latest`)
- Client: `claude-code` CLI (`claude --print` with tool definitions)
- Format pipeline: claude → openai (for deepseek) / claude → claude (for GLM, MiniMax)

## Summary

When a claude-code request carries tool definitions (10 tools is the typical case), the response body returned to the client is empty/malformed:

```
API Error: API returned an empty or malformed response (HTTP 200)
— check for a proxy or gateway intercepting the request
```

9router itself logs `Model X succeeded` and `STREAM complete`, so the failure is silent on the gateway side. Multiple backends are affected; DeepSeek is broken 100% of the time, GLM-5.1 leaks ~13%.

## Measurements (5,000 lines of my own production logs)

| Backend | Calls | `out < 50 tokens` (≈ empty) | avg input | avg output |
|---------|-------|----------------------------|-----------|------------|
| ds/deepseek-v4-pro-max | 4   | **4 / 4 = 100 %** | 9 983 | 17 |
| glm/glm-5.1           | 212 | **29 / 212 = 13 %** | 1 856 | 952 |
| minimax-cn/MiniMax-M2.7-highspeed | 52 | 0 / 52 = 0 % | 401 | 165 |

DeepSeek output:input ratio is **0.17 %** — essentially nothing comes back. Sample log:

```
[04:46:05] POST /v1/messages | test | 1 msgs | 10 tools
[04:46:05] [COMBO] Trying model 1/3: ds/deepseek-v4-pro-max
[04:46:05] [ROUTING] ds/deepseek-v4-pro-max → deepseek/deepseek-v4-pro-max
[04:46:05] [FORMAT] claude → openai | stream=true
[04:46:05] [REQUEST] DEEPSEEK | deepseek-v4-pro-max | 2 msgs
[04:46:05] [COMBO] Model ds/deepseek-v4-pro-max succeeded         ← 误报成功
[04:46:07] [PENDING] END | provider=deepseek
[04:46:07] [USAGE] DEEPSEEK | in=19882 | out=25                   ← 输出 25 token
[04:46:08] [STREAM] DEEPSEEK | 2445ms | complete                  ← gateway 标完成
```

Client side simultaneously: `is_error: true`, `result: "API returned an empty or malformed response (HTTP 200)"`.

## Reproducer

```bash
# Sanity: same backend WITHOUT tools — works fine, streams normally
  -H 'anthropic-version: 2023-06-01' \
  -d '{"model":"<your-deepseek-combo>","max_tokens":50,
       "messages":[{"role":"user","content":"reply ok"}]}'

# Trigger: same backend WITH tools — gateway logs success, body is empty
  -H 'anthropic-version: 2023-06-01' \
  -d '{"model":"<your-deepseek-combo>","max_tokens":2048,
       "messages":[{"role":"user","content":"echo ok"}],
       "tools":[
         {"name":"Bash","description":"run shell","input_schema":{"type":"object","properties":{"command":{"type":"string"}},"required":["command"]}},
         {"name":"Read","description":"read file","input_schema":{"type":"object","properties":{"path":{"type":"string"}},"required":["path"]}}
       ]}'
```

## Likely related

Same class of bug as several open translator issues:

- #1371 — `custom_tool_call` lost in translator for `/v1/responses`
- #1368 — antigravity tool schema `required` field stripped
- #1321 — thinking-model `reasoning_content` not passed back

All three are translator content loss. This report adds another data point: **deepseek + tools via claude→openai translation returns ~zero useful content**, and the gateway falsely reports success.

## Suggested fixes

1. When upstream's `out_tokens < some_threshold` AND request contained tools AND no tool_call emitted, treat the call as failed and **fall through to the next model in the combo** instead of reporting success.
2. Surface a configurable "empty body" detector at the gateway level, returning HTTP 502/503 to the client so claude CLI can apply its built-in retry (currently CLI does not retry on HTTP 200 with empty body).
3. Document which backends are known to mishandle tool definitions, so users can keep them out of tool-heavy combos.

## Workaround

For tool-heavy workloads (claude-code, codex), avoid DeepSeek in combos; prefer providers that handle native `claude → claude` format (GLM, MiniMax). Even then, 13 % empty rate on GLM means app-level retry is required.

Delcado19:
## Summary
- inspect successful combo SSE responses only when the original request contains tools
- fall through to the next combo model when a 200 SSE stream has no client-visible text, reasoning, or tool-call output
- keep normal no-tool streams untouched and keep tool streams that emit tool calls

Fixes #1382.

## Validation
- `node --check open-sse/services/combo.js`
- `npx vitest run --config ./tests/vitest.config.js --reporter=verbose tests/unit/combo-routing.test.js` (7 passed)
- `npm run build`
- `git diff --check`

## Notes
- Scope is intentionally limited to combo fallback. Direct single-model requests still return the provider response; this avoids buffering every streaming request globally.
- The stream inspector reads only until the first visible output/tool-call, then releases the cloned stream. It reads to the end only for streams that are actually empty/malformed.

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -5,6 +5,8 @@
 import { checkFallbackError, formatRetryAfter } from "./accountFallback.js";
 import { unavailableResponse } from "../utils/error.js";
 
+const EMPTY_TOOL_STREAM_STATUS = 502;
+
 /**
  * Track rotation state per combo (for round-robin strategy)
  * @type {Map<string, { index: number, consecutiveUseCount: number }>}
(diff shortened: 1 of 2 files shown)


zierocode:
I added a draft executable companion contract in #2655. It preserves the useful empty/tool-heavy cases here and adds the missing single-reader, byte-bound, first-action timeout/cancellation, exact replay, and no-post-commit-fallback boundaries. The intent is to converge on one stream owner rather than compete with this PR.

Fresh cross-check on 2026-07-23:

- upstream base: `79918c7830695bbca4a45c9fea4a42c3e9fd73d1` (`v0.5.40`)
- this PR head: `72178e9faeef705db4c85664004e0132342df004`
- this PR's focused suite: 7/7 passed
- applying the two test-only commits from #2655 to vanilla: 2 passed, 10 expected failures
- applying the same contract to this PR: 2 previously expected failures now pass, 2 baseline tests pass, and 8 expected failures remain

The two newly satisfied cases are:

1. fallback when a successful tool-bearing SSE response terminates before an action;
2. withholding the response until the first action while replaying the exact prefix once.

Reproduction:

```bash
NODE_PATH=/tmp/node_modules /tmp/node_modules/.bin/vitest run \
  --reporter=verbose --config tests/vitest.config.js \
  tests/unit/combo-preaction-streaming.contract.test.js
```

This confirms that #1395 is the correct implementation owner for its current scope. The broader timeout, ownership, cancellation, action-classification, and non-LLM boundaries in #2655 still need convergence before the production compatibility guard can be removed.

PR #1401, opened 2026-05-24
Fix /v1/messages non-streaming JSON mode

### LINKED ISSUE #1396: [Bug] /v1/messages stream=false returns SSE-tainted JSON body (data: [DONE] appended)
## Environment
- 9router v0.4.59 (Docker: `decolua/9router:latest`)
- Combo: "opus" → GLM (glm-5.1) as first fallback
- Client: Claude Code CLI (`claude --print`)

## Problem

When a client sends `stream=false` to `/v1/messages`, the response body is valid JSON **with `data: [DONE]` appended directly after it**, making the entire body unparseable as JSON:

```
HTTP/1.1 200 OK
Content-Type: text/event-stream    ← should be application/json

{"id":"msg_...","type":"message","role":"assistant","model":"glm-5.1","content":[{"type":"text","text":"Hi there!"}],"stop_reason":"end_turn",...}data: [DONE]
```

The `data: [DONE]` SSE terminator is concatenated without any separator. `JSON.parse()` fails with `Unexpected token 'd'`.

Claude Code CLI reports: **"API Error: API returned an empty or malformed response (HTTP 200)"**

## Steps to Reproduce

```bash
# Non-streaming request to /v1/messages (Anthropic format)
curl -s -X POST http://localhost:20128/v1/messages \
  -d '{"model":"opus","max_tokens":50,"messages":[{"role":"user","content":"say hi"}]}'

# Response body (note the trailing "data: [DONE]"):
# {"id":"msg_...","content":[...],...}data: [DONE]
```

Streaming (`"stream":true`) works correctly — returns proper SSE format.

## Expected Behavior

For `stream=false`, response should be:
- Body: pure JSON object, no SSE markers

## Impact

This breaks all non-streaming `/v1/messages` calls. Claude Code CLI makes both streaming and non-streaming API calls internally (e.g., haiku session initialization uses `stream=false`). When the non-streaming call fails, the entire session dies.

## Workaround

Local proxy that strips `data: [DONE]\s*$` from non-streaming responses.

## Related

- #1025 — same error message with NVIDIA NIM provider
- Docs mention "SSE→JSON fallback for non-streaming calls" in local runtime, but it doesn't appear to work for `/v1/messages` (Anthropic format) — possibly only implemented for `/v1/chat/completions` (OpenAI format)?

Delcado19:
## Summary
- keep Claude-format /v1/messages non-streaming by default unless the client explicitly requests SSE
- preserve existing explicit stream:true and Accept: text/event-stream behavior
- add focused stream-mode regression coverage for #1396

Fixes #1396

## Tests
- npx vitest run --config ./tests/vitest.config.js tests/unit/chat-stream-mode.test.js
- node --check open-sse/handlers/chatCore.js
- git diff --check
- npm run build

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -20,6 +20,43 @@ import { dedupeTools } from "../utils/toolDeduper.js";
 import { injectCaveman } from "../rtk/caveman.js";
 import { compressMessages, formatRtkLog } from "../rtk/index.js";
 
+function getHeaderValue(headers, name) {
+  if (!headers) return "";
+  if (typeof headers.get === "function") return headers.get(name) || "";
+  const lowerName = name.toLowerCase();
+  for (const [key, value] of Object.entries(headers)) {
+    if (key.toLowerCase() === lowerName) return String(value || "");
+  }
+  return "";
+}
+
+export function resolveChatStreamMode({ body = {}, sourceFormat, provider, clientHeaders = {}, clientTool = null }) {
+  const acceptHeader = getHeaderValue(clientHeaders, "accept").toLowerCase();
+  const clientPrefersJson = acceptHeader.includes("application/json");
+  const clientPrefersSSE = acceptHeader.includes("text/event-stream");
+  const formatAlwaysStreams = sourceFormat === FORMATS.ANTIGRAVITY || sourceFormat === FORMATS.GEMINI || sourceFormat === FORMATS.GEMINI_CLI;
+  const clientRequestedStreaming = body.stream === true || (body.stream !== false && clientPrefersSSE) || formatAlwaysStreams;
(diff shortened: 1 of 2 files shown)


heishen6:
Hi — I'm the reporter of #1396. This bug is blocking my Claude Code CLI sessions (non-streaming /v1/messages calls fail on every haiku init). The workaround (local proxy stripping `data: [DONE]`) works but is fragile.

This PR looks clean and focused. Could a maintainer review and merge when possible? Thanks! 🙏

the-long-ride (approved):

audacioustux:
hi, just a follow-up... is this going to be merged soon?

the-long-ride:
I have seen a some refactoring commit were pushed by maintainer last night, so this PR has some conflicts already

PR #1416, opened 2026-05-25
fix: add Codex subagent role description

nguyenha935:
## Summary
- add the required Codex CLI agent role description for agents.subagent
- reuse that role shape in the Codex settings generator
- update the dashboard manual config preview and parser to handle the new description line

## Tests
- ./node_modules/.bin/vitest run --reporter=verbose --config ./tests/vitest.config.js tests/unit/codex-settings.test.js
- npm run build

### DIFF
--- a/src/app/(dashboard)/dashboard/cli-tools/components/CodexToolCard.js
+++ b/src/app/(dashboard)/dashboard/cli-tools/components/CodexToolCard.js
@@ -58,7 +58,7 @@ export default function CodexToolCard({ tool, isExpanded, onToggle, baseUrl, api
       if (modelMatch) setSelectedModel(modelMatch[1]);
 
       // Parse subagent settings
-      const subagentModelMatch = codexStatus.config.match(/\[agents\.subagent\]\s*\n\s*model\s*=\s*"([^"]+)"/m);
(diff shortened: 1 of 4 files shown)


nguyenha935:
This still applies to current master. The Codex settings generator still writes agents.subagent with only model, while Codex requires agent roles to include a description. This PR keeps the generated config valid and adds a small unit test for the role shape.

tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

The Codex subagent description is extracted into a shared builder (`buildCodexSubagentRole`) so role text lives in one place. The regex for parsing the INI-style `[agents.subagent]` block is also hardened against multiline config sections.

- `src/app/api/cli-tools/codex-settings/config.js` — added `buildCodexSubagentRole(model)` returning `{ description, model }`.
- `CodexToolCard.js` — regex uses `[\s\S]*?` to span multiline config sections instead of only matching within a single line.
- `tests/unit/codex-settings.test.js` — new unit test guards the role build output.

### ✅ Looks Good
- Single-responsibility refactor with a focused change set.
- Regex fix directly addresses the parsing corner case.
- Test coverage for the new helper.

Issue #1408, opened 2026-05-25
[Bugs] New Codex Update make 9router "break"


nhanwork:
Timeout all codex:

failed" retry 2/3 after 3s
[09:33:33] 🔍 [RETRY] network "fetch failed" retry 1/3 after 3s
[09:34:21] 🌊 [STREAM] CODEX | gpt-5.5 | 789865ms | error: stream stall timeout
[09:34:21] [PENDING] END | provider=codex | model=gpt-5.5
[09:36:38] 🌊 [STREAM] CODEX | gpt-5.5 | 486032ms | error: stream stall timeout
[09:36:38] [PENDING] END | provider=codex | model=gpt-5.5
[09:38:59] 🌊 [STREAM] CODEX | gpt-5.5 | 464302ms | error: stream stall timeout
[09:38:59] [PENDING] END | provider=codex | model=gpt-5.5

gede-cahya:
same problem bro

failed" retry 2/3 after 3s
[09:33:33] 🔍 [RETRY] network "fetch failed" retry 1/3 after 3s

Henrywalker2002:
same problem 
[12:22:55] 🔍 [REQUEST] CODEX | gpt-5.5 | 154 msgs
[12:27:57] 🔍 [RETRY] network "fetch failed" retry 1/3 after 3s

quanla93:
The official issue codex#24260, opened on 2026-05-24 — right around the same time you encountered the problem — reports the exact same pattern:

The official Codex CLI v0.133.0 with gpt-5.5 submits the turn successfully, the request goes through, but the stream stays silent for 380s / 671s / 984s before the first token appears. The logs show time.busy=ms while time.idle lasts for hundreds of seconds. This happens on both gpt-5.5 and gpt-5.4, and affects both Desktop and CLI. If you interrupt and resubmit, it usually completes within a few seconds.

nhanwork:
Fixed with # v0.4.62

PR #1418, opened 2026-05-25
fix(codex): abort stalled initial upstream responses

rama1933:
## Summary
- add a Codex-specific initial upstream response deadline while `executor.execute()` is still waiting for `fetch()` to return
- default the deadline to 7 seconds, with `CODEX_INITIAL_RESPONSE_TIMEOUT_MS=0` available to disable it
- return this deadline as HTTP 504 so the existing account and combo fallback paths rotate away from a stalled Codex request
- document the environment setting and add executor regression coverage for deadline aborts and caller abort preservation

## Why this is separate from #576
PR #576 applies a TTFT deadline after `providerResponse` exists and races the first stream chunk. This PR covers the earlier failure mode where the Codex upstream request has not returned an initial response at all, so there is no response body to race yet.

This behavior was observed while investigating #1408: Codex calls could remain pending before an initial upstream response, keeping the account selected rather than allowing fallback to proceed.

## Verification
- `./node_modules/.bin/vitest run --config tests/vitest.config.js tests/unit/codex-image-fetch.test.js tests/unit/combo-routing.test.js --reporter=verbose`
- local runtime validation with an intentionally shortened timeout: two stalled Codex accounts returned 504 and an existing combo proceeded to its next provider, which returned HTTP 200
- local usage validation with the 7000 ms default: two stalled Codex attempts were aborted at 7022 ms and 7133 ms; the subsequent attempts on alternate Codex connections completed in 3218 ms and 3402 ms

Related to #1408. Complementary to #576.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,6 +9,8 @@ DATA_DIR=/var/lib/9router
 # Recommended runtime variables
 PORT=20128
 NODE_ENV=production
+# Initial Codex upstream response deadline in milliseconds (0 disables timeout)
+CODEX_INITIAL_RESPONSE_TIMEOUT_MS=7000
 
 # Recommended security and ops variables
 API_KEY_SECRET=endpoint-proxy-api-key-secret
--- a/README.md
+++ b/README.md
@@ -1102,6 +1102,7 @@ docker pull decolua/9router:latest   # update to latest
 | `PORT` | framework default | Service port (`20128` in examples) |
 | `HOSTNAME` | framework default | Bind host (Docker defaults to `0.0.0.0`) |
 | `NODE_ENV` | runtime default | Set `production` for deploy |
+| `CODEX_INITIAL_RESPONSE_TIMEOUT_MS` | `7000` | Abort a Codex request that has not returned an initial upstream response within this time; set `0` to disable |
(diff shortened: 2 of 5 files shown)


gede-cahya:
image
still same

gede-cahya:
I changed it to CODEX_INITIAL_RESPONSE_TIMEOUT_MS=0 and it worked.

afu-it:
Tested on Docker deployment (decolua/9router:latest)

Applied the patch manually to the running container:
- open-sse/executors/codex.js — timeout logic working
- open-sse/handlers/chatCore.js — 504 error handling working
- Set CODEX_INITIAL_RESPONSE_TIMEOUT_MS=7000 in env

Note on test file: The two new test cases in codex-image-fetch.test.js fail with vitest v4.1.7 (ESM fake timer issue — vi.useFakeTimers() does not intercept setTimeout in ESM modules properly in vitest 4). Tests pass with a workaround but not 100% as written. May need vi.useFakeTimers({ toFake: [...] }) adjustment for vitest 4 compatibility.

Workaround: Setting CODEX_INITIAL_RESPONSE_TIMEOUT_MS=0 also resolves the stall as noted by others, but disables the fallback rotation benefit entirely.

PR #1424, opened 2026-05-25
Guard MCP routes to localhost

### LINKED ISSUE #1114: Security Note for Maintainers
Issue: MCP custom plugin routes (/api/mcp/*) can spawn local processes and were reachable when requireLogin is disabled or if middleware is bypassed, leading to unauthenticated RCE.
Fix: Enforce localhost-only access at the route handler level for /api/mcp/[plugin]/sse and /api/mcp/[plugin]/message in addition to middleware checks. This provides defense‑in‑depth in case proxy/middleware configuration changes or new routes are added.

Impact: Remote attackers could interact with MCP stdio plugins and potentially execute arbitrary commands via plugin arguments or payloads.

Mitigation: Always require loopback Host/Origin for MCP routes; reject non-local requests with HTTP 403.

✅ Test Case Suggestion (minimal)
Add tests that ensure MCP endpoints reject non‑localhost requests even when authentication is disabled.

Example (pseudo-test)
Arrange: set requireLogin=false in settings or mock settings read.
Act: call GET /api/mcp/{plugin}/sse and POST /api/mcp/{plugin}/message with:
Host: example.com
Origin: https://example.com
Assert: response is 403 with { error: "Local only: MCP requires localhost access" }
Optional positive case
Act: same endpoints with:
Host: localhost
(no Origin or Origin: http://localhost/)
Assert: not 403 (should be 200/202 depending on route)

Delcado19:
## Summary
- enforce localhost-only Host/Origin checks inside the MCP SSE and message route handlers
- reject forwarded remote hosts before plugin lookup, session registration, or child-process writes
- add route-level regression coverage alongside the existing dashboard guard tests

Fixes #1114

## Tests
- npx vitest run --config tests/vitest.config.js tests/unit/mcp-local-only.test.js tests/unit/dashboard-guard.test.js --reporter=verbose
- node --check src/lib/security/localOnly.js
- node --check tests/unit/mcp-local-only.test.js
- node --check src/app/api/mcp/[plugin]/sse/route.js
- node --check src/app/api/mcp/[plugin]/message/route.js
- git diff --check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# Unreleased
+
+## Fixes
+- MCP: enforce localhost-only access inside the SSE and message route handlers as defense in depth for local stdio plugins (#1114).
+
 # v0.4.59 (2026-05-21)
 
 ## Fixes
--- a/src/app/api/mcp/[plugin]/message/route.js
+++ b/src/app/api/mcp/[plugin]/message/route.js
@@ -1,10 +1,15 @@
 import { NextResponse } from "next/server";
 import { sendToChild, findPlugin } from "@/lib/mcp/stdioSseBridge";
+import { isLocalOnlyRequest, localOnlyJsonError } from "@/lib/security/localOnly";
 
 export const runtime = "nodejs";
 export const dynamic = "force-dynamic";
 
 export async function POST(request, { params }) {
+  if (!isLocalOnlyRequest(request)) {
+    return localOnlyJsonError("Local only: MCP requires localhost access");
+  }
+
   const { plugin } = await params;
   if (!findPlugin(plugin)) {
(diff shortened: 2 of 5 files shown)


Delcado19:
@
Closing as superseded by the current master behavior.

Both MCP routes (`message/route.js`, `sse/route.js`) now return an unconditional `403 "Cowork is disabled"` *before* any handler logic, since the stdio bridge can spawn arbitrary processes (RCE risk). That fully neutralizes the attack surface this PR targeted — the localhost guard (`isLocalOnlyRequest`) would sit behind the early `return` and never execute, i.e. dead code.

So the underlying concern is already addressed, just more aggressively than a localhost-only guard.

Caveat for the future: if Cowork/MCP is ever re-enabled, a localhost-only guard like the one here (`src/lib/security/localOnly.js`) becomes relevant again. Happy to revive this PR at that point.
@

Issue #1343, opened 2026-05-21
[Bug] json_schema response_format rejected with 400 Invalid JSON body


Xpos587:
## Updated Finding

**9Router DOES pass `json_schema` response_format through** — the initial 400 errors I saw were due to curl shell-escaping issues, not 9Router rejecting the request. I apologize for the inaccurate initial report.

The real issue is more nuanced:

### Actual Problem

When `json_schema` response_format is sent through 9Router to models that don't natively support Structured Output (e.g., DeepSeek via Ollama), the model returns empty or malformed content instead of following the schema. This is expected model behavior.

However, 9Router already has a solution for this in the **OpenAI to Claude translator** — when targeting Claude, `json_schema` is converted to a system prompt injection:

```javascript
// open-sse/translator/request/openai-to-claude.js (lines 108-121)
if (responseFormat.type === "json_schema" && responseFormat.json_schema?.schema) {
  const schemaJson = JSON.stringify(responseFormat.json_schema.schema, null, 2);
  systemParts.push(`You must respond with valid JSON that strictly follows this JSON schema:
\`\`\`json
${schemaJson}
\`\`\`
Respond ONLY with the JSON object, no other text.`);
}
```

This same approach should be applied in the **DefaultExecutor** (for OpenAI-compatible providers like Ollama, Wafer, etc.) as a fallback: when the target model likely doesn't support `json_schema` natively, inject the schema into the system prompt and downgrade `response_format` to `{"type": "json_object"}`.

### Evidence

```bash
# json_schema through 9Router to ollama/deepseek-v4-flash: model returns empty content
{"choices":[{"message":{"content":""},"finish_reason":"length"}]}

# json_object through 9Router to ollama/deepseek-v4-flash: model returns valid JSON
{"choices":[{"message":{"content":"{\"extracted_entities\": []}"}}]}
```

### Proposal

Add a fallback in the DefaultExecutor (or OpenAI-compatible request path) that:
1. Detects `response_format.type === "json_schema"`
2. Injects the JSON schema into the system prompt (same pattern as Claude translator)
3. Replaces `response_format` with `{"type": "json_object"}`

This would make Structured Output work transparently across all providers, not just Claude.

### Why This Matters

[...]

PR #1432, opened 2026-05-25
fix: implement json_schema fallback for OpenAI-compatible providers (#1343)

mustafabozkaya:
This PR addresses Issue #1343 by adding a fallback mechanism for json_schema in the DefaultExecutor.

### Problem
When json_schema response format is sent to models that do not natively support Structured Output (e.g., DeepSeek via Ollama), they often return empty or malformed content.

### Solution
Inspired by the Claude translator, this PR:
1. Detects 
esponse_format.type === 'json_schema'.
2. Injects the JSON schema into the system prompt.
3. Downgrades the 
esponse_format to json_object.

This ensures Structured Output works transparently across all OpenAI-compatible providers.

### Changes
- Modified open-sse/executors/default.js to handle the transformation.
- Added unit tests in 	ests/unit/default-executor.test.js to verify the logic.

Verified with itest.

### DIFF
--- a/open-sse/executors/default.js
+++ b/open-sse/executors/default.js
@@ -12,7 +12,33 @@ export class DefaultExecutor extends BaseExecutor {
   }
 
   transformRequest(model, body) {
-    return injectReasoningContent({ provider: this.provider, model, body });
+    let transformedBody = { ...body };
+
+    // Handle json_schema response_format fallback for OpenAI-compatible providers
+    // (Models that don't natively support Structured Output will fail if passed json_schema)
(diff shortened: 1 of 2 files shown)


decolua:
Thanks for the contribution! The fix has been adapted and merged into master (commit `30e1497`) with a small adjustment: the fallback is gated to `provider.startsWith("openai-compatible-")` to avoid downgrading providers that natively support Structured Output (OpenAI, Gemini, Claude, GLM, Kimi, MiniMax, etc.). Closing this PR — your authorship is preserved in the commit. 🙏

PR #1458, opened 2026-05-26
fix: validate HuggingFace tokens via whoami

Delcado19:
## Summary
- Validate HuggingFace API keys against https://huggingface.co/api/whoami-v2 instead of task/model media probes
- Apply the same auth-only check to the provider connection test path
- Avoid marking valid fine-grained Inference Provider tokens invalid when model/task endpoints are unavailable or incompatible

## Validation
- node --check src/app/api/providers/validate/route.js
- node --check src/app/api/providers/[id]/test/testUtils.js
- git diff --check

Notes: I did not find an existing HuggingFace validation issue or PR upstream before opening this.

### DIFF
--- a/src/app/api/providers/[id]/test/testUtils.js
+++ b/src/app/api/providers/[id]/test/testUtils.js
@@ -17,6 +17,8 @@ import {
 } from "@/lib/oauth/constants/oauth";
 import { buildClineHeaders } from "@/shared/utils/clineAuth";
 
+const HUGGINGFACE_WHOAMI_URL = "https://huggingface.co/api/whoami-v2";
+
 // OAuth provider test endpoints
 const OAUTH_TEST_CONFIG = {
   claude: { checkExpiry: true, refreshable: true },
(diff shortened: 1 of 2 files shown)


yicone:
I hit the same HF token-validation problem locally, and this PR matches the behavior I needed.

One small note that may be useful for follow-up scope (not necessarily for this PR): on current `master`, HuggingFace still seems to have two adjacent gaps beyond token validation itself:

- `/api/providers/[id]/test` can still fall through to `Provider test not supported` for `huggingface` because the connection-test switch does not appear to include an HF case yet.
- the runtime/media config still points at the legacy `api-inference.huggingface.co/models` base in a few places, while current HF provider docs and runtime behavior now center around `router.huggingface.co/hf-inference/...`.

So from my testing, the whoami-based validation here is the right first fix for the false-`Invalid` problem, even if a small HF follow-up may still be needed later for connection test/runtime endpoint parity.

PR #1460, opened 2026-05-26
Preserve reasoning effort for Codex translations

### LINKED ISSUE #1459: Codex translations drop Claude Code reasoning effort
## Problem

Claude Code requests sent through the Claude Messages endpoint can carry an explicit reasoning effort, but when the request is routed to Codex the effort is lost during format translation.

Observed flow from logs:

```text
POST /v1/v1/messages | gpt-5.5
[FORMAT] claude → openai-responses | stream=true
[ROUTING] cx/gpt-5.5 → codex/gpt-5.5
```

The Claude → OpenAI translator currently keeps messages, tools, max tokens, temperature, and tool choice, but does not preserve `reasoning_effort` or `reasoning.effort`. The OpenAI Chat → Responses translator also does not map `reasoning_effort` into Responses `reasoning.effort`.

As a result, Codex receives no explicit reasoning config and falls back in `open-sse/executors/codex.js` to its default effort instead of the client-requested level such as `xhigh`.

## Expected behavior

Explicit reasoning effort should survive the full translation path:

```text
Claude request reasoning.effort / reasoning_effort
→ OpenAI Chat reasoning_effort
→ OpenAI Responses reasoning.effort
→ Codex executor
```

## Impact

Claude Code users who configure/request higher reasoning effort through 9router can silently get lower-effort Codex responses, making debugging confusing because routing succeeds but effort does not match the client configuration.

ntdung6868:
## Summary
- Preserve explicit `reasoning_effort` / `reasoning.effort` when translating Claude-format requests into OpenAI Chat format.
- Map OpenAI Chat `reasoning_effort` into OpenAI Responses `reasoning.effort` so Codex receives the requested effort instead of falling back to its default.

Fixes #1459

## Test plan
- [x] Verified with a Node import smoke test that Claude `reasoning.effort: \"xhigh\"` becomes OpenAI Chat `reasoning_effort: \"xhigh\"` and then Responses `reasoning.effort: \"xhigh\"`.
- [x] Confirmed PR diff against upstream `master` only includes the translator fix.

### DIFF
--- a/open-sse/translator/request/claude-to-openai.js
+++ b/open-sse/translator/request/claude-to-openai.js
@@ -70,6 +70,16 @@ export function claudeToOpenAIRequest(model, body, stream) {
     result.tool_choice = convertToolChoice(body.tool_choice);
   }
 
+  if (body.reasoning_effort !== undefined) {
+    result.reasoning_effort = body.reasoning_effort;
+  } else if (body.reasoning?.effort !== undefined) {
+    result.reasoning_effort = body.reasoning.effort;
+  }
+
+  if (body.reasoning !== undefined) {
+    result.reasoning = body.reasoning;
+  }
+
   return result;
 }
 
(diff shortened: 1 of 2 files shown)


punal100:
Same root cause on translation path. My [pull request](https://github.com/decolua/9router/pull/2096) keeps `reasoning_effort` / `reasoning.effort` intact through Claude -> OpenAI Chat -> OpenAI Responses, so Codex receives requested effort instead of defaulting.

This should cover the OpenAI/Codex issue from my side as well.

PR #1463, opened 2026-05-26
fix(cursor): resolve cu/default empty responses

### LINKED ISSUE #1077: Cursor doesn't respond with default
Image
Image
Image

Duycan17:
Fixes #1077

## Summary

- Fix `cu/default` returning 0 output tokens by promoting visible text from thinking protobuf frames (not only Composer models).
- Resolve `default`/`auto` to a concrete Cursor upstream model (`claude-4.5-sonnet`, overridable via `CURSOR_DEFAULT_UPSTREAM_MODEL`).
- Force Agent mode for `default`/`auto` and map bare `default`/`auto` model aliases to the `cursor` provider.
- Return `502 empty_completion` when Cursor yields no text and no tool calls (instead of silent empty 200).
- Dashboard model ping: use `stream: true` for `cu/*` models and validate non-empty content.
- Add `cursorModel` helpers with legacy id normalization and unit tests.

## Root cause

Cursor Auto often responds with thinking-only frames. The executor only surfaced thinking text for Composer models, so clients (Claude Code, dashboard) saw empty completions with input tokens billed.

## Test plan

- [x] `tests/unit/cursor-default.test.js`
- [x] `tests/unit/cursor-composer-thinking.test.js`
- [ ] Dashboard: ping `cu/default` → non-zero output tokens
- [ ] `curl` / Claude Code: `claude --model cu/default "hi"` returns text
- [ ] `cu/kimi-k2.5` still works

Made with [Cursor](https://cursor.com)

### DIFF
--- a/open-sse/executors/cursor.js
+++ b/open-sse/executors/cursor.js
@@ -7,6 +7,12 @@ import {
   extractTextFromResponse
 } from "../utils/cursorProtobuf.js";
 import { buildCursorHeaders } from "../utils/cursorChecksum.js";
+import {
+  normalizeCursorModelId,
+  resolveCursorUpstreamModel,
+  shouldPromoteThinkingToContent,
+  visibleContentFromThinking
+} from "../utils/cursorModel.js";
 import { estimateUsage } from "../utils/usageTracking.js";
 import { FORMATS } from "../translator/formats.js";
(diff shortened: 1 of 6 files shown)


SscSPs:
Why isn't this merged btw?

PR #1488, opened 2026-05-27
fix responses max_tokens mapping

### LINKED ISSUE #1489: OpenAI Responses translator should map max_tokens to max_output_tokens
## Problem

When integrating 9router with sub2api-style OpenAI-compatible clients, requests may send the token limit as `max_tokens`. The OpenAI Responses API expects the equivalent field to be `max_output_tokens`, so forwarding `max_tokens` directly can make Responses requests incompatible.

## Expected behavior

- If `max_output_tokens` is provided, preserve it.
- Otherwise, map OpenAI-compatible `max_tokens` to `max_output_tokens`.
- Do not forward `max_tokens` in the Responses request payload.

## Related PR

A fix is proposed in #1488.

dangzitou:
## Summary
- Map OpenAI-compatible `max_tokens` to Responses API `max_output_tokens`.
- Preserve explicit `max_output_tokens` when provided.
- Make CLI standalone build detection handle nested standalone app output.

Fixes #1489.

## Context
This came up when integrating 9router with sub2api-style OpenAI-compatible clients that send `max_tokens`.

## Test plan
- Added unit coverage for `openaiToOpenAIResponsesRequest` mapping `max_tokens` to `max_output_tokens`.
- Attempted to run `npm --prefix /Users/ann/Desktop/workspace/9router/tests test -- ../tests/unit/translator-request-normalization.test.js`, but the local test script depends on `/tmp/node_modules/.bin/vitest`, which is missing in this environment.


kacipbuah:
Friendly maintainer ping 🙏

This token mapping fix for responses API seems broadly impactful for OpenAI-compatible flows. If possible, please review when you have a slot.

cc @decolua

Benson-mk:
Hi maintainers, could you please review and merge this PR when possible?

It fixes max_tokens being dropped instead of mapped to max_output_tokens for Responses API requests.

It would be great to get this merged into the next release. Thanks!

dangzitou:
Rebased onto current `master` (commit `8488b84`) and resolved the conflict. The branch now contains only the focused Responses API mapping plus its regression test: `max_output_tokens` is preserved, otherwise `max_tokens` is mapped, and `max_tokens` is not forwarded.

`git diff --check` passes. I could not run the Vitest command in this environment because Node/npm is not installed.

KarmicDaoist:
@decolua ping — this is still the live Chat→Responses bug on `0.5.55` / `latest` (no newer image).

`openaiToOpenAIResponsesRequest` still does `result.max_tokens = body.max_tokens`. OpenCode (and other Responses-only upstreams) then 400:

```
unknown parameter `max_tokens`
```

They want `max_output_tokens`. Cursor and other OpenAI-compatible clients send `max_tokens` on `/v1/chat/completions`, so every Muse Spark / Codex-style Responses hop through 9router fails unless the client omits the field.

This PR already has the mapping + a regression test. `master` still has the passthrough. Please merge so we do not have to patch running images.

Repro: Chat Completions request with `max_tokens` to a custom OpenAI-compatible node whose API type is Responses (e.g. `https://opencode.ai/zen/go/v1`). Same request without `max_tokens` succeeds.

dangzitou:
Rebased the PR branch onto current `master` (now at `90b52e0`) and resolved the conflict in the token-limit mapping while preserving upstream `max_completion_tokens` handling. The PR now carries the original regression test and the updated source change. Please let GitHub refresh mergeability and review when convenient.

dangzitou:
Local validation after rebase: the added `openaiToOpenAIResponsesRequest maps max_tokens` test passes (1 passed). The broader translator normalization file currently has four pre-existing failures on current `master` (Claude text flattening and raw NDJSON expectations), unrelated to this PR's token-limit mapping.

PR #1499, opened 2026-05-27
feat: add param compatibility adaptation for upstream API

kenlin8827:
## Problem

Third-party providers often don't support new OpenAI parameters (e.g., max_completion_tokens, reasoning_effort, thinking), causing requests to fail with 400 errors and poor user experience.

## Solution

Automatically detect unsupported parameter 400 errors, cache results, adapt parameters, and retry requests.

- **Transparent to users**: Auto-fix without manual troubleshooting
- **Smart caching**: Learn parameter support, faster subsequent requests
- **Minimal changes**: Only 9 lines in chatCore.js
- **Backward compatible**: No side effects for non-OpenAI protocols

## Flow

Request → 400 param error → Detect → Cache → Adapt → Retry (max 3) → Success

## Supported Parameters

- max_completion_tokens ↔ max_tokens
- reasoning_effort
- thinking / budget_tokens
- stream_options
- response_format.json_schema
- temperature
- logprobs
- frequency_penalty
- presence_penalty

## Scope

OpenAI-compatible protocols. Non-OpenAI protocols unaffected (parameter name mismatch → no cache → normal error return).

## Files Changed

- `open-sse/executors/base.js` - 9 lines (added adaptTransformedBody hook)
- `open-sse/handlers/chatCore.js` - 9 lines
- `open-sse/services/paramProbe.js` - 282 lines (async probe service)
- `src/lib/paramAdapter.js` - 189 lines (adaptation logic)
- `src/lib/paramSupportCache.js` - 194 lines (cache service)
- `tests/unit/paramAdapter.test.js` - 46 lines (unit tests)

## Note

Executors overriding execute without calling super.execute may need similar post-transform adaptation (antigravity, cursor, kiro, etc.). Can be addressed in follow-up PR if needed.

### DIFF
--- a/open-sse/executors/base.js
+++ b/open-sse/executors/base.js
@@ -96,7 +96,7 @@ export class BaseExecutor {
     return { status: response.status, message: bodyText || `HTTP ${response.status}` };
   }
 
-  async execute({ model, body, stream, credentials, signal, log, proxyOptions = null }) {
+  async execute({ model, body, stream, credentials, signal, log, proxyOptions = null, adaptTransformedBody = null }) {
     const fallbackCount = this.getFallbackCount();
     let lastError = null;
     let lastStatus = 0;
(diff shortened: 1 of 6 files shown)


kenlin8827:
@decolua This PR addresses a common pain point where third-party providers reject new OpenAI parameters. The solution is minimal (9 lines in chatCore.js) with automatic detection, caching, and retry. Would appreciate your review!

PR #1506, opened 2026-05-27
fix: reject unsupported Kiro 1m context suffix

### LINKED ISSUE #1503: kr/* (Kiro/Bedrock) silently forwards [1m] context suffix to upstream → 400 Improperly formed request
## Summary

`kr/*` (Kiro provider, AWS CodeWhisperer/Bedrock backed) returns `400 "Improperly formed request"` for any model name with a `[1m]` suffix. Example:

```
[kiro/claude-opus-4.7-thinking-agentic[1m]] [400]: {"message":"Improperly formed request.","reason":null}
```

After analyzing the codebase and 36 PRs, the suffix `[1m]` is never parsed in the `kr/*` request path. The literal string is forwarded to AWS upstream as-is, which rejects it. Even if a future PR added suffix-stripping plus `anthropic-beta: context-1m-2025-08-07` header injection, the request would still terminate at AWS Bedrock — and AWS Bedrock's contract for Claude models does not appear to honor Anthropic's 1M-context beta header.

This was already explored and abandoned by a prior contributor in PR #1485 (`feat: Add 1M context window toggle in Settings`), self-closed with the note: *"1M context window toggle not fully effective. The model-native limits override 9router metadata."*

## Reproduction (sanitized)

```bash
curl -sS -w '\n[%{http_code}]\n' \
  -H "x-api-key: <api-key>" \
  -H "content-type: application/json" \
  <your-9router-endpoint>/v1/messages \
  -d '{
        "model": "kr/claude-opus-4.7-thinking-agentic[1m]",
        "max_tokens": 30,
        "messages": [{"role":"user","content":"hi"}]
      }'

# → 400 "Improperly formed request"
```

Sanity check (same request, no `[1m]` suffix) returns `200`, so the model itself is reachable; only the suffix routing fails.

## Root cause

The Kiro suffix parser at `open-sse/config/kiroConstants.js` (introduced via PR #1166, on master via commit `4a575f21`) recognizes only `-thinking` and `-agentic`:

```js
export const KIRO_AGENTIC_SUFFIX = "-agentic";
export const KIRO_THINKING_SUFFIX = "-thinking";
```

`resolveKiroModel()` strips these two suffixes and returns the upstream id. There is no branch for `[1m]`, no fallback, no error — the bracketed segment travels untouched into `userInputMessage.modelId` of the Kiro payload. AWS Bedrock then rejects it.

Repo-wide search confirms zero hits for `[1m]`, `context-1m`, `1000000`, or long-context parsing logic. PR #1273 (open) extends `-thinking-agentic` variants but does not add a `[1m]` axis.

## Why this is upstream-blocked, not a 9router bug

The `kr/*` path terminates at AWS CodeWhisperer/Bedrock. Anthropic's 1M context window is gated by a beta header (`anthropic-beta: context-1m-2025-08-07`) on Anthropic's direct API. Public AWS Bedrock documentation for Claude models does not list this beta as supported, and Bedrock's request schema is independent of Anthropic's headers in any case. So a hypothetical patch in 9router that strips `[1m]` and injects the beta header would likely still receive a 400 from Bedrock for unrelated reasons (model-id whitelist, schema validation).

## Proposal

Three lightweight changes, no upstream dependency:

1. **Fast-fail in `kr` executor** — when the requested model contains `[1m]`, return a clear error before forwarding:
   > `[kr/*] '[1m]' suffix is not supported by Kiro upstream. Kiro is AWS Bedrock-backed and does not honor Anthropic's context-1m beta. Use a direct-Anthropic provider for 1M-context routing.`

2. **Inline doc comment** in `open-sse/config/providerModels.js` near the `kr` namespace explaining the limitation, mirroring the pattern of the existing `// The translator strips these suffixes...` comment near the `-thinking`/`-agentic` rows.

3. *(Optional, lower priority)* — accept the suffix at parse time, drop it silently, and emit a one-time warning in logs that the request was downgraded to the model's native context window. This trades fast-fail for backwards-compatibility if any external tooling appends `[1m]` blindly.

PR #386 (`fix: treat Kiro 400 'improperly formed request' as model-unavailable`, merged) currently classifies this 400 as `paymentRequired` and triggers fallback. That works around the symptom but masks the root cause and locks accounts incorrectly when the actual cause is a malformed model id rather than a tier issue.

## Environment

- 9router `0.4.62` (npm `9router@latest` global)
- Node 18+
- Linux x86_64
- Kiro provider configured with valid AWS credentials

## Scope offer

Happy to submit a PR for items (1) and (2) — both are small, no behavior change for existing users, and surface a clear contract for `[1m]`-suffix consumers. Confirming the design direction before opening one.

Delcado19:
## Summary
- reject `kr/*` model IDs containing `[1m]` before building the Kiro/Bedrock payload
- document the Kiro namespace limitation near the provider model list
- add a regression test for the unsupported Kiro `[1m]` suffix

Fixes #1503.

## Validation
- `node --check open-sse/config/kiroConstants.js`
- `node --check open-sse/translator/request/openai-to-kiro.js`
- `node --check open-sse/config/providerModels.js`
- `git diff --check`
- `npx vitest run --config ./tests/vitest.config.js tests/unit/openai-to-kiro.test.js`
- `npm run build`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Fixes
 - proxyFetch: restore missing `Readable` import causing runtime `ReferenceError` in DNS-bypass fetch path
+- Kiro: reject unsupported `[1m]` context suffixes before forwarding malformed model IDs to AWS Bedrock (#1503)
 
 ## Improvements
 - Lower stream stall timeout from 60s → 35s for faster hang detection
--- a/open-sse/config/kiroConstants.js
+++ b/open-sse/config/kiroConstants.js
@@ -17,6 +17,9 @@
 
 export const KIRO_AGENTIC_SUFFIX = "-agentic";
 export const KIRO_THINKING_SUFFIX = "-thinking";
+export const KIRO_UNSUPPORTED_CONTEXT_1M_SUFFIX = "[1m]";
+export const KIRO_UNSUPPORTED_CONTEXT_1M_MESSAGE =
+  "[kr/*] '[1m]' suffix is not supported by Kiro upstream. Kiro is AWS Bedrock-backed and does not honor Anthropic's context-1m beta. Use a direct-Anthropic provider for 1M-context routing.";
 
 export const KIRO_THINKING_BUDGET_DEFAULT = 16000;
 
@@ -180,6 +183,18 @@ export function stripThinkingSuffix(model) {
 }
 
 /**
+ * Kiro is AWS Bedrock-backed, so Anthropic's direct-provider `[1m]` context
+ * beta cannot be forwarded as part of a `kr/*` model id.
+ *
+ * @param {string} model
+ * @returns {boolean}
+ */
+export function hasUnsupportedKiroContextSuffix(model) {
+  return typeof model === "string"
+    && model.toLowerCase().includes(KIRO_UNSUPPORTED_CONTEXT_1M_SUFFIX);
+}
+
+/**
  * Resolve a 9router model id to the real upstream Kiro model id, plus flags
  * describing which behaviours the suffixes implied.
  *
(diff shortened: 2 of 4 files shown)


steygagency (approved):

tonydwb (approved):
## Code Review Summary

**Verdict: Approved** — PR #1506

---
*Reviewed by Hermes Agent*

Issue #1169, opened 2026-05-15
"HTTP 401: Unauthorized" when turning on "Require login" option in setting


mra00000:
Step to reproduce:
1. Go to Setting page > ensure that "Require Login" option is enabled
2. Go to Providers page, select any provider, press "test" on several model (tested on Claude Code, Gemini, Kiro)

Actual response:
- Get "HTTP 401: Unauthorized" error message

Version: 0.4.45 and 0.4.46

DmitryKn:
I had the same problem today. The providers connected, but after testing the models, I always get HTTP 401: Unauthorized (v0.4.46).

steveepreston:
Have this problem too.
Maybe related: https://github.com/decolua/9router/issues/1160

cybercode88:
same problem here, but in Console Log normal

913883232:
Has anyone solved it?

mra00000:
The issue has been fix this issue since version 0,4,49, closing the issue

<img width="851" height="893" alt="Image" src="https://github.com/user-attachments/assets/6359ede0-cbe4-4ff5-bf4d-2d1efabc9194" />

PR #1539, opened 2026-05-29
fix: distinguish dashboard-session 401 from upstream auth error (#1160)

### LINKED ISSUE #1160: Dashboard "HTTP 401: Unauthorized" for provider model list when upstream returns 200
## Summary

The dashboard's provider detail page shows **"HTTP 401: Unauthorized"** under "可用模型 / Available Models" for OAuth-based providers (Codex, Kilo Code), even though the upstream API returns **HTTP 200** when called with the same OAuth token.

The root cause appears to be that the dashboard's error display does **not distinguish** between:
1. **Dashboard internal auth failure** (browser session cookie missing for `/api/providers/<id>/models`)
2. **Upstream provider auth failure** (token rejected by ChatGPT / Kilo Code)

Both surface as identical `HTTP 401: Unauthorized` to the user, leading to wasted time re-running OAuth flows when the token is actually valid.

## Version

- `9router` v0.4.46
- Platform: macOS 24.0.0
- Install via `npm install -g 9router` (custom prefix `~/.hermes/node/lib/node_modules`)

## Reproduction

1. Add a Codex (or Kilo Code) provider via OAuth — connection succeeds, status: `active`
2. Open the provider detail page in dashboard
3. "可用模型 / Available Models" section shows: `HTTP 401: Unauthorized`

## Evidence the upstream is fine

Decoded JWT from `providerConnections.data.accessToken`:

```json
{
  "exp": <future timestamp · ~10 days from "now">,
  "https://api.openai.com/auth": {
    "chatgpt_plan_type": "team",
    ...
  },
  ...
}
```
Token is **not expired**, plan is **valid Team subscription**.

Direct upstream `curl` with the same token + headers that `route.js` would send:

```bash
curl -sS -o /dev/null -w "HTTP %{http_code}\n" \
  "https://chatgpt.com/backend-api/codex/models?client_version=1.0.0" \
  -H "Authorization: Bearer <JWT from db>" \
  -H "Accept: application/json"

# Result: HTTP 200 (with ~199KB models payload — gpt-5.5, etc.)
```

Calling `9router`'s own `/api/providers/<id>/models` endpoint **with a valid dashboard `auth_token` cookie**:

```bash
# login → set-cookie auth_token=<JWT>
COOKIE=$(curl -sSi -X POST http://127.0.0.1:20128/api/auth/login \
  -H "Content-Type: application/json" -d '{"password":"<password>"}' \
  | grep -i "^set-cookie:" | sed 's/^set-cookie: //I' | cut -d';' -f1)

curl -sS -o /dev/null -w "HTTP %{http_code} · %{size_download}B\n" \
  "http://127.0.0.1:20128/api/providers/<codex-id>/models" \
  -H "Cookie: $COOKIE"

# Result: HTTP 200 · 364479B  (with full model list)
```

So 9router's backend correctly fetches upstream and returns 200. The "401" the user sees in dashboard is **not** coming from upstream.

## Suspected root cause

`app/.next-cli-build/server/middleware.js` requires `auth_token` cookie for `/api/providers/*` paths (not in the `/v1`, `/v1beta` allow-list). When the dashboard SPA does client-side `fetch('/api/providers/<id>/models')` but the cookie is missing/expired (e.g., user opened dashboard in a new browser profile, behind a webview with isolated cookie jar, or session expired), the middleware returns `{"error":"Unauthorized"}` with HTTP 401.

The dashboard's provider detail page then displays this generic `HTTP 401: Unauthorized` under "可用模型" — but the user reasonably interprets it as **OAuth token rejection by upstream**, because that's where the "HTTP 401" pattern usually originates in LLM proxy setups.

`app/.next-cli-build/server/app/api/providers/[id]/models/route.js` returns errors like:

```js
return v.NextResponse.json(
  { error: `Failed to fetch models: ${j.status}` },
  { status: j.status }
);
```

— this is fine when an actual upstream error occurs. But when the **middleware** returns 401 before route.js even runs, the user has no way to distinguish "you need to log in to the dashboard" from "your provider token is bad".

## Suggested fix

Add a `source` field to error responses so the UI can tell the user which auth layer failed:

```js
// middleware.js — dashboard auth fail
return NextResponse.json(
  { error: "Unauthorized", source: "dashboard", hint: "Login to dashboard required" },
  { status: 401 }
);

// route.js — upstream auth fail
return NextResponse.json(
  { error: `Failed to fetch models: ${j.status}`, source: "upstream", hint: "Provider token may be expired/insufficient scope" },
  { status: j.status }
);
```

UI provider detail page then renders:
- `Dashboard auth: HTTP 401 — please login again` (with login link)
- `Upstream auth: HTTP 401 — re-authorize this provider` (with edit/re-OAuth link)

## Additional notes

This is especially confusing for users who embed the 9router dashboard inside another Electron app (e.g., via `WebContentsView` with a separate cookie partition). The host app may have done a successful upstream OAuth, but the embedded dashboard cookie partition starts fresh — every model list shows "401", and there's no signal that the fix is "log into the dashboard with `123456`".

Happy to help with a PR if this direction sounds right.

Delcado19:
## Problem

The dashboard's provider detail page and quota panel show **"HTTP 401: Unauthorized"** for OAuth providers (Codex, Kilo Code) even when the upstream API returns 200 with the same token.

Root cause: a `401` was rendered identically for two different failures:
1. **Dashboard session expired** — the middleware rejects the browser's request to our own `/api/...` route before it runs (no JSON `error` body).
2. **Upstream provider auth failure** — our API route returns an explicit `{ error: "Failed to fetch models: 401" }`.

Both surfaced as a bare `HTTP 401: Unauthorized`, so users blamed their provider token.

## Fix

Add `describeFetchError(status, statusText, dataError)` (`src/shared/utils/fetchError.js`):
- prefer our API's explicit `error` message (already identifies an upstream failure);
- otherwise treat a bodyless `401`/`403` as a **dashboard session/auth** issue and tell the user to reload / log in again.

Applied at the two fetch-error sites that produced the bare message:
- `ProviderLimits/index.js` (quota fetch)
- `CompatibleModelsSection.js` (model import from `/models`)

## Tests

Adds `tests/unit/fetch-error.test.js` (5 tests, all passing) covering: explicit upstream error preferred, bodyless 401 → session message (no longer "Unauthorized"), bodyless 403, and HTTP fallback.

Fixes #1160

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v0.4.63 (2026-05-26)
 
 ## Fixes
+- Dashboard: distinguish an expired dashboard session from an upstream provider auth failure instead of showing a bare "HTTP 401: Unauthorized" in the provider model list / quota panels (#1160)
 - proxyFetch: restore missing `Readable` import causing runtime `ReferenceError` in DNS-bypass fetch path
 
 ## Improvements
--- a/src/app/(dashboard)/dashboard/providers/[id]/CompatibleModelsSection.js
+++ b/src/app/(dashboard)/dashboard/providers/[id]/CompatibleModelsSection.js
@@ -3,6 +3,7 @@
 import { useState } from "react";
 import PropTypes from "prop-types";
 import { Button } from "@/shared/components";
+import { describeFetchError } from "@/shared/utils/fetchError";
 function CompatibleModelRow({ modelId, fullModel, copied, onCopy, onDeleteAlias, onTest, testStatus, isTesting }) {
   const borderColor = testStatus === "ok"
     ? "border-green-500/40"
@@ -149,9 +150,10 @@ export default function CompatibleModelsSection({ providerStorageAlias, provider
     setImporting(true);
     try {
       const res = await fetch(`/api/providers/${activeConnection.id}/models`);
-      const data = await res.json();
+      const data = await res.json().catch(() => ({}));
       if (!res.ok) {
-        alert(data.error || "Failed to import models");
+        // Session-401 vs upstream/provider error are now distinguished (#1160).
+        alert(describeFetchError(res.status, res.statusText, data.error));
         return;
       }
       const models = data.models || [];
--- a/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js
+++ b/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js
@@ -8,6 +8,7 @@ import { parseQuotaData, calculatePercentage } from "./utils";
 import Card from "@/shared/components/Card";
 import { EditConnectionModal } from "@/shared/components";
 import { USAGE_SUPPORTED_PROVIDERS } from "@/shared/constants/providers";
+import { describeFetchError } from "@/shared/utils/fetchError";
 
 function getConnectionLabel(connection) {
   const isEmail = (value) =>
@@ -317,7 +318,8 @@ export default function ProviderLimits() {
 
       if (!response.ok) {
         const errorData = await response.json().catch(() => ({}));
-        const errorMsg = errorData.error || response.statusText;
+        // Distinguish a dashboard-session 401 from an upstream/provider error (#1160).
+        const errorMsg = describeFetchError(response.status, response.statusText, errorData.error);
 
         // Handle different error types gracefully
         if (response.status === 404) {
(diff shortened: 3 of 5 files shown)


Delcado19:
Closing: this only improved the error *message*; the underlying "HTTP 401" root cause for the dashboard model list (#1160, reported on v0.4.46) is the same class as #1169, which was already fixed in v0.4.49 ("model check blocked by dashboardGuard: pass machineId-based CLI token in internal self-calls"). Withdrawing to avoid a redundant cosmetic change.

PR #1542, opened 2026-05-29
fix: include free OpenCode models without -free suffix in suggested models (#1535)

### LINKED ISSUE #1535: fix: opencode-free filter excludes big-pickle (non-free-suffix models)
[file: src/app/api/providers/suggested-models/route.js]
`opencode-free` filter only includes models ending in `-free`:

```js
.filter((m) => m.id?.endsWith("-free"))
```

But `big-pickle` is a free OpenCode model that doesn't use that suffix (already routed in `open-sse/executors/opencode.js`). So it never shows in the suggested-models UI.

**Fix:**

```js
const KNOWN_FREE_MODELS = new Set(["big-pickle"]);
...
.filter((m) => m.id?.endsWith("-free") || KNOWN_FREE_MODELS.has(m.id))
```

Add to the set as other non-`-free` free models appear.

Delcado19:
## Problem

The `opencode-free` suggested-models filter (`src/app/api/providers/suggested-models/route.js`) only kept models whose id ends in `-free`:

```js
.filter((m) => m.id?.endsWith("-free"))
```

`big-pickle` is a free OpenCode model that does **not** use that suffix — it's already routed in `open-sse/executors/opencode.js` (`MESSAGES_MODELS`). As a result it never showed up in the suggested-models import UI.

## Fix

- Extract the filters into `src/app/api/providers/suggested-models/filters.js` so they are unit-testable.
- Add a `KNOWN_FREE_OPENCODE_MODELS` allowlist (`big-pickle`) and include it alongside the `-free` suffix check.

Future free-but-unsuffixed OpenCode models can be added to the same allowlist.

## Tests

Adds `tests/unit/suggested-models-filters.test.js` (5 tests, all passing): `big-pickle` inclusion, `-free` models still included, paid models excluded, `{ id, name }` shape, and null-safety.

Fixes #1535

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v0.4.63 (2026-05-26)
 
 ## Fixes
+- Suggested models: include free OpenCode models without a `-free` suffix (e.g. `big-pickle`) so they show up for import (#1535)
 - proxyFetch: restore missing `Readable` import causing runtime `ReferenceError` in DNS-bypass fetch path
 
 ## Improvements
--- a/src/app/api/providers/suggested-models/filters.js
+++ b/src/app/api/providers/suggested-models/filters.js
@@ -1,0 +1,27 @@
+// Suggested-model filters for the "Import suggested models" dashboard flow.
+// Extracted from route.js so the filter logic is unit-testable (#1535).
+
+// Free OpenCode models that are routable but do NOT use the conventional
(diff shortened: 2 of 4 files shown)


decolua:
Thanks for the clean fix, @Delcado19! Extracting the filters into a testable module and adding the `KNOWN_FREE_OPENCODE_MODELS` allowlist for `big-pickle` is exactly the right approach, and the unit tests are appreciated.

Merged. Closing this PR — thank you for the contribution! 🙏

PR #1560, opened 2026-05-30
Add Crof AI provider support

anandghegde:
## Summary
- Add Crof AI as a first-class API-key provider.
- Route Crof AI through the existing OpenAI-compatible executor at `https://crof.ai/v1/chat/completions`.
- Add a focused config regression test for provider metadata, runtime routing, endpoint display, and passthrough models.

## Test Plan
- [x] `npm install --prefix tests && tests/node_modules/.bin/vitest run --config tests/vitest.config.js tests/unit/crof-provider-config.test.js --reporter=verbose`


tonydwb (approved):
## Code Review Summary

**Verdict: Approved** — PR #1560

---
*Reviewed by Hermes Agent*

anandghegde:
Rebased onto `master` and rewritten against the provider registry that landed since this was opened.

The original diff no longer applies at all: `open-sse/config/providers.js` is now a barrel over `providers/index.js`, `src/shared/constants/providers.js` derives its lists with `byCategory()` instead of holding literals, and `PROVIDER_ENDPOINTS` in `src/shared/constants/config.js` is gone. So rather than resolve the conflicts, this is the same provider expressed the new way.

It is now one registry entry plus its registration:

- `open-sse/providers/registry/crof-ai.js` — id, alias, `category: "apikey"`, display, `transport.baseUrl`, `passthroughModels`, `serviceKinds`
- registered in `registry/index.js` (appended as `p124` rather than renumbering the existing imports)
- `tests/unit/crof-ai-provider-config.test.js` rewritten to assert against the registry entry

`format: "openai"` and the UI provider list are both registry defaults now, so neither is declared. Verified `PROVIDERS["crof-ai"]` still resolves to `{ baseUrl: "https://crof.ai/v1/chat/completions", format: "openai" }`, which is what the original PR set explicitly.

One note: `vitest` isn't in `package.json` and there's no test script, so I ran the new test with `npx vitest run` (3 passed). Happy to adjust if the suite is meant to be run some other way.

Issue #1258, opened 2026-05-18
feat(dashboard): show which API key was used for each recent request


yogadwipayana:
## Problem

The **Recent Requests** table on the dashboard shows `Model`, `In / Out`, and `When`, but it doesn't show **which API key** was used to make each request.

When you have multiple API keys configured (e.g. one per CLI tool, or one per teammate), there's no way to tell from the dashboard which key sent which request. This makes it harder to:

- Debug which tool/integration a request came from
- Audit usage per key
- Spot a leaked or misbehaving key

## Current state

- `apiKey` is already stored on every row in `usageHistory` (see `src/lib/db/repos/usageRepo.js`)
- It's already aggregated into `stats.byApiKey` for the Usage Analytics view
- However, the `recentRequests` array built in `getActiveRequests()` and `getUsageStats()` does **not** include the `apiKey` field, so the UI has nothing to render

## Proposed change

1. In `src/lib/db/repos/usageRepo.js`, include `apiKey` (resolved to its friendly `keyName` via the existing `apiKeyMap`) in the `recentRequests` objects returned by both `getActiveRequests()` and `getUsageStats()`.
2. Add a new **API Key** column to the Recent Requests table on the dashboard so each row shows which key was used (or \"Local\" when no key is set).

## Would like to work on this

Happy to send a PR if a maintainer agrees this is a good direction. Will follow the existing commit-message style (\`feat(dashboard): ...\`).

PR #1561, opened 2026-05-30
feat(dashboard): show which API key was used per recent request (#1258)

Delcado19:
## What

Implements #1258: adds an **API Key** column to the **Recent Requests** table on the Usage dashboard, so operators running multiple keys (one per CLI tool, or one per teammate) can see which key sent each request.

The displayed key is the **inbound 9router API key** — the one a CLI tool puts in its `Authorization` header and that gets validated against the configured keys (`isValidApiKey`). It is **not** the upstream provider credential (those are resolved separately via `getProviderCredentials`).

## Security

The plaintext key never leaves the server, in **both** dashboard surfaces that expose key usage:

1. **Recent Requests** — `resolveRecentRequestApiKey` returns only `apiKeyId` (stable non-secret id) and `keyName` (configured name, or masked `••••last4`). Previously the full key was placed in `apiKey`/`apiKeyKey` and streamed to the client.
2. **`byApiKey` aggregate** ("Usage by API Key") — this used the plaintext key as its **map key** *and* stored it in `apiKey`/`apiKeyKey`. A single `sanitizeByApiKey` pass runs just before `getUsageStats` returns: it re-keys each entry by the non-secret `apiKeyId` and drops the secret fields. Internal aggregation is unchanged, so dedup/`lastUsed` keep working.

## Layout

The Recent Requests side panel was narrow for an extra column, so it was widened (`1.4fr` / min `380px`) and **Model + API Key** are stacked in a single column (model on top, key name below) instead of two hard-truncated columns — both stay readable, no horizontal overflow.

## Tests

- `tests/unit/usage-recent-requests.test.js`
- `tests/unit/usage-recent-requests-stats.test.js`

All green. They cover name resolution, the masked fallback, the local (no-key) case, "no plaintext in the recent-requests payload", "no plaintext in the byApiKey payload (keys or values)", and "two known keys stay distinct rows with summed tokens".

### DIFF
--- a/src/lib/db/repos/usageRepo.js
+++ b/src/lib/db/repos/usageRepo.js
@@ -136,6 +136,36 @@ export function resolveRecentRequestApiKey(apiKey, apiKeyMap = {}) {
   };
 }
 
+// Strip plaintext API keys from the byApiKey aggregate before it leaves the
+// server. Internally the plaintext key is used as the map key and stored in
+// `apiKey`/`apiKeyKey`, but the client only renders `keyName`. Re-key each entry
+// by the key's stable non-secret id (uuid for known keys, ••••last4 otherwise)
+// and drop the secret fields. Two distinct known keys keep distinct uuid ids, so
+// they stay distinct rows; only unknown keys sharing a ••••last4 fallback merge,
+// which is acceptable and no worse than the previous slice(0,8) label. (#1258)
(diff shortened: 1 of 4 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved** — PR #1561

---
*Reviewed by Hermes Agent*

PR #1568, opened 2026-05-30
fix(sse): prevent false stall aborts on large-context reasoning streams

thomasvan:
## Problem

Large-context requests to Kiro (e.g. Claude Opus thinking, 100+ msgs / many tools / 40k+ input tokens) intermittently fail with:

> API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request

The proxy returns `200 OK` + SSE headers but an empty body. Logs show:

```
🌊 [STREAM] KIRO | claude-opus-4.8 | first chunk | ttft=3ms
🌊 [STREAM] KIRO | claude-opus-4.8 | ABORT stall timeout | budget=30000ms | chunks=54 | bytes=11730 | sinceLast=30001ms
🌊 [STREAM] KIRO | claude-opus-4.8 | error: stream stall timeout
```

## Root cause

Two issues compounded:

1. **One watchdog gating two very different phases.** The single `STREAM_STALL_TIMEOUT_MS` covered both time-to-first-token (prompt prefill, can be long for big contexts) and the inter-chunk gap. A long prefill, or a mid-stream reasoning pause, both tripped the same short timer.
2. **The stall default was too aggressive.** It had been lowered to 30s "for faster hang detection", which regressed reasoning models that legitimately go silent 60s+ between tokens on large contexts (observed real pause: `sinceLast=60002ms`).

For Kiro specifically, the AWS EventStream→SSE transform runs inside the executor, so the downstream stall watchdog only sees *enqueued client chunks* — frames that produce no chunk (partial-frame buffering, metering/context events, reasoning pauses) look like silence even though raw upstream bytes are still arriving.

## Fix

- **Split the watchdog into two phases** — `STREAM_FIRST_CHUNK_TIMEOUT_MS` (120s) covers prefill until the first byte; `STREAM_STALL_TIMEOUT_MS` (restored to 180s) covers inter-chunk gaps once tokens flow.
- **Kiro keepalive** — emit an SSE comment (`: ka`) when an upstream frame yields no client chunk, so the watchdog reflects genuine upstream activity. Comment lines are ignored by the downstream SSE parser and clients.
- **Env-overridable** — all three timeouts (`STREAM_STALL_TIMEOUT_MS`, `STREAM_FIRST_CHUNK_TIMEOUT_MS`, `FETCH_CONNECT_TIMEOUT_MS`) now read from env with positive-integer validation.

## Testing

Built the Docker image and ran it in production mode (`NODE_ENV=production`). Confirmed via live logs that large-context Kiro/Opus requests that previously aborted with `stream stall timeout` now stream to completion. No behavior change for fast requests.

### DIFF
--- a/open-sse/config/runtimeConfig.js
+++ b/open-sse/config/runtimeConfig.js
@@ -31,11 +31,23 @@ export const MEMORY_CONFIG = {
   proxyDispatchersMaxSize: 20,
 };
 
-// Stream stall timeout: abort if no chunk received within this duration
-export const STREAM_STALL_TIMEOUT_MS = 30 * 1000;
+// Parse a positive integer env override, falling back to a default.
+function envMs(name, def) {
+  const raw = process.env[name];
+  if (raw == null || raw === "") return def;
+  const n = parseInt(raw, 10);
+  return Number.isFinite(n) && n > 0 ? n : def;
+}
+
+// Inter-chunk stall timeout (once tokens are flowing). Generous headroom so
+// slow reasoning models aren't aborted mid-stream. Env: STREAM_STALL_TIMEOUT_MS.
+export const STREAM_STALL_TIMEOUT_MS = envMs("STREAM_STALL_TIMEOUT_MS", 180 * 1000);
+
+// Time-to-first-token timeout (prompt prefill). Env: STREAM_FIRST_CHUNK_TIMEOUT_MS.
+export const STREAM_FIRST_CHUNK_TIMEOUT_MS = envMs("STREAM_FIRST_CHUNK_TIMEOUT_MS", 120 * 1000);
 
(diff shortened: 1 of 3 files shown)


mftzk:
+1 — ran into the same hardcoded `STREAM_STALL_TIMEOUT_MS`/`FETCH_CONNECT_TIMEOUT_MS` causing `fetch connect timeout` → 502 and aborted streams on long Claude Opus reasoning requests in a self-hosted deploy. I opened #1644 to make them env-configurable before finding this PR, which already covers that (plus the first-chunk/inter-chunk split and the Kiro keepalive root-cause fix) — so I closed mine in favor of this. Would be great to see this land. Thanks!

tonydwb (commented):
## Code Review Summary

**Verdict: Comment** — PR #1568

---
*Reviewed by Hermes Agent*

PR #1569, opened 2026-05-30
feat: per-key token quota + request attribution for shared gateways

vanminhph:
## Why

9Router is often used as a **single gateway feeding many AI tools** — Claude Code, Codex, Cursor, Cline, Copilot, etc. — all sharing one upstream account pool, with **one API key per tool**. At that scale two gaps show up:

1. **One heavy tool can drain the shared pool.** When a single tool's API key burns through the pool's quota, *every other tool* starts getting `429`s through no fault of its own. There is currently no way to bound a single key's consumption.
2. **You can't tell which tool used what.** `requestDetails` stores `connectionId` (the upstream account) but not the **API key** that made the call, so usage/MCP/tooling can't be attributed back to the tool that issued it.

This PR adds two small, **opt-in** governance features to close those gaps — without changing default behavior for anyone who doesn't turn them on.

### 1. Per-key token quota (disabled by default)
- A **fixed 5h window anchored to the key's first request** (mirrors upstream reset semantics — not a rolling window), enforced per API key in `handleChat` **before** combo/non-combo routing so every path is covered.
- Returns **`429` + `Retry-After`** with a human-readable reset time when the budget is exceeded.
- Per-key override via `apiKeys.tokenBudget5h` (e.g. a larger budget for your main coding tool, tighter for background agents); sensible global default otherwise.
- Counts `promptTokens` of the canonical (endpoint-null) usage rows only, so it stays correct alongside the existing dual-write.

### 2. Per-key request attribution
- Persists `apiKey` on every `requestDetails` row (previously `connectionId` only), so usage, MCP/tool calls and quota debugging can be traced back to the specific tool/key by an external process.

## Safe by design
- **Opt-in, off by default** — `userQuotaEnabled` resolves `DB setting > env > built-in default (false)`. Existing deployments behave exactly as before until explicitly enabled.
- **Additive schema only** — applied automatically by `syncSchemaFromTables` (`apiKeys.tokenBudget5h`, `requestDetails.apiKey`, `userQuotaWindow` table, supporting indexes). No migration file, no destructive change, no data loss.
- **Config follows existing conventions** — same `setting || env || default` resolution used by `observability*`; new env vars `USER_QUOTA_ENABLED`, `USER_TOKEN_BUDGET_5H`.
- **Backward compatible** — request attribution is a single nullable column; rows written before the upgrade simply have `apiKey = null`.

## Footprint
- 9 files, ~120 lines added. No new dependencies. Follows repo style (2-space, double quotes, `await getAdapter()`, dynamic repo imports, local-tz date formatting like `getLocalDateKey`). `eslint` clean.

## Testing
- Unit-level logic verified for the quota window (open/reset, under/over budget, per-key override, enabled resolution).
- Verified end-to-end on a live instance: `429` returned with correct `Retry-After`/reset time when a key is over budget, requests served normally when under budget, and `apiKey` correctly recorded on `requestDetails`.

## How to enable
```yaml
# docker-compose env
environment:
  - USER_QUOTA_ENABLED=true
  - USER_TOKEN_BUDGET_5H=25000000   # tokens per 5h per key, optional
```
Or via DB settings (`userQuotaEnabled`, `userTokenBudget5hDefault`); per-key limit via `apiKeys.tokenBudget5h`.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -187,7 +187,7 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     trackPendingRequest(model, provider, connectionId, false, true);
     appendRequestLog({ model, provider, connectionId, status: `FAILED ${error.name === "AbortError" ? 499 : HTTP_STATUS.BAD_GATEWAY}` }).catch(() => { });
     saveRequestDetail(buildRequestDetail({
-      provider, model, connectionId,
+      provider, model, connectionId, apiKey,
       latency: { ttft: 0, total: Date.now() - requestStartTime },
       tokens: { prompt_tokens: 0, completion_tokens: 0 },
       request: extractRequestConfig(body, stream),
@@ -233,7 +233,7 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     const { statusCode, message, resetsAtMs } = await parseUpstreamError(providerResponse, executor);
     appendRequestLog({ model, provider, connectionId, status: `FAILED ${statusCode}` }).catch(() => { });
     saveRequestDetail(buildRequestDetail({
-      provider, model, connectionId,
+      provider, model, connectionId, apiKey,
       latency: { ttft: 0, total: Date.now() - requestStartTime },
       tokens: { prompt_tokens: 0, completion_tokens: 0 },
       request: extractRequestConfig(body, stream),
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -201,7 +201,7 @@ export async function handleNonStreamingResponse({ providerResponse, provider, m
 
   const totalLatency = Date.now() - requestStartTime;
   saveRequestDetail(buildRequestDetail({
-    provider, model, connectionId,
+    provider, model, connectionId, apiKey,
     latency: { ttft: totalLatency, total: totalLatency },
     tokens: usage || { prompt_tokens: 0, completion_tokens: 0 },
     request: extractRequestConfig(body, stream),
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -159,7 +159,8 @@ export async function handleNonStreamingResponse({ providerResponse, provider, m
 
   const usage = extractUsageFromResponse(responseBody);
   appendLog({ tokens: usage, status: "200 OK" });
-  saveUsageStats({ provider, model, tokens: usage, connectionId, apiKey, endpoint: clientRawRequest?.endpoint });
+  // Only write for this request → canonical quota row
+  saveUsageStats({ provider, model, tokens: usage, connectionId, apiKey, endpoint: clientRawRequest?.endpoint, countsTowardQuota: true });
 
   const translatedResponse = needsTranslation(targetFormat, sourceFormat)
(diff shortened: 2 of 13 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved** — PR #1569

---
*Reviewed by Hermes Agent*

muhfauziazhar:
Took a careful pass through this — traced the write paths down to the callers to verify the core assumption. The direction is right and the implementation is clean (opt-in, enforced before combo/non-combo branching, additive schema that auto-syncs via `syncSchemaFromTables`, sensible `idx_uh_apikey_ts` index, masked key in logs). A few things worth addressing before merge:

## 🔴 Quota only counts `endpoint IS NULL` rows → non-streaming requests are never charged

The window sum is:

```sql
SELECT COALESCE(SUM(promptTokens),0) FROM usageHistory
WHERE apiKey = ? AND endpoint IS NULL AND timestamp >= ?
```

`usageHistory` is written only by `saveRequestUsage` (`appendRequestLog` is now a no-op). Tracing the two callers:

- **Streaming** (`open-sse/utils/stream.js`): calls `logUsage()` → writes a row with `endpoint = NULL`, **and** `saveUsageStats()` via `onStreamComplete` → writes a row with `endpoint` set. So a streaming request is genuinely logged twice, and `endpoint IS NULL` correctly dedupes it to one. ✅
- **Non-streaming** (`open-sse/handlers/chatCore/nonStreamingHandler.js:162`): only calls `saveUsageStats({ ..., endpoint: clientRawRequest?.endpoint })`. `clientRawRequest.endpoint` is always populated (`url.pathname`, see `src/sse/handlers/chat.js:41`), so **there is no `endpoint = NULL` row at all.**

Net effect: every `stream:false` request bypasses the quota. The enforcement check covers all paths, but the accounting only sees streaming traffic. The "each request is logged twice" assumption holds only for the streaming path through `stream.js`.

Suggestion: instead of relying on `endpoint IS NULL` (which works by accident because streaming happens to write an extra row), add an explicit marker — e.g. a `countsTowardQuota` column, or dedupe by a request id — so both streaming and non-streaming get counted exactly once.

## 🟡 `tokenBudget5h` per-key override is currently a dead column

`checkUserQuota` reads `apiKeys.tokenBudget5h`, but `apiKeysRepo` (`createApiKey` / `updateApiKey` / `rowToKey`) never reads or writes it. As-is the per-key override can only be set by editing SQLite by hand. It needs an API + dashboard surface to be usable.

## Minor

- **Burst race:** usage is written async (`.catch()`) after the response completes, while the check reads a committed `SUM`. Concurrent requests can all pass the check before any usage row lands, so the budget can be overshot during a spike. Eventually consistent, but worth a comment.
- **Prompt-only accounting:** only `promptTokens` is summed — completion tokens and cost are excluded. "token budget" is a bit misleading; it's really a prompt-token budget.
- **`resetAtLocal` uses server timezone** (`d.getHours()`), typically UTC in a container — may confuse users in other zones. Consider including the offset or using UTC explicitly.
- **No tests:** the repo has a decent suite (`db-concurrent`, `db-sqlite-vs-lowdb`); the window/reset math would benefit from unit coverage.

Context: this overlaps with #1154 and PRs #1107 / #1158 — opened #1677 to track the broader per-key governance picture. This PR is the cleanest-scoped of the three; fixing the non-streaming accounting gap would make it solid to merge.

vanminhph:
Superseded by #1681 — same feature consolidated into a single clean commit, with all review findings addressed (incl. switching quota accounting from prompt tokens to cost-based to mirror upstream credit weighting).

PR #1573, opened 2026-05-30
fix(kiro): emit valid concatenable tool_calls.arguments deltas

wahyuzero:
## Summary

Fixes a streaming protocol bug in the Kiro executor that caused random `Unterminated string` JSON parse errors in downstream consumers (opencode, claude-code, etc.) for tool calls with substantial argument content.

## Root cause

`open-sse/executors/kiro.js` reads `toolUseEvent.input` and emits its `JSON.stringify` as an OpenAI `tool_calls[].function.arguments` delta on **every** event. Kiro/AWS CodeWhisperer streams `input` as **partial objects that grow** across events:

```
event 1: { command: "cat /home" }
event 2: { command: "cat /home/wxsys" }
event 3: { command: "cat /home/wxsys/Project/naskin/.impeccable.md" }
```

Stringifying each one produces overlapping prefixes:

```
{"command":"cat /home"}
{"command":"cat /home/wxsys"}
{"command":"cat /home/wxsys/Project/naskin/.impeccable.md"}
```

The OpenAI streaming protocol requires `arguments` deltas to be **concatenable string fragments** that join into one valid JSON. Concatenating the three above produces:

```
{"command":"cat /home"}{"command":"cat /home/wxsys"}{"command":"cat /home/wxsys/Project/naskin/.impeccable.md"}
```

→ `JSON.parse` fails with `Unexpected non-whitespace character after JSON at position N` or `Unterminated string` depending on where the parser bails out.

The truncation position is **non-deterministic** because it tracks the prefix length of whichever partial object Kiro happened to send. This makes the bug present as random "the network truncated my tool call" failures — `{"comm`, `{"comman`, `{"command":` — when in fact the protocol contract was violated upstream.

## Reproduction

Any Kiro session that streams a tool whose arguments grow over multiple `toolUseEvent`s will hit this. Common triggers:
- File paths longer than a few words
- Multi-line `content` parameters in `write` tools
- Shell commands that get progressively expanded

I encountered it in opencode while writing a 1500-byte markdown file via the `write` tool routed through 9router → Kiro → Claude Opus 4.7. The `write` call failed with `Unterminated string` at variable byte offsets across retries.

## Fix

1. **String-form `input` payloads** are already concatenable deltas — emit them incrementally as before, tracking accumulated text per `toolCallId` in `state.toolArgsEmitted`.

2. **Object-form `input` payloads** are buffered into `state.toolArgsBuffered` keyed by `toolCallId`. The latest canonical (full JSON of the most recent partial object) overwrites earlier buffered values — earlier partials are noise once a more complete one arrives.

3. New helper `flushBufferedToolArgs()` is invoked at **all three** finish boundaries to emit each buffered canonical exactly once as a complete arguments string:
   - `messageStopEvent`
   - The dual-event finish path (`hasMeteringEvent && hasContextUsage`)
   - The TransformStream `flush()` callback

The result: downstream consumers receive arguments deltas that concatenate into valid JSON regardless of whether Kiro chose to send string fragments or growing partial objects.

## Tests

**New regression coverage** (`tests/unit/kiro-tool-args-streaming.test.js`, 7 tests):
- Single buffered tool call → valid flush
- 4-event partial-object stream (the exact reported repro) → 1 valid flush
- Multiple concurrent tool calls → indices preserved
- Idempotent flush (canonical equals already-emitted) → no re-emit
- Empty buffer → no-op
- Contract test: re-stringifying partials is unparseable
- Contract test: emitting once is parseable

**No regression**: all 14 existing kiro tests (`openai-to-kiro.test.js`, `rtkKiro.test.js`) continue to pass.

**End-to-end verified**: built the patched CLI tarball locally, reinstalled `9router@0.4.66`, restarted the proxy, and reproduced the original failure case (substantial `write` tool call ~1500 bytes) — now succeeds. Multiple test writes confirmed.

## Notes

- The fix is **executor-only** — `translator/response/kiro-to-openai.js` has the same shape of bug but is a separate (non-streaming-executor) code path. If this PR lands, that file deserves a follow-up; happy to include it here if preferred.
- `state` initialization adds two new Maps (`toolArgsEmitted`, `toolArgsBuffered`) — no change to existing field types.
- `flushBufferedToolArgs` is `export`ed so the unit tests can target it directly without standing up a full TransformStream pipeline.

### DIFF
--- a/open-sse/executors/kiro.js
+++ b/open-sse/executors/kiro.js
@@ -5,6 +5,45 @@ import { refreshKiroToken } from "../services/tokenRefresh.js";
 import { proxyAwareFetch } from "../utils/proxyFetch.js";
 import { HTTP_STATUS, RETRY_CONFIG, DEFAULT_RETRY_CONFIG, resolveRetryEntry } from "../config/runtimeConfig.js";
 
+// Flush buffered tool arguments at finish boundaries.
+// Kiro/CodeWhisperer streams toolUseEvent.input as PARTIAL OBJECTS that grow over
+// time (e.g. {command:"cat /home"} then {command:"cat /home/wxsys"}). Re-stringifying
+// each one and emitting it as an OpenAI argument delta produces overlapping prefixes
+// that concatenate into unparseable garbage downstream (Unterminated string).
+//
+// Fix: defer object-form payloads into state.toolArgsBuffered keyed by toolCallId,
+// keep only the latest canonical, and emit ONCE here as the complete arguments
+// string (the final object is the source of truth — intermediate states are noise).
+// String-form payloads are already concatenable deltas and emitted incrementally.
+export function flushBufferedToolArgs(state, controller, ctx) {
+  if (!state.toolArgsBuffered || state.toolArgsBuffered.size === 0) return;
+  const { responseId, created, model } = ctx;
+  for (const [toolCallId, info] of state.toolArgsBuffered) {
+    const alreadyEmitted = state.toolArgsEmitted.get(toolCallId) || "";
+    if (info.canonical && info.canonical !== alreadyEmitted) {
+      const argsChunk = {
+        id: responseId,
+        object: "chat.completion.chunk",
+        created,
+        model,
+        choices: [{
+          index: 0,
+          delta: {
+            tool_calls: [{
+              index: info.toolIndex,
+              function: { arguments: info.canonical }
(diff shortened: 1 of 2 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved** — PR #1573

---
*Reviewed by Hermes Agent*

PR #1582, opened 2026-05-31
fix: handle Kiro 400 on tool-bearing history without client tools

simonsmh:
# Summary

Kiro's (AWS CodeWhisperer) schema validator requires a non-empty currentMessage.userInputMessageContext.tools array whenever the conversation history references any tool use. If that array is missing, the API rejects the request with "Improperly formed request" (HTTP 400) — even when the current turn carries no tool call at all.

A client trips this by omitting the tools array on a follow-up request, which commonly happens after client-side compaction (e.g. OpenCode dropping the original tool definitions on a continuation).

# Approach

One principle drives both halves of the fix: never discard tool content — only downgrade its structured form to text when the structure itself would 400.

1. Client sent no tools → flatten. Collapse the conversation's tool calls/results into plain text. The request stays honest about its capabilities, and because no structured tool content survives, the validator's "tools required" rule never fires. We avoid the alternative of fabricating stub tool specs, which would advertise tool-calling capability the client never requested and may not handle — risking a phantom tool call on an otherwise plain text turn.

2. Client sent tools → reconcile orphans. When compaction drops the assistant tool_use but keeps the tool_result, the dangling reference 400s. Rather than discard the orphan (the client deliberately kept it through compaction), fold its content back into the user text as a salvaged [Tool result: ...] line — the same shape flattening produces. This covers both the history items and the popped-out currentMessage, and guards against deleting a co-located tools array when removing the orphaned reference.

The two paths are mutually exclusive (reconcile runs only when the client provided tools) and share a single toolResultToText renderer.

# Changes
- flattenToolInteractions() — collapse tool_use / tool_result / tool_calls to readable text ([Tool call: name(args)], [Tool result: ...]), only when the client sent no tools
- reconcileOrphanedToolResults() — salvage orphaned results as text across both history and currentMessage; keep results that still have a matching toolUse; guard the co-located tools array
- safeJSONParse() — guard tool-call argument parsing against malformed JSON
- Merge consecutive user messages' userInputMessageContext so toolResults/tools aren't silently dropped
- currentMessage null-guard for assistant-only input

# Testing
- npx vitest run tests/unit/openai-to-kiro.test.js — 11/11 pass (4 new: flatten OpenAI, flatten Claude, structured-tools-present, orphan-salvage)
- Manually verified multi-step agent history and assistant-only input produce clean alternating user/assistant history with no structured tool fields leaking; and that orphan removal preserves a co-located tools array
- docker build succeeds

It's a must have fix for Kiro. Better approach than #1209 

Reproduce: Claude Code prompt
```
/goal use tool to calculate 1+1 
```

### DIFF
--- a/open-sse/translator/request/openai-to-kiro.js
+++ b/open-sse/translator/request/openai-to-kiro.js
@@ -12,22 +12,182 @@ import {
   KIRO_AGENTIC_SYSTEM_PROMPT
 } from "../../config/kiroConstants.js";
 
+/** Render a single tool call as a readable text line. */
+function toolCallToText(name, input) {
+  let argStr;
+  try {
+    argStr = typeof input === "string" ? input : JSON.stringify(input ?? {});
+  } catch {
+    argStr = "{}";
+  }
+  return `[Tool call: ${name || "unknown"}(${argStr})]`;
+}
+
+/** Render a tool result (string or content-block array) as a text line. */
+function toolResultToText(content) {
+  const text = Array.isArray(content)
+    ? content.map(c => (typeof c === "string" ? c : c.text || "")).join("\n")
+    : (typeof content === "string" ? content : "");
+  return `[Tool result: ${text}]`;
+}
+
+/**
+ * Flatten all tool calls/results in a conversation into plain text.
+ *
+ * Kiro's schema validator requires a non-empty
+ * currentMessage.userInputMessageContext.tools array whenever the history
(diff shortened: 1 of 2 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved** — PR #1582

---
*Reviewed by Hermes Agent*

decolua:
Thanks @simonsmh for the contribution! Merged into master. 🙏

PR #1606, opened 2026-06-01
fix(runtime): save runtime deps so sibling install can't prune them

### LINKED ISSUE #1605: [Bug] runtime better-sqlite3 pruned by tray install (--no-save) → "No SQLite driver available"
### Summary

The SQLite and tray runtime installers both `npm install ... --no-save` into the **same** user-writable dir (`~/.9router/runtime`), whose `package.json` declares **no dependencies**. Because postinstall runs them in sequence (SQLite first, tray second), the tray install treats the just-installed `better-sqlite3` as an *extraneous* package and npm prunes it. The runtime ends up with `systray2` present but `better-sqlite3` gone, so the app falls through the whole driver chain and dies with:

```
[DB] No SQLite driver available (bun/better/node/sql.js all failed)
```

### Where

- `cli/hooks/sqliteRuntime.js` — `npmInstall()` uses `--no-save` (via `opts.optional`).
- `cli/hooks/trayRuntime.js` — `npmInstall()` uses `--no-save`.
- `cli/hooks/postinstall.js` — calls `ensureSqliteRuntime()` then `ensureTrayRuntime()`.

`ensureRuntimeDir()` writes a `package.json` with no `dependencies`, so anything installed with `--no-save` is extraneous and gets pruned by the next install into the same dir.

### Reproduce (end-to-end, with the real packages)

Runs the shipped hooks against an isolated runtime dir via `DATA_DIR` — no need to touch your real `~/.9router`. Requires `9router` installed globally. Copy-paste the whole block as-is (macOS arm64, Node 20.18.3, 9router 0.4.66):

```bash
export DATA_DIR="$(mktemp -d /tmp/9r-repro.XXXXXX)"
export HOOKS="$(npm root -g)/9router/hooks"   # global install: hooks live at the package root
cat > "$DATA_DIR/repro.js" <<'JS'
const fs = require("fs");
const sq = require(process.env.HOOKS + "/sqliteRuntime.js");
const tr = require(process.env.HOOKS + "/trayRuntime.js");
const p  = process.env.DATA_DIR + "/runtime/node_modules/better-sqlite3";
sq.ensureSqliteRuntime({ silent: true });
console.log("better-sqlite3 after sqlite install:", fs.existsSync(p) ? "PRESENT" : "absent");
tr.ensureTrayRuntime({ silent: true });
console.log("better-sqlite3 after tray install: ", fs.existsSync(p) ? "PRESENT" : "PRUNED");
console.log("runtime node_modules:", fs.readdirSync(process.env.DATA_DIR + "/runtime/node_modules").filter(x => !x.startsWith(".")).join(", "));
JS
node "$DATA_DIR/repro.js"
```

Output:

```
better-sqlite3 after sqlite install: PRESENT
better-sqlite3 after tray install:  PRUNED
runtime node_modules: debug, fs-extra, graceful-fs, jsonfile, ms, systray2, universalify
```

`better-sqlite3` is gone after the tray install; the next launch prints `No SQLite driver available`.

### Minimal mechanism (npm only, no 9router)

The runtime dir is a deps-less manifest, so `--no-save` prunes the previous package as extraneous:

```bash
d=$(mktemp -d); cd "$d"
printf '{"name":"r","version":"1.0.0","private":true}' > package.json
npm install ms@2.1.3    --no-save     # ms installed
npm install debug@4.3.4 --no-save     # -> "removed 1 package": ms pruned
ls node_modules/ms                    # missing
```

Saving the deps instead keeps both (`--save-exact`, or any non-`--no-save` install):

```bash
d=$(mktemp -d); cd "$d"
printf '{"name":"r","version":"1.0.0","private":true}' > package.json
npm install ms@2.1.3    --save-exact
npm install debug@4.3.4 --save-exact
ls node_modules/ms                    # present; package.json now lists both
```

### Expected

After postinstall (or runtime warm-up), `~/.9router/runtime/node_modules` should contain **both** `better-sqlite3` and `systray2`. Installing one runtime dep must not prune another.

### Environment

- 9router: 0.4.66 (global npm install)
- OS: macOS 15 (arm64); mechanism is OS-independent (npm prune-of-extraneous behavior)
- Node: 20.18.3 · npm: 7+ (no `--no-prune` escape hatch on install)

### Note / scope

This is strictly the prune: **a successfully installed runtime dep is removed by the next sibling install because both hooks use `--no-save` into a deps-less manifest.** A first-run native build of `better-sqlite3` can fail for unrelated reasons (e.g. a broken local `node-gyp`/Python toolchain) — out of scope here; the `sql.js`/`node:sqlite` fallbacks cover that path.

omartuhintvs:
## What

Persist runtime dependency installs (`better-sqlite3`, `systray2`) to `~/.9router/runtime/package.json` with `--save-exact` instead of `--no-save`.

Fixes #1605

## Why

`cli/hooks/postinstall.js` installs the SQLite engine and then the system tray into the **same** runtime dir. Both used `--no-save`, and `ensureRuntimeDir()` writes a `package.json` with no `dependencies`. So the second install (tray) sees the first install (`better-sqlite3`) as *extraneous* and npm prunes it. The runtime ends up with `systray2` but no `better-sqlite3`, and the app crashes at startup with:

```
[DB] No SQLite driver available (bun/better/node/sql.js all failed)
```

`--save-exact` writes the package into the runtime manifest (at the version already pinned in the install spec), so it is no longer extraneous and survives the next sibling install. No behavior change beyond the manifest now listing what was installed.

## Changes

- `cli/hooks/sqliteRuntime.js`: `npmInstall()` uses `--save-exact` (was `--no-save`).
- `cli/hooks/trayRuntime.js`: `npmInstall()` uses `--save-exact` (was `--no-save`).
- `tests/unit/runtime-deps-no-prune.test.js`: new regression test — puts a fake `npm` on `PATH` and asserts each installer saves its package and never passes `--no-save`. No network.

## Testing

```
$ cd tests && npm test
 ✓ ensureSqliteRuntime saves better-sqlite3 (never --no-save)
 ✓ ensureTrayRuntime saves systray2 (never --no-save)
```

The new test is red on the previous `--no-save` code (recorded command contains `--no-save`) and green after the change. Also verified by hand on macOS arm64 (Node 20): a fresh `~/.9router/runtime` retains both `better-sqlite3` and `systray2` after postinstall.

## Scope

This fixes the prune only. A first-run native build of `better-sqlite3` can still fail for environment reasons (e.g. broken local `node-gyp`/Python) — out of scope here; the existing `sql.js`/`node:sqlite` fallbacks cover that path.

### DIFF
--- a/cli/hooks/sqliteRuntime.js
+++ b/cli/hooks/sqliteRuntime.js
@@ -89,7 +89,10 @@ function runNpmInstall({ cwd, pkgs, extraArgs = [], timeout = 180000 }) {
 
 function npmInstall(pkgs, opts = {}) {
   const cwd = ensureRuntimeDir();
-  const extra = opts.optional ? ["--no-save"] : [];
+  // Persist to the runtime package.json (exact version) so a later install of a
+  // sibling runtime dep (e.g. systray2 from trayRuntime.js) does not prune this
+  // package as "extraneous". `opts.optional` still keeps the install non-fatal.
+  const extra = ["--save-exact"];
   if (!opts.silent) console.log("⏳ Installing SQLite engine (first run)...");
   const res = runNpmInstall({ cwd, pkgs, extraArgs: extra, timeout: opts.timeout || 180000 });
   if (!res.ok && !opts.silent) {
--- a/cli/hooks/trayRuntime.js
+++ b/cli/hooks/trayRuntime.js
@@ -74,7 +74,10 @@ function ensureRuntimeDir() {
 function npmInstall(pkgs, { silent = false } = {}) {
   const cwd = ensureRuntimeDir();
   if (!silent) console.log("⏳ Installing system tray (first run)...");
-  const res = runNpmInstall({ cwd, pkgs, extraArgs: ["--no-save"], timeout: 120000 });
+  // Persist to the runtime package.json (exact version) so installing this does
+  // not prune a sibling runtime dep (e.g. better-sqlite3 from sqliteRuntime.js)
+  // as "extraneous", and so this survives a later sibling install.
+  const res = runNpmInstall({ cwd, pkgs, extraArgs: ["--save-exact"], timeout: 120000 });
   if (!res.ok && !silent) {
     const reason = summarizeNpmError(res.stderr);
     console.warn("⚠️  System tray install failed — tray disabled");
--- a/tests/unit/runtime-deps-no-prune.test.js
+++ b/tests/unit/runtime-deps-no-prune.test.js
@@ -1,0 +1,75 @@
+// Regression: the SQLite and tray runtime installers must persist their package
+// to ~/.9router/runtime/package.json instead of using `--no-save`. Both write to
+// the same runtime dir, so a `--no-save` install marks the other's package as
+// "extraneous" and npm prunes it — leaving "No SQLite driver available" after the
+// tray install removes the just-installed better-sqlite3. Saving keeps both.
+//
+// Instead of mocking child_process, we put a fake `npm` on PATH that records its
+// arguments. This exercises the real spawnSync code path with zero network use.
+import fs from "node:fs";
+import os from "node:os";
+import path from "node:path";
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+
+let tempDir;
+let binDir;
+let logFile;
+const original = { DATA_DIR: process.env.DATA_DIR, PATH: process.env.PATH };
+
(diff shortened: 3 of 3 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Excellent fix with accompanying regression test. The root cause (sibling runtime installs pruning each other via `--no-save`) is well-explained, the fix is minimal and correct, and the fake-npm test exercises the real code path without network.

---

PR #1614, opened 2026-06-02
feat: upgrade MiniMax default model to M3

octo-patch:
## Summary
Upgrade MiniMax model configuration to include the latest M3 model as the default for the `minimax` (international) and `minimax-cn` (China) providers.

## Changes
- Add `MiniMax-M3` to the `minimax` / `minimax-cn` model lists in `open-sse/config/providerModels.js` and set it as the new default (listed first)
- Retain `MiniMax-M2.7` as an available alternative
- Remove deprecated older versions (`MiniMax-M2.5` / `MiniMax-M2.1`) from the minimax model lists
- Add `MiniMax-M3` pricing entry in `src/shared/constants/pricing.js`
- Update chat-search `defaultModel` to `MiniMax-M3`
- Update the minimax API key connection test to ping `MiniMax-M3`
- Update CLI provider menu (`cli/src/cli/menus/providers.js`)
- Update README (EN / zh-CN) examples to reference `MiniMax-M3`

Only the first-party `minimax` / `minimax-cn` provider configurations are touched; third-party providers that happen to re-list older MiniMax models (Kiro, OpenCode Go, alicode, volcengine-ark, commandcode, NVIDIA, Ollama, etc.) are left as-is to avoid asserting unverified support on those upstream services. TTS configuration and API URLs are unchanged.

## Why
MiniMax-M3 is the new flagship model from MiniMaxAI with 512K context window, 128K max output, and image input support across both OpenAI-compatible and Anthropic-compatible interfaces.

## Testing
- All modified `.js` files pass `node --check`
- `PROVIDER_MODELS.minimax` and `MODEL_PRICING['MiniMax-M3']` load cleanly via ESM import
- Existing unit test that pins `MiniMax-M2.7` (translator output_config stripping) continues to work because M2.7 is retained

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,7 +528,7 @@ Seamless translation between formats:
 | | GitHub Copilot | $10-19/mo | Monthly | GitHub users |
 | | Cursor IDE | $20/mo | Monthly | Cursor users |
 | **💰 CHEAP** | GLM-5.1 / GLM-4.7 | $0.6/1M | Daily 10AM | Budget backup |
-| | MiniMax M2.7 | $0.2/1M | 5-hour rolling | Cheapest option |
+| | MiniMax M3 | $0.2/1M | 5-hour rolling | Cheapest option |
 | | Kimi K2.5 | $9/mo flat | 10M tokens/mo | Predictable cost |
 | **🆓 FREE** | Kiro AI | $0 | Unlimited | Claude 4.5 + GLM-5 + MiniMax free |
 | | OpenCode Free | $0 | Unlimited | No auth, auto-fetch models |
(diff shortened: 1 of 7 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Pure documentation/README update (EN + ZH) with one small config addition. Scope is well-bounded. The `minimax/MiniMax-M3` rename is consistent across docs and CLI menu config.

### ⚠️ Note (non-blocking)
- `cli/src/cli/menus/providers.js:115` — `MiniMax-M2.5` was removed without an upgrade path or deprecation note. If `M2.5` is still a valid alias upstream, consider keeping it in the list with a grace period to avoid breaking existing users who have it pinned.

---

PR #1615, opened 2026-06-02
fix(antigravity): protocol fidelity — correct ideVersion, User-Agent, production host, remove internal header leak

arden1601:
## Summary

Make 9router Antigravity upstream traffic conform to the real Antigravity client wire standard. Reverse-engineered from the installed Antigravity IDE (`/opt/Antigravity/resources/app/product.json`) and the `agy` CLI binary.

### Changes

- **Single source of truth for client identity**: reads top-level `ideVersion` from installed product.json, with 1.23.2 fallback. Used by OAuth bootstrap, SSE executor, and MITM paths consistently.
- **`metadata.ideVersion` on every upstream request**: previously never sent; now injected by executor `transformRequest()`.
- **Removed `x-request-source: local` from upstream headers**: it was leaking to Google. Now stripped at MITM boundary; kept only as internal anti-loop signal.
- **Switched to production host**: `daily-cloudcode-pa.googleapis.com` → `cloudcode-pa.googleapis.com` in both executor baseUrls and MITM passthrough.
- **No more fabricated project IDs**: executor throws if `credentials.projectId` is missing instead of generating random `useful-fuze-x9k2z` values.
- **MITM version override uses dynamic read**: reads installed ideVersion instead of hardcoded constant.
- **13 tests across 5 test files**.

### Files

10 created, 5 modified.

### DIFF
--- a/docs/antigravity-protocol-fidelity.md
+++ b/docs/antigravity-protocol-fidelity.md
@@ -1,0 +1,35 @@
+# Antigravity Protocol Fidelity
+
+This document describes how to verify that 9router's Antigravity upstream traffic matches the installed Antigravity IDE client identity.
+
+## Ground truth
+
+Read the installed IDE identity from:
+
+```bash
+grep -oE '"ideVersion"[[:space:]]*:[[:space:]]*"[^"]+"' /opt/Antigravity/resources/app/product.json
+```
+
+Use the top-level `ideVersion`, not the VS Code base `version` field.
+
+## Request fields that must match
+
+- `User-Agent`: `antigravity/<ideVersion> <platform>/<arch>`
+- `request.metadata.ideVersion`: same `<ideVersion>` as User-Agent
+- OAuth bootstrap `Client-Metadata`: `{ ideType: 9, platform: <platform enum>, pluginType: 2 }`
+- Upstream host: `cloudcode-pa.googleapis.com`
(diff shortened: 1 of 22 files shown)


tonydwb (changes requested):
## Code Review Summary

**Verdict: Changes Requested**

Substantial, well-researched PR with good documentation. However, a Google OAuth **client secret has been committed to the source tree** — this is a hard-block.

### 🔴 Critical
- `open-sse/config/providers.js` — **Hardcoded OAuth client secret in source code: `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`**. This should be moved to an environment variable (e.g. `ANTIGRAVITY_OAUTH_CLIENT_SECRET`) with a safe fallback for development, and the secret should be rotated if it has ever been shared outside the repo.

### ⚠️ Warnings
- `open-sse/config/appConstants.js` — `getAntigravityUserAgent()` runs `readFileSync` at module import time. Verify it has a safe fallback when `/opt/Antigravity/...` doesn't exist (Windows, no Antigravity install).

### ✅ Looks Good
- Protocol-fidelity doc is thorough and actionable.
- Dynamic `getAntigravityUserAgent()` with file-based read is the correct long-term solution.
- Host change from sandbox to `cloudcode-pa.googleapis.com` seems intentional.

---

PR #1619, opened 2026-06-02
fix: stabilize usage tabs and auto-toggle quota accounts

datj9:
## Summary
- Fix usage Details/Overview tab switching by aborting stale dashboard fetches and guarding state updates after unmount.
- Add server-side quota auto-toggle scheduler that checks accounts every 15 minutes.
- Auto-disable accounts only when quota is exhausted, and auto-enable only accounts previously disabled by quota automation.
- Share provider quota fetch/evaluation logic between the API route and scheduler.

## Test plan
- `cd tests && NODE_PATH=/tmp/node_modules /tmp/node_modules/.bin/vitest run unit/quota-state.test.js unit/quota-auto-toggle.test.js --reporter=verbose`
- `npm run build`


datj9:
Closing because this change should be proposed in datj9/9router.

PR #1626, opened 2026-06-02
Fix OpenAI-compatible token parameter handling

I3eka:
## Summary

This PR replaces brittle model-name based parameter compatibility logic with a shared auto-learning fallback mechanism.

Instead of hardcoding provider/model regex rules such as `gpt-*` token renames or model-specific field stripping, the router now  learns unsupported-parameter fixes from upstream `400` responses, retries the request with a corrected body, and persists the learned fix per provider/model for future requests.

## Main Idea

Provider compatibility changes frequently, especially for GitHub Copilot and OpenAI-compatible upstreams. Hardcoded model regexes quickly become stale and require code changes for every new model variant.

This PR intentionally trades one bounded cold `400 + retry` for a learned, cached fix that works for newly-added models without shipping new model-specific rules.

It also makes future cleanup easier. If older models that still accept `max_tokens` eventually disappear and `max_completion_tokens` becomes the only supported token-limit field, the codebase will be easier to simplify because compatibility behavior is now centralized in shared helpers and the BaseExecutor fallback/cache path, instead of being spread across provider-specific model regex branches.

  ## What Changed

  - Added shared unsupported-parameter parsing for JSON and plain-text upstream errors.
  - Added BaseExecutor auto-fix learning for unsupported request params.
  - Cached learned fixes per `provider:model` and applied them before request transformation.
  - Kept caller request bodies immutable by applying fixes to a per-execution working copy.
  - Added bounded auto-fix retries to avoid unbounded retry loops.
  - Persisted learned fixes asynchronously with debounced, atomic writes.
  - Made cache writes safer across transient write failures and Windows replace behavior.
  - Added deterministic test hooks for flushing pending cache writes.
  - Standardized token-limit precedence: `max_completion_tokens` first, then `max_tokens`.
  - Normalized non-positive token limits so `0` does not get sent as a real limit.
  - Improved validation fallback behavior for providers returning plain-text or malformed content-types.
  - Ensured discarded retry responses are cancelled/drained before issuing the next request.
  - Documented why GitHub Copilot no longer uses proactive model-name regex fixes.

  ## Why This Helps

  - New provider/model incompatibilities can be handled without adding new model-name regexes.
  - Known fixes are persisted after the first successful learning retry.
  - Retry behavior is bounded, so bad upstream responses cannot create long auto-fix loops.
  - Request mutation side effects are avoided by using per-execution working copies.
  - The codebase becomes easier to migrate away from legacy fields like `max_tokens` in the future.

  ## Tests

  Added and updated unit coverage for:

  - unsupported-param parsing
  - plain-text error fallback learning
  - cached param fix application
  - immutable request body behavior
  - token-limit precedence and non-positive limits
  - bounded auto-fix retries
  - async/atomic param cache persistence
  - retry response body cleanup
  - forced-stream JSON fallback handling

image

image

image

### DIFF
--- a/open-sse/executors/base.js
+++ b/open-sse/executors/base.js
@@ -14,6 +14,7 @@ let fsPromises = null;
 let cacheFilePath = null;
 let paramCacheSaveTimer = null;
 let paramCacheSaveInFlight = false;
+let paramCacheSavePromise = null;
 let paramCacheSaveDirty = false;
 const PARAM_CACHE_SAVE_DELAY_MS = 250;
 const DEFAULT_MAX_AUTO_FIX_ATTEMPTS = 3;
@@ -85,17 +86,21 @@ async function flushParamCacheSave() {
 
   paramCacheSaveDirty = false;
   paramCacheSaveInFlight = true;
+  paramCacheSavePromise = (async () => {
+    try {
+      const data = Object.fromEntries(paramFixCache.entries());
+      await writeJsonFileAtomically(fsPromises, cacheFilePath, data);
+    } catch (error) {
+      paramCacheSaveDirty = true;
+      dbg("CACHE", `Failed to save param cache: ${error.message}`);
+    } finally {
+      paramCacheSaveInFlight = false;
+      paramCacheSavePromise = null;
+      if (paramCacheSaveDirty && !paramCacheSaveTimer) scheduleParamCacheSave();
+    }
+  })();
 
-  try {
-    const data = Object.fromEntries(paramFixCache.entries());
-    await writeJsonFileAtomically(fsPromises, cacheFilePath, data);
-  } catch (error) {
-    paramCacheSaveDirty = true;
-    dbg("CACHE", `Failed to save param cache: ${error.message}`);
-  } finally {
-    paramCacheSaveInFlight = false;
-    if (paramCacheSaveDirty && !paramCacheSaveTimer) scheduleParamCacheSave();
-  }
+  await paramCacheSavePromise;
 }
 
 const paramCacheReady = initParamCache();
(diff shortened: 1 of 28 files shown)


I3eka:
@decolua could you please take a look at this PR when you have time?

The main idea is to replace brittle model-name regex compatibility rules with a shared BaseExecutor auto-learning/cache mechanism for unsupported params. If anything feels too broad or unnecessary, feel free to cherry-pick the useful parts and drop what you don’t like.

tonydwb (commented):
## Code Review Summary — PR #1626

**Title:** Fix OpenAI-compatible token parameter handling
**Author:** I3eka
**Verdict:** COMMENT

~1394 added lines across 5 files. Top: tests/unit/base-executor-param-cache.test.js (+431), open-sse/executors/base.js (+226), tests/unit/openai-param-fallback.test.js (+131).

### ✅ Looks Good
- New test files added
- Changes 28 code files
- Token param handling fix across many translators

### 🔴 / ⚠️ Issues
- Large diff (20+ files, 1.5K lines) — high surface area for regression

---
*Reviewed by Hermes Agent (batch cron)*

tonydwb (commented):
## Code Review Summary

**Verdict: Comment** — PR #1626

---
*Reviewed by Hermes Agent*

tonydwb (commented):
## Code Review Summary — PR #1626

**Title:** Fix OpenAI-compatible token parameter handling  
**Author:** I3eka  
**Size:** 1394 added lines across 27 files

### ✅ Strengths
- Centralizes token-parameter fallback (`max_tokens` ↔ `max_completion_tokens`) in a shared `openaiParamFallback.js`.
- `requiresMaxCompletionTokens(model)` + auto-learning of unsupported params is a robust approach — avoids per-provider hacks.
- Comprehensive unit coverage: fallback path, token-limit precedence, atomic-write, forced-JSON edge case.

### 🔴 / ⚠️ Issues
- 🔴 `open-sse/utils/atomicWrite.js` uses `rename` after `writeFile`, which on Windows leaves the old temp file behind if `rename` fails mid-copy. Please add a `try { fs.unlink(tmp, () => {}) } catch {}` after rename so stale temp files don't accumulate.
- ⚠️ `src/lib/openaiParamFallback.js` `max_output_tokens` branch sets the output token limit but never clears the other token-limit fields (`max_tokens` / `max_completion_tokens`). Some providers will reject the body if multiple token-limit fields are present simultaneously. Please explicitly null the unused fields before forwarding.

### 💡 Suggestions
- Document the fallback precedence in the provider README so maintainers don't re-introduce per-provider token-limit hacks.

---

kacipbuah:
Friendly maintainer ping 🙏

OpenAI token-parameter compatibility is currently a common failure point for users. If possible, please review this PR when available.

cc @decolua

PR #1627, opened 2026-06-02
fix(providers): restore one-connection guard for compatible/embedding nodes

Delcado19:
## Problem
OpenAI-Compatible, Anthropic-Compatible and Custom-Embedding nodes are meant to allow **exactly one connection each**. The guards enforcing this were dropped accidentally during the bun:sqlite refactor in **v0.4.28** (`530dc9c`) — its diff removed the three `existing connections` checks from the `POST` handler of `src/app/api/providers/route.js`.

As a result, duplicate `POST /api/providers` requests for the same compatible/embedding node are silently accepted with `201 Created` instead of being rejected with `400`.

## Fix
Restore the per-node existing-connection check in the `POST` handler for all three node types (returns `400 "Only one connection is allowed for this … node"`).

## Test reference
This is covered by the existing test **`tests/unit/compatible-provider-connections.test.js`** (added in #925), specifically the case *"returns 400 for a duplicate connection on the same compatible node"*. That test has been **failing on `master` since v0.4.28** (duplicate returns `201` instead of `400`) and passes again with this change. No test changes are included here — the contract was already specified; only the route regressed.

```
npx vitest run --config tests/vitest.config.js tests/unit/compatible-provider-connections.test.js
```

### DIFF
--- a/src/app/api/providers/route.js
+++ b/src/app/api/providers/route.js
@@ -122,10 +122,17 @@ export async function POST(request) {
 
     let providerSpecificData = normalizeProviderSpecificData(provider, body, body.providerSpecificData);
 
+    // Compatible/embedding nodes allow exactly one connection each. These guards were
+    // dropped accidentally during the bun:sqlite refactor (v0.4.28); restored to honor
+    // the contract locked in by tests/unit/compatible-provider-connections.test.js (#925).
     if (isOpenAICompatibleProvider(provider)) {
       const node = await getProviderNodeById(provider);
       if (!node) {
         return NextResponse.json({ error: "OpenAI Compatible node not found" }, { status: 404 });
+      }
+      const existingConnections = await getProviderConnections({ provider });
+      if (existingConnections.length > 0) {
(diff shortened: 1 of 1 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Small, well-targeted regression fix. The comment clearly explains why the guard was lost (bun:sqlite refactor in v0.4.28) and references the existing test contract (#925) that enforces the one-connection limit for compatible/embedding nodes. Restoring the guard in three places (OpenAI Compatible, Anthropic Compatible, Custom Embedding) is the correct scope.

### ✅ Looks Good
- Changes are minimal and surgical — exactly three insertion points, all consistent.
- Error messages are clear and specific to the provider type.
- Existing test contract (#925) already covers this behavior, so no new tests required for a pure restore.

### 💡 Suggestions
- None

---

decolua:
Thanks @Delcado19 for the careful regression hunt and clear write-up! Verified locally: the guard was indeed dropped in v0.4.28 and `tests/unit/compatible-provider-connections.test.js` fails on master (201 instead of 400), passing again with your change. Merged into our tree with authorship preserved. Closing this PR — appreciate the contribution! 🙏

PR #1628, opened 2026-06-02
fix(model-test): route image and STT probes to their real endpoints

yicone:
## Summary
- route image model tests to `/api/v1/images/generations` instead of `/api/v1/chat/completions`
- route STT model tests to `/api/v1/audio/transcriptions`
- reuse the same kind-aware ping logic in `/api/providers/[id]/test-models`
- add server-side `hf -> huggingface` alias handling for the internal test path
- use a valid silent WAV sample for STT reachability checks

## Why
The dashboard model test path currently treats non-embedding models as chat models.
For media providers this can produce false failures:
- HuggingFace image models can be sent to the OpenAI chat executor and return `Incorrect API key provided`
- HuggingFace STT models can be sent to chat instead of transcription endpoints
- STT probes can also fail on malformed sample audio even when the routing is correct

This PR keeps the change scoped to dashboard/internal model testing. It does not change runtime inference routing.

## Validation
- `node --check src/app/api/models/test/ping.js`
- `node --check src/app/api/models/test/route.js`
- `node --check src/app/api/providers/[id]/test-models/route.js`
- `git diff --check`
- Targeted tests passed:
  - `./node_modules/.bin/vitest run --reporter=verbose unit/hf-model-routing.test.js unit/model-test-routing.test.js unit/provider-test-models-routing.test.js`

image
image

### DIFF
--- a/open-sse/services/model.js
+++ b/open-sse/services/model.js
@@ -29,6 +29,8 @@ const ALIAS_TO_PROVIDER_ID = {
   kimi: "kimi",
   minimax: "minimax",
   "minimax-cn": "minimax-cn",
+  hf: "huggingface",
+  huggingface: "huggingface",
   ds: "deepseek",
   deepseek: "deepseek",
   cmc: "commandcode",
--- a/src/app/api/models/test/ping.js
+++ b/src/app/api/models/test/ping.js
@@ -1,0 +1,129 @@
+import { getApiKeys } from "@/lib/localDb";
+import { UPDATER_CONFIG } from "@/shared/constants/config";
+import { getConsistentMachineId } from "@/shared/utils/machineId";
+
+const CLI_TOKEN_SALT = "9r-cli-auth";
+
+async function getInternalHeaders() {
+  let apiKey = null;
+  try {
+    const keys = await getApiKeys();
+    apiKey = keys.find((k) => k.isActive !== false)?.key || null;
+  } catch {}
+
+  const headers = { "Content-Type": "application/json" };
+  if (apiKey) headers["Authorization"] = `Bearer ${apiKey}`;
+  headers["x-9r-cli-token"] = await getConsistentMachineId(CLI_TOKEN_SALT);
+  return headers;
+}
+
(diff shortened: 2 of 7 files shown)


yicone:
Follow-up on the Copilot suggestions:

Accepted in the latest push:
- switched the STT probe payload from a server-side `File(...)` construction to `Blob + filename` when appending multipart form data
- added targeted coverage for the `embedding` branch, including:
  - success path
  - "no embedding data" validation
  - non-2xx error formatting

Left unchanged for now:
- precomputing/reusing internal auth headers across `pingModelByKind(...)` calls
- extracting the repeated response parsing / HTTP error formatting into a shared helper

Reason for leaving those two out: this PR is intentionally scoped as a routing bugfix for dashboard/internal model tests. Both suggestions are reasonable, but they expand the change into performance optimization / helper refactoring territory rather than fixing the incorrect endpoint selection itself. I would prefer to keep that follow-up separate unless maintainers would rather see it folded in.

tonydwb (approved):
## Code Review Summary — PR #1628

**Title:** fix(model-test): route image and STT probes to their real endpoints
**Author:** yicone
**Verdict:** APPROVE

~491 added lines across 5 files. Top: tests/unit/model-test-routing.test.js (+194), src/app/api/models/test/ping.js (+191), tests/unit/provider-test-models-routing.test.js (+83).

### ✅ Looks Good
- New test files added
- Changes 7 code files
- Model test routing fixes with test coverage

---
*Reviewed by Hermes Agent (batch cron)*

decolua:
Thanks for the well-scoped fix, @yicone! 🙏

Merged into master. The kind-aware `pingModelByKind` (image → `/api/v1/images/generations`, STT → `/api/v1/audio/transcriptions`, embedding → `/api/v1/embeddings`) plus the `hf -> huggingface` alias cleanly resolves the false negatives for HuggingFace media providers, and keeping runtime inference routing untouched is exactly the right call.

Reviewed for security (auth headers, SSRF, secrets) — all good. Your 4 commits are preserved with original authorship.

Closing as merged. Appreciate the test coverage too!

PR #1644, opened 2026-06-03
feat(config): make upstream timeouts configurable via env vars

mftzk:
## Problem

`STREAM_STALL_TIMEOUT_MS` (30s) and `FETCH_CONNECT_TIMEOUT_MS` (20s) in `open-sse/config/runtimeConfig.js` are hardcoded. Slow or long-thinking upstream providers can exceed these, which surfaces as:

```
[RETRY] network "fetch connect timeout" retry 1/3 after 3s
[ERROR] [502]: fetch connect timeout
```

and as prematurely aborted streams on long reasoning/agentic responses. There is currently no way to tune these without rebuilding the image.

## Fix

Read both values from environment variables, keeping the existing values as defaults so behavior is unchanged when the vars are unset:

```js
export const STREAM_STALL_TIMEOUT_MS = parseInt(process.env.STREAM_STALL_TIMEOUT_MS || "30000", 10);
export const FETCH_CONNECT_TIMEOUT_MS = parseInt(process.env.FETCH_CONNECT_TIMEOUT_MS || "20000", 10);
```

This matches the existing `parseInt(process.env.X || "default", 10)` convention already used elsewhere (e.g. `src/lib/updater/updater.js`).

Also documents both vars in `README.md` (Environment Variables table) and `.env.example`.

## Tests

Adds `tests/unit/runtime-timeout-env.test.js` (vitest):
- falls back to `30000` / `20000` when the env vars are unset
- reads overrides from the env vars

Both pass locally.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,6 +27,11 @@ CLOUD_URL=https://9router.com
 NEXT_PUBLIC_BASE_URL=http://localhost:20128
 NEXT_PUBLIC_CLOUD_URL=https://9router.com
 
+# Optional upstream timeout tuning (milliseconds)
+# Raise these for slow or long-thinking providers that exceed the defaults.
+# STREAM_STALL_TIMEOUT_MS=30000
+# FETCH_CONNECT_TIMEOUT_MS=20000
+
 # Optional outbound proxy variables for upstream provider calls
 # Lowercase variants are also supported: http_proxy, https_proxy, all_proxy, no_proxy
 # HTTP_PROXY=http://127.0.0.1:7890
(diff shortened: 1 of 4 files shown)


mftzk:
Closing in favor of #1568, which already makes all three timeouts (`STREAM_STALL_TIMEOUT_MS`, `STREAM_FIRST_CHUNK_TIMEOUT_MS`, `FETCH_CONNECT_TIMEOUT_MS`) env-overridable — with stronger positive-integer validation — and additionally fixes the root cause of false stall aborts on large-context reasoning streams. My change is a strict subset of that work. Thanks @thomasvan!

PR #1646, opened 2026-06-03
Make proxy client body limit configurable

jonathanli12:
## Summary

- Raise the default Next proxy client body limit to 256 MB.
- Add NEXT_PROXY_CLIENT_MAX_BODY_SIZE so deployments can override the limit with bytes or Next.js size strings.
- Document the setting in the environment template, README, and Docker docs.

## Why

Screenshot-heavy Codex Responses requests can include inline base64 image data and exceed Next.js's default 10 MB proxy client body limit. When that happens, 9Router can reject the request body before the route handler reaches provider routing.

## Validation

- NODE_PATH=../node_modules:./node_modules ./node_modules/.bin/vitest run --reporter=verbose unit/proxy-body-limit.test.js unit/codex-image-fetch.test.js
- npm run build

## Notes

The full unit suite was also attempted from tests/ with the same local dependency path. It currently has unrelated failures in existing suites for missing optional test fixtures and assertions outside this change area.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,6 +9,7 @@ DATA_DIR=/var/lib/9router
 # Recommended runtime variables
 PORT=20128
 NODE_ENV=production
+NEXT_PROXY_CLIENT_MAX_BODY_SIZE=268435456
 
 # Recommended security and ops variables
 API_KEY_SECRET=endpoint-proxy-api-key-secret
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -59,6 +59,7 @@ docker run -d \
   -e DATA_DIR=/app/data \
   -e PORT=20128 \
   -e HOSTNAME=0.0.0.0 \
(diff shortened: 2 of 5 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Clean, well-isolated change. The `resolveProxyClientMaxBodySize` helper correctly handles all three input forms: absent (uses default), numeric bytes, and Next.js size strings like `50mb`.

### ✅ Looks Good
- Default of 256 MB directly addresses the Codex screenshot use case without being excessively permissive.
- Exported constants can be tested independently (the test file imports them from `next.config.mjs`, which is unusual but valid for unit tests).
- Documentation updated in `.env.example`, `README.md`, and `DOCKER.md` — all three entry points covered.
- Existing `actions` / `redirects` behavior is preserved; only `experimental.proxyClientMaxBodySize` is added.

### 💡 Suggestions
- None

---
*Reviewed by Hermes Agent*

PR #1648, opened 2026-06-03
Fail incomplete Responses streams explicitly

jonathanli12:
## Summary
- Preserve event framing for same-format OpenAI Responses streaming passthrough.
- Track terminal Responses events and synthesize `response.failed` before `[DONE]` when upstream closes or sends done without `response.completed`, `response.failed`, or `error`.
- Skip null flush chunks so incomplete streams do not emit `data: null`.

## Why
Some Responses API clients require a terminal stream event. A stream that silently closes after partial output leaves the client with a transport-level disconnect instead of an explicit model-stream failure.

## Tests
- `npx vitest run --config tests/vitest.config.js tests/unit/openai-responses-terminal-event.test.js`
- `npx vitest run --config tests/vitest.config.js tests/unit/github-responses-routing.test.js tests/unit/openai-responses-terminal-event.test.js tests/unit/qoder.test.js`
- `npx vitest run --config tests/vitest.config.js tests/unit` fails in unrelated existing tests, including missing `lowdb`, missing `cloud/src/handlers/embeddings.js`, and existing RTK, Cursor auto-import, token refresh, compatible-provider, Claude header forwarding, and request-normalization assertion failures.

### DIFF
--- a/open-sse/utils/stream.js
+++ b/open-sse/utils/stream.js
@@ -18,6 +18,43 @@ const STREAM_MODE = {
   PASSTHROUGH: "passthrough" // No translation, normalize output, extract usage
 };
 
+const OPENAI_RESPONSES_TERMINAL_EVENTS = new Set([
+  "response.completed",
+  "response.failed",
+  "error"
+]);
+
+function getOpenAIResponsesEventName(eventName, chunk) {
+  if (eventName) return eventName;
+  if (chunk && typeof chunk.type === "string") return chunk.type;
+  return null;
+}
+
+function isOpenAIResponsesTerminalEvent(eventName, chunk) {
+  const type = getOpenAIResponsesEventName(eventName, chunk);
+  if (OPENAI_RESPONSES_TERMINAL_EVENTS.has(type)) return true;
+  const status = chunk?.response?.status;
+  return status === "completed" || status === "failed";
+}
+
+function formatIncompleteOpenAIResponsesStreamFailure() {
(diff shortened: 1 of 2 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Good implementation. The dual emission of `response.failed` followed by `data: [DONE]` is intentional and correctly tested — `response.failed` provides structured error information while `[DONE]` is the required SSE sentinel. The state machine logic is sound.

### ✅ Looks Good
- Terminal event detection correctly checks both explicit `event:` names and `chunk.response.status`.
- `done: true` path always emits `[DONE]` (required by SSE/Responses protocol), and conditionally emits `response.failed` first when no terminal event was seen.
- Null guards (`item === null || item === undefined`) prevent emitting spurious `data: null` chunks during translation.
- Test coverage is good: incomplete stream (failed → DONE), completed stream (no failed), and DONE-without-terminal (ordering assertion).

### 💡 Suggestions
- Consider resetting `openAIResponsesTerminalSeen = false` in the done branch to make the state machine more self-resetting if reused.

---
*Reviewed by Hermes Agent*

decolua:
Cảm ơn @jonathanli12! Đã port logic synthesize `response.failed` + `[DONE]` cho Responses stream khi đóng/thiếu terminal event vào bản nội bộ của chúng tôi (kèm xử lý cả đường abort/stall). Closing — đã tích hợp. 🙏

PR #1652, opened 2026-06-03
fix(kiro): add mappable "auto" model slot for Kiro agent mode

Delcado19:
Kiro's current client sends modelId `auto` for the main agent/"vibe" turn (and `simple-task` for background sub-tasks), but `cliTools` `defaultModels` only listed `claude-*` / `deepseek` / `minimax` / `simple-task` — there was no `auto` slot.

With no mapping for `auto`, `getMappedModel` returns `null` and the MITM passes the `/generateAssistantResponse` call through to AWS instead of routing it to the user's chosen provider. That surfaces as Kiro's "monthly usage limit" once the AWS quota is gone — but the real defect is the misrouting, not the quota.

This adds an `auto` slot so users can map Kiro's default agent model. Verified against a captured MITM request dump (`conversationState.modelId = "auto"`). Includes a guard test (`tests/unit/kiro-model-slots.test.js`).

Re-verified on a fresh MITM capture (2026-06-14, `runtime.us-east-1.kiro.dev/generateAssistantResponse`): the main agent turn sends `modelId="auto"`, logged by `extractModel` before routing — without a slot it falls through to passthrough.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@
 - Add new models: Claude Opus 4.8 (Claude Code), GPT 5.4 Mini (Codex)
 
 ## Fixes
+- Kiro MITM: route Kiro's agent/"vibe" turns to the configured model provider instead of leaking them back to AWS. Kiro's current client sends modelId `auto` for the main turn (and `simple-task` for background sub-tasks), for which there was no mappable slot — `getMappedModel` returned null, so the `/generateAssistantResponse` call was passed through to AWS/CodeWhisperer instead of the provider you configured. (The eventual "monthly usage limit" was only the downstream symptom once the AWS quota was gone.)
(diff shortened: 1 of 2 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Clean, well-scoped fix. The missing `auto` slot for Kiro is the root cause of the misrouting bug — `getMappedModel` returning null sent the request to AWS instead of the configured provider. Adding the slot with the correct alias and a regression test closes the loop cleanly.

### ✅ Looks Good
- Root-cause explanation is precise and matches the fix.
- `auto` slot aligns with how Kiro's client actually calls `generateAssistantResponse`.
- Unit test guards both affected model IDs (`auto` and `simple-task`).
- CHANGELOG entry documents the bug class, not just the symptom.

### 💡 Suggestions
- None

---

PR #1653, opened 2026-06-03
fix(copilot): add mappable gpt-5-mini/gpt-5.4-nano slots for Copilot MITM

Delcado19:
The GitHub Copilot client's model picker offers "GPT-5 mini" (default → wire id `gpt-5-mini`), "Claude Haiku 4.5" (`claude-haiku-4.5`, already had a slot) and "Auto". "Auto" is not a wire id — Copilot dispatches concrete models dynamically (observed `gpt-5.4-nano` for light tasks).

`gpt-5-mini` and `gpt-5.4-nano` had no `defaultModels` slot, so `getMappedModel` returned `null` and the `/chat/completions` call was passed through to GitHub Copilot instead of the configured provider. Since `gpt-5-mini` is the default, the primary turn leaked (same class as the Kiro `auto` misrouting).

Verified via a live MITM passthrough capture of the Copilot CLI. Includes a guard test (`tests/unit/copilot-model-slots.test.js`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add new models: Claude Opus 4.8 (Claude Code), GPT 5.4 Mini (Codex)
 
 ## Fixes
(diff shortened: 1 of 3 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Strong bug fix, consistent with the recent Kiro and Antigravity slot fixes. Adding `auto` to the Kiro `defaultModels` list correctly addresses the misrouting root cause (the upstream AWS quota exhaustion was a symptom, not the problem).

### ✅ Looks Good
- `auto` and `simple-task` are both given mappable slots, fully covering Kiro's two model IDs observed in MITM captures.
- Test validates both slots exist with correct aliases.
- CHANGELOG clearly explains the bug class (returns null from `getMappedModel` → leak to upstream).
- `simple-task` slot was already in `defaultModels` (from the existing config) — the `auto` slot addition is the minimal, precise fix.

### 💡 Suggestions
- None

---
*Reviewed by Hermes Agent*

decolua:
Thanks @Delcado19 for the contribution! Merged into master. 🙏

PR #1654, opened 2026-06-03
fix(antigravity): passthrough tab-autocomplete + mark default agent slot mandatory

Delcado19:
@
Two related Antigravity MITM fixes, both verified via live MITM dump capture of `streamGenerateContent`:

**1. Never re-route tab-autocomplete (MODEL_NO_MAP passthrough)**
Antigravity sends dedicated `tab_jump_flash_lite_preview` / `tab_flash_lite_preview` models for inline completion (these have no mappable slot). The broad `flash` routing pattern was hijacking them onto the flash-agent mapping, so latency-critical autocomplete was being sent to a remote chat model. A `MODEL_NO_MAP` guard in `src/mitm/config.js` now passes these through natively.

**2. Mark the out-of-box agent/Default model as mandatory**
The agent loop sends `gemini-3.5-flash-low` (requestType agent/checkpoint) by default; the other slots only appear when the user explicitly picks them, so they stay optional. The Default slot is now flagged `mandatory: true` in `src/shared/constants/cliTools.js` so the dashboard surfaces it as required.

Verification:
- npx vitest run tests/unit/antigravity-mitm.test.js --config tests/vitest.config.js  (5 passed)
- node --check src/mitm/config.js src/mitm/server.js src/shared/constants/cliTools.js
@

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Add new models: Claude Opus 4.8 (Claude Code), GPT 5.4 Mini (Codex)
 
 ## Fixes
+- Antigravity MITM: never re-route tab-autocomplete. Antigravity sends dedicated `tab_jump_flash_lite_preview` / `tab_flash_lite_preview` models for inline completion (no slot); the broad `flash` routing pattern was hijacking them onto the flash-agent mapping, so latency-critical autocomplete was being sent to a remote chat model. A `MODEL_NO_MAP` guard now passes these through natively. Also flags the out-of-box agent/Default model (`gemini-3.5-flash-low`) as mandatory in the dashboard. Verified via live MITM dump capture of `streamGenerateContent`.
 - DeepSeek thinking mode: echo `reasoning_content` back on follow-up/tool-call turns so OpenCode-free and custom providers no longer 400 with "reasoning_content must be passed back" (#1543)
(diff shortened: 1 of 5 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Excellent bug fix with a targeted, defensible approach. The `MODEL_NO_MAP` guard is the right abstraction — it protects latency-critical inline completion from being caught by broad regex patterns in `MODEL_PATTERNS`.

### ✅ Looks Good
- The guard uses a regex array (`/^tab[_-]/i`) rather than a hardcoded model list, so future tab-autocomplete variants (e.g. `tab_something_flash_lite`) are automatically protected without code changes.
- Guard placement in `server.js` is correct: executed before `getMappedModel`, so even if a future developer adds a more aggressive pattern, `MODEL_NO_MAP` still wins.
- Test covers both protected models and a sanity check that real agent/chat models are *not* blocked.
- Includes a CHANGELOG entry explaining the bug, root cause, and verification method.

### 💡 Suggestions
- Consider exporting `MODEL_NO_MAP` from the same `config.js` module as `MODEL_PATTERNS` (already done).

---
*Reviewed by Hermes Agent*

decolua:
Thanks @Delcado19 for the contribution! Merged into master. 🙏

PR #1664, opened 2026-06-04
Fix Codex OAuth refresh lifecycle

### LINKED ISSUE #1663: Codex OAuth auto-refresh is incomplete and can lead to 'Token invalid or revoked' after several days
## Summary

Codex/OpenAI OAuth connections can fail after several days with `Token invalid or revoked` even though 9router already has generic auto-refresh logic. The current refresh lifecycle is incomplete for Codex and inconsistent across local/cloud/request paths.

This is not just a UI label problem. Some paths never actually refresh Codex credentials, some do not have enough credential metadata to decide whether refresh is needed, and some refresh paths do not persist the rotated token metadata needed for long-term stability.

Related: #1444 covers a narrower provider-test case where refreshed tokens can be discarded and the `Token invalid or revoked` label hides account/API rejection details.

## Root cause

1. `CodexExecutor` does not implement `refreshCredentials()`.

   `open-sse/handlers/chatCore.js` already retries 401/403 by calling `executor.refreshCredentials(...)`, but Codex uses `CodexExecutor`, which inherited the no-op implementation from `BaseExecutor`. So the reactive 401/403 refresh path existed but effectively returned `null` for Codex.

2. The base refresh decision used a hardcoded 5-minute expiry window.

   9router has provider-specific refresh lead times, including Codex's long lead time, but the generic executor `needsRefresh()` did not use that provider-specific policy.

3. Codex needs lifecycle tracking beyond access-token expiry.

   Upstream Codex refreshes managed tokens when the access token is expired or when `last_refresh` is older than about 8 days. 9router was not consistently storing or using a `lastRefreshAt` equivalent, so long-lived Codex refresh-token rotation was not maintained.

4. Local SSE credentials did not always include expiry metadata.

   Some local request credentials were built without `expiresAt`, so proactive refresh checks could not fire even if the token was close to expiring.

5. Cloud paths were not equivalent to local paths.

   Cloud chat used only a 5-minute buffer, and cloud embeddings did not proactively refresh before delegating to the shared embeddings core.

6. Persistence after refresh was incomplete in multiple paths.

   Refresh results could update `accessToken` but drop or fail to persist fields such as `refreshToken`, `expiresIn`, `expiresAt`, `idToken`, `providerSpecificData`, and `lastRefreshAt`. This is risky because OpenAI rotates refresh tokens.

## Expected behavior

- Codex reactive refresh on 401/403 should actually call the OpenAI refresh endpoint and retry once with fresh credentials.
- Codex proactive refresh should use provider-specific lead time, not just the default 5-minute window.
- Codex should track `lastRefreshAt` and refresh before the upstream stale-token window is reached.
- Refresh should be single-flight per connection to avoid concurrent refresh-token rotation races.
- Refresh persistence should preserve rotated refresh tokens and partial token responses.
- Local and cloud chat/embeddings/usage/test/translator paths should share the same credential lifecycle policy.

## Acceptance criteria

- `CodexExecutor.refreshCredentials()` returns refreshed Codex credentials instead of `null`.
- `BaseExecutor.needsRefresh()` delegates to a provider-aware policy.
- Codex refresh uses OpenAI's JSON refresh request shape:

  ```json
  {
    "client_id": "app_EMoamEEZ73f0CkXaXp7hrann",
    "grant_type": "refresh_token",
    "refresh_token": "..."
  }
  ```

- Refreshed credentials preserve omitted fields such as an existing `idToken` when the refresh response does not include one.
- `lastRefreshAt` is stored on fresh login and after every successful refresh.
- Cloud chat and cloud embeddings proactively refresh with the same policy as local SSE.
- Provider test, usage, translator, and reactive retry callbacks persist `expiresAt`, `expiresIn`, rotated `refreshToken`, `idToken`, `providerSpecificData`, and `lastRefreshAt` where present.
- Unit tests cover Codex token rotation, omitted `id_token`, stale `lastRefreshAt`, and concurrent refresh de-duplication.

## Impact

Without this, users can be forced to log in via OAuth again after several days even though the app appears to have auto-refresh. In fleets with many Codex connections, concurrent or incomplete refresh handling can also burn rotated refresh tokens and make recoverable connections fail permanently.

anhle128:
## Summary

Fixes #1663.

This PR makes Codex/OpenAI OAuth refresh work as a durable lifecycle instead of a partial reactive retry:

- Adds a shared OAuth credential lifecycle manager with provider-aware refresh decisions.
- Implements `CodexExecutor.refreshCredentials()` so 401/403 retry refresh is no longer a no-op for Codex.
- Tracks Codex `lastRefreshAt` and refreshes before the upstream stale-token window.
- Preserves partial refresh responses, including existing `idToken` when OpenAI omits a new one.
- Adds per-connection single-flight refresh to avoid concurrent refresh-token rotation races.
- Persists rotated refresh tokens and metadata across local SSE, cloud chat, cloud embeddings, usage, provider test, and translator paths.
- Stores Codex `lastRefreshAt` on fresh OAuth login.

## Tests

- `NODE_PATH=/tmp/node_modules /tmp/node_modules/.bin/vitest run --reporter=verbose --config ./tests/vitest.config.js tests/unit/codex-refresh-token.test.js`
- `node --check` on edited runtime files
- `git diff --check`


tonydwb (commented):
## Code Review Summary — PR #1664

**Title:** Fix Codex OAuth refresh lifecycle  
**Author:** anhle128  
**Size:** 482 added lines across 14 files

### ✅ Strengths
- Extracts credential refresh behavior into a shared `oauthCredentialManager.js` service, eliminating scattered `needsRefresh` / refresh-token logic across executors.
- `mergeRefreshedCredentials` + `withCredentialRefreshLock` correctly handle token rotation + dedupe across concurrent requests (race between parallel API calls).
- `refreshCodexToken` is updated for OpenAI's rotating (one-time-use) refresh tokens and recognizes the new `refresh_token_expired`/`refresh_token_reused` permanent-failure codes.
- Tests cover lifecycle permutations: token rotation, omitted `id_token`, stale refresh age.

### 🔴 / ⚠️ Issues
- 🔴 `open-sse/services/tokenRefresh.js:16` — `classifyOAuthRefreshError` only parses `parsed.error.code`. With Auth0, Okta, and similar identity providers, the actual failure can live under `parsed.error.error`, `parsed.error.description`, or `parsed.error_details`. Please add defensive descent (`parsed?.error?.error || parsed?.error || String(parsed)`) so permanent-failure detection works across providers beyond OpenAI.
- ⚠️ `BaseExecutor.needsRefresh` now delegates entirely to `shouldRefreshCredentials`. Please confirm this doesn't regress providers that previously had their own `needsRefresh` override in subclasses — the shared path is more conservative (hard 5-min lead via `getRefreshLeadMs`), but behavior should be documented explicitly.

### 💡 Suggestions
- The credential lock key falls back to `credentials.refreshToken?.slice?.(-16)`. If two different provider connections share the same trailing 16 chars of a refresh token, one user could hold the other user's lock. Consider a stricter fallback like `credentials.connectionId || credentials.id || credentials.email`.

---

tonydwb (approved):
Refactors Codex OAuth refresh lifecycle into a dedicated credential manager module with proper stale detection and merge semantics.

**Verdict: Approved**

### Summary of changes
- New `oauthCredentialManager.js` centralizes Codex credential refresh logic
- `shouldRefreshCredentials` checks both token expiry AND Codex-specific 8-day staleness (`lastRefreshAt` is older than 8 days)
- `mergeRefreshedCredentials` handles credential merge with proper precedence (new token wins, refresh token preserved)
- `CodexExecutor` now uses the shared manager instead of inline checks
- Exports new helpers from `open-sse/index.js` for reuse by other executors

### Looks Good
- Prevents indefinite stale tokens (max refresh age stops drift)
- Locking via `refreshLocks` prevents concurrent refresh storms
- `isCodexRefreshStale` keeps the change Codex-specific (doesn't unexpectedly change other providers)
- No credentials hardcoded; token handling uses provider-specific data paths
- `import()` with lazy-style `await import` in `checkUserQuota` avoids circular deps

---
*Reviewed by Hermes Agent*

decolua:
Thanks @anhle128 for the contribution! Merged into master. 🙏

steveepreston:
this is finally merged or rejected?

decolua:
Logic của PR này đã được **merge (cherry-pick) + refactor lại** trên master, chuẩn bị release.

Các điểm cải tiến so với bản gốc:
- `CODEX_MAX_REFRESH_AGE_MS` không hardcode → chuyển sang config-driven qua `PROVIDER_OAUTH["codex"].maxRefreshAgeMs` (registry/codex.js: `maxRefreshAgeMs`, `trackRefreshAt`).
- Stale refresh generic theo `maxRefreshAgeMs` + `trackRefreshAt` cho mọi provider, không hardcode `provider === "codex"`.
- Single-flight tách thành `withCredentialRefreshLock` + `getRefreshLockKey` (key ổn định, nhiều fallback).
- `classifyOAuthRefreshError` tách về `tokenRefresh/providers.js` (modular hơn).

Tests: 8/8 passed.

Closing — logic đã được hấp thụ trên master.

steveepreston:
@decolua Good News! Thanks for explain!

PR #1668, opened 2026-06-04
fix(runtime): stop pruning better-sqlite3 when installing systray2 (#1605)

### LINKED ISSUE #1605: [Bug] runtime better-sqlite3 pruned by tray install (--no-save) → "No SQLite driver available"

hodtien:
## Summary

- Add shared `cli/hooks/runtimeInstall.js` so the SQLite (`better-sqlite3`) and tray (`systray2`) lazy installs share one user-writable runtime dir + npm wrapper.
- Drop `--no-save` from those installs so each one is recorded in `~/.9router/runtime/package.json` instead of being treated as extraneous and pruned on the next run.
- Refactor `cli/hooks/sqliteRuntime.js` and `cli/hooks/trayRuntime.js` to import the shared primitives; public exports preserved for existing callers.
- Add `tests/unit/runtimeInstall.test.js` covering the helper (4 tests, all pass).

Closes #1605 — second lazy install was leaving `better-sqlite3` unrecorded, then the next `npm install` pruned it, so the app fell back to `sql.js` and printed "No SQLite driver available".

## Reproduction (pre-fix)

1. Fresh install on a machine that triggers both lazy installs in sequence (SQLite first, then tray on macOS/Linux, or vice versa).
2. Second install runs `npm install systray2 --no-save` → npm prunes the previously-installed `better-sqlite3` (not in `package.json` → extraneous → removed).
3. App falls back to `sql.js`, prints "No SQLite driver available", loses native speed.

## Test plan

- [x] `npm test -- --run unit/runtimeInstall.test.js` → 4/4 pass
- [x] Manual repro of the prune: install two packages with the new helper, confirm both survive a third `npm install` in the same runtime dir.
- [x] On macOS: install order SQLite → tray, then a no-op `npm install` → both `better-sqlite3` and `systray2` still present under `~/.9router/runtime/node_modules/`.
- [x] Public API of `sqliteRuntime` and `trayRuntime` unchanged: `ensureSqliteRuntime`, `ensureTrayRuntime`, `buildEnvWithRuntime`, `getRuntimeDir`, `getRuntimeNodeModules`, `runNpmInstall`, `summarizeNpmError` still exported.

## Notes for reviewer

- The npm install is no longer `optional: true`. The SQLite engine is still a speed optimization (sql.js remains the fallback), so a hard failure here would be a regression — but it now fails the same way the tray install already did, and both end with the existing "using fallback" / "tray disabled" warning path. Happy to revisit if maintainers prefer to keep the SQLite install best-effort.
- `cli/hooks/trayRuntime.js` now imports directly from `./runtimeInstall` (not via `./sqliteRuntime`) to avoid a CJS circular import.
- The 4 spec/plan markdown docs in this PR's local development are intentionally not included; only implementation, tests, and changelog.


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Refactors runtime dependency installs to share one npm wrapper (`cli/hooks/runtimeInstall.js`) between SQLite and systray hooks, removing duplicate `--no-save` usage that caused sibling package pruning. Tests cover the new helper end-to-end with a fake npm shim.

### ✅ Looks Good
- Single install wrapper enforces consistent flags across runtime deps
- `ensureRuntimeDir()` bootstraps a minimal `package.json` only when absent
- Tests use a fake npm shim with the same PATH injection for production parity
- `vitest.config.testIsolation: false` is the right call to avoid macOS `/private` symlink skew
- No stale `debugger`/`TODO`/secret patterns
- Covers installer error classification and warning hooks

---
*Reviewed by Hermes Agent*

tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

### Looks Good
- Root cause fix is clean and minimal: shared `runtimeInstall.js` owns npm flag policy so no single caller can accidentally reintroduce `--no-save`
- Tests explicitly verify `--no-save` is absent from the npm argv
- No side effects outside `DATA_DIR/runtime`

### Tiny Polish Suggestion
- `cli/hooks/runtimeInstall.js` creates `package.json` with an empty `description`. Not a bug, but a placeholder string would be slightly cleaner.

---
*Reviewed by Hermes Agent*

PR #1670, opened 2026-06-04
feat(provider): add Command Code CLI provider

Spidy092:
## Summary

Adds a local Command Code CLI provider using the authenticated `cmd` CLI in non-interactive print mode.

This provider is separate from the existing `commandcode` API provider and routes requests through the local Command Code CLI session.

## Features

- Adds `commandcode-cli` provider
- Supports `cccli/<model-id>` model prefix
- Supports OpenAI-compatible SSE streaming
- Uses `spawn()` with argument array and `shell:false`
- Handles missing CLI, auth failure, timeout, and upstream errors
- Supports dynamic/static Command Code CLI model discovery

## Requirements

Users need Command Code CLI installed and authenticated:

```bash
npm i -g command-code
cmd login


Spidy092:
Tested locally with Command Code CLI and 9Router streaming.

Direct CLI:

```bash
cmd --model xiaomi/mimo-v2.5-pro -p "Say hello" --skip-onboarding --trust --max-turns 4
cmd --model deepseek/deepseek-v4-pro -p "Say hello" --skip-onboarding --trust --max-turns 4

Spidy092:
curl -i http://localhost:20128/api/v1/chat/completions \
  -H "Authorization: Bearer <9ROUTER_KEY>" \
  -d '{
    "model":"cccli/xiaomi/mimo-v2.5-pro",
    "messages":[{"role":"user","content":"Say hello"}],
    "stream":true
  }'

Spidy092:
Additional context on why this provider is useful:

The existing `commandcode` provider uses the Command Code Provider API. Some users can use Command Code successfully through the local `cmd` CLI, but do not have Provider API chat access.

For example, on a Go/$1 plan, direct CLI works:

```bash
cmd --model xiaomi/mimo-v2.5-pro -p "Say hello" --skip-onboarding --trust --max-turns 4


Provider API endpoint return

{
  "code": "upgrade_required"
}

tonydwb (commented):
Approved for review: please confirm the global SSE header change in streamingHandler.js is intentional; the message layer currently drops non-text parts so multimodal messages will be truncated while that is in scope.

tonydwb (commented):
## Command Code CLI provider review

Overall the implementation is well-structured: shell:false spawn, typed args array, timeout/abort handling, and stream/non-stream response shapes are all correct.

Spidy092:
Thanks for the review.

Yes, the SSE header change in `streamingHandler.js` is intentional. The `commandcode-cli` provider returns OpenAI-compatible SSE chunks for `stream:true`, and the header change ensures clients like Cline can parse the response as `text/event-stream`.

For multimodal messages: yes, the current implementation intentionally supports text-only prompt conversion for the initial CLI provider scope. Non-text content is not sent to the `cmd -p` prompt because Command Code CLI print mode is text-based in this implementation. I can add a note in the provider docs/comments that `commandcode-cli` currently supports text-only message content, and multimodal support can be handled in a follow-up if needed.

erikdemarco:
when it will be merged?

tonydwb (approved):
Adds Command Code CLI as a first-class provider in 9Router. Clean executor implementation with good fallback behavior for local CLI discovery.

**Verdict: Approved**

### Summary of changes
- New `cccli` provider with 27-model static catalog
- `CommandCodeCLIExecutor` spawning `cmd` CLI, parsing stdout to OpenAI-compatible JSON
- Model ID prefix convention: `cccli/<model>` → passed directly to `cmd --model`
- Dual discovery: API-based (via COMMAND_CODE_API_KEY) → CLI `--list-models` → static fallback
- `noAuth: true` provider (local CLI handles its own session)
- Debug logging with `maskArgForLog` and `sanitizeLog` to protect sensitive CLI output

### Looks Good
- `child_process` spawn with `signal` support — respects upstream cancellation
- Timeout bound via `DEFAULT_TIMEOUT_MS = 180s` (3 min) — prevents indefinite hangs
- `log?.debug?.` optional chaining avoids console noise in production
- Static fallback catalog ensures the provider degrades gracefully if CLI discovery fails
- `maskArgForLog` applied to user prompts before logging

### Minor Nits (non-blocking)
- `toUpstreamModel()` only strips the `cccli/` prefix when present but does not trim whitespace on unprefixed models (e.g. `"  gpt-4  "` passes through) — consider adding `.trim()` unconditionally
- Model catalog uses `contextLength: 1000000` for all entries — this is a placeholder; consider aligning with actual provider context limits or using `null` to defer to upstream response

---
*Reviewed by Hermes Agent*

tonydwb (commented):
## Code Review Summary

**Verdict: Comment** — needs one fix before merge

### ⚠️ Warning
- `.env.example:10` — `INITIAL_PASSWORD=123456` uses an obvious guessable default. Example envs should always use non-obvious values like `change-me-to-a-long-random-secret` (the previous value) so deployments don't accidentally ship with a trivially-brute-forced initial password.

### 💡 Suggestion
- Consider adding test coverage for the provider validation branch in `src/app/api/providers/validate/route.js` (`CommandCodeCliValidate` provider test) alongside the executor test.

### Looks Good
- New `CommandCodeCliExecutor` is cleanly isolated with no credential leakage (authorization headers redacted on error)
- Provider registration in `PROVIDERS` and `PROVIDER_MODELS` is correctly scoped (`noAuth: true`, passthrough flow)
- `commandCodeApiKey` env/credential lookup is optional with sensible fallback to local CLI
- README documents install steps, CLI flow, model ID convention, and discovery order
---
*Reviewed by Hermes Agent*

Spidy092:
Thanks for the review.

Fixed the required `.env.example` issue by changing `INITIAL_PASSWORD=123456` to `INITIAL_PASSWORD=change-me-to-a-long-random-secret`.

I also added minimal route-level test coverage for the `commandcode-cli` provider validation branch:

- mocks local `cmd` validation
- verifies it does not call the Command Code Provider API
- covers success, missing CLI, and unauthenticated CLI failure

Tested with:

```bash
./node_modules/.bin/vitest run --reporter=verbose unit/provider-validation.test.js unit/commandcode-cli-validation.test.js

hamsa0x7:
***

**Hey! Testing this out on Windows and found a few critical bugs that cause the CLI integration and model auto-discovery to fail. Here is how to fix them:**

**1. `spawn("cmd")` breaks on Windows and causes huge prompts to crash**
In both `open-sse/executors/commandcode-cli.js` and `commandCodeCliModels.js`, the code uses `spawn("cmd", ...)`. On Windows, this literally launches the native OS `C:\Windows\System32\cmd.exe` instead of the Command Code CLI, which immediately exits with an error. 
Even if you change it to `spawn("commandcode")`, it will still run via `cmd.exe`'s wrapper, which has a strict **8,191 character limit**. For any workflows that pass full conversation histories or large context, the prompt instantly crashes with `The command line is too long`. 
*Fix:* Use `execSync("npm root -g")` to dynamically locate `command-code/dist/index.mjs` and spawn it directly with `node` instead of using `cmd` or `commandcode`. This completely bypasses `cmd.exe` and gives you a massive 32,767 character prompt limit on Windows.

**2. Auto-discovery `/v1/models` breaks because of the wrong alias mapping**
In `open-sse/config/providerModels.js`, the fallback static models are mapped under the key `cccli`. But the provider ID used in `PROVIDERS` is `"commandcode-cli"`. Because the keys don't match, `9router` can't find the models to inject them into the auto-discovery `/v1/models` endpoint for external API clients.
*Fix:* Change the key in `PROVIDER_MODELS` from `cccli` to `"commandcode-cli"`. 

*** 

P.S Use "commandcode" instead of "cmd login" to both login and start commandcode cli in windows..

hamsa0x7:
***

**Follow-up from further live testing:**

Hey again! After running this locally in a live environment, I uncovered a few more structural edge cases that happen when wrapping `commandcode` as a background process. 

Also, regarding point #2 in my previous comment—I actually dug deeper and realized the alias mapping wasn't the root cause of the auto-discovery bug. Here are the actual remaining issues and how to fix them:

**1. The Auto-Discovery Catch-22 (Correction to #2 above)**
The real reason `cccli/*` models never appear in the `/v1/models` auto-discovery dropdown isn't the alias key. It's because `commandcode-cli` is registered in `FREE_PROVIDERS` as a `noAuth` tool. The 9router frontend intentionally hides the "Save Connection" button for `noAuth` tools since they don't require API keys. However, the backend `/v1/models` loop *only* iterates over providers that have an explicit saved connection record in the local database. Since the user can't click "Save", the loop skips the CLI entirely.
* **Fix:** In `src/app/api/v1/models/route.js`, dynamically auto-inject `noAuth` providers into the `activeConnectionByProvider` map right before building the model list so they are exposed out-of-the-box without requiring a database record.

**2. The Context Leak & "Exit Code 6" Crash**
Because `commandcode` is a smart coding tool, it aggressively indexes the local Git repository of its Current Working Directory (CWD). Because `9router` spawns the CLI natively inside the `9router` source directory (and , the CLI silently bundles thousands of lines of `9router` code and injects it into the prompt behind the scenes. 
When used with downstream API clients that send naturally large prompts (13k+ tokens), the combined payload exceeds the maximum upstream API limit. The server rejects the massive POST body, causing the CLI to instantly crash with `Exit code 6: Error: Unable to connect to the API.` (Or, if it doesn't crash, the LLM hallucinates answers about the `9router` codebase instead of answering the actual prompt).
* **Fix:** Aggressively sandbox the CLI. Pass `cwd: require("os").tmpdir()` in the `spawn()` options inside `open-sse/executors/commandcode-cli.js` (and the `listCliModels` fetcher). This blinds the CLI to the local repository, making it act as a pure, blazing-fast LLM endpoint and preventing payload crashes.

**3. Nested Parameter Auto-Fix Failures**
If the CLI throws a `400 Bad Request` because an API client passed a `max_tokens` value that is too high, 9router's built-in `tryAutoFixBadRequest` logic attempts to clamp the value and retry. However, the auto-fix logic is hardcoded to look for flat OpenAI structures (e.g., `body.max_tokens`). Because this PR translates the payload into Command Code's proprietary nested structure (`body.params.max_tokens`), the auto-fix silently fails to locate the property and the request perma-fails.
* **Fix:** Update `BaseExecutor` to use recursive getter/setter helpers so the clamping logic can traverse into `body.params` to adjust nested payload parameters dynamically.

***
P.S. On the topic of the cwd sandbox fix for the CLI—adding cwd: require("os").tmpdir() is going to be absolutely critical for the public NPM release.

These are just the edge-cases I found during my live testing, but I think I'm mostly sure of these: because the child process automatically inherits the working directory of the user's terminal, a global npm install -g 9router makes this bug highly unpredictable based on where the user types 9router start.

If they launch it from inside a random work project, the LLM will suddenly start hallucinating answers about that specific project's source code instead of answering their actual OMP prompts.
If they launch it from their root ~ Home directory, the CLI might try to index their entire computer, hanging for several minutes before inevitably crashing with a payload size error.
By forcing it into an isolated temp folder, we guarantee that no matter where or how the user launches 9router, the CLI is permanently trapped in a sterile environment where it can't read anything.


****  

Given how many edge cases pop up when running the CLI as a background process, please test this thoroughly on a fresh install before merging!

hamsa0x7:
@tonydwb @Spidy092 please review again

Spidy092:
Resolved the merge conflict with latest `upstream/master`.

Kept the new upstream executor registrations and preserved the `commandcode-cli` executor registration.

Also re-ran the Command Code CLI tests:

```bash
tests/node_modules/.bin/vitest run --config tests/vitest.config.js --reporter=verbose \
  tests/unit/commandcode-cli-executor.test.js \
  tests/unit/commandcode-cli-models.test.js \
  tests/unit/models-noauth-commandcode-cli.test.js \
  tests/unit/commandcode-cli-validation.test.js

Spidy092:
Thanks for the detailed Windows and live-testing feedback.

I pushed a follow-up commit addressing the scoped `commandcode-cli` issues you mentioned:

- Added shared `COMMAND_CODE_CLI_BIN` resolver
  - env override wins
  - Windows defaults to `commandcode`
  - Linux/macOS defaults to `cmd`
- Kept `shell:false` and argument-array spawning
- Added `cwd: os.tmpdir()` for Command Code CLI execution, model discovery, and validation so the CLI does not accidentally index the 9Router repo or the launch directory
- Added noAuth `/v1/models` inclusion so `cccli/...` models can appear without requiring a saved `commandcode-cli` connection
- Left the broader nested `max_tokens` auto-fix behavior untouched because that affects shared/global executor behavior and is better handled separately unless maintainers want it in this PR

Tests run:

```bash
tests/node_modules/.bin/vitest run --config tests/vitest.config.js --reporter=verbose \
  tests/unit/commandcode-cli-executor.test.js \
  tests/unit/commandcode-cli-models.test.js \
  tests/unit/models-noauth-commandcode-cli.test.js \
  tests/unit/commandcode-cli-validation.test.js

robzlabz:
waiting this to be merged

nadimtuhin:
waiting for this to be merged

Spidy092:
Hi @tonydwb @hamsa0x7, just checking in.

I addressed the latest review feedback, resolved the merge conflict with `upstream/master`, and re-ran the Command Code CLI tests successfully.

Current status:
- no merge conflicts
- Windows-safe CLI resolver added
- `cwd: os.tmpdir()` sandboxing added
- noAuth `/v1/models` discovery fixed
- validation/model/executor tests passing

Please let me know if any further changes are needed.

agarwalbikram:
@decolua Please Merge this

PR #1672, opened 2026-06-04
fix(validate): increase validation max_tokens to 10 to satisfy strict provider requirements

rifuki:
### Summary
Increases the hardcoded `max_tokens` from `1` to `10` across several validation and testing endpoints.

### Context / Problem
Some upstream providers (such as B.AI / Minimax) enforce strict parameter constraints (e.g., `max_tokens must be greater than 2` or similar minimums). Using a hardcoded `max_tokens: 1` during credential and model validation checks causes API key validation to fail with a `500 Internal Server Error` even when the credentials are correct.

### Solution
- Updated `max_tokens` from `1` to `10` in all validation and test files.
- This is still a negligible token amount to minimize costs, but avoids parameter threshold errors from stricter APIs.

### Files Changed
- `src/app/api/models/test/route.js`
- `src/app/api/provider-nodes/validate/route.js`
- `src/app/api/providers/[id]/test-models/route.js`
- `src/app/api/providers/[id]/test/testUtils.js`
- `src/app/api/providers/validate/route.js`

### DIFF
--- a/src/app/api/models/test/route.js
+++ b/src/app/api/models/test/route.js
@@ -2,6 +2,7 @@ import { NextResponse } from "next/server";
 import { getApiKeys } from "@/lib/localDb";
 import { UPDATER_CONFIG } from "@/shared/constants/config";
 import { getConsistentMachineId } from "@/shared/utils/machineId";
+import { getTestMaxTokensForModel } from "@/shared/constants/providers";
 
 const CLI_TOKEN_SALT = "9r-cli-auth";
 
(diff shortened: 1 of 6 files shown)


tonydwb (approved):
Approved: narrow validation param bump with cleanup of redundant override.

tonydwb (approved):
Bumps the validation max_tokens ceiling to 10 and adds MiniMax-M3 model support with pricing. This directly addresses provider-side strict requirements that were causing validation failures.

**Verdict: Approved**

### Changes
- max_tokens validation limit raised to accommodate providers requiring higher values
- MiniMax-M3 added to model list and pricing constants

### Risk
Low. Validation ceiling increase is conservative (only to 10), and new model entry follows existing pricing structure.

---
*Reviewed by Hermes Agent*

Issue #1677, opened 2026-06-05
feat: Per-API-key quota enforcement (rate limit, token budget, model allowlist)


muhfauziazhar:
## Problem

9router is often used as a shared gateway serving multiple tools (Claude Code, Codex, Cursor, Cline, Copilot) through one API key per tool. Currently there's no way to enforce usage boundaries per API key:

1. **One heavy tool drains the shared pool** — a single key can exhaust upstream quota, causing 429s for all other keys
2. **No per-key visibility** — `requestDetails` tracks `connectionId` (upstream account) but not which API key made the call
3. **No model access control** — any key can access any model, including expensive ones (Claude Opus, GPT-4o)

### 1. Per-Key Token Budget
- Configurable token quota per API key (daily, weekly, 5h window, or hard cap)
- Returns `429` + `Retry-After` when budget exceeded
- Global default with per-key override

### 2. Model Allowlist per Key
- Admin specifies which models each key can access
- Prevents junior keys from accessing expensive models
- Filters `/v1/models` endpoint per key

### 3. Request Attribution
- Log which API key was used per request
- Enables per-key usage dashboard (consumption, remaining quota)

## Existing Work

This extends #1154 (Vietnamese proposal for per-key quota). Three PRs already address parts of this:

| PR | Scope | Status |
|---|---|---|
| #1569 | Per-key 5h token quota + request attribution | OPEN |
| #1107 | Token control: daily/weekly/hard caps, key expiry, SSRF hardening | OPEN |
| #1158 | Daily token limits, expiry, model allowlists, dashboard progress | OPEN |

**Recommendation:** Review #1569 first (cleanest scope, opt-in, no default behavior change). Consider merging best parts from #1107 and #1158 into a comprehensive solution.

## Use Case

Shared 9router instance for a team: 3 developers, each with their own API key. Want to ensure one developer's Claude Code session doesn't exhaust the entire team's upstream quota, and junior developers can only access cost-effective models.

muhfauziazhar:
## Status Update

All three proposed features now have approved PRs:

| Feature | PR | Status |
|---|---|---|
| Per-key token budget (cost-based, 5h window) | #1681 | ✅ Approved |
| Request attribution (apiKey on usage/requestDetails) | #1681 | ✅ Approved |
| Model allowlist per connection | #1683 | ✅ Approved |

**Notes:**
- #1681 supersedes #1569 — consolidates all review fixes into a single clean commit
[...]

PR #1681, opened 2026-06-05
feat: per-key cost quota + request attribution for shared gateways

vanminhph:
## Why

9Router is often used as a **single gateway feeding many AI tools** — Claude Code, Codex CLI, Cursor, Cline, Copilot, etc. — all sharing one upstream account pool, with **one API key per tool**. At that scale two gaps show up:

1. **One heavy tool can drain the shared pool.** When a single tool's key burns through the pool's quota, every other tool starts getting `429`s through no fault of its own. There is no way to bound a single key's consumption.
2. **You can't tell which tool used what.** `requestDetails` stores `connectionId` (the upstream account) but not the **API key** that made the call, so usage can't be attributed back to the tool that issued it.

This PR adds two small, **opt-in** governance features to close those gaps — without changing default behavior for anyone who doesn't turn them on.

> Supersedes #1569 — consolidates the original feature plus all fixes from the review there (thanks @muhfauziazhar) into a single clean commit.

### 1. Per-key cost quota (disabled by default)

- **Fixed 5h window anchored to the key's first request** (`userQuotaWindow` table) — mirrors upstream reset semantics, not a rolling window. Enforced in `handleChat` **before** combo/non-combo routing so every path is covered.
- **Cost-based accounting, not raw tokens.** The budget is `SUM(cost)` (USD-equivalent), where `cost` is already computed per row by `calculateCost` with per-model pricing — weighting **non-cached input, cached input and output separately**. This mirrors how upstream actually accounts usage (codex credits: input 1×, cached ~0.1×, output ~6×; Claude weights similarly). A prompt-token budget would overcharge cache reads ~10× and ignore output entirely.
  - Models without pricing data have `cost = 0` and are **intentionally not charged** (no token fallback).
- **Each request charged exactly once** via an explicit `usageHistory.countsTowardQuota` marker, set at the single canonical write of each path:
  - streaming → `logUsage()` (the `onStreamComplete` duplicate stats row stays unflagged)
  - non-streaming → `nonStreamingHandler`
  - forced SSE→JSON → both branches of `sseToJsonHandler`
- Over budget → **`429` + `Retry-After`** with a human-readable reset time incl. explicit UTC offset (e.g. `14:30 05/06 (UTC+7)`).
- **Budget resolution:** `apiKeys.costBudget5h` (per-key, set via `PUT /api/keys/[id]`; `null` clears) → `userCostBudget5hDefault` DB setting → `USER_COST_BUDGET_5H` env → built-in default (`50`).
- **Eventually consistent by design:** usage rows land async after the response completes, so a burst can overshoot the soft budget. Documented in the code.

### 2. Per-key request attribution

Persists `apiKey` on every `requestDetails` and `usageHistory` row (previously `connectionId` only), so usage and tool calls can be traced back to the specific tool/key.

## Safe by design

- **Opt-in, off by default** — `userQuotaEnabled` resolves `DB setting > env > built-in default (false)`. Existing deployments behave exactly as before until explicitly enabled.
- **Additive schema only** — applied automatically by `syncSchemaFromTables`: `apiKeys.costBudget5h`, `requestDetails.apiKey`, `usageHistory.apiKey/endpoint/countsTowardQuota`, `userQuotaWindow` table, supporting indexes (`idx_uh_apikey_ts`, `idx_rd_apikey`). No migration file, no destructive change.
- **Backward compatible** — attribution columns are nullable; pre-upgrade rows simply have `apiKey = null` / `countsTowardQuota = 0`.

## Testing

`tests/unit/user-quota.test.js` — 12 cases:
- accounting dedupe: streaming dual-write charged exactly once; non-streaming charged
- weighting: output and cached-input weighted per pricing; model without pricing charged 0 (documented no-fallback)
- over-budget blocking with correct `Retry-After`/reset info
- window math: anchor on first request, reset after 5h
- enabled-flag resolution; per-key override and its CRUD path

Also verified end-to-end on a live instance: `429` with correct reset time when over budget, normal service under budget, `apiKey` recorded on `requestDetails`.

## How to enable

```yaml
environment:
  - USER_QUOTA_ENABLED=true
  - USER_COST_BUDGET_5H=50   # USD-equivalent per 5h per key, optional
```

Or via DB settings (`userQuotaEnabled`, `userCostBudget5hDefault`); per-key limit via `PUT /api/keys/[id]` with `{"costBudget5h": 20}`.

## Follow-ups (out of scope)

- Dashboard UI surface for `costBudget5h` (API is in place)
- Broader per-key governance tracked in #1677

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -187,7 +187,7 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     trackPendingRequest(model, provider, connectionId, false, true);
     appendRequestLog({ model, provider, connectionId, status: `FAILED ${error.name === "AbortError" ? 499 : HTTP_STATUS.BAD_GATEWAY}` }).catch(() => { });
     saveRequestDetail(buildRequestDetail({
-      provider, model, connectionId,
+      provider, model, connectionId, apiKey,
       latency: { ttft: 0, total: Date.now() - requestStartTime },
       tokens: { prompt_tokens: 0, completion_tokens: 0 },
       request: extractRequestConfig(body, stream),
@@ -233,7 +233,7 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     const { statusCode, message, resetsAtMs } = await parseUpstreamError(providerResponse, executor);
     appendRequestLog({ model, provider, connectionId, status: `FAILED ${statusCode}` }).catch(() => { });
     saveRequestDetail(buildRequestDetail({
-      provider, model, connectionId,
+      provider, model, connectionId, apiKey,
       latency: { ttft: 0, total: Date.now() - requestStartTime },
       tokens: { prompt_tokens: 0, completion_tokens: 0 },
       request: extractRequestConfig(body, stream),
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -159,7 +159,8 @@ export async function handleNonStreamingResponse({ providerResponse, provider, m
 
   const usage = extractUsageFromResponse(responseBody);
   appendLog({ tokens: usage, status: "200 OK" });
-  saveUsageStats({ provider, model, tokens: usage, connectionId, apiKey, endpoint: clientRawRequest?.endpoint });
+  // Only write for this request → canonical quota row
+  saveUsageStats({ provider, model, tokens: usage, connectionId, apiKey, endpoint: clientRawRequest?.endpoint, countsTowardQuota: true });
 
   const translatedResponse = needsTranslation(targetFormat, sourceFormat)
     ? translateNonStreamingResponse(responseBody, targetFormat, sourceFormat)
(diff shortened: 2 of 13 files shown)


tonydwb (approved):
Implements per-key cost quota for shared gateway pools. Comprehensive implementation covering DB migration, usage accounting, quota enforcement, and testing.

**Verdict: Approved**

### Summary of changes
- New `countsTowardQuota` flag on usage rows (streaming writes once, flags canonical row)
- Per-key `costBudget5h` field on apiKeys (override global default per key)
- New `userQuotaWindow` table tracking 5h anchored windows
- `checkUserQuota()` enforces budget with `429 Too Many Requests` and `Retry-After`
- All relevant paths updated: non-streaming, streaming, requestDetail
- 9+ unit tests covering edge cases (cached tokens, no-pricing models, over-budget, window reset, override)

### Looks Good
- Quota is disabled by default (`USER_QUOTA_ENABLED` env/DB flag) — operator opt-in
- Rollout-safe: existing rows have `countsTowardQuota=0`, so no data migration needed
- Timezone-aware `resetAtLocal` string in 429 response (explicit UTC offset, avoids confusion in containers)
- `saveRequestUsage` catch-all absorbs quota check failures (no quota = no path breakage)
- Test comments mix English + Vietnamese (natural from original author, not a code quality issue)

### Minor Nits
- `defaultTimeout` for the abort controller is now unbounded if `quota.allowed===false` because `unavailableResponse` uses `setTimeout` — but this is scoped to the quota path only and 429s are short; acceptable
- The 429 body hardcodes `[quota] Usage quota exceeded` + `Unavailable` payload — consistent with existing quota-style responses

---
*Reviewed by Hermes Agent*

muhfauziazhar (approved):
LGTM! Clean consolidation of #1569. Cost-based quota is the right approach — cache-weighted accounting avoids the 10x overcharge problem of raw token budgets. Opt-in with additive schema, no default behavior change. Ship it. 🚀

PR #1688, opened 2026-06-05
fix: make stream stall timeout configurable

nguyenha935:
## Summary
- make the stream stall watchdog configurable via `STREAM_STALL_TIMEOUT_MS`
- set the default stall timeout to 60 seconds
- document the env var and add unit coverage for default, configured, and invalid values

## Why
The current hard-coded 30s stall timeout is useful for detecting genuinely hung upstream streams quickly, but it can cut off valid long-context / reasoning / agentic requests when the upstream stays quiet for more than 30s before sending the next bytes.

This keeps fast failure behavior bounded while restoring the previous 60s tolerance and gives self-hosted deployments a way to tune the value without patching source.

## Verification
- `./node_modules/.bin/vitest run --reporter=verbose --config ./tests/vitest.config.js tests/unit/runtime-config.test.js`


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Small, focused configurable timeout change.

### What changed
- Adds `STREAM_STALL_TIMEOUT_MS` env var with a safe integer parser
- Raises default from 30s to 60s
- Includes regression tests for defaults, env override, and invalid values

### Looks Good
- Input validation on env values prevents `0`, negatives, and non-numbers
- Test coverage is concise and exercises all branches
- Docs updated in `.env.example` and README

No issues found.

---
*Reviewed by Hermes Agent*

Xpos587:
Really glad to see this PR! We're self-hosting 9router in production and the 30s stall timeout has been cutting off reasoning models (GLM-5.1, deepseek-v4-pro) mid-thought. They legitimately need 30-50s of silence before the first token sometimes, so bumping the default to 60s is a great call.

One thing I noticed — `FETCH_CONNECT_TIMEOUT_MS` (20s, hardcoded in the same file) isn't covered here. We're seeing about 85% of our errors come from `fetch connect timeout` — ollama providers occasionally take 25-40s to establish the connection, and the 20s timeout kills them every time. Then 3 retries × 20s = 60s of frozen UI followed by a 502.

Would love to see `FETCH_CONNECT_TIMEOUT_MS` get the same `parsePositiveIntegerEnv` treatment. The same pattern from this PR would work perfectly — env var with a higher default (maybe 60s to match the new stall timeout). Filed #1692 with the full details.

Thanks for the quick fix on the stall timeout though!

decolua:
Cảm ơn @nguyenha935! Đã áp nâng stream stall timeout lên 60s vào bản nội bộ. Closing — đã tích hợp. 🙏

PR #1693, opened 2026-06-05
feat: Core stability, atomic writes, and security middlewares

hamsa0x7:
This PR introduces several isolated, high-value improvements to core stability and API security that I've been running locally:

### 1. Runtime & Config Stability
- **Atomic File Writes (`atomicWrite.js`)**: Prevents `settings.json` from corrupting to 0 bytes if the node process crashes mid-save.
- **SQLite OAuth Locking (`oauthCredentialManager.js`)**: Adds a database lock to prevent concurrent write collisions during token refreshes.
- **Next.js Caching Fixes**: Adds `force-dynamic` and fallback patches to prevent the CLI tools from seeing stale model data.

### 2. Security Validations
- **SSRF Protection (`localOnly.js`)**: Middleware ensuring sensitive local endpoints are only accessible via loopback.
- **Infinite Loop Prevention (`comboValidation.js`)**: Validates Combo models to prevent circular dependencies that can crash the routing loop.

### DIFF
--- a/cli/hooks/runtimeInstall.js
+++ b/cli/hooks/runtimeInstall.js
@@ -1,0 +1,107 @@
+// Shared runtime install helper.
+//
+// Owns the user-writable runtime directory under DATA_DIR (or the platform
+// default), the npm install wrapper, and the package-specific install logging
+// used by cli/hooks/sqliteRuntime.js and cli/hooks/trayRuntime.js.
+//
+// Keeping a single install wrapper ensures the --no-save flag is never passed
+// for runtime installs: each install writes to package.json dependencies, so
+// later installs no longer treat earlier runtime packages as extraneous and
+// do not prune them.
+const { spawnSync } = require("child_process");
+const fs = require("fs");
+const os = require("os");
+const path = require("path");
+
+function getDataDir() {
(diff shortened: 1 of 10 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Root-cause fix for #1605: sibling `npm install` of `systray2` was pruning `better-sqlite3` because the shared runtime wrapper lacked a common `--no-save` guard.

### What changed
- Extracted `ensurePackageJson()` + `installRuntimeDeps()` into `cli/hooks/runtimeInstall.js`
- SQLite and systray hooks now call the same wrapper — single source of truth for npm flags
- Tests verify `--no-save` is NOT passed to npm (preventing sibling pruning)

### Looks Good
- Failure path clearly split: npm ENOENT → warning hook (no crash), other errors → fatal
- Fake npm shim in tests uses the same PATH injection as production — good parity
- `createPackageJsonIfMissing` only writes when absent — idempotent and safe
- Debug artifacts / console.log: only the benign `npm install` progress output (existing), no new stray logs
- No hardcoded secrets, SQL injection, or auth changes

---

PR #1697, opened 2026-06-06
fix: preserve tool context in combo and Kiro routing

kaseru:
## Summary
- Keep combo requests that include tools in priority order instead of round-robin rotation.
- Preserve the original request body across combo fallback so messages/tools/tool_choice stay intact.
- Preserve OpenAI-compatible tool message content arrays when translating tool results to Kiro.

## Test Plan
- `NODE_PATH=/root/src/9router-src/node_modules:/root/src/9router-src/tests/node_modules /root/src/9router-src/tests/node_modules/.bin/vitest run --reporter=verbose tests/unit/openai-to-kiro.test.js tests/unit/combo-routing.test.js`
- Result: 2 files passed, 14 tests passed

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -106,8 +106,17 @@ export function getComboModelsFromData(modelStr, combosData) {
  * @returns {Promise<Response>}
  */
 export async function handleComboChat({ body, models, handleSingleModel, log, comboName, comboStrategy, comboStickyLimit = 1 }) {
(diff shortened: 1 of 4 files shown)


kaseru:
Đóng PR: hướng fix combo cho coding agent đang được thiết kế lại (sticky 1 model sống + bỏ rotation khi có tool). Sẽ mở PR mới sau khi chốt spec.

PR #1701, opened 2026-06-06
fix: forward connection-level proxy to Gemini/OpenAI embedding requests

nic562:
The embeddings handler called raw global fetch() without passing proxyOptions, so the connection-level proxy (proxy pool, legacy connectionProxyUrl, or Vercel/Cloudflare/Deno relay) configured on a provider connection was ignored — the request went direct even when the chat path applied the proxy correctly.

Root cause:
- open-sse/handlers/embeddingsCore.js used the bare global fetch for both the primary request and the 401/403 retry.
- proxyAwareFetch (which honors connection-level proxy) was never invoked, and the executor.refreshCredentials call on 401/403 was also missing the proxyOptions argument.

Fix:
- Build proxyOptions from credentials.providerSpecificData (mirrors chatCore.js:144-149) and pass to proxyAwareFetch on every fetch.
- Pass proxyOptions to executor.refreshCredentials so token refresh also respects the connection proxy.
- The global outbound proxy (env HTTPS_PROXY) keeps working unchanged.

Tests:
- Refactored existing tests to mock proxyAwareFetch as a named export.
- Added 5 regression tests covering: connectionProxyEnabled+Url, vercelRelayUrl, no proxy configured, missing providerSpecificData, and proxyOptions forwarded on 401/403 retry.

All 43 unit tests pass.

### DIFF
--- a/open-sse/handlers/embeddingsCore.js
+++ b/open-sse/handlers/embeddingsCore.js
@@ -3,6 +3,22 @@ import { HTTP_STATUS } from "../config/runtimeConfig.js";
 import { getExecutor } from "../executors/index.js";
 import { refreshWithRetry } from "../services/tokenRefresh.js";
 import { getEmbeddingAdapter } from "./embeddingProviders/index.js";
+import { proxyAwareFetch } from "../utils/proxyFetch.js";
+
+/**
+ * Build proxy options from a connection's providerSpecificData.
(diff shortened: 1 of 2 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Connects per-account proxy pool setting through to Gemini/OpenAI embedding routes. Low risk plumbing fix — uses existing connection.proxyPoolId.

---

PR #1704, opened 2026-06-06
feat(providers): support Vercel AI Gateway embeddings and images

newnol:
Reopened from dedicated feature branch (previously #1268).

This change adds support for using Vercel AI Gateway for embeddings and image generation, extending the existing provider integration beyond text/chat completion workflows.


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Adds Vercel AI Gateway as a built-in provider with embedding and image generation support. Includes correct OpenAI-compatible passthrough configuration, and test fixtures use placeholder keys only (vag-test-key).

---

decolua:
Thanks @newnol for the contribution! 🙏

Merged the Vercel AI Gateway **embeddings, image generation, credit/usage, retry-on-429 and models fetcher** into `master`.

I held back the **Reasoning/Thinking** part (the Claude→OpenAI `reasoning_effort` mapping and the provider-settings Thinking dropdown) for now — I am reworking the thinking handling into a shared, project-wide mechanism so it stays consistent across all providers rather than per-provider. It will land in a separate change.

Heads up: the `providers/[id]/page.js` edit in this PR had an unbalanced JSX comment (the `{/*` opener was left in while the `*/}` closer was removed), which would break the build, so that file was left untouched here.

PR #1711, opened 2026-06-06
feat: auth session persistence + antigravity fallbacks + codex OAuth refresh + models envelope

zuhabul:
## Problem

The dashboard session JWT expires after **24h** and the cookie has no `maxAge`, making it a browser session cookie. Result: re-login required every day *and* every time the browser is closed.

## Changes

**`src/lib/auth/dashboardSession.js`**

- Single canonical TTL in **seconds** drives both the JWT `exp` claim and the cookie `maxAge` — no parser differential between the two.
- Configurable via `AUTH_SESSION_TTL` env var (format: `Nd`, `Nh`, `Nm`, or integer seconds). Defaults to `30d`.
- Clamped to ≤ 30 days to prevent operators from accidentally setting extreme values.
- `setExpirationTime(epoch)` instead of `setExpirationTime("30d")` so both the JWT and cookie use exactly the same computed expiry.

## Security notes

- Hard ceiling of 30 days (`MAX_SESSION_S`). Operators can reduce via `AUTH_SESSION_TTL`.
- `httpOnly: true` + `secure` (auto-detected) + `sameSite: lax` unchanged.
- No change to password hashing, token verification, or login logic.

## Test plan

- [ ] Log in → close browser → reopen → still logged in (cookie persisted)
- [ ] Log in → wait >24h (or mock time) → still logged in (JWT not expired)
- [ ] Set `AUTH_SESSION_TTL=1h` → session expires in 1h
- [ ] Set `AUTH_SESSION_TTL=999d` → clamped to 30d

### DIFF
--- a/open-sse/config/providerModels.js
+++ b/open-sse/config/providerModels.js
@@ -82,15 +82,18 @@ export const PROVIDER_MODELS = {
     { id: "iflow-rome-30ba3b", name: "iFlow ROME" },
   ],
   ag: [  // Antigravity - special case: models call different backends
+    // Latest AG entries.
+    { id: "gemini-3.5-flash-extra-low", name: "Gemini 3.5 Flash (Extra Low)" },
     { id: "gemini-3-flash-agent", name: "Gemini 3.5 Flash (High)" },
     { id: "gemini-3.5-flash-low", name: "Gemini 3.5 Flash (Medium)" },
-    { id: "gemini-3.5-flash-extra-low", name: "Gemini 3.5 Flash (Low)" },
-    { id: "gemini-pro-agent", name: "Gemini 3.1 Pro (High)" },
-    { id: "gemini-3.1-pro-low", name: "Gemini 3.1 Pro (Low)" },
+    // Compatibility aliases for existing account mappings.
+    { id: "gemini-3.1-pro-high", name: "Gemini 3.1 Pro (High)", upstreamModelId: "gemini-2.5-flash" },
+    { id: "gemini-pro-agent", name: "Gemini Pro Agent", upstreamModelId: "gemini-2.5-flash" },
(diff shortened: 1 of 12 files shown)


tonydwb (commented):
## Code Review Summary

**Verdict: Comment**

Changes dashboard session cookie lifetime to 30 days and unifies the cookie parser. Auth boundary change — worth verifying that downstream consumers of the auth token handle the longer TTL correctly.

### Observations
* Diff shows `MAX_ANTIGRAVITY_OUTPUT_TOKENS` cap change alongside cookie parser logic. If 1711 and 1712 share the same author and touch the same translator files, confirm they are independent changes and not a split-PR dependency (otherwise one may need to be rebased after the other).
---

kacipbuah:
Friendly maintainer ping 🙏

Session/auth persistence improvements here seem important for stability. Could you review when convenient?

cc @decolua

zuhabul:
Thanks for the ping @kacipbuah / @decolua — following up now.

**Consolidation:** #1711 and #1712 were cut from the same working branch and overlap almost entirely (#1711 is a strict superset of #1712 — identical changes plus `dashboardSession.js`). To make this reviewable as a single unit I've **retitled this PR to cover the full bundle and am closing #1712 as a duplicate.** Everything from #1712 (antigravity model fallbacks, subagent routing interceptor, codex OAuth refresh, models envelope) is already contained here, so there's nothing to rebase.

**Re the `max_tokens` truthiness change** flagged on #1712 (`if (body.max_tokens)` vs `!== undefined` in `openai-to-gemini.js`): this is intentional. `max_tokens: 0` is treated as unset and falls back to the model default rather than being forwarded as a literal zero-token cap (a zero-length completion isn't a meaningful request, and upstream Gemini rejects it anyway). The change also adds `Math.min(body.max_tokens, 8192)` clamping that the old branch lacked.

Happy to split into focused PRs instead if you'd prefer that for review — just let me know.

PR #1712, opened 2026-06-06
feat: antigravity model fallbacks, subagent routing interceptor, codex OAuth refresh

zuhabul:
## Summary

Three independent improvements bundled together:

### 1. Antigravity — model fallback chain + upstream ID resolution

**`open-sse/executors/antigravity.js`**
- `resolveModelCandidates(model)` builds a try-list: `upstreamModelId` first, then fallback chain (configurable via `ROUTER_AG_PRIMARY_REPLACEMENT` / `ROUTER_AG_SECONDARY_REPLACEMENT` env vars).
- Retries on 404 / 503 with the next candidate instead of surfacing an error.
- Drops `sessionId` from the request body (Antigravity sends it in the `X-Machine-Session-Id` header — including it in the body caused `INVALID_ARGUMENT` errors).
- Caps `maxOutputTokens` at 8 192 (was 16 384, which frequently hit AG limits).
- Sanitizes buffered tool-call arguments before flushing.

**`open-sse/config/providerModels.js`**
- Adds `gemini-3.1-pro-high` alias with `upstreamModelId: "gemini-2.5-flash"` for accounts that reference the old model name.

### 2. Subagent routing interceptor

**`src/sse/handlers/chat.js`**
- Detects explicit subagent-spawn language in the latest user message (`"spawn a background agent"`, `"delegate to a subagent"`, etc.).
- Routes those requests to `ROUTER_SUBAGENT_MODEL` (default `ag/gemini-3.1-pro-high`) instead of the main model.
- Skips interceptor for compaction/skill payloads and messages >4 000 chars to avoid false positives.

### 3. Codex import-token — OAuth refresh token support

**`src/app/api/oauth/codex/import-token/route.js`**
- Accepts optional `refreshToken` in the POST body.
- Stores `authType: "oauth"` when a refresh token is present (vs `"access_token"` for static tokens).
- Persists `expiresAt` from the JWT `exp` claim so the token refresh service can proactively refresh.

### 4. Models list envelope

**`src/app/api/v1/models/route.js`, `[kind]/route.js`**
- Adds `models` alias alongside `data` in the list response for compatibility with clients that expect the field under that name.

### 5. Translator / error hardening

Minor fixes in `openai-to-gemini.js`, `openai-responses.js`, `error.js`, `claudeCloaking.js`.

## Test plan

- [ ] `ag/gemini-3.1-pro-high` request → completes without `INVALID_ARGUMENT` sessionId error
- [ ] AG request on a deprecated model → automatically retries with fallback
- [ ] Message containing "spawn a background agent" → routed to `ROUTER_SUBAGENT_MODEL`
- [ ] Compaction payload mentioning "subagent" → NOT intercepted
- [ ] Codex import-token with `refreshToken` → `authType: "oauth"` in DB
- [ ] `GET /v1/models` → response includes both `data` and `models` arrays

### DIFF
--- a/open-sse/config/providerModels.js
+++ b/open-sse/config/providerModels.js
@@ -82,15 +82,18 @@ export const PROVIDER_MODELS = {
     { id: "iflow-rome-30ba3b", name: "iFlow ROME" },
   ],
   ag: [  // Antigravity - special case: models call different backends
+    // Latest AG entries.
+    { id: "gemini-3.5-flash-extra-low", name: "Gemini 3.5 Flash (Extra Low)" },
     { id: "gemini-3-flash-agent", name: "Gemini 3.5 Flash (High)" },
     { id: "gemini-3.5-flash-low", name: "Gemini 3.5 Flash (Medium)" },
-    { id: "gemini-3.5-flash-extra-low", name: "Gemini 3.5 Flash (Low)" },
-    { id: "gemini-pro-agent", name: "Gemini 3.1 Pro (High)" },
-    { id: "gemini-3.1-pro-low", name: "Gemini 3.1 Pro (Low)" },
+    // Compatibility aliases for existing account mappings.
+    { id: "gemini-3.1-pro-high", name: "Gemini 3.1 Pro (High)", upstreamModelId: "gemini-2.5-flash" },
+    { id: "gemini-pro-agent", name: "Gemini Pro Agent", upstreamModelId: "gemini-2.5-flash" },
+    { id: "gemini-3.1-pro-low", name: "Gemini 3.1 Pro (Low)", upstreamModelId: "gemini-2.5-flash" },
     { id: "claude-sonnet-4-6", name: "Claude Sonnet 4.6 (Thinking)" },
     { id: "claude-opus-4-6-thinking", name: "Claude Opus 4.6 (Thinking)" },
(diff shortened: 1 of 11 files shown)


tonydwb (commented):
## Code Review Summary

**Verdict: Comment**

Reduces Antigravity max_output_tokens cap from 16384 to 8192 and updates fallback routing behavior. Needs one clarification before merge.

### Warnings
* `if (body.max_tokens)` was previously `if (body.max_tokens !== undefined)`. The new truthiness check silently skips clamping when `max_tokens = 0` — possibly intentional, but verify that zero is treated as unset in your contract. If a caller explicitly requests 0 output tokens, the cap is not enforced.
* Same pattern appears in 1711 (fix(auth): persistent session cookie) — if these PRs are meant to be independent, confirm they don t depend on each other.
---

zuhabul:
Closing as a duplicate of #1711.

These two PRs were cut from the same working branch and overlap entirely — every file/change in this PR is byte-for-byte identical to #1711, which additionally includes the auth `dashboardSession.js` change. Consolidating everything (antigravity model fallbacks, subagent routing interceptor, codex OAuth refresh, models envelope) into **#1711** so it can be reviewed and merged as one unit.

The `max_tokens: 0` review note is answered over on #1711. Follow there.

PR #1720, opened 2026-06-07
feat(kiro): direct Claude<->Kiro translation route

thienpvt:
## Summary

Adds a **direct `claude:kiro` / `kiro:claude` translation route** so Anthropic Messages API traffic to Kiro (AWS CodeWhisperer) no longer pivots through the OpenAI format. `translateRequest` / `translateResponse` now use a translator registered on the exact `source:target` pair when one exists, falling back to the OpenAI two-hop otherwise — so every other provider is unaffected.

Previously, `register(CLAUDE, KIRO, ...)` / `register(KIRO, CLAUDE, ...)` were registered but **never reachable** — the translator only ever looked up `{src}:openai` and `openai:{tgt}`. A Claude client hitting `/v1/messages` for a `kr/...` model went `claude -> openai -> kiro` and back, a lossy double-hop.

## Changes

- **`open-sse/translator/request/claude-to-kiro.js`** — builds the Kiro `conversationState` payload straight from Claude messages. Carries over the two "Improperly formed request" (HTTP 400) guards from `openai-to-kiro.js`:
  - flatten tool_use/tool_result blocks to text when the client sent no `tools` (avoids the "tools required" validator)
  - reconcile orphaned `tool_result`s (fold dangling references back into user text)
  - plus `-agentic` / `-thinking` suffix resolution and thinking-mode injection.
- **`open-sse/translator/response/kiro-to-claude.js`** — converts the OpenAI-shaped `chat.completion.chunk` objects that `KiroExecutor.transformEventStreamToSSE` emits into Claude SSE events (text, thinking via `reasoning_content`, tool_use with buffered `input_json_delta`, usage, finish).
- **`open-sse/translator/index.js`** — direct-route lookup ahead of the OpenAI pivot in both `translateRequest` and `translateResponse`.
- **stream timeouts** (`runtimeConfig.js`, `streamHandler.js`, `executors/kiro.js`) — split the single stall timeout into an env-overridable first-chunk (TTFT) timeout vs inter-chunk stall timeout, and emit a Kiro keepalive frame so slow prompt-prefill is not aborted mid-stream (`STREAM_FIRST_CHUNK_TIMEOUT_MS`, `STREAM_STALL_TIMEOUT_MS`, `FETCH_CONNECT_TIMEOUT_MS`).

## Why the payload is equivalent

The direct request payload was diffed against the existing `claude -> openai -> kiro` output for identical input and is **byte-identical** except the random `conversationId` / timestamp. On the response side, the old `kiro -> openai` step was a pass-through of the executor's already-OpenAI-shaped chunks, so the direct `kiro -> claude` translator is functionally `openai -> claude` fed the same chunks.

## Testing

- `tests/translator/claude-kiro-direct.test.js` — 8 cases covering both routes and both 400 guards (registered in `registerAll.js`).
- Verified end-to-end against a live Kiro account: a real agent session (40+ msgs, tools, streaming) routes through the direct path and streams successfully; the prior `400 Improperly formed request` no longer occurs.


tonydwb (commented):
## Code Review Summary

**Verdict: Comment**

Large architectural PR (+3,261 / -82) adding Claude<->Kiro direct translation route plus extensive codebase documentation. Mixed scope — core routing code is clean, but doc additions dominate the diff and the logger hard-code deserves a follow-up.

### What changed
* New direct translator: `open-sse/translator/request/claude-to-kiro.js` and `open-sse/translator/response/kiro-to-claude.js`
* Kiro auth flow expanded with API-key route and profile-ARN handling
* .planning/codebase/ docs added (ARCHITECTURE/CONCERNS/CONVENTIONS/INTEGRATIONS/STACK/STRUCTURE/TESTING)
* Proxy-pools API routes extended

### Warnings
* `src/sse/utils/logger.js` hard-codes DEBUG level and emits ANSI color via console.log on every request. Mentioned in the PR notes as known; should be extracted to a follow-up fix (env-driven LOG_LEVEL + structured logger).

### Suggestions
* The .planning docs are valuable but could be reviewed separately — mixing them with 334 lines of live code diffs makes the PR harder to audit. Consider splitting into a docs-only PR.
* Dedicated test files for the new direct translator path are included — good. Verify coverage for error/empty-tool-call branches.
---

PR #1722, opened 2026-06-07
feat(kiro): enable multi-endpoint failover for GenerateAssistantResponse

thienpvt:
## Summary

Enables multi-endpoint failover for the Kiro provider's `GenerateAssistantResponse` calls by wiring `KiroExecutor` into the `BaseExecutor` `baseUrls` fallback mechanism that already exists for other providers.

A Kiro request now advances across the three CodeWhisperer surfaces on `429` and on network/`5xx` errors:

1. `runtime.us-east-1.kiro.dev` (current Kiro IDE endpoint)
2. `codewhisperer.us-east-1.amazonaws.com` (legacy)
3. `q.us-east-1.amazonaws.com` (legacy)

## Changes

- **`open-sse/config/providers.js`** — add `baseUrls[]` to the kiro config (newest endpoint first); keep `baseUrl` as the single-value default for backward compat.
- **`open-sse/executors/kiro.js`** — replace the hand-rolled single-endpoint loop with `super.execute()` delegation (inherits endpoint rotation + retry + connect-timeout), transforming the binary AWS EventStream → SSE only on `response.ok`. Matches the `super.execute()`+post-process idiom already used in `codex.js` / `github.js`. Removes two now-unused imports.

## Important caveat on 429s

The three hosts are **alternate DNS surfaces of one regional service**, not separate quota buckets. AWS throttles per authenticated identity (access token + `profileArn`), **not** per hostname — so this is **edge-level failover** (degraded surface / 5xx / connect timeout), and does **not** multiply 429 headroom for a single account.

To actually spread 429 load, add multiple Kiro accounts: per-account rotation already exists generically in `src/sse/handlers/chat.js` (retry with next account on fallback-worthy error) backed by `src/sse/services/auth.js` (selection + per-model cooldown locks with exponential backoff).

The headers `X-Amz-Target`, `User-Agent`, `X-Amz-User-Agent` do **not** affect rate limiting — `X-Amz-Target` is AWS operation routing (required for correctness), the UA strings are client identifiers AWS may check for request acceptance but not for quota bucketing.

## Testing

- `node --check` passes on both edited files.
- Verified the delegation pattern matches existing executors (`codex.js:268`, `github.js:189`) and that both `chatCore.js` call sites pass the full args object the new `execute(args)` expects.
- Not exercised against a live Kiro account in this change.

### DIFF
--- a/open-sse/config/providers.js
+++ b/open-sse/config/providers.js
@@ -192,7 +192,19 @@ export const PROVIDERS = {
     clientId: "Iv1.b507a08c87ecfe98"
   },
   kiro: {
-    baseUrl: "https://codewhisperer.us-east-1.amazonaws.com/generateAssistantResponse",
+    // All three hosts resolve to the same regional CodeWhisperer streaming service
+    // (GenerateAssistantResponse). They are alternate DNS surfaces, NOT separate quota
+    // buckets — AWS throttles per authenticated identity (token + profileArn), not per
+    // hostname. Listing them enables edge-level failover (5xx / connect timeout / a
+    // degraded surface); it does NOT multiply 429 headroom. To actually spread 429 load,
+    // add multiple Kiro accounts — account rotation in sse/handlers/chat.js handles that.
+    // Order: newest Kiro IDE endpoint first, legacy AWS domains as fallback.
+    baseUrl: "https://runtime.us-east-1.kiro.dev/generateAssistantResponse",
+    baseUrls: [
+      "https://runtime.us-east-1.kiro.dev/generateAssistantResponse",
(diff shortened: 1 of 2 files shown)


tonydwb (approved):
## Code Review Summary

**Verdict: Approved**

Small, targeted enhancement to Kiro endpoint failover — adds multi-endpoint rotation for GenerateAssistantResponse. Net reduction in lines; low risk change.

### Looks Good
* Failover bucket logic is additive and guarded — no change to existing single-endpoint behavior when only one endpoint is configured.
* Token refresh path preserved in the failover loop.
---

decolua:
Thanks @thienpvt for the contribution! Merged into master. 🙏

PR #1727, opened 2026-06-07
fix(github): resolve 401 error on models discovery by proactively checking/refreshing copilot token

v01dgrace:
### Description
This PR resolves the 401 Unauthorized errors encountered when fetching models from the GitHub Copilot provider. The root cause was that Copilot tokens expire every 25-30 minutes, and the models endpoint was reading the cached token without proactively checking for expiration or missing values.

### Key Changes
- **Proactive Token Refresh**: Integrated `checkAndRefreshToken` at the entry point of the `/api/providers/[id]/models` route handler for `github`.
- **Enhanced Expiration Guard**: Updated `checkAndRefreshToken` in `tokenRefresh.js` to proactively fetch a new token if the cached `copilotToken` or `copilotTokenExpiresAt` is missing entirely (not just soon-to-expire).
- **ID Normalization**: Standardized ID mappings (`creds.connectionId = creds.id`) inside the credentials object to prevent database update failures during refresh persistence.

### Verification
- Verified that the backend successfully refreshes the Copilot token whenever it expires.
- Ran Next.js production build (`npm run build`) successfully with no compilation errors.
- Ran all unit tests in `tests/unit/github-adaptive-routing.test.js` and confirmed they pass.

### DIFF
--- a/src/sse/services/tokenRefresh.js
+++ b/src/sse/services/tokenRefresh.js
@@ -220,6 +220,9 @@ export async function updateProviderCredentials(connectionId, newCredentials) {
  */
 export async function checkAndRefreshToken(provider, credentials) {
   let creds = { ...credentials };
+  if (!creds.connectionId && creds.id) {
+    creds.connectionId = creds.id;
+  }
 
   // ── 1. Regular access-token expiry ────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @v01dgrace for the contribution! Merged into master. 🙏

Issue #1639, opened 2026-06-03
Usage bug


JESVN:
1.switching between overview and details may fail.
<img width="1584" height="744" alt="Image" src="https://github.com/user-attachments/assets/cac7fd81-d06e-431c-a9f5-e093ee142d55" />

2.Click somewhere else and then click [Usage]. This will inevitably cause the "Overview" data to disappear.
<img width="1842" height="715" alt="Image" src="https://github.com/user-attachments/assets/211dc11f-4628-4dff-ae40-72eedd3ef552" />

PR #1738, opened 2026-06-08
fix: count alternate usage token fields

Jordannst:
## Description
Addresses the Overview zero-token case reported in #1639.

Usage Overview can show `0` tokens even when Recent Requests contains non-zero usage rows. The today/24h aggregation only reads OpenAI-style `prompt_tokens` / `completion_tokens` from the stored token JSON, while some providers/store paths use `input_tokens` / `output_tokens`.

This PR focuses on the data aggregation path for Overview cards. It does not attempt to change the Overview / Details tab switching behavior.

## Changes
- Add `input_tokens` / `output_tokens` fallback support in `getUsageStats()` for today/24h aggregation.
- Fall back to normalized `usageHistory.promptTokens` / `usageHistory.completionTokens` columns when token JSON fields are missing.
- Add a focused unit test covering usage rows stored with `input_tokens` / `output_tokens`.
- Add `vitest` as a devDependency so existing Vitest unit tests can run from a fresh install.

## Verification
- `./node_modules/.bin/vitest run --config tests/vitest.config.js tests/unit/usage-stats-token-fallback.test.js`
- `./node_modules/.bin/vitest run --config tests/vitest.config.js tests/unit/dashboard-guard.test.js tests/unit/gemini-usage-projectid.test.js tests/unit/minimax-usage.test.js tests/unit/usage-stats-token-fallback.test.js`
- `npx eslint src/lib/db/repos/usageRepo.js tests/unit/usage-stats-token-fallback.test.js`
- `git diff --check`

## Note
`tests/unit/translator-request-normalization.test.js` currently fails in unrelated normalization / `parseSSELine` assertions, so it was excluded from the scoped verification for this Usage fix.

Refs #1639

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,6 +54,7 @@
     "eslint": "^9",
     "eslint-config-next": "16.1.6",
     "postcss": "^8.5.6",
-    "tailwindcss": "^4"
+    "tailwindcss": "^4",
+    "vitest": "^4.1.8"
   }
 }
--- a/src/lib/db/repos/usageRepo.js
+++ b/src/lib/db/repos/usageRepo.js
@@ -545,8 +545,8 @@ export async function getUsageStats(period = "all") {
 
     for (const r of filtered) {
       const tokens = parseJson(r.tokens, {}) || {};
-      const promptTokens = tokens.prompt_tokens || 0;
-      const completionTokens = tokens.completion_tokens || 0;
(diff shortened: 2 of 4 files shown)


Jordannst:
Rebased this PR onto latest upstream/master and re-ran validation.

Verification:
- `npx vitest run tests/unit/usage-stats-token-fallback.test.js` ✅
- `npm run build` ✅

Also added a minimal Vitest config so tests resolve the existing `@/*` alias from `jsconfig.json`.

Issue #1749, opened 2026-06-09
Bug: Gemma 4 thinking output is included in final response content when routed through 9router


zvwgvx:
### Description

When using Gemma 4 31B directly in Google AI Studio, the model output is displayed correctly:

* The thinking/thoughts section is separated from the final response.
* The final response only contains the actual assistant answer.

However, when using the same model through 9router, the thinking output appears to be included in the final assistant response content.

I observed this issue through two different front-end clients connected to 9router:

* OpenWebUI
* Hermes Agent

Both clients display thinking-like text as part of the normal assistant response, before the final answer.

### Reproduction

Model:

```text
Gemma 4 31B
```

Provider:

```text
Google AI Studio / Gemini API
```

Gateway:

```text
9router
```

Front-end clients used for testing:

```text
OpenWebUI
Hermes Agent
```

Input:

```text
Hello World
```

### Expected behavior

The response content shown in OpenWebUI or Hermes Agent should only contain the final assistant answer, for example:

```text
Hello! How can I help you today?
```

or:

```text
Hello World! How can I help you today?
```

### Actual behavior

When using Google AI Studio directly, the thoughts are shown separately from the final response.

Example from AI Studio:

```text
Thoughts:

The user said "Hello World".
This is a classic introductory phrase, often used in programming.

* Greeting: A friendly response.
* Acknowledgment: Recognizing the "Hello World" trope.
* Call to action: Asking how I can help.
```

Final response:

```text
Hello! How can I help you today?
```

But when routed through 9router and displayed in OpenWebUI, the thinking-like content is included in the normal assistant message:

```text
The user said "Hello World".

This is a standard greeting, often used as the first program in learning a new language. It's a friendly, open-ended start to a conversation.

* Respond politely.
* Acknowledge the "Hello World" trope (optional but adds personality).
* Offer assistance.

"Hello! How can I help you today?" or "Hello World! Welcome. What's on your mind?"Hello World! How can I help you today?
```

When using Hermes Agent with 9router, similar thinking-like content is also included in the normal response:

```text
The user [Zang Vũ] said "Hello World".
The user's language preference is Vietnamese (from memory).
I should respond in Vietnamese.Chào Zang Vũ! Rất vui được gặp bạn. Mình có thể giúp gì cho bạn hôm nay?
```

### Screenshots

I attached screenshots showing:

1. Google AI Studio, where thoughts and final response are separated correctly.
[...]

PR #1752, opened 2026-06-09
Fix Gemini thought parts leaking into assistant content (issue: #1749)

zvwgvx:
## Summary

This PR fixes an issue where Gemini/Gemma streaming responses could expose model thinking text as normal assistant content in OpenAI-compatible clients.

This was observed with Gemma 4 from Google AI Studio when routed through 9router and displayed in clients such as OpenWebUI and Hermes Agent.

ISSUE: #1749 

## Problem

Gemini can stream model thinking as text parts marked with `thought: true`.

Previously, the Gemini-to-OpenAI streaming translator only separated thought content in cases where a thought signature was present. Some Google AI Studio / Gemini API responses can emit thought parts without a `thoughtSignature`, so those parts were treated as regular text and forwarded as `delta.content`.

As a result, OpenAI-compatible clients displayed the thinking text directly in the assistant message before the final answer.

## Changes

This PR updates the streaming Gemini-to-OpenAI conversion logic so that text parts marked as `thought: true` are no longer emitted as normal assistant content.

Instead:

* normal text parts are mapped to `delta.content`
* thought text parts are mapped to `delta.reasoning_content`

This keeps model reasoning/thinking separate from the final assistant response for clients that support reasoning fields, while preventing thinking text from leaking into normal chat content.

## Validation

I verified the fix with the following checks:

* Confirmed the fix is present in the latest HEAD
* Added/updated unit coverage for Gemini thinking conversion
* `gemini-to-openai-thinking.test.js` passes: 2/2 tests
* Confirmed the non-streaming handler already separates thought content into `reasoning_content`
* Built the project locally after the fix and confirmed the generated bundle includes the updated behavior

## Notes

The issue mainly affected streaming responses. Non-streaming conversion already handled thought separation correctly.

### DIFF
--- a/open-sse/translator/response/gemini-to-openai.js
+++ b/open-sse/translator/response/gemini-to-openai.js
@@ -92,7 +92,10 @@ export function geminiToOpenAIResponse(chunk, state) {
         continue;
       }
 
-      // Text content (non-thinking)
+      // Text content. Gemini marks model-internal thinking with `thought: true`.
+      // Some responses include a thoughtSignature, but Google AI Studio/Gemini API
+      // can also stream thought parts without a signature; those must not be
+      // surfaced as normal assistant content in OpenAI-compatible clients.
       if (part.text !== undefined && part.text !== "") {
         results.push({
           id: `chatcmpl-${state.messageId}`,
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @zvwgvx for the contribution! Merged into master. 🙏

PR #1765, opened 2026-06-10
fix: strip claude billing header from openai requests

dajinglingpake:
## Summary
- strip the Anthropic `x-anthropic-billing-header` from Claude system prompts when translating requests to OpenAI format
- preserve the remaining system prompt content
- add unit coverage for the header stripping behavior

## Why
Claude clients may include a dynamic Anthropic billing header in the system prompt. When forwarded to OpenAI-compatible providers, this metadata is not useful and can change the prompt prefix, reducing prompt cache hits for routed models.

## Tests
- `npx eslint "open-sse/translator/request/claude-to-openai.js" "tests/unit/claude-to-openai-billing-header.test.js"`
- `npx vitest run "tests/unit/claude-to-openai-billing-header.test.js"`

### DIFF
--- a/open-sse/translator/request/claude-to-openai.js
+++ b/open-sse/translator/request/claude-to-openai.js
@@ -2,6 +2,11 @@ import { register } from "../index.js";
 import { FORMATS } from "../formats.js";
 import { adjustMaxTokens } from "../helpers/maxTokensHelper.js";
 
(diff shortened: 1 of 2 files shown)


kidboy-man (approved):

decolua:
Thanks @dajinglingpake for the contribution! Merged into master. 🙏

PR #1767, opened 2026-06-10
fix: avoid partial usage stats on initial load

dajinglingpake:
## Summary
- avoid creating partial usage stats from SSE before the initial REST stats load completes
- keep real-time SSE merges limited to existing full stats
- remove a stale eslint-disable comment that is no longer needed

## Why
The usage dashboard opens both a REST stats request and an SSE stream. If the SSE message arrives first, the previous code creates a partial stats object containing only real-time fields and marks loading as complete. That can make overview cards and tables briefly render zero values until another refresh.

## Tests
- `npx eslint "src/shared/components/UsageStats.js"`
- `npm run build`


decolua:
Thanks @dajinglingpake for the contribution! Merged into master. 🙏

PR #1770, opened 2026-06-10
fix: show provider node name instead of connection name for custom providers

### LINKED ISSUE #1769: Custom providers show connection name instead of provider node name
## Bug Description

Custom providers (openai-compatible, anthropic-compatible) display the **connection record name** instead of the **provider node name** in two places:

1. **Usage page — React Flow topology nodes** (ProviderTopology.js)
2. **Add Model to Combos — group header labels** (ModelSelectModal.js)

When two custom providers have different endpoints but share the same connection name (e.g. both named "Default Key"), they appear **identical** in both the topology and the model picker, making them indistinguishable.

## Steps to Reproduce

1. Create two custom provider nodes with different names (e.g. "My GPT Endpoint" and "My Claude Endpoint")
2. Add a connection to each — use the same connection name (e.g. "Default Key")
3. Go to **Usage** page → topology shows both nodes as "Default Key"
4. Open **Add Model to Combos** → both custom providers show as "Default Key" group headers

## Expected Behavior

Both locations should display the **provider node name** (e.g. "My GPT Endpoint", "My Claude Endpoint") — the user-assigned name that identifies the endpoint — not the connection/API-key entry name.

## Root Cause

**ModelSelectModal.js (line 211):**
```js
// BEFORE (bug): connection.name wins over node name
const displayName = connection?.name || matchedNode?.name || providerInfo.name;
```

**ProviderTopology.js (line 152):**
```js
// BEFORE (bug): p.name (connection name) used for custom providers
label: (config.name !== p.provider ? config.name : null) || p.name || p.provider
```

For custom providers, `AI_PROVIDERS` has no entry, so the first term is `null`, and the label falls to `p.name` — which is the **connection name**, not the provider node name.

## Fix

1. **ModelSelectModal.js**: Swap priority to `matchedNode?.name` first
2. **UsageStats.js**: Fetch `/api/provider-nodes` in parallel, merge `nodeName` into providers array
3. **ProviderTopology.js**: Prefer `p.nodeName` before `p.name` in label logic

See PR for the implementation.

ricatix:
## Summary

Fixes custom providers displaying the connection record name instead of the provider node name in:

- **Usage page** — React Flow topology nodes
- **Add Model to Combos** — group header labels

When two custom providers have different endpoints but share the same connection name, they now correctly show their distinct provider node names instead of appearing identical.

## Changes

| File | Change |
|------|--------|
| `ModelSelectModal.js` | Swap displayName priority: `matchedNode?.name` first, then `connection?.name` |
| `UsageStats.js` | Fetch `/api/provider-nodes` in parallel with `/api/providers`, merge `nodeName` into providers array |
| `ProviderTopology.js` | Prefer `p.nodeName` before `p.name` in label logic |

## Root Cause

For custom providers (openai-compatible, anthropic-compatible), `AI_PROVIDERS` has no static entry. The label/displayName resolution chain fell through to `connection.name` (the API-key entry name) instead of the provider node name (the user-assigned endpoint name).

## Testing

- [ ] Create two custom providers with different node names but same connection name
- [ ] Verify Usage topology shows distinct node names
- [ ] Verify Add Model to Combos shows distinct node names
- [ ] Verify known providers (OpenAI, Anthropic, etc.) still display correctly

Closes #1769

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,3 +72,7 @@ ecosystem.config.*
 scripts/agSniffer/*
 gitbooks/*
 gitbook/README.md
+
+# Local agent knowledge files
+AGENTS.md
+**/AGENTS.md
--- a/src/app/(dashboard)/dashboard/usage/components/ProviderTopology.js
+++ b/src/app/(dashboard)/dashboard/usage/components/ProviderTopology.js
@@ -149,7 +149,7 @@ function buildLayout(providers, activeSet, lastSet, errorSet) {
     const error = !active && errorSet.has(p.provider?.toLowerCase());
     const nodeId = `provider-${p.provider}`;
     const data = {
-      label: (config.name !== p.provider ? config.name : null) || p.name || p.provider,
+      label: (config.name !== p.provider ? config.name : null) || p.nodeName || p.name || p.provider,
       color: config.color || "#6b7280",
       imageUrl: getProviderImageUrl(p.provider),
       textIcon: config.textIcon || (p.provider || "?").slice(0, 2).toUpperCase(),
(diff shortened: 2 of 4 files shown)


decolua:
Thanks @ricatix for the contribution! The fix has been applied to master. 🙏

PR #1773, opened 2026-06-10
fix: include noAuth providers (opencode) in GET /v1/models

### LINKED ISSUE #1046: OpenCode Free models missing from /v1/models endpoint
## Problem
Models from the **OpenCode Free** provider (e.g. `oc/minimax-m2.5-free`, `oc/hy3-preview-free`, `oc/nemotron-3-super-free`, `oc/deepseek-v4-flash-free`) do not appear on the `/v1/models` endpoint, even though they show correctly on the provider detail page in the dashboard.
### Suspected Root Cause
`/v1/models` builds its list from:
1. Active provider connections (stored in DB)
2. Static `PROVIDER_MODELS` definitions
OpenCode Free is configured as a `passthroughModels` + `noAuth` provider with a `modelsFetcher` that dynamically pulls models from `https://opencode.ai/zen/v1/models`. Since there are no stored connections for no-auth providers, the models are never included in the `/v1/models` response.

omartuhintvs:
## Problem

\`oc/\` (OpenCode Free) models never appear in \`GET /v1/models\`. This makes it impossible to discover them via the API or include them in combos automatically.

**Root cause:** \`buildModelsList()\` in \`src/app/api/v1/models/route.js\` builds an \`activeConnectionByProvider\` map from stored DB connections. NoAuth providers like \`opencode\` never have a stored connection, so they are silently skipped in the \`else\` branch (when at least one other provider IS connected).

Closes #1046

## Reproduce the bug (before fix)

**Requires:** at least one other provider connected (e.g. Kiro). OpenCode is noAuth — nothing to connect.

```bash
curl -s http://localhost:20128/v1/models | jq '[.data[].id | select(startswith("oc/"))]'
# → []   (empty — bug)
```

Confirm OpenCode models DO work when called directly:

```bash
curl -s http://localhost:20128/v1/chat/completions \
  -d '{"model":"oc/big-pickle","messages":[{"role":"user","content":"hi"}],"max_tokens":10}'
# → returns a valid response (models work, just invisible in /v1/models)
```

### `src/app/api/v1/models/route.js`
Inside the `else` branch (triggered when stored connections exist), inject a synthetic entry for every `noAuth` provider that has static models defined in `PROVIDER_MODELS`. This makes the existing iteration logic emit their models normally.

```js
// noAuth providers (e.g. opencode, uncloseai) have no stored connections but
// should appear in /v1/models when they have static models defined.
for (const [providerId, providerConfig] of Object.entries(AI_PROVIDERS)) {
  if (providerConfig.noAuth && !activeConnectionByProvider.has(providerId)) {
    const staticAlias = PROVIDER_ID_TO_ALIAS[providerId] || providerId;
    if (PROVIDER_MODELS[staticAlias]?.length) {
      activeConnectionByProvider.set(providerId, { provider: providerId, providerSpecificData: {} });
    }
  }
}
```

**Note:** this also affects `uncloseai` (the only other `noAuth: true` provider in `providers.js`), which is correct — its models were similarly invisible.

### `open-sse/services/opencodeModels.js` (new)
Live resolver that fetches \`https://opencode.ai/zen/v1/models\` (no auth required), filters to free-tier models, and caches for 5 min with a 3-second timeout. Falls back to the static \`PROVIDER_MODELS\` list on any failure.

Free model detection: id ends with \`-free\` **or** is in the known-free set (currently: \`big-pickle\`).

### `open-sse/config/providerModels.js`
Uncommented and updated the \`oc:\` model list with current live free models (verified via \`opencode.ai/zen/v1/models\`):

\`big-pickle\`, \`deepseek-v4-flash-free\`, \`mimo-v2.5-free\`, \`minimax-m3-free\`, \`nemotron-3-ultra-free\`, \`north-mini-code-free\`, \`qwen3.6-plus-free\`

Two models (\`qwen3.6-plus-free\`, \`minimax-m3-free\`) were missing from the original commented list.

### Option A — run the unit tests locally

```bash
git clone https://github.com/omartuhintvs/9router.git /tmp/9router-fix
cd /tmp/9router-fix
git checkout fix/opencode-noauth-models-v1
cd tests && npm install
./node_modules/.bin/vitest run --reporter=verbose \
  unit/opencode-models.test.js \
  unit/noauth-models-injection.test.js
# → 17 passed
```

### Option B — against a running 9router instance

After applying the patch:

```bash
# Should now list oc/ models
curl -s http://localhost:20128/v1/models | jq '[.data[].id | select(startswith("oc/"))]'
# → ["oc/big-pickle","oc/deepseek-v4-flash-free","oc/mimo-v2.5-free",...]

# Live API that drives the resolver (public, no auth)
curl -s https://opencode.ai/zen/v1/models \
  -H "x-opencode-client: desktop" \
  | jq '[.data[] | select(.id | endswith("-free") or . == "big-pickle") | .id]'
```

## Test coverage

- \`tests/unit/opencode-models.test.js\` — 12 tests: free-model filtering, cache hit/expiry, HTTP errors, network failures, display-name generation (with and without \`-free\` suffix), request headers, missing-data-key fallback
- \`tests/unit/noauth-models-injection.test.js\` — 5 tests: bug scenario (oc/ appears when kiro connected), coexistence with connected providers, non-noAuth providers NOT injected, live resolver override (with \`toHaveBeenCalled\` guard), empty PROVIDER_MODELS guard (uses \`Object.defineProperty\` + \`try/finally\` for safe isolation)

### DIFF
--- a/open-sse/services/opencodeModels.js
+++ b/open-sse/services/opencodeModels.js
@@ -1,0 +1,51 @@
+/**
+ * Live model resolver for OpenCode Free (noAuth).
+ * Fetches https://opencode.ai/zen/v1/models — no credentials needed.
+ */
+
+const MODELS_URL = "https://opencode.ai/zen/v1/models";
+const TIMEOUT_MS = 5000;
+
+let _cache = null;
+let _cacheAt = 0;
+const CACHE_TTL_MS = 5 * 60 * 1000; // 5 min
+
+export function clearOpencodeModelsCache() {
+  _cache = null;
+  _cacheAt = 0;
+}
+
+/**
+ * @returns {{ models: { id: string }[] } | null}
+ */
+export async function resolveOpencodeModels() {
+  const now = Date.now();
+  if (_cache && now - _cacheAt < CACHE_TTL_MS) return _cache;
+
+  try {
+    const controller = new AbortController();
+    const tid = setTimeout(() => controller.abort(), TIMEOUT_MS);
+    const res = await fetch(MODELS_URL, {
+      headers: { "x-opencode-client": "desktop" },
+      cache: "no-store",
+      signal: controller.signal,
+    });
+    clearTimeout(tid);
+    if (!res.ok) return null;
+
+    const data = await res.json();
+    // Response shape: { data: [{ id, ... }] } (OpenAI-style) or [{ id }]
+    const raw = Array.isArray(data) ? data : (data?.data || data?.models || []);
+    const models = raw
+      .map((m) => ({ id: m?.id || m?.name }))
+      .filter((m) => typeof m.id === "string" && m.id.trim() !== "");
+
+    if (!models.length) return null;
+
+    _cache = { models };
+    _cacheAt = now;
+    return _cache;
+  } catch {
+    return null;
+  }
+}
--- a/src/app/api/v1/models/route.js
+++ b/src/app/api/v1/models/route.js
@@ -286,6 +286,20 @@ export async function buildModelsList(kindFilter) {
       });
     }
   } else {
+    // noAuth providers (e.g. opencode, uncloseai) have no stored connections but
+    // should appear in /v1/models when they have static models OR a live resolver
+    // (e.g. opencode fetches its catalog live, so its static list is empty).
+    for (const [providerId, providerConfig] of Object.entries(AI_PROVIDERS)) {
+      if (!providerConfig.noAuth) continue;
+      if (activeConnectionByProvider.has(providerId)) continue;
+      const staticAlias = PROVIDER_ID_TO_ALIAS[providerId] || providerId;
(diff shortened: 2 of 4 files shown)


kacipbuah:
Friendly maintainer ping 🙏

Including noAuth providers in `/v1/models` seems important for model discovery/UX. Could you review when available?

cc @decolua

arpicato:
Gentle maintainer bump 🙏

This looks like a small but useful fix for model discovery: noAuth providers like OpenCode currently work for completions but are invisible in `/v1/models`, which makes them hard to discover or use in downstream tooling.

Would appreciate a review when you have time. Thanks!

@decolua

omartuhintvs:
Rebased onto master (v0.5.18). Cherry-picked the two fix commits — resolved conflict in `src/app/api/v1/models/route.js` (noAuth injection block) and the modify/delete on `tests/unit/noauth-models-injection.test.js`. Branch is clean and mergeable.

PR #1774, opened 2026-06-10
fix: remove single-connection restriction for compatible provider nodes

### LINKED ISSUE #1718: can't add bulk key
"only one connection is allowed" , thats the error when i try to add multiple keys

### LINKED ISSUE #2119: List Model from custom Provider OpenAI Compatile Not Show in Combo
Trying to add some provider Agent, tested OK, model list is imported autmatically and everything is connected. when made combo, the model from openAi Compatible custom provider not showing.
the Model is Ok when the agent directly pointing with model name.

mrnasil:
## Summary

Remove the "Only one connection is allowed" restriction for compatible provider nodes (OpenAI Compatible, Anthropic Compatible, Custom Embedding).

Multiple API keys can now be connected to the same compatible node for round-robin usage.

## Changes

1. **`route.js`** — Dropped `existingConnections` duplicate guards for all three compatible provider types. 404 on missing node still works. `providerSpecificData` assignment unchanged.

2. **`page.new.js`** — Removed "Only one connection is allowed" warning text. Removed `!isCompatible` guards from Add/Add Connection buttons so they appear for compatible nodes.

3. **`page.js`** (old version) — Same button visibility fix: removed `isCompatible` from the connection bar buttons and header button label.

4. **`tests/unit/compatible-provider-connections.test.js`** — Changed duplicate-guard test to verify multiple connections succeed (201 + 2 connections stored).

Closes #1718

Closes #2119


mrnasil:
Closes #2119 — compatible provider custom models not showing in combo (ModelSelectModal). Same root cause as the broader compatible provider fix.

PR #1784, opened 2026-06-11
fix: resolve ReferenceError for allProviders in ModelSelectModal

mustafabozkaya:
This PR fixes a runtime ReferenceError that prevented the Combos page and potentially other pages from rendering properly.

### Problem
In \src/shared/components/ModelSelectModal.js\, \llProviders\ was being used in the dependency array of a \useEffect\ hook *before* it was declared with \useMemo\. This caused a classic Temporal Dead Zone (TDZ) error in modern environments (Next.js 16.x).

Additionally, \kiloFreeModels\ was present in the dependency array of the \groupedModels\ hook but wasn't defined anywhere in the component, while the actually used state variable \dynamicModels\ was missing from dependencies.

### Solution
- Moved the \llProviders\ declaration above the \useEffect\ hook that depends on it.
- Replaced the undefined \kiloFreeModels\ dependency with the correct \dynamicModels\ dependency in the \groupedModels\ useMemo hook.

These changes resolve the build and runtime errors while ensuring hooks have the correct dependencies.

### DIFF
--- a/src/shared/components/ModelSelectModal.js
+++ b/src/shared/components/ModelSelectModal.js
@@ -45,6 +45,7 @@ export default function ModelSelectModal({
   const [providerNodes, setProviderNodes] = useState([]);
   const [customModels, setCustomModels] = useState([]);
   const [disabledModels, setDisabledModels] = useState({});
+  const [dynamicModels, setDynamicModels] = useState({});
 
   const fetchCombos = async () => {
     try {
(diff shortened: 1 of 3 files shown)


kacipbuah:
Friendly maintainer ping 🙏

This resolves a `ReferenceError` in `ModelSelectModal`, so it may impact frontend stability directly. Could you review when convenient?

cc @decolua

PR #1817, opened 2026-06-15
fix(codex): normalize max reasoning effort

ryanngit:
## Summary
- maps Codex reasoning effort max to xhigh in the Codex executor only
- preserves nested reasoning.effort and adds missing summary auto
- leaves provider-native max support for other providers untouched

## Checks
- direct Codex transform assertion for top-level and nested max
- git diff --check


ryanngit:

PR #1818, opened 2026-06-15
fix(cli): detect standalone server path

ryanngit:
## Summary
- Detect current .next/standalone/server.js output layout
- Preserve support for older nested standalone app layouts
- Improve failure output when server.js is missing

## Tests
- node --check cli/scripts/build-cli.js


ryanngit:
Closing as stale after re-validating the v0.5.18 live patch set. The current verified local diff has no CLI standalone-path patch to upstream.

PR #1820, opened 2026-06-15
fix(codex): normalize fast service tier

ryanngit:
## Summary
- maps Codex service_tier fast to upstream priority
- strips unsupported Codex service_tier values before request forwarding

## Checks
- direct Codex transform assertion for fast and unsupported tiers
- git diff --check


ryanngit:

PR #1383, opened 2026-05-24
fix(antigravity-to-openai): preserve required fields in tool schemas (closes #1368)

anuragg-saxenaa:
## Summary

When OpenCode (or any AI SDK client) sends tool schemas with JSON Schema Draft 2020-12 `required` arrays, 9router's Antigravity translator was stripping the `required` constraint — causing the model to call `glob`/`grep` tools without the mandatory `pattern` argument.

**Root cause:** The translator used `cleanJSONSchemaForAntigravity` (geminiHelper) which removes `required` when all its entries are referenced in constraint blocks (`allOf`/`$defs`/`$ref`) that get flattened away by the schema normalization pipeline.

**Fix:** Added `cleanSchemaPreservingRequired()` — a targeted cleaner that:
1. Strips only genuinely irrelevant metadata (`$schema`, `$defs`, `$ref`, `enumDescriptions`, etc.)
2. **Preserves** `required` arrays that reference fields still present in `properties`
3. Recursively handles nested schema objects

## Also fixed: NVIDIA `nv-embedqa-e5-v5` requires `input_type`

Added `input_type: "query"` to the NVIDIA embeddings body builder for the `nv-embedqa-e5-v5` asymmetric model (closes #1378).

## Changes

- `open-sse/translator/request/antigravity-to-openai.js` — new `cleanSchemaPreservingRequired()` + switched tool schema cleaning to use it
- `open-sse/handlers/embeddingProviders/openai.js` — add `input_type: "query"` for NVIDIA nv-embedqa-e5-v5

Closes #1368
Closes #1378

PR #1826, opened 2026-06-15
fix: preserve Antigravity tool property names

### LINKED ISSUE #1368: [bug] [provider:antigravity] glob/grep schemas lose required 'pattern' parameter in translation
When using `OpenCode` with `ag/gemini-3-flash` (Antigravity provider), 9router's translation layer strips the `pattern` property and the `required` constraint from the `glob` and `grep` tool definitions. 

This causes the model to call the tool without providing a `pattern` argument, triggering schema errors.

### Schema Comparison

**Original Schema (from OpenCode):**
```json
"parameters": {
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "type": "object",
  "properties": {
    "pattern": {
      "type": "string",
      "description": "The glob pattern to match files against"
    },
    "path": {
      "type": "string",
      "description": "The directory to search in..."
    }
  },
  "required": [
    "pattern"
  ]
}
```

**Translated Schema (sent to Provider):**
```json
"parameters": {
  "type": "object",
  "properties": {
    "path": {
      "type": "string",
      "description": "The directory to search in..."
    }
  }
}
```
*Note: `"pattern"` is missing from `properties`, and the `"required"` constraint is completely removed.*

### Error Output
```
The glob tool was called with invalid arguments: SchemaError (Missing key at ["pattern"]).
Please rewrite the input so it satisfies the expected schema.
```

CtriXin:
## Summary

Fixes #1368.

OpenCode sends tools such as `grep`/`glob` with a required parameter named `pattern`. The Antigravity schema cleaner treated every key named `pattern` as an unsupported JSON Schema keyword, including parameter names under `properties`, so it removed `properties.pattern`. The later required-field cleanup then removed `required: ["pattern"]` too.

This patch preserves user-defined parameter names inside JSON Schema `properties` maps while still removing unsupported schema constraints from each property's nested schema.

## What changed

- Preserve property names under `properties`, including `pattern`, `default`, or other names that can also be JSON Schema keywords.
- Continue removing unsupported constraints such as a nested string schema's regex `pattern`.
- Add regression coverage for the OpenCode `grep` shape through:
  - `OpenAI -> Antigravity` request translation
  - the Antigravity executor's second schema sanitization pass

## Relation to #1383

This is a smaller targeted fix for the OpenCode/Antigravity path in #1368. The important path for OpenCode using `ag/gemini-*` is `OpenAI -> Antigravity` plus the executor's final sanitization pass; this PR covers both with tests.

## Validation

```bash
# Test Files 1 passed (1)
# Tests 3 passed | 2 expected fail (5)

npx vitest run --config tests/vitest.config.js tests/translator/
# Test Files 9 passed | 1 skipped (10)
# Tests 123 passed | 21 expected fail | 1 skipped (145)
```

### DIFF
--- a/open-sse/translator/helpers/geminiHelper.js
+++ b/open-sse/translator/helpers/geminiHelper.js
@@ -119,9 +119,12 @@ export function generateProjectId() {
   return `${adj}-${noun}-${crypto.randomUUID().slice(0, 5)}`;
 }
 
-// Helper: Remove unsupported keywords recursively from object/array
-// Also strips all vendor extension fields (x- prefixed) not supported by Gemini
-function removeUnsupportedKeywords(obj, keywords) {
+// Helper: Remove unsupported schema keywords recursively from object/array.
+// Keys inside a JSON Schema `properties` map are user-defined parameter names,
+// not schema keywords. Preserve them even when they are named `pattern`,
+// `default`, etc., and only clean the schema nested under each parameter.
+// Also strips all vendor extension fields (x- prefixed) not supported by Gemini.
+function removeUnsupportedKeywords(obj, keywords, context = {}) {
   if (!obj || typeof obj !== "object") return;
 
   if (Array.isArray(obj)) {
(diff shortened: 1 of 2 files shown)


CtriXin:
Closing: opened against the wrong upstream repo by mistake. I am withdrawing this PR.

PR #1832, opened 2026-06-15
fix: strip temperature for Anthropic models that reject it (claude-opus-4)

### LINKED ISSUE #1748: `temperature` parameter causes HTTP 400 on claude-opus-4 and newer Anthropic models
## Bug Description

When routing requests to `claude-opus-4` (and potentially other new Anthropic models), 
the router forwards a `temperature` parameter that Anthropic has deprecated for these models. 
This causes a `400 Bad Request` error.

## Error Message

```
HTTP 400: [claude/claude-opus-4-8] [400]: {
  "type": "error",
  "error": {
    "type": "invalid_request_error",
    "message": "temperature is deprecated for this model."
  }
}
```

## Steps to Reproduce

1. Configure 9router with a route pointing to `claude-opus-4` or `claude-opus-4-8`
2. Send any request that includes a `temperature` field in the payload

## Expected Behavior

The router should detect that the target model does not support `temperature` and strip 
(or ignore) that parameter before forwarding the request.

## Proposed Fix

Add a model-aware parameter filter: for Anthropic models that do not support `temperature` 
(e.g. `claude-opus-4`, `claude-opus-4-5`), remove `temperature` from the request body 
before proxying upstream.

## Environment

- Router: 9router (decolua/9router)
- Model: `claude-opus-4` / `claude-opus-4-8`
- Provider: Anthropic

0xSkybreaker:
@
## Summary

Fixes #1748 — the `temperature` parameter is deprecated for newer Anthropic models (`claude-opus-4`, `claude-opus-4-8`, etc.), causing HTTP 400:

```
"temperature is deprecated for this model."
```

## Fix

`DefaultExecutor` now strips `temperature` from the request body for `claude-opus-4` series models, following the same `supportsTemperature()` pattern already used in `GithubExecutor`.

Only `claude-opus-4*` models are affected. All other models continue to pass `temperature` as before.

## Files changed

- `open-sse/executors/default.js` — add `supportsTemperature()` + stripping in `transformRequest()`

@

### DIFF
--- a/open-sse/executors/default.js
+++ b/open-sse/executors/default.js
@@ -11,12 +11,25 @@ export class DefaultExecutor extends BaseExecutor {
     super(provider, PROVIDERS[provider] || PROVIDERS.openai);
   }
 
+  // Anthropic models that reject the deprecated `temperature` parameter
+  supportsTemperature(model) {
+    if (!model) return true;
+    // claude-opus-4 series: temperature is deprecated
+    if (/claude-opus-4/i.test(model)) return false;
+    return true;
+  }
+
   transformRequest(model, body) {
     const transformed = this.applyJsonSchemaFallback(body);
 
     if (transformed && typeof transformed === "object") {
       if (this.provider === "cerebras" || this.provider === "mistral") {
         delete transformed.client_metadata;
+      }
+
+      // Strip temperature for models that reject it (e.g. claude-opus-4)
+      if (!this.supportsTemperature(model) && transformed.temperature !== undefined) {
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @0xSkybreaker for the contribution and for tracking down #1748! 🙏 We refactored the fix into a config-driven `stripUnsupportedParams` helper (in `translator/concerns/paramSupport.js`) so the default and github executors share one rule table, and merged it into master. The `claude-opus-4` temperature stripping is included.

PR #1837, opened 2026-06-15
fix(usage): account for cached token pricing

ryanngit:
## Summary
- update OpenAI Standard API pricing and GPT fallback ordering for cache-aware estimates
- calculate usage with uncached input, cached input, cache write, output, reasoning, and cache-savings buckets
- preserve cache/reasoning metadata, avoid double-saving streaming usage, and surface breakdowns in usage views
- add focused pricing/cache regression coverage

## Verification
- checked current official OpenAI pricing page for Standard API rates
- node --check open-sse/providers/pricing.js
- node --check open-sse/utils/usageTracking.js
- node --check src/lib/db/repos/usageRepo.js
- node --check open-sse/handlers/chatCore/streamingHandler.js
- direct Node pricing/cache assertions passed
- git diff --check
- NEXT_DIST_DIR=.next-upstream-pricing-build NEXT_TRACING_ROOT_MODE=workspace npm run build

## Note
- local Vitest config could not run in this checkout because node_modules is symlinked to the live patch repo and does not include vitest; the added test matches the existing Vitest test style and direct Node assertions cover the same pricing cases.

### DIFF
--- a/open-sse/handlers/chatCore/streamingHandler.js
+++ b/open-sse/handlers/chatCore/streamingHandler.js
@@ -5,7 +5,7 @@ import { pipeWithDisconnect } from "../../utils/streamHandler.js";
 import { PROVIDERS } from "../../config/providers.js";
 import { STREAM_STALL_TIMEOUT_MS } from "../../config/runtimeConfig.js";
 import { buildAbortedResponsesTerminalBytes } from "../../utils/responsesStreamHelpers.js";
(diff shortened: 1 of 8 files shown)


ryanngit:
Closing as stale after re-cutting local patches against v0.5.18. The verified live diff no longer contains cache-pricing changes; current upstream/local patch work is tracked in the clean PR set.

PR #1841, opened 2026-06-16
fix: strip optional from Antigravity tool schemas

### LINKED ISSUE #1537: antigravity and gemini  400 error to fix
**[Bug Report] antigravity ang gemini provider: `optional` field in tool schema causes 400 error on all models**

---

**Description**

All models routed through the `antigravity` provider fail with a 400 error. The root cause is that the request transformer does not strip the `optional` field from tool parameter properties before forwarding to the Gemini backend. The Gemini API does not recognize `optional` as a valid field in `function_declarations`, causing `INVALID_ARGUMENT` on every request that includes tools.

This affects **all models** under antigravity, including OpenAI-style named models (e.g. `gpt-oss-120b-medium`), confirming the backend is Gemini regardless of model naming.

---

**Steps to Reproduce**

1. Configure 9router with antigravity provider
2. Use Claude CLI with any model routed to antigravity (e.g. `ag/gemini-3-flash`, `ag/gpt-oss-120b-medium`)
3. Send any request that includes tool definitions
4. Observe 400 error

---

**Error**

```
[ERROR] [400]: {
  "error": {
    "code": 400,
    "message": "Invalid JSON payload received. Unknown name \"optional\" at 'request.tools[0].function_declarations[63].parameters.properties[2].value': Cannot find field.",
    "status": "INVALID_ARGUMENT"
  }
}
```

---

**Root Cause**

Claude CLI sends tool definitions with an `optional` field per property:

```json
{
  "parameters": {
    "properties": {
      "myParam": {
        "type": "string",
        "optional": true
      }
    }
  }
}
```

Gemini API does not accept `optional` as a field. It uses the standard JSON Schema `required` array instead. The antigravity transformer currently does not remove `optional` before forwarding.

---

**Expected Behavior**

The transformer should strip `optional` from all parameter properties and (optionally) reconstruct the `required` array before sending to Gemini.

---

**Suggested Fix**

```javascript
function stripOptionalFields(tools) {
  return tools.map(tool => {
    tool?.function_declarations?.forEach(fn => {
      const props = fn?.parameters?.properties || {};
      Object.values(props).forEach(prop => delete prop.optional);
    });
    return tool;
  });
}
```

---

**Environment**
- Tool: 9router / Claude CLI
- Provider: antigravity
- Affected models: all (gemini-3-flash, gpt-oss-120b-medium, etc.)
- Error code: 400 INVALID_ARGUMENT

nkgotcode:
Fixes #1537.

## Summary

This removes the Antigravity-specific `optional` marker from nested tool schemas before forwarding requests to Gemini, matching the existing cleanup path for unsupported schema keywords.

## Tests

```bash
```

Passed: 1 test file, 3 tests, 1 expected failure.

### DIFF
--- a/open-sse/translator/formats/gemini.js
+++ b/open-sse/translator/formats/gemini.js
@@ -19,7 +19,7 @@ export const UNSUPPORTED_SCHEMA_CONSTRAINTS = [
   // Dependency keywords (not supported)
   "dependencies", "dependentSchemas", "dependentRequired",
   // Other unsupported keywords
-  "title", "if", "then", "else", "contentMediaType", "contentEncoding",
+  "title", "optional", "if", "then", "else", "contentMediaType", "contentEncoding",
   // UI/Styling properties (from Cursor tools - NOT JSON Schema standard)
   "cornerRadius", "fillColor", "fontFamily", "fontSize", "fontWeight",
   "gap", "padding", "strokeColor", "strokeThickness", "textColor"
@@ -375,4 +375,3 @@ export function cleanJSONSchemaForAntigravity(schema) {
 
   return cleaned;
 }
-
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @nkgotcode for the contribution! Merged into master. 🙏

PR #1844, opened 2026-06-16
fix(claude-to-openai): handle OpenAI-format responses in non-streaming path

### LINKED ISSUE #1836: Bug: Empty content when provider returns reasoning_content (claude-to-openai conversion)
## Bug Description

When using a model with `-claude` suffix (e.g., `xmtp/mimo-v2.5-pro-claude`), the provider correctly returns `content` and `reasoning_content` in the response, but 9Router converts this to `content: null` and `thinking: null`.

## Steps to Reproduce

1. Use model `xmtp/mimo-v2.5-pro-claude` (routes to xiaomi-tokenplan)
2. Send a simple chat completion request
3. Provider returns valid response with `content` and `reasoning_content`
4. 9Router returns `content: null`

## Provider Response (Raw)
```json
{
  "choices": [{
    "message": {
      "content": "Hello there! How can I help you today?",
      "reasoning_content": "The user simply asked Say hello..."
    }
  }]
}
```

## 9Router Client Response
```json
{
  "choices": [{
    "message": {
      "content": "",
      "role": "assistant"
    }
  }],
  "usage": {
    "prompt_tokens": 2000,
    "completion_tokens": 0
  }
}
```

## Request Details (from 9Router dashboard)
- Format conversion: `openai → claude` (stream=false)
- Provider: xiaomi-tokenplan
- Model: mimo-v2.5-pro-claude
- Status: success
- But response.content is null

## Environment
- 9Router version: 0.4.80
- Provider: xiaomi-tokenplan (Xiaomi MiMo Token Plan)

## Expected Behavior

The `content` field should contain the actual response text from the provider, not null.

## Additional Context

This issue is similar to #1189 (DeepSeek reasoning_content), but affects any provider that returns `reasoning_content` in OpenAI format response.

The provider response shows `model: mimo-v2.5-pro` and the content IS present in providerResponse, but gets lost in the conversion back to client response.

0xSkybreaker:
## Summary

Fixes #1836 — Empty content when provider returns reasoning_content in non-streaming claude-to-openai conversion.

## Root Cause

When a provider (e.g. xiaomi-tokenplan with `-claude` suffix models) returns an OpenAI-format response even though the request was translated to Claude format, the non-streaming response handler failed to recognize it. The early-return check on line 79 only looked for `responseBody.content` as a non-array, missing the case where the response has `choices[0].message.content` (OpenAI format). The handler would then process it as a Claude body, iterating over `responseBody.content || []` (which is `[]`), producing empty content.

### 1. Early return for OpenAI-format responses
Added `responseBody.choices` check to the early-return condition. If the response already has a `choices` array, it's in OpenAI format and needs no further translation.

### 2. Conditional reasoning_content strip
Changed the unconditional `delete choice.message.reasoning_content` to only strip when `choice.message.content` is non-empty. This matches the existing behavior in `sseToJsonHandler.js` and preserves reasoning_content for thinking models where it may be the only output.

### DIFF
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -76,7 +76,12 @@ export function translateNonStreamingResponse(responseBody, targetFormat, source
     // missing/null (e.g. M3 with max_tokens:1 spends the budget on thinking
     // and returns `content: null`). Returning the raw body would leave the
     // OpenAI client without a `choices` array and surface as a UI test error.
-    if (responseBody.content && !Array.isArray(responseBody.content)) return responseBody;
+    // Early return if the response is already in OpenAI format (has choices array)
+    // or if it has content as a non-array value (likely a different non-Claude format).
+    // Some providers (e.g. xiaomi-tokenplan) return OpenAI-format responses even when
+    // the request was translated to Claude format — the targetFormat is Claude but the
+    // actual response is OpenAI-native and needs no further translation.
+    if (responseBody.choices || (responseBody.content && !Array.isArray(responseBody.content))) return responseBody;
 
     let textContent = "", thinkingContent = "";
     const toolCalls = [];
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @0xSkybreaker for the contribution! Merged into master. 🙏

PR #1856, opened 2026-06-17
fix(image): prevent compatible nodes from shadowing provider aliases

lowkruc:
Problem:
Requests such as cf/@cf/black-forest-labs/flux-2-klein-9b could be routed to a user-defined openai-compatible node when that node used cf as its prefix. The image handler then rejected the request because openai-compatible-chat-* has no image adapter.

Cause:
The app-level model resolver checked provider-node prefixes after the core resolver had already mapped built-in aliases like cf to cloudflare-ai. That allowed custom compatible prefixes to override built-in provider ids and aliases.

Fix:
Build a reserved prefix set from registry provider ids, alias, aliases[], plus local aliases, and skip compatible-node prefix matching when the request uses one of those reserved prefixes. Built-in routes like cf/... and cloudflare-ai/... now stay on Cloudflare, while non-reserved compatible prefixes still work.

Tests:
Add regression coverage for a cf compatible-node collision and for a normal non-reserved compatible prefix.

### DIFF
--- a/src/sse/services/model.js
+++ b/src/sse/services/model.js
@@ -1,12 +1,20 @@
 // Re-export from open-sse with localDb integration
 import { getModelAliases, getComboByName, getProviderNodes } from "@/lib/localDb";
 import { parseModel as parseModelCore, resolveModelAliasFromMap, getModelInfoCore } from "open-sse/services/model.js";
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @lowkruc for the contribution! Merged into master. 🙏

PR #1857, opened 2026-06-17
fix: show custom models in LLM selector

Jordannst:
## Description
Fixes selector/model-list handling for user-added custom models, including vision-capable custom models typed as `imageToText`.

## Changes
- Keep custom user-added models visible in the default LLM selector/model list when appropriate.
- Expose custom `imageToText` models in the default `/v1/models` list because they are vision-capable chat models.
- Preserve typed filtering for `/v1/models/{kind}` (`image`, `tts`, `stt`, `embedding`, `image-to-text`, `web`).
- Map custom model service kinds to runtime capabilities, e.g. `imageToText` → `vision`.

## Verification
- `npx vitest run tests/unit/capabilities-service-kind.test.js`
- `npm run build`

Related to #1855

Note: the issue discussion also mentions image fallback for vision models not processing images. This PR improves selector/model-list/capability exposure for custom vision models, but does not claim to fully cover every runtime fallback/translator path.

### DIFF
--- a/open-sse/providers/capabilities.js
+++ b/open-sse/providers/capabilities.js
@@ -51,6 +51,21 @@ export const DEFAULT_CAPABILITIES = {
   maxOutput: 64000,
 };
 
+// User-added model metadata can carry dashboard service kinds instead of the
+// runtime capability names used here. Map those typed model kinds into input /
+// output capabilities so custom vision models are not treated as text-only.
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @Jordannst for the contribution! Merged into master. 🙏

PR #1859, opened 2026-06-17
fix(combo/fusion): strip tools and messages history in panel expert calls to prevent 503

### LINKED ISSUE #1858: [Bug] Fusion combo returns 503 when client request contains tools
## Bug

When a combo with the `fusion` strategy receives a request that includes `tools`, the panel models respond with `tool_calls` instead of prose. `extractPanelText` returns an empty string for all answers, so the engine hits the `"All panel models failed"` 503 branch and the user never gets a fused response.

## Reproduction

1. Create a combo, e.g. `...side` with members `ag/gemini-3-flash-agent` and `minimax/MiniMax-M3`.
2. Set its strategy to `fusion` with judge `..glm`.
3. Send a request from a tool-using client (e.g. Antigravity IDE) that has `tools` defined in the body.
4. Observe `503 "All fusion panel models failed"` and 0 judge calls in the logs.

## Evidence

From the local SQLite `requestDetails` table, both panel calls succeed (`status: success`, `200 OK`) but their bodies look like:

\`\`\`json
{
  "providerResponse": {
    "choices": [{
      "finish_reason": "tool_calls",
      "message": {
        "content": "",
        "tool_calls": [{ "function": { "name": "grep_search" } }]
      }
    }]
  }
}
\`\`\`

`providerRequest.tools` shows 15 tools were forwarded to the panel model.

## Root cause

`handleFusionChat` strips `tools` and `tool_choice` from the body in `open-sse/services/combo.js:469-470`, but the dispatch layer in `src/sse/handlers/chat.js:105` still passes the **original** `clientRawRequest` down to `handleChatCore`. The chat core / translator pipeline can re-attach tools from the raw request.

In addition, even if the `tools` array is removed, the `messages` history still contains previous `tool_calls` from the assistant and `role: "tool"` results, which causes agentic models to continue the tool loop instead of producing prose.

## Expected behavior

The fusion panel must be a "prose-only" sandbox. `tools` and previous tool history must not leak into panel calls. The judge (and any direct fallback) keeps the client's original `tools` and `stream` flag.

## Fix sketch

1. `open-sse/services/combo.js` — pass `isPanel=true` to `handleSingleModel` and strip tool-related history from `panelBody.messages` / `panelBody.input` before fan-out (filter out `role: "tool"` / `role: "function"`, drop the `tool_calls` property from assistant messages).
2. `src/sse/handlers/chat.js` — when the call is a panel call (`isPanel=true`), build a sanitized `clientRawRequest` whose `body.tools` and `body.tool_choice` are removed before forwarding to `handleSingleModelChat`, so `handleChatCore` cannot re-attach them.

## Test

Added `combo-fusion.test.js` test: \`strips previous tool history and assistant tool_calls from panel calls\`. All 7 unit tests pass.

warelik:
Closes #1858

## Summary

When using the `fusion` strategy, the panel expert models must be executed in a prose-only sandbox. If the original request contains `tools` or previous `tool_calls` / `role: tool` messages in the history, agentic panel models (such as gemini-3.5-flash-agent, MiniMax-M3) keep executing tool loops and emit `tool_calls` instead of prose. `extractPanelText()` then returns an empty string for every panel answer and the engine falls into the 503 "All fusion panel models failed" branch.

## Changes

1. **`open-sse/services/combo.js`** — pass `isPanel = true` to `handleSingleModel` for fan-out calls and clean up `panelBody.messages` / `panelBody.input` before fanning out:
   - drop `role: "tool"` and `role: "function"` messages
   - drop the `tool_calls` property from assistant messages (keeping the turn so positional context is preserved)

2. **`src/sse/handlers/chat.js`** — when `isPanel` is true, build a sanitized clone of `clientRawRequest` with `tools` and `tool_choice` removed from `clientRawRequest.body` before forwarding to `handleSingleModelChat`. This prevents `handleChatCore` from re-attaching the original client tools during the translateRequest pass.

3. **`tests/unit/combo-fusion.test.js`** — new test `strips previous tool history and assistant tool_calls from panel calls` verifies both layers. The judge call still receives the unmodified history + the synthesis prompt.

All 7 unit tests pass.

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -468,8 +468,32 @@ export async function handleFusionChat({ body, models, handleSingleModel, log, c
   // 1. Fan out to the panel in parallel: non-streaming, tools stripped (we want prose).
   const { tools, tool_choice, ...rest } = body;
   const panelBody = { ...rest, stream: false };
+
+  // Strip tool history from messages so models don't get tempted to emit tool_calls.
+  if (Array.isArray(panelBody.messages)) {
+    panelBody.messages = panelBody.messages
+      .filter((msg) => msg && msg.role !== "tool" && msg.role !== "function")
+      .map((msg) => {
+        if (msg.role === "assistant" && msg.tool_calls) {
+          const { tool_calls, ...cleanMsg } = msg;
+          return cleanMsg;
+        }
+        return msg;
+      });
+  } else if (Array.isArray(panelBody.input)) {
+    panelBody.input = panelBody.input
+      .filter((msg) => msg && msg.role !== "tool" && msg.role !== "function")
+      .map((msg) => {
+        if (msg.role === "assistant" && msg.tool_calls) {
+          const { tool_calls, ...cleanMsg } = msg;
+          return cleanMsg;
+        }
+        return msg;
+      });
+  }
+
   const t0 = Date.now();
-  const calls = panel.map((m) => withTimeout(handleSingleModel(panelBody, m), cfg.panelHardTimeoutMs));
+  const calls = panel.map((m) => withTimeout(handleSingleModel(panelBody, m, true), cfg.panelHardTimeoutMs));
   const settled = await collectPanel(calls, { ...cfg, minPanel });
   log.info("FUSION", `fan-out collected in ${Date.now() - t0}ms`);
 
--- a/src/sse/handlers/chat.js
+++ b/src/sse/handlers/chat.js
@@ -102,7 +102,14 @@ export async function handleChat(request, clientRawRequest = null) {
       return handleFusionChat({
         body,
         models: comboModels,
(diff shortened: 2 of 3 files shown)


decolua:
Thanks @warelik for the contribution! 🙏

Merged into `master` with two refinements on top of your patch:
- **Always forward `request`** to combo sub-calls — the original `cleanRawReq ? null : request` dropped `userAgent` and `sourceFormatOverride` for every combo sub-call, not just panels.
- **Flatten tool turns instead of deleting them** — tool results / `tool_calls` are now turned into assistant prose (`[Tool result: ...]`, `[Called tools: ...]`) so panel models keep the context but still can't loop on tools. The judge keeps the unmodified history.

All 7 unit tests pass. Appreciate the fix!

PR #1878, opened 2026-06-18
fix(kiro): honor thinking effort budgets

SemonCat:
## Summary
- resolve Kiro thinking budgets from OpenAI reasoning_effort, OpenAI Responses reasoning.effort, Claude output_config.effort, and Claude thinking.budget_tokens
- pass the resolved budget into the Kiro thinking system prefix while preserving the default budget for synthetic -thinking model aliases
- keep explicit none/off/disabled/non-positive thinking configs from injecting the Kiro thinking prefix

## Tests
- npx vitest run --reporter=verbose --config ./tests/vitest.config.js tests/unit/openai-to-kiro.test.js tests/translator/claude-kiro-direct.test.js
- npx vitest run --reporter=verbose --config ./tests/vitest.config.js tests/translator/thinking-unified.test.js
- npx eslint open-sse/config/kiroConstants.js open-sse/translator/request/openai-to-kiro.js open-sse/translator/request/claude-to-kiro.js tests/unit/openai-to-kiro.test.js tests/translator/claude-kiro-direct.test.js

### DIFF
--- a/open-sse/config/kiroConstants.js
+++ b/open-sse/config/kiroConstants.js
@@ -14,6 +14,7 @@
  * Kiro upstream does not advertise `-agentic` model IDs; they are a 9router
  * fiction. The suffix is stripped before the request leaves this process.
  */
+import { effortToBudget } from "../translator/concerns/thinking.js";
 
 export const KIRO_AGENTIC_SUFFIX = "-agentic";
 export const KIRO_THINKING_SUFFIX = "-thinking";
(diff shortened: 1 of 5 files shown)


decolua:
Thanks a lot for this contribution, @SemonCat! 🙏 The thinking-budget fix is genuinely useful and the test coverage was great.

We merged it into `master` with one adjustment to keep things DRY: instead of adding a dedicated effort/budget parser inside `kiroConstants.js`, we reused the existing shared `thinkingUnified.extractThinking()` helper, which already normalizes every client shape (OpenAI `reasoning_effort` / Responses `reasoning.effort`, Claude `output_config.effort` / `thinking.budget_tokens`, plus Gemini/Qwen) into a single `{ mode, budget, level }` config.

So `resolveKiroThinkingBudget()` is now a thin mapper on top of that shared parser:
- `mode: "none"` → no prefix injected
- `mode: "budget"` → use the budget directly
- `mode: "level"` → `effortToBudget(level)` (clamped to Kiro's 1..32000 in `buildThinkingSystemPrefix`)
- `mode: "auto"` / synthetic `-thinking` aliases → default budget (16000)

We also wired the real inbound `credentials.rawHeaders` into the resolver so the `Anthropic-Beta: interleaved-thinking` header is honored too. Verified end-to-end against a live Kiro credential — `low`, `high`, `thinking.budget_tokens=4096`, and `output_config.effort=high` all return 200 with reasoning output, while `reasoning_effort=none` correctly skips thinking.

Your authorship is preserved on the merged commit. 🚀

PR #1889, opened 2026-06-18
feat(provider): add CodeBuddy CN provider (copilot.tencent.com)

osyduck:
## Add CodeBuddy CN (`codebuddy-cn`) OAuth provider

Adds Tencent **CodeBuddy** (the China deployment, `copilot.tencent.com`) as a first-class OAuth provider, with end-to-end support: login, token refresh, model catalog, inference, streaming, reasoning, and per-model capabilities. Everything was derived from the official `@tencent-ai/codebuddy-code` CLI (decompiled `dist/`) and verified against live captured traffic.

### Why a dedicated `codebuddy-cn` provider (CN ≠ AI)

CodeBuddy ships as **two separate deployments that are not interchangeable** — they differ in host, OAuth realm, inference endpoint, and the model catalog. This is explicit in the CLI's `product.json` (`authentication.attributes`):

| | **codebuddy.cn** (this PR) | **codebuddy.ai** (intl.) |
|---|---|---|
| Inference / API host | `copilot.tencent.com` | `www.codebuddy.ai` |
| Auth realm (`iss`) | `www.codebuddy.cn` | `www.codebuddy.ai` |
| `X-Domain` | `www.codebuddy.cn` | `www.codebuddy.ai` |
| Model catalog | GLM 5.x / MiniMax / Kimi / DeepSeek / Hunyuan | (different per-account, intl. models) |
| Account | CN account | separate AI account |

A token minted against the `.cn` realm is **not valid** on the `.ai` host, the model ids/credits differ, and the per-model context windows/capabilities are deployment-specific. Collapsing them into one provider would mean one of: wrong endpoint, wrong realm, or a wrong/empty model list. So this PR scopes everything to `codebuddy-cn`, leaving room for a future `codebuddy-ai` provider that reuses the same flow shape with the `.ai` host and its own catalog. (The `.ai` variant is intentionally **not** included here — its model list is fetched per-account from the server and can't be enumerated without an `.ai` login.)

### What's included

**OAuth (browser polling flow)** — `src/lib/oauth/providers.js`
- Poll the token endpoint via **`GET /v2/plugin/auth/token?state=…`** (state in the query string), matching the official CLI. The earlier `POST` + JSON body never received the approved token, so login hung on "Waiting for authorization" forever.
- Use the server-provided `expiresIn` instead of a hardcoded value.

**Token refresh** — `open-sse/services/tokenRefresh*`
- New `refreshCodebuddyToken`: `POST /v2/plugin/auth/token/refresh` with the refresh token in the `X-Refresh-Token` header (`X-Auth-Refresh-Source: plugin`), parsing `{ code: 0, data: <token> }`. Wired into `REFRESH_HANDLERS`. Without it, access tokens silently expired (~24h) and forced a re-login.

**Inference** — `open-sse/providers/registry/codebuddy-cn.js`, `open-sse/executors/codebuddy-cn.js`
- Correct endpoint **`POST /v2/chat/completions`** (was `/v1/...` → `404 Route Not Found`).
- CodeBuddy rejects non-stream requests (`400` `"Non-stream chat request is currently not supported"`). The provider sets `forceStream`, and a dedicated `CodeBuddyExecutor` forces `body.stream = true` on the wire (the same-format openai→openai path otherwise leaves the client's value), while 9router still re-aggregates SSE for non-streaming clients.
- Static fingerprint headers from the CLI (`User-Agent`, `X-Product`, `X-IDE-*`, `x-requested-with`, `x-codebuddy-request`).

**Reasoning** — registry `thinkingFormat: "openai"` + executor
- CodeBuddy is a unified OpenAI-compatible gateway: **every** model (GLM/Kimi/MiniMax/DeepSeek/Hunyuan) takes reasoning via OpenAI-style `reasoning_effort` + `reasoning_summary: "auto"`, not its vendor-native thinking shape. 9router previously mapped each model to its native format (`zai`/`kimi`/`minimax`/…), which the gateway ignores — so reasoning never appeared.
- `transport.thinkingFormat: "openai"` routes all models through `reasoning_effort`, and the executor injects `reasoning_summary: "auto"` + defaults `reasoning_effort` to `"medium"` when the client sends none (respects an explicit `none`/`off`). This mirrors the CLI exactly. Verified live: `usage.reasoning_tokens > 0`.

**Per-model capabilities** — `open-sse/providers/capabilities.js`
- `PROVIDER_CAPABILITIES["codebuddy-cn"]` encodes the authoritative metadata from the gateway's model config: real context windows (GLM-5.2 / DeepSeek-V4 = 1M, etc.), max output, vision (`supportsImages`), and `thinkingCanDisable: false` for `onlyReasoning` models (clamped to minimal instead of disabled). Scoped to `codebuddy-cn` so it overrides the global name-pattern caps without affecting other providers (e.g. `glm-5v-turbo` now keeps images; `glm-5.2` context is 1M, not the generic `glm-5` pattern's 200k).

**Models + branding**
- Default model list = the 15 models exposed by the CLI's `cli` agent (`glm-5.2 … deepseek-v3-2-volc`).
- Official CodeBuddy logo at `public/providers/codebuddy-cn.png`.

### Testing

- Login, refresh, inference, streaming, and reasoning all verified against the live `copilot.tencent.com` backend using a real CN account.
- Reasoning confirmed via `usage.completion_tokens_details.reasoning_tokens`; reasoning text streams as `delta.reasoning_content` (non-stream strips it when `content` is present, per existing 9router behavior).
- Baseline/snapshot fixtures (`tests/__baseline__`, golden snapshot) updated for the renamed id and `/v2` endpoint.

### Notes for reviewers

- This started as a provider named `codebuddy`; it was renamed to `codebuddy-cn` across the registry, OAuth flow, token refresh, executor, and OAuth constants once the CN/AI split was clear. Existing `codebuddy` credentials (if any) will need a one-time re-auth as **CodeBuddy CN**.


decolua:
Thanks @osyduck for the contribution! Merged into master. 🙏

PR #1896, opened 2026-06-18
fix: preserve 'pattern' in antigravity tool schema translation (#1368)

### LINKED ISSUE #1368: [bug] [provider:antigravity] glob/grep schemas lose required 'pattern' parameter in translation

mustafabozkaya:
This PR fixes a bug where the `pattern` property and its `required` constraint were being stripped from `glob` and `grep` tool schemas when using the antigravity provider.

### Problem
The `cleanJSONSchemaForAntigravity` function in `geminiHelper.js` was using a list of unsupported keywords that included `pattern`. This caused the property to be removed from the translated schema, leading to errors when models tried to call these tools without the required `pattern` argument.

### Solution
- Removed `"pattern"` from the `UNSUPPORTED_SCHEMA_CONSTRAINTS` array in `open-sse/translator/helpers/geminiHelper.js`.
- Added a new unit test in `tests/unit/gemini-helper-schema.test.js` to verify that `pattern` and `required` are correctly preserved.

This fix ensures that tool schemas are correctly translated for the antigravity provider without losing essential properties.

Fixes #1368

### DIFF
--- a/open-sse/translator/helpers/geminiHelper.js
+++ b/open-sse/translator/helpers/geminiHelper.js
@@ -4,7 +4,7 @@
 export const UNSUPPORTED_SCHEMA_CONSTRAINTS = [
   // Basic constraints (not supported by Gemini API)
   "minLength", "maxLength", "exclusiveMinimum", "exclusiveMaximum",
-  "pattern", "minItems", "maxItems", "format",
+  "minItems", "maxItems", "format",
   // Claude rejects these in VALIDATED mode
   "default", "examples",
   // JSON Schema meta keywords
--- a/src/shared/components/ModelSelectModal.js
+++ b/src/shared/components/ModelSelectModal.js
@@ -45,6 +45,7 @@ export default function ModelSelectModal({
   const [providerNodes, setProviderNodes] = useState([]);
   const [customModels, setCustomModels] = useState([]);
   const [disabledModels, setDisabledModels] = useState({});
+  const [dynamicModels, setDynamicModels] = useState({});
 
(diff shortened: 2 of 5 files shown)


decolua:
Cảm ơn @mustafabozkaya! Fix đã được apply vào repo chính (app) tại commit f6c2f7c — `pattern` đã được giữ lại trong tool schemas cho antigravity provider. Đóng PR #1368.

PR #1897, opened 2026-06-18
fix(cli-tools): tolerate JSONC configs in CLI tool settings routes

Zireael:
## Summary

CLI tool detection fails when a tool's config file contains trailing commas (JSONC). The `readConfig`/`readSettings`/`readJson` functions used `JSON.parse()` but only caught `ENOENT` — a `SyntaxError` from trailing commas was re-thrown, causing the settings API to return HTTP 500. The UI interpreted the 500 response (missing `installed` field) as "tool not installed."

This affected 8 CLI tool routes: opencode, openclaw, kilo, droid, cowork, copilot, claude, and cline.

To explain it simply, if CLI .jsonc config file contains syntax that is invalid as per .json specs, 9router will end up saying thay the CLI is not installed.

## What changed

Each affected route's config-reading function now:

1. **Strips trailing commas** before parsing (`/,(\s*[}\]])/g` → `$1`), so JSONC-formatted configs parse without error.
2. **Returns `null` on any remaining parse error** instead of re-throwing, so the UI shows "installed but not configured" (accurate) rather than "not installed" (misleading).

The opencode fix also retains the explicit `ENOENT` check (for clarity), while the other 7 routes use a blanket `return null` in catch — both are functionally equivalent since `null` is already the ENOENT fallback.

## Verification
I removed trailing commas from my `opencode.jsonc` config file and 9router started detecting opencode as installed again.

### DIFF
--- a/src/app/api/cli-tools/claude-settings/route.js
+++ b/src/app/api/cli-tools/claude-settings/route.js
@@ -41,12 +41,12 @@ const readSettings = async () => {
   try {
     const settingsPath = getClaudeSettingsPath();
     const content = await fs.readFile(settingsPath, "utf-8");
-    return JSON.parse(content);
+    // Tolerate JSONC (trailing commas) and treat unparseable files as "no config"
+    // rather than throwing a 500 that the UI misreads as "tool not installed".
+    const stripped = content.replace(/,(\s*[}\]])/g, "$1");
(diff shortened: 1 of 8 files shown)


decolua:
Thanks @Zireael for the contribution! Merged into master. 🙏

PR #1898, opened 2026-06-18
fix(usage): eliminate zero-token duplicate entries in streaming usage

Zireael:
## Problem

The Usage dashboard shows alternating zero-token and real-token rows for streaming requests. Every second entry has 0 input tokens, 0 output tokens, and 0ms TTFT.

image

## Root Cause

 and  each generated independent  values. Both called  with different IDs, creating two rows per streaming request:

1. **Placeholder row** (from ): saved immediately with  and 
2. **Final row** (from ): saved when the stream finishes with actual tokens and content

Since the IDs differed, the SQL  never merged them — both rows persisted.

## Fix

Share a single  from  through to  so both saves target the same row. The final entry (with real tokens) overwrites the placeholder via the existing  clause.

### Files Changed
-  — destructure and pass 
-  — accept  parameter, use it instead of generating a new one

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -296,8 +296,8 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   }
 
   // Streaming response
-  const { onStreamComplete } = buildOnStreamComplete({ ...sharedCtx });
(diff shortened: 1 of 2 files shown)


Zireael:
This bug affects token metrics calculations, estimated cost calculations etc. Please have a look to fix.

PR #1909, opened 2026-06-19
fix(codex): preserve tool_search tools

TerenceXin:
## Summary

- Preserve Responses `tool_search` tools in the Codex executor normalization allowlist.
- Add a regression test that exercises `CodexExecutor.transformRequest()` with `tool_search`, a deferred namespace tool, and a normal function tool.

## Why

Codex Desktop/App can expose deferred tools through `tool_search`. 9Router currently removes `tool_search` before forwarding the Responses request, so deferred app/thread/multi-agent tools cannot be discovered by the model.

## Test Plan

- PASS: `node --input-type=module` assertion against `CodexExecutor.transformRequest()` confirmed transformed tool types are `["tool_search", "namespace", "function"]`.
- NOT RUN: `npm exec --package vitest@4.1.9 -- vitest run tests/unit/codex-tool-search.test.js --config tests/vitest.config.js` could not load `tests/vitest.config.js` because this checkout does not declare a local `vitest` dependency, so `vitest/config` could not be resolved.

### DIFF
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -22,7 +22,8 @@ const SERVER_ID_PATTERN = /^(rs|fc|resp|msg)_/;
 // Hosted tool types that Codex/OpenAI Responses executes server-side
 const CODEX_HOSTED_TOOL_TYPES = new Set([
   "image_generation", "web_search", "web_search_preview", "file_search",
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @TerenceXin for the contribution! Merged into master. 🙏

PR #1910, opened 2026-06-19
fix(combo/fusion): flatten Anthropic style tool messages in panel calls

warelik:
## Summary

This is a follow-up to #1859. When using the `fusion` strategy with an Anthropic-compatible client (e.g., Claude Code, or other Anthropic-compatible integrations pointing to `/v1/messages`), the client sends previous tool usage in the Anthropic format, where tool invocations and results are nested inside the message `content` array as blocks (`type: "tool_use"` and `type: "tool_result"`). 

The `flattenToolHistory` helper only recognized OpenAI-style tool calls (`msg.role === "tool"` and `msg.tool_calls` array on assistant messages). As a result, Anthropic's tool blocks were left unmodified, and since the `tools` definitions are stripped from the panel calls (as implemented in #1859), the panel expert models would receive structured tool history without schemas, causing them to fail or misbehave (leading to empty responses and 503 errors).

This PR extends `flattenToolHistory` to correctly recognize and flatten Anthropic-style tool blocks within the `content` array into prose text, keeping panel expert execution robust and independent of the client's API format.

## Changes

1. **`open-sse/services/combo.js`** — Update `flattenToolHistory` to inspect `content` arrays for `tool_use` and `tool_result` blocks. If present, extract any interleaved text, join the tool names/results into plain text, and format them using the predefined `TOOL_CALL_PREFIX` and `TOOL_RESULT_PREFIX` strings.

2. **`tests/unit/combo-fusion.test.js`** — Add a new unit test `flattens Anthropic-style tool_use and tool_result blocks in arrays` verifying the correct extraction and formatting of both block types.

All 8 unit tests pass successfully.

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -31,6 +31,29 @@ function flattenToolHistory(messages) {
         const base = extractTextContent(rest.content) || (typeof rest.content === "string" ? rest.content : "");
         return { ...rest, content: `${base}${base ? "\n" : ""}${TOOL_CALL_PREFIX}${names}]` };
       }
+      if (Array.isArray(msg.content)) {
+        const hasToolUse = msg.content.some((c) => c.type === "tool_use");
+        const hasToolResult = msg.content.some((c) => c.type === "tool_result");
+        if (hasToolUse || hasToolResult) {
+          const textParts = [];
+          const toolNames = [];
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @warelik for the contribution! Merged into master. 🙏

PR #1912, opened 2026-06-19
fix: record non-stream usage for provider-native responses

nguyenha935:
## Summary
- Expand non-stream usage extraction to support provider-native response shapes such as Ollama, Kiro, CommandCode, nested OpenAI Responses usage, and nested Gemini usage metadata.
- Estimate missing prompt tokens from the original request when a provider reports output tokens only, preserving an estimated flag for visibility.
- Preserve cache/reasoning usage metadata when saving usage history so Recent Requests and cost accounting have the same token context.

## Test Plan
- ./node_modules/.bin/vitest run --config vitest.config.js --reporter=verbose unit/nonstream-usage-extraction.test.js unit/usage-concern.test.js
- npm run build


nguyenha935:
Superseded by #1994. The new PR keeps the Recent Requests UI filtering clean, records token-bearing CLI usage through the usageHistory path, avoids duplicate stream usage rows, and also fixes Responses API field cleanup for OpenCode Go GLM.

PR #1916, opened 2026-06-19
fix(test): use POST /v1/messages to validate anthropic-compatible connections

rehanchrl:
## Problem

The connection test for `anthropic-compatible` providers uses `GET /models` to validate API keys:

```js
const res = await fetchWithConnectionProxy(`${modelsBase}/models`, {
  headers: { "x-api-key": connection.apiKey, ... },
});
return { valid: res.ok, ... };
```

`GET /models` is not part of the Anthropic API spec. Many compatible proxies don't implement it, so it returns a 404 — causing the dashboard indicator to always show red even when the key and endpoint are valid.

## Fix

Switch to `POST /v1/messages` with `max_tokens: 1`, matching what the built-in `anthropic` provider already does for its own connection test. Treat any response that isn't 401 or 403 as valid — a 400 or 529 still confirms the key was accepted.

Also uses `connection.defaultModel` when set, so the test respects the configured model instead of always hardcoding `claude-3-haiku-20240307`.

## Relation to #1892

PR #1892 fixes the same issue for the built-in `anthropic` provider (adding `baseUrl` override support). This PR fixes the parallel `anthropic-compatible` branch which was missed.

## Test

Tested against a self-hosted Anthropic-compatible proxy — connection test now returns green with a valid key.

### DIFF
--- a/src/app/api/providers/[id]/test/testUtils.js
+++ b/src/app/api/providers/[id]/test/testUtils.js
@@ -354,11 +354,14 @@ async function testApiKeyConnection(connection, effectiveProxy = null) {
     if (!modelsBase) return { valid: false, error: "Missing base URL" };
     try {
       modelsBase = modelsBase.replace(/\/$/, "");
-      if (modelsBase.endsWith("/messages")) modelsBase = modelsBase.slice(0, -9);
-      const res = await fetchWithConnectionProxy(`${modelsBase}/models`, {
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @rehanchrl for the contribution! Switched both the validate and test-connection paths over to `POST /v1/messages` with `max_tokens: 1` and the configured `defaultModel`, so the indicator is now green for self-hosted Anthropic-compatible proxies. Merged into master. 🙏

PR #1922, opened 2026-06-19
fix(cli): auto-install sql.js when wasm binary missing

kacipbuah:
## Summary
- npm publish silently strips `.wasm` files from nested `node_modules/`, so `sql-wasm.wasm` is missing after `npm install -g 9router`
- On servers where `better-sqlite3` and `node:sqlite` also fail (e.g. Node 18 without build tools), all four SQLite drivers fail → app can't start
- `ensureSqliteRuntime()` now checks if `sql-wasm.wasm` exists, and if missing, installs `sql.js` into `~/.9router/runtime/node_modules/` (same approach as `better-sqlite3`)

## Test plan
- [ ] `npm pack` and verify `sql-wasm.wasm` still missing from tarball (confirms the bug exists)
- [ ] Install on Node 18 server without build tools → verify sql.js auto-installs to `~/.9router/runtime/`
- [ ] Verify `[DB] Driver: sql.js` or `better-sqlite3` in logs after startup
- [ ] Verify login page loads and DB operations work

### DIFF
--- a/cli/hooks/sqliteRuntime.js
+++ b/cli/hooks/sqliteRuntime.js
@@ -101,21 +101,32 @@ function npmInstall(pkgs, opts = {}) {
   return res.ok;
 }
 
-// Public: ensure better-sqlite3 native module is installed in user-writable
-// runtime dir. sql.js is bundled in bin/app already; node:sqlite is built-in.
-// This is purely a *speed optimization* — app works without it via fallbacks.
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @kacipbuah for the contribution! This fix has already been merged into master (with the version extracted to a `SQL_JS_VERSION` constant for DRY). Closing as resolved. 🙏

PR #1923, opened 2026-06-19
fix(models): store provider custom models by provider scope

nguyenha935:
## Summary
- Store provider-added custom models through the provider-scoped customModels store instead of the global modelAliases map.
- Keep legacy alias-backed models visible/deletable so existing user data is not broken.
- Include passthrough providers in ModelSelectModal customModels merging so newly added models remain selectable.

## Root cause
Provider detail Add Model and import flows were persisting provider-added models as global aliases keyed by model ID. When two providers expose the same model ID, for example minimax-m2.5 under Ollama and OpenCode Go, the second write overwrites the first global alias.

## Test Plan
- /opt/9router/tests/node_modules/.bin/vitest run --config tests/vitest.config.js --reporter=verbose tests/unit/provider-custom-models.test.js tests/unit/capabilities-service-kind.test.js tests/unit/model-routing.test.js
- npm run build

### DIFF
--- a/src/app/(dashboard)/dashboard/providers/[id]/CompatibleModelsSection.js
+++ b/src/app/(dashboard)/dashboard/providers/[id]/CompatibleModelsSection.js
@@ -3,6 +3,7 @@
 import { useState } from "react";
 import PropTypes from "prop-types";
 import { Button } from "@/shared/components";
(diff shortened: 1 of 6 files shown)


decolua:
Thanks @nguyenha935 for the contribution! Merged into master. 🙏

PR #1931, opened 2026-06-20
feat(opencode-go): align Go models with official endpoints

nguyenha935:
## Summary
- Update OpenCode Go model catalog to match the official Go docs.
- Route Qwen 3.7 and MiniMax models through `/v1/messages` with Anthropic-compatible headers instead of OpenAI-compatible chat.
- Keep GLM, Kimi, DeepSeek, and MiMo on `/chat/completions`; use `kimi-k2.7-code` because the live Go API rejects `kimi-k2.7` for chat while the docs config example uses the code model ID.

## Test Plan
- `npm --prefix tests test -- unit/opencode-go-models.test.js`
- `npm --prefix tests test -- translator/coverage-all-models.test.js translator/thinking-unified.test.js`
- `npm run build`
- Local smoke test through 9router: `ocg/qwen3.7-max`, `ocg/qwen3.7-plus`, `ocg/minimax-m3`, and `ocg/kimi-k2.7-code` all returned HTTP 200.

Docs: https://opencode.ai/docs/go#models


decolua:
Thanks @nguyenha935 for the contribution! Merged into master. 🙏

PR #1941, opened 2026-06-20
feat(codebuddy-cn): add API key auth + credit quota tracker

osyduck:
## CodeBuddy CN: API key auth + credit quota tracker

Follow-up to #1889 (the CodeBuddy CN OAuth provider, now merged). Two additions on top of the merged provider:

1. **Connect with a direct API key**, not only the OAuth browser flow.
2. **A usage-dashboard quota tracker** that surfaces CodeBuddy credit balances (refill + bonus) for both OAuth and API-key connections.

Both verified end-to-end against the live `copilot.tencent.com` backend with a real CN API key.

### What's included

**API key auth** — `open-sse/providers/registry/codebuddy-cn.js`, `src/app/api/providers/route.js`
- Registry: add `authModes: ["oauth", "apikey"]` + `hasOAuth: true` (still `category: "oauth"`). The dashboard already renders dual-auth providers with both an **OAuth** and an **API Key** button generically off `authModes`, so no UI change is needed.
- No executor change: the transport `auth` is already `combined` Bearer, so an API key is forwarded as `Authorization: Bearer <key>` exactly like an access token. Token refresh is a no-op for key connections (no refresh token), so they're skipped automatically.
- `POST /api/providers`: the create-connection validation only accepted providers in `APIKEY_PROVIDERS` (i.e. `category: "apikey"`), so dual-auth providers that live under `category: "oauth"` were rejected as **"Invalid provider"**. Widened to also accept any provider whose `authModes` includes `"apikey"`. This was a latent gap that also blocked `xai`'s API-key path; this fixes both.
- CodeBuddy also accepts an `X-API-Key` header, but `Authorization: Bearer` alone is sufficient, so no extra header is sent.

**Quota tracker** — `open-sse/providers/registry/codebuddy-cn.js`, `open-sse/services/usage/codebuddy-cn.js`, `open-sse/services/usage.js`
- Registry: add `transport.usage.url` (Tencent billing meter endpoint) + `features.usage` / `features.usageApikey` so the connection is quota-eligible for **both** OAuth and API-key auth.
- New CN-scoped handler `services/usage/codebuddy-cn.js`: `POST /v2/billing/meter/get-user-resource`, unwrapping the doubly-wrapped `data.Response.Data.Accounts[]`. The payload mixes two credit types that must **not** be merged:
  - **Refill / base** (`基础体验包`): a recurring allowance whose cycle resets long before the resource itself expires (`CycleEndTime` << `DeductionEndTime`). The live numbers live in the `*Cycle*` fields (e.g. `CycleCapacityUsed` 6.54 / 500); `resetAt` is the next refresh. Cadence-labelled (Monthly / Weekly / Daily).
  - **Bonus** (`活动赠送包`): one-shot credits that run a single cycle and expire for good at `CycleEndTime`. Reads the plain `Capacity` fields. Labelled `Bonus Pack N`.
  - One quota row per package, soonest-expiring first. Aggregating would mislabel every credit with the earliest package's expiry.
- Registered under `"codebuddy-cn"` in `USAGE_HANDLERS`. Named `codebuddy-cn` (not a generic `codebuddy`) to leave room for a future `codebuddy-ai` handler, consistent with the CN/AI split from #1889.
- No frontend change — `parseQuotaData`'s generic branch already renders the `{ quotas: { <name>: { used, total, resetAt } } }` shape.

### Testing

- **API key:** added a CodeBuddy CN connection via API key, confirmed inference works (Bearer auth) and the connection validates. OAuth connections continue to work unchanged.
- **Quota:** verified live against `copilot.tencent.com` with a real CN API key. The handler returns a `Monthly` refill row (e.g. `6.54 / 500`) plus `Bonus Pack 1/2` rows (e.g. `0 / 3000`, `0 / 1000`), matching CodeBuddy's own billing UI (`基础体验包` + `活动赠送包` breakdown). Refill-vs-bonus classification (`CycleEndTime` vs `DeductionEndTime`) confirmed against the live response.

### Notes for reviewers

- The providers **grid** (`src/app/(dashboard)/dashboard/providers/page.js`) has a pre-existing limitation for dual-auth providers: each card hardcodes `authType="oauth"` for its status badge and disable toggle, so an API-key connection isn't counted/toggled from the grid (the per-connection toggle on the provider **detail** page works correctly, and inference enforcement of `isActive` is correct). This affects `xai` too and is intentionally left out of this PR to keep its scope to API-key auth + quota; it can be addressed separately by passing `["oauth","apikey"]` for `authModes` providers (mirroring the existing Kiro handling).

### DIFF
--- a/open-sse/providers/registry/codebuddy-cn.js
+++ b/open-sse/providers/registry/codebuddy-cn.js
@@ -12,6 +12,8 @@ export default {
     },
   },
   category: "oauth",
+  authModes: ["oauth", "apikey"],
+  hasOAuth: true,
   transport: {
     baseUrl: "https://copilot.tencent.com/v2/chat/completions",
     forceStream: true,
--- a/open-sse/providers/registry/codebuddy-cn.js
+++ b/open-sse/providers/registry/codebuddy-cn.js
@@ -34,6 +34,11 @@ export default {
       header: "Authorization",
       scheme: "bearer",
     },
+    // Quota endpoint differs from the chat gateway: POST returns nested Tencent
+    // billing payload (data.Response.Data.Accounts[]). See services/usage/codebuddy.js.
+    usage: {
+      url: "https://copilot.tencent.com/v2/billing/meter/get-user-resource",
+    },
   },
   models: [
     { id: "glm-5.2", name: "GLM-5.2" },
@@ -60,5 +65,9 @@ export default {
     userAgent: "CLI/2.63.2 CodeBuddy/2.63.2",
     platform: "CLI",
     pollInterval: 5000,
+  },
+  features: {
+    usage: true,
+    usageApikey: true,
   },
 };
--- a/open-sse/services/usage.js
+++ b/open-sse/services/usage.js
@@ -10,6 +10,7 @@ import { getCodexUsage, consumeCodexRateLimitResetCredit } from "./usage/codex.j
 export { consumeCodexRateLimitResetCredit };
 import { getKiroUsage } from "./usage/kiro.js";
 import { getMiniMaxUsage } from "./usage/minimax.js";
+import { getCodeBuddyCnUsage } from "./usage/codebuddy-cn.js";
 import {
   getQwenUsage,
   getIflowUsage,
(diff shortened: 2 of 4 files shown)


osyduck:
image

decolua:
Thanks @osyduck for the contribution! Merged into master. 🙏

Issue #1907, opened 2026-06-19
Codex Responses proxy drops tool_search during request normalization


TerenceXin:
## Summary

When 9Router is used as a Codex Responses API proxy, `tool_search` is removed from the outgoing request before the request is forwarded upstream.

This breaks Codex deferred tool discovery. In Codex Desktop, tools such as app automation, thread-management tools, and `multi_agent_v1` tools are normally discovered through `tool_search`. Once the proxy drops `tool_search`, the model has no entrypoint to discover those deferred tools, so they appear to be missing in the session.

This is a request-normalization issue in 9Router, not a model capability issue.

## Affected code

`open-sse/executors/codex.js`

`normalizeCodexTools(body)` currently keeps:

- `function`
- `namespace`
- tool types listed in `CODEX_HOSTED_TOOL_TYPES`

But `tool_search` is neither a `function` nor in `CODEX_HOSTED_TOOL_TYPES`, so it is filtered out.

## Minimal reproduction

Run this from the 9Router source checkout:

```bash
node --input-type=module <<'EOF'
import { CodexExecutor } from './open-sse/executors/codex.js';

const executor = new CodexExecutor();
const body = {
  model: 'gpt-5.5',
  input: [
    { type: 'message', role: 'user', content: [{ type: 'input_text', text: 'probe' }] }
  ],
  tools: [
    { type: 'tool_search', execution: 'sync', description: 'Discover deferred tools', parameters: { type: 'object', properties: {} } },
    { type: 'namespace', name: 'codex_app', description: 'app tools', tools: [
      { type: 'function', name: 'automation_update', description: 'automation', parameters: { type: 'object', properties: {} }, defer_loading: true }
    ] },
    { type: 'function', name: 'plain_fn', description: 'plain', parameters: { type: 'object', properties: {} } }
  ],
  stream: true
};

const before = body.tools.map(t => t.type + ':' + (t.name || ''));
executor.transformRequest('gpt-5.5', body, true, { connectionId: 'probe', providerSpecificData: {} });
const after = body.tools.map(t => t.type + ':' + (t.name || ''));

console.log(JSON.stringify({ before, after, dropped: before.filter(x => !after.includes(x)) }, null, 2));
EOF
```

Actual output:

```json
{
  "before": [
    "tool_search:",
    "namespace:codex_app",
    "function:plain_fn"
  ],
  "after": [
    "namespace:codex_app",
    "function:plain_fn"
  ],
  "dropped": [
    "tool_search:"
[...]

Issue #1873, opened 2026-06-18
Codex subagent compatibility: multi_agent_v2 tool exposure and encrypted tool schema support / Codex 子代理兼容性问题


bradford54:
### Summary

When using 9router as a third-party relay for Codex CLI, Codex subagent / multi-agent workflows do not work reliably.

This is related to #1758, but this issue describes the full compatibility path more explicitly. Codex subagents are not only a model text capability. They depend on Responses API tool negotiation and support for Codex-specific tool schemas.

There are currently two possible failure modes:

1. **Subagent tools are not exposed to the model**
   - The current Codex session does not receive tools such as `spawn_agent`, `send_message`, `wait_agent`, `close_agent`, or `list_agents`.
   - In this state, the model cannot call subagents even if the underlying model behaves exactly like native GPT, because the tool entrypoints are not available.

2. **Subagent tools are exposed, but the backend rejects the request**
   - This is the failure mode already described in #1758.
   - When `multi_agent_v2` is enabled, Codex includes tools such as `spawn_agent` in the Responses API request.
   - The `spawn_agent.message` parameter schema includes `"encrypted": true`.
   - If the relay/backend does not support encrypted tool parameters correctly, the request fails with an error like:

```text
Invalid Value: 'tools'. Function 'functions.spawn_agent' declares encrypted parameters
but is not configured for encrypted tool use by this model.
```

### Why this is not only a model-name issue

Codex subagent availability depends on runtime tool availability, not only on model text capability.

The Codex-side behavior is roughly:

- If model metadata includes `multi_agent_version: "v2"`, Codex uses the multi-agent v2 tool surface.
- If model metadata is absent, Codex falls back to local feature flags such as `multi_agent_v2` / `multi_agent`.
- If the final resolved mode is disabled, Codex does not add subagent tools to the request.
- If the final resolved mode is v2, Codex adds tools such as `spawn_agent`, `send_message`, `followup_task`, `wait_agent`, `close_agent`, and `list_agents` to the Responses API request.

So even if 9router routes to a model that is identical to native GPT, subagents still cannot work unless 9router fully supports the Responses API tool schema, encrypted parameters, and relevant model capability metadata used by Codex.

### Expected Behavior

Please either support Codex subagent / `multi_agent_v2` explicitly, or document that it is unsupported.

If supported, 9router should:

1. correctly forward Responses API tools;
2. correctly support tool parameter schemas with `encrypted: true`;
[...]

Issue #1758, opened 2026-06-09
Codex subagent tools fail when multi_agent_v2 is enabled: spawn_agent.message uses encrypted parameters


JESVN:
<img width="855" height="272" alt="Image" src="https://github.com/user-attachments/assets/5114cf28-4264-4ba5-8d5a-78e45993099b" />
When using Codex CLI with multi_agent_v2 enabled through a custom OpenAI-compatible Responses API backend, Codex exposes subagent tools such as spawn_agent, wait_agent, send_message, followup_task, close_agent, and list_agents.

However, any request that includes these tools fails immediately with a 400 error before the subagent starts.

The error says that functions.spawn_agent declares encrypted parameters, but the selected model/backend is not configured for encrypted tool use.

Environment:
- Codex CLI: 0.138.0
- OS: Windows
- API shape: Responses API
- Model: gpt-5.5
- Feature flags:
  - multi_agent = true
  - multi_agent_v2 = true

Minimal config shape:
    model = "gpt-5.5"
    model_provider = "custom_responses_proxy"

    [model_providers.custom_responses_proxy]
    name = "Custom Responses Proxy"
    base_url = "https://<redacted>/v1"
    wire_api = "responses"
    requires_openai_auth = true

    [features]
    multi_agent = true
    multi_agent_v2 = true

Steps to reproduce:
1. Configure Codex CLI to use a custom OpenAI-compatible Responses API backend.
2. Enable multi_agent and multi_agent_v2.
3. Start a Codex session.
4. Ask Codex to spawn a subagent, for example:

    Spawn 1 explorer subagent and wait for it to finish.

Actual behavior:
The request fails immediately with a 400 error:

    Invalid Value: 'tools'. Function 'functions.spawn_agent' declares encrypted parameters
    but is not configured for encrypted tool use by this model.

This happens before the subagent can start.

Expected behavior:
Codex subagent workflows should work through a compatible Responses API backend.

If encrypted tool parameters are not supported by the backend/model, the proxy should either:
1. support/forward encrypted tool parameter schemas correctly, or
2. fail gracefully with a clear compatibility message, or
3. document that Codex multi_agent_v2 / subagent tools are currently unsupported.

Additional details:
I captured the outgoing Responses API request shape locally with a diagnostic proxy.

When multi_agent_v2 = true, Codex includes these tools in the request:
    spawn_agent
    send_message
    followup_task
    wait_agent
    close_agent
    list_agents

[...]

Issue #1371, opened 2026-05-23
Codex apply_patch broken via /v1/responses: custom_tool_call not preserved in translator


andrew05060414:
## Summary

When **Codex Desktop** sends `POST /v1/responses` through 9Router, the `apply_patch` tool call reaches Codex as **`function_call` with empty JSON arguments (`{}`)**, not **`custom_tool_call` with patch text in `input`**. Codex rejects/aborts the edit; agents fall back to shell/heredoc.

This is **not** a Codex patch-format problem. The proxy translation layer drops the freeform tool input shape Codex requires.

**Update (2026-05-23):** The originally suggested workaround — **oa-first combos with no `cc/*` first hop — does not fix the bug.** Failure reproduces on the OpenAI/oa path (`call_*` IDs), so the primary issue is the Responses translator, not only Claude combo routing.

## Environment

- Client: Codex Desktop (`wire_api = "responses"`, `model_provider = "9router"`)
- 9Router: local proxy on `:20128` (`decolua/9router`)
- Models tested: `gpt-5.4`, `gpt-5.5`
- Combos in use (oa-first):

```text
gpt-5.5      → oa/gpt-5.5 → ds/deepseek-v4-pro-max
gpt-5.4      → oa/gpt-5.4 → ds/deepseek-v4-pro
gpt-5.4-mini → oa/gpt-5.4-mini → cx/gpt-5.4-mini → ds/deepseek-v4-flash
```

Bare-name API probes return HTTP 200. Chat works. **`apply_patch` file edits do not.**

### Case A — Claude first hop (combo used to start with `cc/*`)

Rollout `rollout-2026-05-23T10-15-24-019e529d-...`:

```json
{"type":"function_call","name":"apply_patch","arguments":"{}","call_id":"toolu_01MboJRvmEnajA7FpkymQMuS"}
```

- `toolu_*` → Claude hop served the turn
- Patch text sometimes appeared in assistant message instead of tool input

### Case B — oa-first combo (still broken)

Rollout `rollout-2026-05-23T15-14-05-019e53af-1b92-7732-80c4-6acb5e0e45b6.jsonl`, model `gpt-5.4`, combo `gpt-5.4 → oa/gpt-5.4 → ds/deepseek-v4-pro`:

```json
{"type":"function_call","name":"apply_patch","arguments":"{}","call_id":"call_WCveAMYK64CeGkNWf02a74VC"}
```

- `call_*` → OpenAI/oa path (not Claude)
[...]

PR #1945, opened 2026-06-20
fix(codex): preserve custom tools during request normalization

Jordannst:
## Summary

This PR tightens Codex Responses request normalization so Responses-native tools are not accidentally dropped before the request is forwarded upstream.

The original #1907 report focused on `tool_search` being filtered out, which breaks deferred tool discovery in Codex Desktop. On current `master`, `tool_search` is already included in `CODEX_HOSTED_TOOL_TYPES`, so that part of the repro is now covered before this PR. This PR addresses the remaining related gap from the same normalization path: `custom` / freeform tools were still filtered out because they have a `name` field but are not `function`, `namespace`, or hosted tools.

## What changed

- Preserves Responses-native `custom` tools during `normalizeCodexTools(body)`.
- Keeps the full `custom` tool payload intact, including `name`, `description`, and `format` grammar fields.
- Adds regression coverage showing these tools survive Codex request normalization:
  - `tool_search` deferred-tool discovery
  - `namespace` tools
  - normal `function` tools
  - hosted Responses tools (`web_search`, `image_generation`, `mcp`, `local_shell`, `code_interpreter`, `computer`)
  - `custom` freeform tools such as grammar-backed `apply_patch`

## Why

Codex clients use Responses-native tools for capabilities that are not plain function tools. If request normalization removes those tool entries, Codex Desktop can lose access to deferred tool discovery or freeform tool calls, making tools appear missing even though the model/provider is capable.

This covers the `custom` / freeform-tool observation in #1907:

```js
{ type: 'custom', name: 'apply_patch', description: 'patch', format: { type: 'grammar', syntax: 'lark', definition: 'start: /.+/' } }
```

## Related issues

- Related to #1371, but this PR targets the request-time filtering path in `open-sse/executors/codex.js`; #1371 tracks response translation for `custom_tool_call` output.
- Related to broader Codex tool/subagent compatibility work in #1873 and #1758, but this PR is intentionally scoped to request normalization.

## Verification

```bash
./node_modules/.bin/vitest run tests/unit/codex-tool-normalization.test.js --reporter=verbose
```

Result:

```text
Test Files  1 passed (1)
Tests       3 passed (3)
EXIT:0
```

Refs #1907

### DIFF
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -26,6 +26,9 @@ const CODEX_HOSTED_TOOL_TYPES = new Set([
   "tool_search"
 ]);
 
+// Responses-native freeform tools carry a name plus format payload and must pass through intact.
+const CODEX_PASSTHROUGH_TOOL_TYPES = new Set(["custom"]);
+
 // Allowlist of fields accepted by Codex Responses API — anything else is stripped
 const RESPONSES_API_ALLOWLIST = new Set([
   "model", "input", "instructions", "tools", "tool_choice", "stream", "store",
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @Jordannst for the contribution! Merged into master. 🙏

Issue #1871, opened 2026-06-18
Feature Request: Integrate Headroom compression layer (60-95% token reduction)


madiajijah11:
## Problem
9Router already has RTK providing ~40% token reduction, but we can go further. Users running Claude Code, Codex, Cursor, or other AI agents through 9Router still pay for full context on every request — tool outputs, logs, RAG chunks, conversation history all consume tokens unnecessarily.

## Solution
Integrate [Headroom](https://github.com/chopratejas/headroom) (31.6k ⭐) as an **optional compression layer** within 9Router. Headroom compresses everything the agent reads before it reaches the LLM — achieving **60–95% token reduction** with preserved accuracy, on top of RTK.

## Why Headroom?
| Feature | Details |
|---------|---------|
| Token savings | 60–95% (RTK gives ~40%) |
| Accuracy | Benchmarks show ±0.000 delta on GSM8K, +0.030 on TruthfulQA |
| Modes | Library, Proxy, MCP server, Agent wrap |
| Reversible (CCR) | Originals cached locally, LLM can retrieve on demand |
| Cross-agent memory | Shared store across Claude, Codex, Gemini |
| `headroom learn` | Mines failed sessions, writes corrections to CLAUDE.md/AGENTS.md |
| Local-first | All data stays on your machine |

Headroom supports agents 9Router already works with: **Claude Code ✅, Codex ✅, Cursor ✅, Aider ✅, Copilot CLI ✅** — and any OpenAI-compatible client via proxy.

## Integration Ideas
1. **Optional pipeline step** — Add Headroom as an optional compression stage in 9Router's proxy chain (similar to RTK but far more aggressive). Users opt in via config or env var like `9ROUTER_COMPRESSION=headroom`.

2. **Headroom proxy as upstream** — Document how to chain `headroom proxy --port 8787` → `9router proxy` so users get compression + routing + auto-fallback in one flow.

3. **Bundled mode** — Offer `headroom` as an installable extra dependency (`npm install 9router headroom-ai`) that 9Router auto-detects and activates.

## Expected Impact
- **Massive cost reduction** for users on pay-per-token providers routed through 9Router
- **Competitive advantage** — no other AI proxy offers 60-95% context compression
- **Happy path**: Claude Code user gets compression + free provider routing + auto-fallback in one CLI

## Resources
- Headroom GitHub: https://github.com/chopratejas/headroom
- Headroom Docs: https://headroom-docs.vercel.app/docs
[...]

PR #1948, opened 2026-06-21
fix(headroom): support Docker sidecar proxy

CarmeloCampos:
## Summary
- Treat a configured Headroom proxy as running when its /health endpoint responds, even if the local headroom CLI is not installed
- Keep dashboard Start/Stop limited to local loopback proxies while allowing external Docker sidecars to be enabled
- Add HEADROOM_URL as the default Headroom URL for Docker deployments and document the sidecar setup
- Add unit coverage for external proxy status detection

## Why
The managed Headroom UI currently assumes the headroom CLI and Python are installed inside the same runtime as 9Router. That works for local installs, but not for the published Docker image: Python is only installed in the Docker builder stage and the final runner image intentionally does not bundle Python or headroom-ai.

For Docker users, the natural deployment model is a separate Headroom service/container reachable as http://headroom:8787 or a host proxy such as http://host.docker.internal:8787. Before this change, the dashboard could show Not installed / Python >= 3.10 required and keep the toggle disabled even when the configured external proxy was healthy.

This keeps Headroom optional and avoids making the 9Router image heavier or more fragile by bundling Python + Headroom into the Node runtime image.

## Testing
- NODE_PATH=node_modules:tests/node_modules tests/node_modules/.bin/vitest run --config tests/vitest.config.js tests/unit/headroom.test.js tests/unit/headroom-detect.test.js --reporter=verbose
- rm -rf .next && rtk npm run build

Fixes the Docker mode issue discussed in #1871.

### DIFF
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -64,6 +64,34 @@ docker run -d \
   decolua/9router:latest
 ```
 
+## Optional Headroom sidecar
+
+The 9Router image does not bundle Python or Headroom. To use Headroom in Docker, run it as a separate service and point 9Router at that proxy:
+
+```yaml
+services:
+  9router:
+    image: decolua/9router:latest
+    ports:
+      - "20128:20128"
+    volumes:
+      - "$HOME/.9router:/app/data"
+    environment:
+      DATA_DIR: /app/data
+      HEADROOM_URL: http://headroom:8787
+    depends_on:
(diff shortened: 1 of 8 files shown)


decolua:
Thanks @CarmeloCampos for the contribution! Merged into master. 🙏

PR #1949, opened 2026-06-21
fix(antigravity): strip Claude/OpenAI thinking fields from top-level request body

Arcfoz:
## Problem

`thinking` field from claude-budget/adaptive format leaked to Google API,
causing 400 on thinking models (e.g. `claude-opus-4-6-thinking`):

400 INVALID_ARGUMENT: Unknown name "thinking": Cannot find field.

## Change

Expanded `ANTIGRAVITY_REQUEST_BLACKLIST` to include all fields
`thinkingUnified.js` can set at body root:

`output_config`, `thinking`, `reasoning_effort`, `reasoning`, `enable_thinking`, `thinking_budget`

## Testing

- ✅ `claude-opus-4-6-thinking` via Antigravity — previously 400, now works

### DIFF
--- a/open-sse/executors/antigravity.js
+++ b/open-sse/executors/antigravity.js
@@ -19,7 +19,8 @@ const MAX_RETRY_AFTER_MS = 10000;
 const MAX_ANTIGRAVITY_OUTPUT_TOKENS = 16384;
 
 // Fields Google generateContent rejects (e.g. Claude adaptive output_config) — stripped from antigravity request envelope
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @Arcfoz for the contribution! 🙏 Merged into master.

Both the expanded `ANTIGRAVITY_REQUEST_BLACKLIST` (now also covers `thinkingConfig`) and the top-level body strip are in. We folded the duplicated strip loops into a single `stripBlacklisted()` helper for DRYness. `claude-opus-4-6-thinking` via Antigravity works now.

PR #1953, opened 2026-06-21
fix: enable vision capability for MiniMax-M3/M2.7 family

swatker:
## Bug

The pattern matcher in `open-sse/providers/capabilities.js` marks all `minimax-*` models as `vision: false`. This causes 9Router to strip image attachments from requests before forwarding to upstream, breaking vision flows for Claude Code / Cursor / Cline / etc. when routing through minimax models.

## Repro

```bash
curl -s http://localhost:20128/v1/chat/completions   -H "Content-Type: application/json"   -d '{"model":"MiniMax-M3","messages":[{"role":"user","content":[{"type":"text","text":"Do you see an image?"},{"type":"image_url","image_url":{"url":"data:image/png;base64,..."}}]}]}'
# Before fix: model responds "NO_IMAGE"
# After fix: model responds "Yes, I see an image"
```

## Root cause

`stripUnsupportedModalities()` in `open-sse/translator/concerns/modality.js` removes image blocks when `caps.vision === false`. The `PATTERN_CAPABILITIES` for `*minimax-m3*`, `*minimax-m2.7*`, and `*minimax*` did not set `vision: true`, so DEFAULT (`vision: false`) applied, causing all MiniMax model requests to have images stripped.

## Fix

Add `vision: true` to the 3 affected minimax patterns (M3, M2.7, catch-all). These models support image input.

## Tested

Patched `app/.next-cli-build/server/chunks/6306.js`, `412.js`, and `static/chunks/5497-*.js` with this one-line change per pattern. Verified via curl: MiniMax-M3 now correctly receives image content instead of stripped placeholder.

### DIFF
--- a/open-sse/providers/capabilities.js
+++ b/open-sse/providers/capabilities.js
@@ -176,11 +176,11 @@ export const PATTERN_CAPABILITIES = [
   { pattern: "*deepseek-chat*", caps: { contextWindow: 128000 } },
   { pattern: "*deepseek*",      caps: { reasoning: true, thinkingFormat: "deepseek", contextWindow: 128000 } },
 
-  // ── MiniMax (M3 = adaptive; M2.x cannot disable) ─────────────────
+  // ── MiniMax (M3 = adaptive + vision; M2.x cannot disable) ─────────
   { pattern: "*minimax*image*", caps: { imageOutput: true } },
-  { pattern: "*minimax-m3*",    caps: { reasoning: true, thinkingFormat: "minimax", contextWindow: 1048576, maxOutput: 512000 } },
-  { pattern: "*minimax-m2.7*",  caps: { reasoning: true, thinkingFormat: "minimax", thinkingCanDisable: false, contextWindow: 204800, maxOutput: 131072 } },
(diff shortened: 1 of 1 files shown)


osyduck:
m3 part looks right, but i don't think we want this on the m2.7 + catch-all patterns. m2.7 is text-only afaik. minimax's own docs and the NIM model card both list it as text input only, no image/video. so flipping vision to true there means we stop stripping and send images to a model that can't read them, which just errors out upstream instead of degrading nicely like it does now.

can we keep `vision: true` scoped to `*minimax-m3*` and leave m2.7 + `*minimax*` at false? that still fixes the m3 stripping bug, which is the actual repro here.

quick check if you have m2.7 access: run the same curl against MiniMax-M2.7 with vision forced on and see if it 400s or just ignores the image. pretty sure it won't come back with "IMG_RECEIVED" like m3 does.

decolua:
Thanks @swatker for the contribution! 🙏

Merged into master, scoped to `*minimax-m3*` only. Per @osyduck's review and MiniMax docs / NVIDIA NIM model card, M2.7 and the catch-all `*minimax*` pattern are text-only — forcing vision there would send images to a model that errors upstream instead of degrading gracefully, so those stay `vision: false`. The M3 stripping bug (the actual repro) is fixed.

PR #1973, opened 2026-06-21
fix(headroom): log skip diagnostics when compression unavailable

### LINKED ISSUE #1956: Headroom "Compress context" toggle shows Running but never invoked for /v1/chat/completions requests
Setup: 9Router v0.5.6 (Docker, ghcr.io/decolua/9router:latest), Headroom installed via pip install "headroom-ai[proxy]", proxy running standalone on port 8787, dashboard shows "Compress context (Headroom)" status as Running, toggle ON.
Steps to reproduce: enable Headroom toggle in Token Saver settings, send a normal request via POST /v1/chat/completions (model alias, e.g. via Codex/Claude Code through 9Router endpoint).
Expected: per README's documented flow (Client → 9Router → Headroom /v1/compress → 9Router → provider), the request should be compressed via Headroom's /v1/compress before forwarding upstream.
Actual: server logs show the normal pipeline stages ([AUTH], [FORMAT], [MODALITY], [CAVEMAN], [PONYTAIL], [PENDING] START) but no log entry referencing Headroom or /v1/compress at any point. Headroom proxy logs (port 8787) show zero requests received (Total requests: 0 in session summary) despite multiple chat completions being sent while the toggle was ON.
Additional note: also tried calling /v1/compress directly from an external client against the public Space URL — this fails because port 8787 isn't exposed externally, suggesting the integration is expected to happen server-side inside 9Router, not client-side.
Question: is this integration still in progress (related to #1871), or is there a required config step (env var, explicit route mapping) not yet documented?

Jordannst:
## Summary

This adds lightweight diagnostics for cases where Headroom compression is skipped while preserving the existing fail-open behavior.

Currently, if the Headroom proxy is unavailable or returns an unusable response, compression can silently fall back to the original request body. This change records the skip reason and logs it from the chat completion path when Headroom is enabled but no compression stats are returned.

## Changes

- Track Headroom skip reasons such as:
  - disabled or missing configuration
  - missing request body
  - unsupported request shape
  - Headroom proxy request failure
  - non-OK proxy response
  - proxy response missing `messages[]`
  - unexpected compression errors
- Include the `/v1/compress` endpoint in diagnostics when a proxy request is attempted.
- Log a warning from `handleChatCore` when Headroom is enabled but compression is unavailable.
- Keep compression fail-open: request handling continues with the original body when Headroom cannot compress.
- Add unit coverage for HTTP errors, request failures, unsupported shapes, and chat handler skip logging.

## Verification

- `npx vitest run tests/unit/headroom-chat-core.test.js --pool=threads --reporter=dot`
- `npx vitest run tests/unit/headroom-chat-core.test.js tests/unit/headroom.test.js tests/unit/headroom-detect.test.js tests/unit/github-responses-routing.test.js --pool=threads`
- `git diff --check`

Closes #1956

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -162,9 +162,11 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   if (rtkLine) console.log(rtkLine);
 
   // Headroom: optional external proxy compression; fail open if proxy is absent.
-  const headroomStats = await compressWithHeadroom(translatedBody, { enabled: headroomEnabled, url: headroomUrl, model: upstreamModel, format: finalFormat, compressUserMessages: headroomCompressUserMessages });
+  const headroomDiagnostics = {};
+  const headroomStats = await compressWithHeadroom(translatedBody, { enabled: headroomEnabled, url: headroomUrl, model: upstreamModel, format: finalFormat, compressUserMessages: headroomCompressUserMessages, diagnostics: headroomDiagnostics });
   const headroomLine = formatHeadroomLog(headroomStats);
   if (headroomLine) log?.info?.("HEADROOM", headroomLine);
+  else if (headroomEnabled) log?.warn?.("HEADROOM", `skipped: ${headroomDiagnostics.reason || "compression unavailable"}${headroomDiagnostics.endpoint ? ` (${headroomDiagnostics.endpoint})` : ""}`);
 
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @Jordannst! This is superseded by #2005 (same author, same headroom diagnostics — #2005 is a superset adding credential masking, payload size snapshots, and phantom-savings warnings), which has been merged. Closing this one. 🙏

Issue #1998, opened 2026-06-22
fix(headroom): provider receives full tokens despite Headroom reporting savings


tmaarcxs:
## Bug

Headroom integration reports token savings (e.g. `[HEADROOM] saved 24380 tokens / 62867 (38.8%) after=38487`) but the provider still receives and bills for the full uncompressed token count.

## Evidence

| Source | Token count |
|---|---|
| Headroom stats (9router log) | 62867 -> 38487 (saved 24380, 38.8%) |
| Provider (OpenCode Go) actual billed | **85,000+** |

The provider usage page shows the full input token count, matching the pre-compression request size. The ~38.8% savings reported by Headroom never materialize in the actual tokens processed by the provider.

## Root Cause Analysis

In `open-sse/rtk/headroom.js`, `compressWithHeadroom` calls Headroom `/v1/compress` and mutates the body in-place:

```javascript
const data = await callCompress(url, body[key], model, timeoutMs, compressUserMessages);
if (!data) return null;
body[key] = data.messages;  // mutates body
return data;
```

The compressed messages replace the original, but the provider still reports the full token count. Two possible causes:

1. **Tokenizer mismatch**: Headroom counts tokens with its own tokenizer (likely tiktoken). The provider uses a different tokenizer. Headroom reports "saved" tokens that the provider still counts because the compressed text tokenizes to the same count under the provider's tokenizer.

2. **Phantom stats**: Headroom returns `tokens_before/saved/after` stats that don't reflect the actual compressed output. Direct testing of `/v1/compress` with various payloads returns `compressed: false` with empty savings, while 9router logs show non-zero savings in production.

## Setup

- 9router v0.5.8 (built from source, master branch)
- Headroom v0.26.0 (standalone proxy on port 8787)
- Provider: OpenCode Go (custom provider)
- `headroomEnabled: true`
- `headroomCompressUserMessages: false`

## How to Verify

1. Enable Headroom in 9router settings
2. Send a large chat completion request (50K+ tokens)
3. Check 9router logs for `[HEADROOM] saved X tokens`
4. Check provider usage page - input tokens will match the pre-compression count, not the post-compression count

## Related

- #1956 (Docker users: Headroom never invoked at all)
- #1974 (same as #1956)
- #1973 (diagnostic logging PR, does not fix the token mismatch)

This is a **different issue** from #1956/#1974. In this case, Headroom IS invoked and returns stats, but the savings are phantom - the provider never sees fewer tokens.

[...]

PR #2005, opened 2026-06-22
fix(headroom): clarify token diagnostics

Jordannst:
## Summary

This PR improves Headroom diagnostics so 9Router no longer implies that Headroom-reported token deltas are the same thing as provider-billed token savings.

It adds payload-size visibility around the actual request body that 9Router sends downstream after Headroom compression, which helps investigate the phantom-savings behavior reported in #1998.

## What changed

- Change Headroom logging from savings wording to neutral reported-token wording:
  - before: `saved 75 tokens / 100 (75.0%) after=25`
  - after: `reported token delta=75 before=100 after=25 (75.0%)`
- Add serialized outbound payload diagnostics after compression:
  - `body=before→after`
  - `messages=before→after`
- Warn when Headroom reports a token delta but the outbound provider payload barely shrinks:
  - this is the exact class of issue described in #1998
- Preserve credentials/query parameters for the actual Headroom request URL.
- Mask credentials/query/hash fragments in diagnostics and skip logs.
- Scrub credentials/query strings from Headroom fetch error messages.
- Add chat-core coverage to verify compressed messages are passed to the provider executor.
- Add privacy coverage so diagnostics do not leak raw prompt content or proxy credentials.

## Why this matters

Headroom reports token counts using its own compression/tokenization path. Providers may bill based on the final serialized request body, their own tokenizer, cached context rules, tools/schema overhead, or provider-side accounting.

Because of that, a log line saying `saved X tokens` can be misleading: it reads like provider billing was reduced, even when the provider still reports near-original usage.

This PR makes the logs distinguish between:

1. **Headroom-reported token delta** — what Headroom says changed.
2. **Outbound payload size delta** — what 9Router actually sends downstream after compression.

## How to interpret the new logs

Example:

```txt
[HEADROOM] reported token delta=24380 before=62867 after=38487 (38.8%) | body=120000B→79000B messages=118000B→76000B
```

This means Headroom reported a token delta and the serialized outbound request also shrank.

If provider usage still shows the original size, the likely cause is outside this mutation path, such as:

- provider tokenizer/accounting differences
- provider-side cache/context accounting
- payload not represented by `messages` alone
- provider usage UI reporting a different accounting basis

If Headroom reports a token delta but the serialized outbound body barely shrinks, 9Router now emits a warning:

```txt
[HEADROOM] reported token delta, but outbound JSON shrank <5%; provider may bill near-original payload | body=...
```

That points to phantom savings or compression stats that do not materially reduce the provider-bound payload.

## Security / privacy

Diagnostics intentionally log only byte counts, not message contents.

Proxy credentials and query strings are preserved for the actual request but masked in logs. Fetch error messages are scrubbed as well.

## Tests

- `npx vitest run tests/unit/headroom-chat-core.test.js tests/unit/headroom.test.js --reporter=verbose`
- `git diff --check`

Covered cases include:

- Headroom skip diagnostics
- neutral token-delta formatting
- compressed messages reaching the provider executor
- phantom-savings warning when outbound body barely shrinks
- proxy endpoint masking in diagnostics
- fetch error URL scrubbing
- no raw request content leakage in Headroom logs

Refs #1998

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -22,7 +22,7 @@ import { dedupeTools } from "../utils/toolDeduper.js";
 import { injectCaveman } from "../rtk/caveman.js";
 import { injectPonytail } from "../rtk/ponytail.js";
 import { compressMessages, formatRtkLog } from "../rtk/index.js";
-import { compressWithHeadroom, formatHeadroomLog } from "../rtk/headroom.js";
+import { compressWithHeadroom, formatHeadroomLog, formatHeadroomSizeLog, isHeadroomPhantomSavings } from "../rtk/headroom.js";
 import { getCapabilitiesForModel } from "../providers/capabilities.js";
 import { stripUnsupportedModalities } from "../translator/concerns/modality.js";
 import { prefetchRemoteImages } from "../translator/concerns/prefetch.js";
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @Jordannst! Merged into master. 🙏

PR #2012, opened 2026-06-23
fix: handle Cloudflare AI param support rule

### LINKED ISSUE #1960: Cloudflare AI provider: HTTP 502 "Cannot read properties of undefined (reading 'test')" on v0.5.8
After upgrading from 0.5.4 to 0.5.8, all (or most) Cloudflare AI models fail with a 502 error on connection test. Confirmed working correctly on 0.5.4, broken on 0.5.8.

Image
Image
Image
Image

Steps to reproduce:
Install 9router@0.5.8 (npm i -g 9router@0.5.8)
Add Cloudflare AI provider with valid API key/account
Go to Providers → Available Models
Test any cf/@cf/meta/llama-* or cf/@cf/mistralai/* model

Error:
HTTP 502: [cloudflare-ai/@cf/meta/llama-3.1-70b-instruct-fp8-fast] [502]: Cannot read properties of undefined (reading 'test') (reset after 27s)

Expected: Model test returns success, as it does on 0.5.4.
Observation: Not all models are affected equally — Llama 3.3 70B Instruct FP8 (different icon/handling) appears unaffected, while Llama 3.1/3.2 family and Mistral Small 3.1 all fail with the same error. This suggests the regression is tied to how model IDs are validated/matched, not the Cloudflare provider connection itself.
Versions tested:
✅ 0.5.4 — all models pass
❌ 0.5.8 — fails with above error
Environment: Ubuntu VPS (Alibaba Cloud), Node.js v22.22.3

gitcommit90:
## Summary
Fixes a crash when testing/using Cloudflare Workers AI models through the OpenAI-compatible path.

The Cloudflare-specific strip rule sets `flattenContent: true` but does not need a `match` predicate or `drop` list. The shared param-support loop assumed both existed, so the Cloudflare rule crashed with:

```text
Cannot read properties of undefined (reading 'test')
```

## Changes
- Treat strip rules without `match` as provider-wide rules
- Treat missing `drop` as an empty list
- Add a focused unit test for Cloudflare content-part flattening
- Keep an existing drop-rule regression test for GitHub params

## Verification
- Watched the new Cloudflare unit test fail first with the reported `reading 'test'` error
- `NODE_PATH=/tmp/node_modules /tmp/node_modules/.bin/vitest run --reporter=verbose tests/unit/param-support.test.js` — 2 passed
- `npm run build` — compiled successfully and generated 123 static pages

Fixes #1960

### DIFF
--- a/open-sse/translator/concerns/paramSupport.js
+++ b/open-sse/translator/concerns/paramSupport.js
@@ -16,6 +16,7 @@ const STRIP_RULES = [
 
 // Test a rule's match (regex or predicate) against the model id.
 function matches(rule, model) {
+  if (!rule.match) return true;
   return typeof rule.match === "function" ? rule.match(model) : rule.match.test(model);
 }
 
@@ -25,7 +26,7 @@ export function stripUnsupportedParams(provider, model, body) {
   for (const rule of STRIP_RULES) {
     if (rule.provider && rule.provider !== provider) continue;
     if (!matches(rule, model)) continue;
-    for (const key of rule.drop) {
+    for (const key of rule.drop || []) {
       if (body[key] !== undefined) delete body[key];
     }
     // CF Workers AI oneOf root schema only accepts content as plain string (#1926)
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @gitcommit90 for the fix! Confirmed the crash (#1960) was caused by the `cloudflare-ai` strip rule lacking `match`/`drop`. The provider-wide guard + `drop || []` fallback is the right approach, and the unit tests cover both the flatten path and the GitHub drop regression. Merged into master. 🙏

PR #2019, opened 2026-06-23
fix(translator): resolve custom provider prefix in debug endpoint

### LINKED ISSUE #1083: Custom Provider Prefix should be optional; when set, it incorrectly prefixes model IDs and causes “no such model” errors
When adding a Custom Provider, the `Prefix` field is marked as:

> Required. Used as the provider prefix for model IDs.

However, I don’t think this field should be required.

If a prefix is set, API requests to the LLM provider include the prefix in the model name. For example:

- Configured model name: `claude-opus-4.7`
- Actual request model name: `prefix/claude-opus-4.7`

This causes the provider API to return an error:

> no such model

For custom providers, some APIs expect the model name to be sent exactly as configured, without any provider prefix. Therefore, the prefix should either be optional or there should be a way to disable prefixing model IDs in API requests.

### LINKED ISSUE #1769: Custom providers show connection name instead of provider node name

hamsa0x7:
## Summary

Fix custom provider prefix resolution in the translator debug endpoint. Also resolves display name priority for custom providers.

### Fix #1083 — Custom Provider Prefix causes "no such model" errors

The translator debug API (`/api/translator/translate`) used `parseModel()` directly from `open-sse/services/model.js`, which doesn't resolve custom provider prefixes. The runtime chat path uses `getModelInfo()` from `@/sse/services/model.js` which does resolve prefixes correctly.

**Fix:** Swapped `parseModel` for `getModelInfo` in `src/app/api/translator/translate/route.js` (lines 21 and 31). Now the debug path aligns with the live chat path.

### Fix #1769 — Custom providers show connection name instead of provider node name

Verified the codebase already has the correct display name priority:
- `ModelSelectModal.js` line 239: `matchedNode?.name || connection?.name || providerInfo.name`
- `ProviderTopology.js` line 152: `(config.name !== p.provider ? config.name : null) || p.nodeName || p.name || p.provider`

Both already prefer node name over connection name. This issue was filed against an older version and is already resolved.

## Test Plan

- [x] Unit tests pass (3 tests covering OpenAI-compatible, Anthropic-compatible, and no-prefix cases)

## Notes

- #1769 is already fixed in current master — the display name priority was updated in a previous commit
- This PR only contains the #1083 translator fix plus regression tests

### DIFF
--- a/src/app/api/translator/translate/route.js
+++ b/src/app/api/translator/translate/route.js
@@ -2,7 +2,7 @@ import { NextResponse } from "next/server";
 import { detectFormat, getTargetFormat } from "open-sse/services/provider.js";
 import { translateRequest } from "open-sse/translator/index.js";
 import { FORMATS } from "open-sse/translator/formats.js";
-import { parseModel } from "open-sse/services/model.js";
+import { getModelInfo } from "@/sse/services/model.js";
 import { getProviderConnections } from "@/lib/localDb.js";
 import { getExecutor } from "open-sse/executors/index.js";
 
@@ -18,7 +18,7 @@ export async function POST(request) {
       case 1: {
         // Detect provider + formats from 1_req_client.json
         const clientBody = body.body || body;
-        const { provider, model } = parseModel(clientBody.model);
+        const { provider, model } = await getModelInfo(clientBody.model);
         const sourceFormat = detectFormat(clientBody);
         const targetFormat = getTargetFormat(provider);
         return NextResponse.json({ success: true, result: { provider, model, sourceFormat, targetFormat } });
@@ -28,7 +28,7 @@ export async function POST(request) {
         // source → OpenAI intermediate (mirrors 3_req_openai.json)
         // Translate source→openai only (half of the pipeline)
         const clientBody = body.body || body;
-        const { provider, model } = parseModel(clientBody.model);
+        const { provider, model } = await getModelInfo(clientBody.model);
         const sourceFormat = detectFormat(clientBody);
         const stream = clientBody.stream !== false;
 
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @hamsa0x7! Clean fix — swapping `parseModel` for `getModelInfo` correctly aligns the debug endpoint with the runtime chat path so custom provider prefixes resolve properly. Merged into master. 🙏

PR #2042, opened 2026-06-24
Fix provider thinking, usage tracking, and smart routing

nguyenxvotanminh3:
## Summary

This draft PR groups several local fixes that improve provider compatibility, routing behavior, and dashboard/runtime responsiveness:

- handle Claude/DeepSeek thinking/tool-use edge cases more defensively
- normalize Gemini/Antigravity thinking levels for provider compatibility
- improve provider connection probing for Cloud Code Assist based providers
- batch console log SSE updates and reduce debug log pressure
- make streaming/non-streaming success bookkeeping non-blocking
- add combo smart routing by task weight while preserving hard capability requirements
- deduplicate usage writes so stats are not double-counted when multiple streaming paths report the same usage

## Local verification

- `node --check` on edited source/runtime files passed locally
- direct Node assertions for combo smart routing passed locally
- local 9Router runtime was patched and smoke-tested via `/api/health` and a real `/v1/chat/completions` request
- local usage database was backed up, deduplicated, and dashboard stats dropped to expected values

## Known caveats

- `npm run build` currently hangs locally after `next build --webpack`, so this is opened as a draft for review rather than ready-to-merge
- Vitest runner also hung in this local environment; combo routing logic was verified with direct Node assertions instead
- This PR may be easier to review if split into smaller PRs: provider thinking fixes, usage/logging fixes, and smart routing

PR #2043, opened 2026-06-24
Fix provider thinking compatibility

nguyenxvotanminh3:
## Summary

- handle Claude-compatible thinking/tool-use blocks more defensively
- normalize Gemini/Antigravity thinking levels so unsupported `max`-style levels do not get forwarded
- improve Cloud Code Assist style provider connection probing
- add focused translator regression coverage

## Verification

- `node --check` passed locally for edited source/test files
- focused translator tests were checked locally, but the full Vitest runner hung in my local environment

This is split out from the larger draft PR #2042 to make review easier.


decolua:
Thanks @nguyenxvotanminh3 for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2044, opened 2026-06-24
Fix usage logging dedupe and reduce stats churn

nguyenxvotanminh3:
## Summary

- batch console log buffer events and support batched SSE log messages
- debounce usage stats update/pending events to reduce UI/runtime churn
- avoid awaiting request-success bookkeeping before returning provider responses
- deduplicate identical usage writes so streaming usage is not double-counted in `usageHistory`/daily aggregates
- reduce default logger verbosity from DEBUG to INFO

## Why

In local testing, streaming paths could record the same usage twice, inflating 7d dashboard totals. Frequent log/stat events also increased UI/runtime pressure while requests were streaming.

## Verification

- `node --check` passed locally for edited files
- local runtime was patched and restarted successfully
- local usage DB was backed up and rebuilt from deduped history; dashboard totals dropped to expected values

This is split out from the larger draft PR #2042 to make review easier.

### DIFF
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -167,7 +167,13 @@ export async function handleNonStreamingResponse({ providerResponse, provider, m
   }
 
   reqLogger.logProviderResponse(providerResponse.status, providerResponse.statusText, providerResponse.headers, responseBody);
-  if (onRequestSuccess) await onRequestSuccess();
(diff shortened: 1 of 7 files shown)


cuongle2503:
are you sure???

nguyenxvotanminh3:
> are you sure???

sure af bro

decolua:
Thanks @nguyenxvotanminh3 for the contribution! Reviewed and accepted — applying it to master. 🙏

Issue #1996, opened 2026-06-22
Bug: Combo fallback not triggering on Qoder stream errors


zjw345:
# Bug Report: Combo Fallback Not Triggering on Qoder Stream Errors

## Description

When a model in a combo returns an error embedded in the SSE stream content (instead of HTTP status code), the combo fallback mechanism does not trigger, causing the request to fail without trying alternative models.

## Affected Component

- **Service**: 9router (AI Router)
- **File**: `open-sse/services/combo.js`
- **Function**: `handleComboChat`

## Steps to Reproduce

1. Configure a combo with multiple models (e.g., Qoder + another provider)
2. Send a chat completion request that triggers a Qoder error (e.g., quota exceeded)
3. Qoder returns HTTP 200 but embeds error in SSE stream: `[qoder error 403: {"code":"112","message":"..."}]`
4. Observe that the combo does not fallback to the next model

## Expected Behavior

When a model returns an error in the SSE stream content, the combo should:
1. Detect the error
2. Log a warning
3. Fallback to the next model in the combo

## Actual Behavior

The combo treats the response as successful because:
1. HTTP status code is 200 (SSE connection established successfully)
2. `result.ok` check passes
3. The error in stream content is not inspected
4. The response is returned immediately without fallback

## Root Cause

Qoder's error handling mechanism returns errors through SSE stream content rather than HTTP status codes. The `handleComboChat` function only checks `result.ok` (HTTP status) and does not inspect the stream content for embedded errors.

### Error Format

Qoder errors appear as:
```
[qoder error 403: {"code":"112","message":"{"pricingUrl":"https://qoder.com/pricing?client=qoder"}"}]
```

Other providers (via OpenAI Responses API) use:
```
[Error] Error message here
```

## Proposed Solution

Add stream error detection logic in `handleComboChat`:

1. After receiving a successful HTTP response (`result.ok`), check if the response is an SSE stream
2. Read the first chunk of the stream
3. Use regex patterns to detect error patterns:
   - `/\[qoder error \d+: (.*?)\]/i` for Qoder errors
   - `/\[Error\] (.*)/i` for general errors
4. If an error is detected, log a warning and continue to the next model

### Code Change

```javascript
// In handleComboChat function, after result.ok check:
if (result.ok) {
  log.info("COMBO", `Model ${modelStr} succeeded`);
  
  const contentType = result.headers.get("content-type") || "";
  if (contentType.includes("text/event-stream")) {
    const clonedResult = result.clone();
[...]

Issue #1930, opened 2026-06-20
对部分供应商的api似乎存在异常


lingchenheiye:
日志如下：
---
[13:58:16] 📥 POST /api/v1/chat/completions | openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9/astron-code-latest | 1 msgs
[13:58:16] 🔍 [AUTH] API Key: sk-c...6cc0
[13:58:16] ℹ️ [ROUTING] Provider: openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9, Model: astron-code-latest
[13:58:16] 🔍 [AUTH] openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9 | total connections: 1, excludeIds: none, model: astron-code-latest
[13:58:16] 🔍 [AUTH] openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9 | available: 1/1
[13:58:16] ℹ️ [AUTH] Using openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9 account: yo
[13:58:16] 🔍 [FORMAT] openai → openai-responses | stream=false
[13:58:16] 🔍 [MODALITY] stripped unsupported media for openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9/astron-code-latest
[13:58:16] [PENDING] START | provider=openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9 | model=astron-code-latest
[13:58:16] 🔍 [REQUEST] OPENAI-COMPATIBLE-RESPONSES-6D80B485-82A3-4D80-80E9-2623C079D2E9 | astron-code-latest | 1 msgs
[13:58:16] [PENDING] END | provider=openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9 | model=astron-code-latest
[ChatCore] Failed to parse JSON from openai-compatible-responses-6d80b485-82a3-4d80-80e9-2623c079d2e9: Unexpected token '<', "<!doctype "... is not valid JSON
[13:58:16] 🌊 [STREAM] OPENAI-COMPATIBLE-RESPONSES-6D80B485-82A3-4D80-80E9-2623C079D2E9 | astron-code-latest | 245ms | complete
[...]

PR #2046, opened 2026-06-24
fix(stream): prevent non-JSON SSE lines and duplicate [DONE] from breaking clients

qianze0628:
## Problem

When using 9router as a translate-mode proxy (Claude-format client → OpenAI-format upstream), clients intermittently receive errors like:

```
oc: decode stream: invalid character 'D' looking for beginning of value
```

This has two root causes:

### 1. Non-JSON data lines forwarded in passthrough mode
In `stream.js` passthrough mode, when `JSON.parse()` fails on a `data:` line, the empty `catch {}` block silently swallows the error — but the fallback code at line 172-178 then **forwards the raw non-JSON line to the client**. Upstream providers sometimes return plain-text errors or HTML error pages embedded in the SSE stream, which break downstream JSON decoders.

### 2. `data: [DONE]` sentinel in translate mode
In translate mode, 9router emits the OpenAI-style `data: [DONE]` sentinel after `message_stop`. Some clients (e.g. Reasonix, which uses a Go JSON decoder) attempt to parse `[DONE]` as JSON, causing: `invalid character 'D' looking for beginning of value`. The `message_stop` event already signals end-of-response; the `[DONE]` sentinel is redundant in Claude-format output.

## Fix

**`open-sse/utils/stream.js`:**
1. **Passthrough mode**: Changed `catch {}` to `catch { continue; }` — non-JSON `data:` lines are now skipped silently instead of being forwarded raw to the client.
2. **Translate mode**: Removed `data: [DONE]` emission from both the transform loop and flush. The `message_stop` event already terminates the response.
3. **Passthrough mode**: Added `streamDoneSent` flag to prevent duplicate `[DONE]` emission (transform + flush were both sending it).

**`open-sse/handlers/chatCore/streamingHandler.js`:**
4. Added diagnostic `console.warn` when upstream returns an unexpected Content-Type (e.g. `text/html` instead of `text/event-stream`) for a streaming response — helps diagnose upstream errors.

## Testing

Tested with **CodeBuddy CN** (`cbcn/glm-5.2`) via 9router translate mode (Claude format client → OpenAI format upstream).

- **Before**: Intermittent `decode stream: invalid character 'D'` errors on every tool-call response
- **After**: Stable, no errors across 50+ requests including tool calls

## Related Issues

- #1996 (Combo fallback not triggering on stream errors — same class of problem: upstream errors embedded in SSE stream)
- #1396 (SSE-tainted JSON body — `data: [DONE]` appended to non-streaming responses)
- #1930 (upstream returning `<!doctype` HTML instead of JSON)

### DIFF
--- a/open-sse/handlers/chatCore/streamingHandler.js
+++ b/open-sse/handlers/chatCore/streamingHandler.js
@@ -46,6 +46,14 @@ function buildTransformStream({ provider, sourceFormat, targetFormat, userAgent,
 export function handleStreamingResponse({ providerResponse, provider, model, sourceFormat, targetFormat, userAgent, body, stream, translatedBody, finalBody, requestStartTime, connectionId, apiKey, clientRawRequest, onRequestSuccess, reqLogger, toolNameMap, streamController, onStreamComplete }) {
   if (onRequestSuccess) onRequestSuccess();
 
+  // Warn when upstream returns unexpected Content-Type for a streaming response.
+  // This often means the provider returned an HTML error page or plain-text error
+  // that the SSE transform stream would forward as garbage to the client.
+  const upstreamContentType = (providerResponse.headers.get('content-type') || '').toLowerCase();
+  if (upstreamContentType && !upstreamContentType.includes('text/event-stream') && !upstreamContentType.includes('application/json')) {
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @qianze0628 for the contribution! Reviewed and accepted — applying it to master. 🙏

Issue #1596, opened 2026-06-01
GitHub Copilot does not support the Opus 4.8 model on 9router.


gyf-dev:
**Provider returned no completion choices for this model**
**Actually, that model is available.**
<img width="302" height="454" alt="Image" src="https://github.com/user-attachments/assets/2fad5749-2db1-4684-a088-0615e5b84686" />
<img width="490" height="303" alt="Image" src="https://github.com/user-attachments/assets/995aaa63-1809-4d4f-90b6-f5dc9fe2bcc7" />

PR #2052, opened 2026-06-24
feat(github): resolve Copilot model catalog from upstream

caiqinzhou:
## Summary

The GitHub Copilot provider exposed a hardcoded model list in `/v1/models`,
so newly released models (e.g. `claude-opus-4.8`, `gpt-5.5`, `gemini-3.5-flash`)
were missing from the catalog until the static registry was manually updated,
even though they were already usable upstream.

This adds a live resolver that fetches the available models from the Copilot
`/models` endpoint, mirroring the existing Kiro and Qoder resolvers. The
catalog is cached per credential (5 min TTL) and the Copilot token is
refreshed on a 401/403 before retrying, so the list stays current without
code changes.

## Changes

- Add `open-sse/services/copilotModels.js` to fetch and cache the live Copilot
  model catalog (chat models only; embeddings and disabled models filtered out).
- Wire it into `LIVE_MODEL_RESOLVERS` in `/v1/models`, persisting any refreshed
  Copilot token back to the connection.
- Raise the connectivity-test token budget from 1 to 16. Claude models on
  Copilot return no completion choices at `max_tokens: 1` (the budget is spent
  before content is emitted), which produced a false negative in the model test.

## Testing

- `GET /v1/models` returns the live catalog including the newly released models.
- Chat requests to `gh/claude-opus-4.8` succeed via both the OpenAI
  (`/v1/chat/completions`) and Anthropic (`/v1/messages`) formats.
- Expired Copilot token triggers a refresh-and-retry; the list resolves correctly.
- Falls back to the static registry when the live fetch fails (offline / error).

### DIFF
--- a/open-sse/services/copilotModels.js
+++ b/open-sse/services/copilotModels.js
@@ -1,0 +1,155 @@
+/**
+ * GitHub Copilot model catalog fetcher.
+ *
+ * Calls Copilot's `GET /models` endpoint to get the live catalog for an
+ * authenticated account, so `/v1/models` reflects what the account can
+ * actually use (e.g. newly shipped `claude-opus-4.8`, `gpt-5.5`) instead of
+ * the hand-maintained static registry, which inevitably lags behind.
+ *
+ * Returns chat-capable models the account's policy allows. Embeddings and
+ * disabled models are filtered out.
+ */
+
+import { proxyAwareFetch } from "../utils/proxyFetch.js";
+import { GITHUB_COPILOT } from "../config/appConstants.js";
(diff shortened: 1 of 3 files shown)


caiqinzhou:
This feat can fix this issue: https://github.com/decolua/9router/issues/1596

decolua:
Thanks @caiqinzhou — clean, well-scoped change. The live Copilot resolver mirrors the Kiro/Qoder pattern, caches per-credential with refresh-on-401/403, and falls back to the static registry, and the max_tokens:16 probe fix is correct. Merged downstream. 🙏

PR #2054, opened 2026-06-24
fix(antigravity): retry transient upstream failures

Jordannst:
## Why

Antigravity upstream occasionally returns short-lived capacity/transient errors for Claude models, especially:

- `503 MODEL_CAPACITY_EXHAUSTED` / “No capacity available...”
- high-traffic or terminated-agent style transient failures
- intermittent 5xx/network-like failures before a response can complete

Before this change, these failures could immediately surface to clients even when a short retry or another Antigravity account could recover successfully.

While testing the retry path with Claude clients, I also hit a separate Antigravity request validation failure:

- `400 invalid_request_error: tools: Tool names must be unique.`

That can happen when client tool names collide after Antigravity name sanitization, e.g. distinct incoming names normalize into the same forwarded tool name.

## What changed

- Adds Antigravity-specific retry delay handling for transient upstream failures.
- Enables the retry hook for Antigravity `500` responses in addition to existing `429`/`503` behavior.
- Keeps bounded backoff behavior:
  - uses `Retry-After` / reset hints when present,
  - skips retry when the wait would be too long,
  - uses short exponential-style retry delays for transient 5xx/capacity failures.
- Preserves non-transient `400` behavior: invalid requests are not retried.
- Deduplicates sanitized Antigravity tool names before emitting the final single `functionDeclarations` group.
- Adds tests for retry behavior, BaseExecutor retry-hook gating, and sanitized tool-name collisions.

## Why this is safe

- Retry remains config-gated by status code; adding `500` only enables the existing Antigravity retry hook for that status.
- The hook can still veto retries, e.g. long `Retry-After` / reset windows.
- Non-transient `400` errors remain non-retryable.
- Tool dedupe only affects duplicate names after sanitization, preventing upstream rejection of the whole request.

## Verification

- `npx vitest run tests/unit/antigravity-retry-hook.test.js tests/unit/base-executor-retry.test.js tests/translator/bugs-antigravity.test.js`
  - 3 files passed, 20 tests passed, 1 expected fail.

Manual e2e via 9Router dev server on `127.0.0.1:20138` with Claude client using `ag/claude-opus-4-6-thinking`:

- observed `503` retries with `2s → 4s → 8s` backoff,
- observed fallback to another Antigravity account,
- observed successful streamed completion after retry/fallback,
- confirmed the previous `tools: Tool names must be unique` error no longer occurred after the dedupe patch and fresh restart.

### DIFF
--- a/open-sse/executors/antigravity.js
+++ b/open-sse/executors/antigravity.js
@@ -188,15 +188,22 @@ export class AntigravityExecutor extends BaseExecutor {
 
     if (tools && tools.length > 0) {
       // Merge all groups into a single functionDeclarations group (Gemini expects 1 group)
-      const allDeclarations = tools.flatMap(group =>
-        (group.functionDeclarations || []).map(fn => ({
-          ...fn,
-          name: sanitizeFunctionName(fn.name),
-          parameters: fn.parameters
-            ? cleanJSONSchemaForAntigravity(structuredClone(fn.parameters))
-            : { type: "object", properties: { reason: { type: "string", description: "Brief explanation" } }, required: ["reason"] }
-        }))
-      );
+      const seenToolNames = new Set();
+      const allDeclarations = [];
+      for (const group of tools) {
+        for (const fn of group.functionDeclarations || []) {
+          const name = sanitizeFunctionName(fn.name);
+          if (seenToolNames.has(name)) continue;
+          seenToolNames.add(name);
+          allDeclarations.push({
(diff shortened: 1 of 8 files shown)


decolua:
Thanks @Jordannst! Reviewed and merged into master. The transient 5xx/capacity retry with bounded backoff, tool-name dedupe, and Headroom phantom-savings diagnostics all look solid. 🙏

PR #2055, opened 2026-06-24
fix(tts): resolve Gemini TTS models from catalog

nguyenha935:
## Summary
- Fix Gemini TTS requests resolving to `models/undefined:generateContent` when the handler has no `ttsConfig.models`.
- Resolve Gemini TTS models from the shared TTS catalog and provider registry, with a safe fallback.
- Add `gemini-3.1-flash-tts-preview` to Gemini TTS catalogs and cover model/voice parsing with unit tests.

## Tests
- `tests/node_modules/.bin/vitest run tests/unit/gemini-tts.test.js --config tests/vitest.config.js`
- `tests/node_modules/.bin/vitest run tests/unit/minimax-tts.test.js --config tests/vitest.config.js`
- `tests/node_modules/.bin/vitest run tests/unit/provider-display-split.test.js --config tests/vitest.config.js`

### DIFF
--- a/open-sse/config/ttsModels.js
+++ b/open-sse/config/ttsModels.js
@@ -96,10 +96,12 @@ export const TTS_MODELS_CONFIG = {
   },
   gemini: {
     models: [
+      { id: "gemini-3.1-flash-tts-preview", name: "Gemini 3.1 Flash TTS", type: "tts" },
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @nguyenha935! Merged into master. 🙏

PR #2059, opened 2026-06-24
fix: support Kiro IDC (organization) token import

enjoyer-hub:
## Problem
When logged in to Kiro IDE as an organization (AWS IAM Identity Center / IDC), the token import into 9router fails:
1. Token refresh fails - IDC tokens need \clientId\/\clientSecret\ from OIDC client registration
2. Wrong profileArn causes 403 or 400 errors at the runtime gateway

## Root Cause
- IDC tokens refresh via \https://oidc.{region}.amazonaws.com/token\ with client credentials, NOT the social endpoint
- The profileArn region must be \us-east-1\ for the \
untime.us-east-1.kiro.dev\ gateway regardless of IDC region
- The client registration credentials are stored in \~/.aws/sso/cache/{clientIdHash}.json\

## Fix
- **auto-import**: Reads \clientIdHash\ from the token file, loads the client registration file for \clientId\/\clientSecret\, reads \profileArn\ from Kiro IDE profile.json, normalizes ARN region to us-east-1
- **import**: Accepts IDC credentials in the request body, uses \KiroService.refreshToken()\ with them, persists credentials for future automatic refreshes
- **KiroAuthModal**: Passes IDC credentials from auto-detect response through to the import endpoint

## Testing
- Tested with Kiro Power subscription (IDC organization, eu-central-1)
- Verified token refresh works via regional OIDC endpoint
- Verified API calls succeed with normalized profileArn
- Social/Google auth tokens continue to work unchanged (they don't send clientId/clientSecret)

### DIFF
--- a/src/app/api/oauth/kiro/auto-import/route.js
+++ b/src/app/api/oauth/kiro/auto-import/route.js
@@ -5,13 +5,14 @@ import { join } from "path";
 
 /**
  * GET /api/oauth/kiro/auto-import
- * Auto-detect and extract Kiro refresh token from AWS SSO cache
+ * Auto-detect and extract Kiro refresh token from AWS SSO cache.
+ * For IDC (organization) tokens, also resolves clientId/clientSecret from the
+ * linked client registration file so token refresh works.
  */
 export async function GET() {
   try {
     const cachePath = join(homedir(), ".aws/sso/cache");
 
-    // Try to read cache directory
     let files;
     try {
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @enjoyer-hub! Merged into master. IDC (organization) token import now works — clientId/clientSecret resolved from SSO cache, profileArn normalized to us-east-1, social/builder-id path unchanged. 🙏

PR #2065, opened 2026-06-24
fix(gemini): support native TTS generateContent endpoint

nguyenha935:
## Summary
- expose Gemini models with Google-native /v1beta model names while keeping existing provider-prefixed entries
- add rewrites so standard /v1beta/* requests reach the existing /api/v1beta/* route handlers
- pass Gemini AUDIO generateContent requests through to Google's native endpoint without converting them to chat
- validate client keys from Bearer, x-goog-api-key, or ?key= while forwarding only the configured Gemini provider credential upstream
- add Gemini 3.1 Flash TTS to the Gemini/TTS catalogs

## Tests
- node_modules/vitest/vitest.mjs run tests/unit/gemini-native-endpoint.test.js tests/unit/minimax-tts.test.js tests/unit/provider-display-split.test.js tests/unit/provider-test-models-routing.test.js --config tests/vitest.config.js
- npm run build

### DIFF
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -67,6 +67,14 @@ const nextConfig = {
         destination: "/api/v1/responses"
       },
       {
+        source: "/v1beta/:path*",
+        destination: "/api/v1beta/:path*"
+      },
+      {
+        source: "/v1beta",
+        destination: "/api/v1beta"
+      },
+      {
         source: "/v1/:path*",
         destination: "/api/v1/:path*"
       },
--- a/open-sse/config/runtimeConfig.js
+++ b/open-sse/config/runtimeConfig.js
@@ -50,7 +50,7 @@ export const STREAM_FIRST_CHUNK_TIMEOUT_MS = envMs("STREAM_FIRST_CHUNK_TIMEOUT_M
 export const FETCH_CONNECT_TIMEOUT_MS = envMs("FETCH_CONNECT_TIMEOUT_MS", 60 * 1000);
 
 // Gemini native TTS fetch timeout: abort if Google does not return response headers in time.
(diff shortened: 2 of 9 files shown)


nguyenha935:
Follow-up added after runtime testing on my Tiny server.

The native Gemini TTS route now catches fetch/header-timeout failures inside the per-credential loop instead of letting them escape to the outer 500 handler. This preserves the native Gemini response shape, maps header/internal timeouts to 504, maps other fetch failures to 502, marks the failed Gemini credential unavailable, and falls back to the next credential when configured.

I also added safe structured logs for native Gemini attempts and regression tests for timeout fallback, fetch failure without fallback, and client abort behavior.

Verified locally:
- npm --prefix tests test -- unit/gemini-native-endpoint.test.js
- npm run build
- runtime test through /v1beta/models/gemini-3.1-flash-tts-preview:generateContent returned 200 with native candidates response.

nguyenha935:
Runtime follow-up: I reduced the default Gemini native TTS fetch timeout from 90s to 45s.

Reason: the 90s timeout worked technically, but it consumed too much of the caller-side TTS timeout before account fallback could help. In production logs, one stuck Gemini credential timed out at ~90s, then fallback started correctly, but the client aborted ~35s later. A separate successful longer TTS request completed in ~28s, so 30s was too tight. 45s gives enough room for slower-but-valid TTS while still leaving time for credential fallback before common 120s client timeouts.

Verified again:
- npm --prefix tests test -- unit/gemini-native-endpoint.test.js
- npm run build
- runtime env on my server currently uses GEMINI_NATIVE_TTS_FETCH_TIMEOUT_MS=45000 and native Gemini TTS requests return 200.

decolua:
Thanks @nguyenha935! Reviewed (code + security): auth gating runs before upstream forward, client `?key=` is stripped before reaching Google, no SSRF/SQLi, credentials never logged. Squashed and accepted into our internal master. 🙏

PR #2070, opened 2026-06-24
fix: preserve forced streaming for json clients

### LINKED ISSUE #2031: forceStream option overridden on non-streaming requests, causing 400 Bad Request on stream-only providers (e.g. Command Code)
When a client (such as Claude Code or Hermes Agent) makes a non-streaming request (such as generating session titles or other auxiliary tasks), it sends headers accepting `application/json` without `text/event-stream`, and sets `stream: false` in the request body.

Even if the configured provider requires streaming (i.e. `forceStream: true` is set in the provider's transport configuration), 9Router's routing logic overrides the streaming flag `ah` to `false` when it sees that the client accepts JSON and did not request a stream.

This results in 9Router sending a non-streaming request (`stream: false`) to the downstream provider. For providers like **Command Code (CommandCode.ai)** which mandate streaming, this yields an API validation error:
`HTTP 400 Bad Request: Invalid input: expected true at "params.stream"`.

---

## Root Cause Analysis
In `server/chunks/8895.js` (or the corresponding source file in the router handler), the variable `ah` (target streaming flag) is computed as follows:

```javascript
let af=!0===a.stream||_===g.h.ANTIGRAVITY||_===g.h.GEMINI||_===g.h.GEMINI_CLI,
    ag=n.xq[Y]?.forceStream===!0,
    ah=!!ag||!1!==a.stream; // ah correctly starts as true if forceStream is enabled
```

However, later in the same block, the client's Accept header is checked:
```javascript
let aj=I?.headers?.accept||"",
    ak=aj.includes("application/json"),
    al=aj.includes("text/event-stream");
ak&&!al&&!0!==a.stream&&(ah=!1); // BUG: This unconditionally overrides ah to false
```
Because the client requested a non-streaming response (`ak && !al && !0 !== a.stream` evaluates to `true`), `ah` is set to `false`, overriding `ag` (`forceStream`).

---

## Suggested Fix
The override should only apply if the provider does **not** force streaming. We can add a check for `!ag` (or `!forceStream`) before overriding `ah` to `false`:

```diff
-ak&&!al&&!0!==a.stream&&(ah=!1);
+ak&&!al&&!0!==a.stream&&!ag&&(ah=!1);
```

This ensures that if `ag` (`forceStream`) is `true`, `ah` remains `true` and the request to the provider is correctly streamed. 9Router's existing stream accumulator (`(0, u.I)`) will then consume the provider's stream and return a standard non-streaming JSON response to the client.

---

## Steps to Reproduce
1. Configure a provider that requires streaming (e.g. Command Code `cmc` with `forceStream: true`).
2. Make a non-streaming POST request to 9Router's `/v1/messages` endpoint:
   ```bash
   curl -X POST http://localhost:20128/v1/messages \
     -H "Authorization: Bearer <your-key>" \
     -H "Accept: application/json" \
     -d '{
       "model": "cmc/deepseek/deepseek-v4-pro",
       "messages": [{"role": "user", "content": "test"}],
       "stream": false,
       "max_tokens": 100
     }'
   ```
3. Observe the `400 Bad Request` validation error from the provider.

gitcommit90:
## Summary
Preserve provider-required streaming even when the client prefers JSON responses.

Some providers require upstream streaming. `handleChatCore` correctly enables streaming for those providers, but then the later `Accept: application/json` branch could flip `stream` back to `false`. That breaks stream-only providers for non-SSE clients such as Hermes / Claude Code / other JSON clients.

## Changes
- Do not let the JSON Accept override disable streaming when the selected provider requires streaming
- Add regression coverage for both `body.stream` omitted and `body.stream: false`
- Keep the existing provider config force-stream test intact

## Verification
- Watched the new regression test fail first with `expected false to be true`
- `npm --prefix tests test -- force-stream-config.test.js` — 3 passed
- `npm run build` — compiled successfully and generated static pages

Fixes #2031

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -89,7 +89,7 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   const acceptHeader = clientRawRequest?.headers?.accept || "";
   const clientPrefersJson = acceptHeader.includes("application/json");
   const clientPrefersSSE = acceptHeader.includes("text/event-stream");
-  if (clientPrefersJson && !clientPrefersSSE && body.stream !== true) {
+  if (clientPrefersJson && !clientPrefersSSE && body.stream !== true && !providerRequiresStreaming) {
     stream = false;
   }
 
--- a/tests/unit/force-stream-config.test.js
+++ b/tests/unit/force-stream-config.test.js
@@ -1,9 +1,136 @@
 // Guards forceStream moved from chatCore hardcode → PROVIDERS schema (#5).
-import { describe, it, expect } from "vitest";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { executeMock } = vi.hoisted(() => ({
+  executeMock: vi.fn(),
+}));
+
+vi.mock("../../open-sse/executors/index.js", () => ({
+  getExecutor: vi.fn(() => ({
+    execute: executeMock,
+    refreshCredentials: vi.fn().mockResolvedValue(null),
+  })),
+}));
+
+vi.mock("../../open-sse/utils/requestLogger.js", () => ({
+  createRequestLogger: vi.fn(async () => ({
+    logClientRawRequest: vi.fn(),
+    logRawRequest: vi.fn(),
+    logTargetRequest: vi.fn(),
+    logError: vi.fn(),
+  })),
+}));
+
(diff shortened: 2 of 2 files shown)


decolua:
Thanks @gitcommit90! Clean, minimal fix — guarding the JSON-Accept branch with `providerRequiresStreaming` correctly preserves forced streaming for stream-only providers. Merged into our tree. 🙏

PR #2073, opened 2026-06-25
fix(codex): preserve Responses text format

tungloong:
## Summary
- Preserve the Responses API `text` field when normalizing Codex requests.
- Add a regression test for `text.format.json_schema` passthrough in `CodexExecutor`.

## Why
Codex Desktop uses `text.format.json_schema` for structured title generation. `CodexExecutor` previously stripped `text` in its final allowlist filter, so the Codex Responses backend did not receive the schema and could return plain text instead of JSON.

Keeping `text` allows structured output requests to reach the upstream Codex Responses API while the executor still strips unsupported fields.

## Tests
- `NODE_PATH=/tmp/node_modules /tmp/node_modules/.bin/vitest run --config tests/vitest.config.js tests/unit/codex-tool-normalization.test.js --reporter=verbose`


decolua:
Thanks @tungloong! Reviewed and verified — adding `text` to the Responses API allowlist correctly preserves `text.format.json_schema` for structured outputs, and `metadata` is still stripped. Tests pass locally. Merging this into our tree manually. 🙏

PR #2080, opened 2026-06-25
fix(i18n): add missing zh-CN endpoint key label

dajinglingpake:
Add the missing Simplified Chinese translation for the `Endpoint & Key` sidebar label.


decolua:
Thanks @dajinglingpake! Merged the zh-CN `Endpoint & Key` label into master. 🙏

PR #2085, opened 2026-06-25
fix(headroom): translate openai-responses input through OpenAI for compression (#1998)

### LINKED ISSUE #1998: fix(headroom): provider receives full tokens despite Headroom reporting savings

anki1kr:
## Summary

`compressWithHeadroom()` treated a Codex (`openai-responses`) request's `body.input` array as if it were OpenAI `messages`: it POSTed the Responses items to Headroom's `/v1/compress` and then assigned the returned OpenAI messages straight back to `body.input`. That breaks the Responses format contract — `body.input` must hold Responses items (`{ type: "message", role, content: [...] }`), not OpenAI messages (`{ role, content: "<string>" }`) — so downstream Codex translation/serialization gets a malformed body. (#1998)

This adds an `openai-responses` branch that mirrors the existing `claude` path: translate `input` -> OpenAI messages, compress, translate back to `input`, using the translators already exported from `open-sse/translator/request/openai-responses.js`.

Fixes #1998

## Changes

| File | Change |
|---|---|
| `open-sse/rtk/headroom.js` | Add `openai-responses` branch: `openaiResponsesToOpenAIRequest` -> compress -> `openaiToOpenAIResponsesRequest`; reassign `body.input` as Responses items |
| `tests/unit/headroom-responses-format.test.js` | Unit test: mocks Headroom, asserts `body.input` stays Responses-shaped after compression |

## Testing

- `unit/headroom-responses-format.test.js` -> 1/1 pass (fails before the fix: `body.input[0]` was a raw OpenAI message)
- `eslint` on changed files -> 0 errors
- Additive change — the existing `claude` and OpenAI-`messages` paths are untouched.

### DIFF
--- a/open-sse/rtk/headroom.js
+++ b/open-sse/rtk/headroom.js
@@ -1,5 +1,9 @@
 import { claudeToOpenAIRequest } from "../translator/request/claude-to-openai.js";
 import { openaiToClaudeRequest } from "../translator/request/openai-to-claude.js";
+import {
+  openaiResponsesToOpenAIRequest,
+  openaiToOpenAIResponsesRequest,
+} from "../translator/request/openai-responses.js";
 
 const DEFAULT_TIMEOUT_MS = 3000;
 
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @anki1kr! Merged into master. 🙏

PR #2086, opened 2026-06-25
fix(models): show custom provider models in combo picker

saptoji:
## Problem
Commit 707a915 added `customModels` merge for the `passthroughModels` provider branch, but the `isCustomProvider` branch (OpenAI/Anthropic compatible) was missed. Custom compatible providers show a placeholder `prefix/model-id` instead of the actual imported models in the "Add Model to Combo" modal.

## Fix
Merge `customModels` from `/api/models/custom` into the model list for custom compatible providers, matching the same pattern already applied to passthrough providers.

Key detail: filter by `providerId` (not `nodePrefix`) because `providerAlias` in the DB uses the raw provider ID, not the display prefix.

## Before / After
- **Before**: Custom compatible providers (e.g. EV evo, bai, tokenrouter) show placeholder `ev/model-id` with pencil icon
- **After**: Actual imported models (e.g. `evomap-kimi-k2.6`, `evomap-deepseek-v4-flash`) appear and can be added to combos

### DIFF
--- a/src/shared/components/ModelSelectModal.js
+++ b/src/shared/components/ModelSelectModal.js
@@ -249,9 +249,22 @@ export default function ModelSelectModal({
             value: `${nodePrefix}/${fullModel.replace(`${providerId}/`, "")}`,
           }));
 
+        // Merge custom models registered via /api/models/custom for this provider
+        // providerAlias in DB uses the raw providerId, not the display prefix
+        const registeredCustom = customModels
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @saptoji! Reviewed and merged locally. This correctly mirrors the passthrough `customModels` merge for the `isCustomProvider` branch, filtering by `providerId` with proper dedup. 🙏

PR #2102, opened 2026-06-25
fix(auth): avoid stale redirects after auth changes

### LINKED ISSUE #2100: Login succeeds but dashboard keeps sending user back to password screen
## Problem

Sometimes after logging in with the dashboard password, 9Router opens the dashboard successfully on the default Endpoint page. However, during dashboard navigation, the app may suddenly redirect back to the password login screen.

After entering the password again, the dashboard opens again, but the same behavior can repeat when navigating to another dashboard page. A hard refresh with `Ctrl+Shift+R` appears to stop the loop.

## Expected behavior

After a successful password login, dashboard navigation should keep the authenticated session and should not redirect back to `/login` unless the session expires or the user logs out.

## Actual behavior

The dashboard intermittently behaves as if the user is unauthenticated during client-side navigation after a successful login.

## Reproduction

This does not happen every time, so there is no fully reliable reproduction yet.

Observed flow:

1. Open the 9Router dashboard.
2. Enter the dashboard password.
3. Dashboard opens on the default Endpoint page.
4. Navigate between dashboard pages/tabs.
5. Sometimes the app redirects back to the password login screen.
6. Re-entering the password may temporarily work.
7. The redirect loop stops after `Ctrl+Shift+R`.

## Notes

Authentication appears to use an HTTP-only `auth_token` cookie, not `localStorage`.

The fact that `Ctrl+Shift+R` clears the issue suggests this may be related to stale client-side navigation/cache state after login.

Relevant files to inspect:

- `src/app/login/page.js`
- `src/app/api/auth/login/route.js`
- `src/lib/auth/dashboardSession.js`
- `src/dashboardGuard.js`

Potential fix direction:

- Avoid client-side router navigation immediately after successful login.
- Use a full-page navigation after login so the next dashboard request is made with the fresh auth cookie.
- Add `Cache-Control: no-store` to auth login/logout responses.

emi-ran:
## Summary
- use full-page navigation after password login so dashboard reloads with the fresh auth cookie
- cover the remote default-password change flow before entering the dashboard
- use full-page navigation after logout from header/profile to avoid stale authenticated client state
- mark login/logout auth responses as no-store

Fixes #2100

## Testing
- npm run build
- .\\tests\\node_modules\\.bin\\vitest.cmd run tests/unit/dashboard-guard.test.js tests/unit/web-cookie-validation.test.js --reporter=verbose

### DIFF
--- a/src/app/(dashboard)/dashboard/profile/page.js
+++ b/src/app/(dashboard)/dashboard/profile/page.js
@@ -1,7 +1,6 @@
 "use client";
 
 import { useState, useEffect, useRef } from "react";
-import { useRouter } from "next/navigation";
 import { Card, Button, Toggle, Input } from "@/shared/components";
 import Modal, { ConfirmModal } from "@/shared/components/Modal";
 import LanguageSwitcher from "@/shared/components/LanguageSwitcher";
@@ -21,7 +20,6 @@ function getLocaleFromCookie() {
 }
 
 export default function ProfilePage() {
-  const router = useRouter();
   const { theme, setTheme, isDark } = useTheme();
   const [locale, setLocale] = useState("en");
   const [langOpen, setLangOpen] = useState(false);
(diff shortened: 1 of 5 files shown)


decolua:
Thanks @emi-ran for the fix! Verified the diff — full-page navigation after login/logout plus no-store headers cleanly resolves the stale auth redirect in #2100. Reviewed and accepted; merging on our side. 🙏

PR #2106, opened 2026-06-25
fix: show compatible custom models in combo picker

nguyenha935:
## Summary
- Include custom models registered for OpenAI/Anthropic-compatible provider nodes in the Combo Add Model picker.
- Keep filtering by the provider node storage id while returning combo values with the user-facing node prefix.
- Add regression coverage for compatible provider custom models and legacy aliases using a display prefix.

## Test Plan
- node node_modules/vitest/vitest.mjs run tests/unit/provider-custom-models.test.js --config tests/vitest.config.js
- git diff --check -- src/shared/components/ModelSelectModal.js src/shared/utils/providerCustomModels.js tests/unit/provider-custom-models.test.js
- npm run build


decolua:
Thanks for your contribution! 🙏 This bug (custom models on openai/anthropic-compatible providers not showing in the combo model picker) has already been fixed on `master`: the `groupedModels` logic in `src/shared/components/ModelSelectModal.js` now merges `customModels` (`registeredCustom` + `mergedModels`) for the custom-provider branch. Closing this PR as it's now redundant. Appreciate the effort!

PR #2108, opened 2026-06-25
feat(quota): add opt-in Codex auto-ping

### LINKED ISSUE #2107: Add opt-in Codex auto-ping for 5h quota windows
﻿## Summary

Add an opt-in Codex auto-ping feature that sends a tiny `gpt-5.5` request after the Codex session quota reset time, so the next 5h quota window starts immediately.

## Problem

Codex session quota windows appear to start when the first request is sent after reset. If a user does not send a request until later, the next reset time shifts later as well.

Example:

- Session quota resets at 11:00
- User first sends a Codex request at 15:00
- Next 5h window resets around 20:00

With auto-ping enabled:

- Session quota resets at 11:00
- 9Router sends a tiny Codex request shortly after reset
- Next 5h window starts immediately
- Next reset is around 16:00

## Proposed Behavior

- Add Codex support to the existing quota auto-ping behavior.
- Keep the feature disabled by default.
- Allow enabling it per Codex OAuth connection.
- Use only `gpt-5.5` for Codex auto-ping.
- Send a tiny request after `usage.quotas.session.resetAt`.
- Do not retry repeatedly for the same reset window.
- Do not run unless the 9Router server process is running.
- Keep existing Claude auto-ping behavior unchanged.

## UI

Add an Auto-ping toggle for Codex OAuth connections in:

- Quota Tracker provider cards
- Provider connection detail rows

Tooltip copy:

> Auto-starts the next 5h Codex window after reset by sending a tiny gpt-5.5 request. Consumes a small amount of quota.

## Safety

- Default off.
- Per-connection opt-in.
- OAuth Codex only.
- Do not log access tokens or request headers.
- Do not mark a reset as pinged unless the ping succeeds.
- Add cooldown for failed ping attempts to avoid request spam.

## Implementation Notes

- Add a provider-generic `quotaAutoPing` scheduler.
- Preserve existing `settings.claudeAutoPing`.
- Add `settings.codexAutoPing`.
- Use `gpt-5.5` only for Codex auto-ping.
- Use existing token refresh and proxy-aware fetch paths.

## Acceptance Criteria

- Existing Claude auto-ping still works.
- Codex auto-ping is disabled by default.
- Codex auto-ping can be enabled per OAuth connection.
- Codex auto-ping sends a tiny `gpt-5.5` request after the session reset time.
- Same `resetAt` is not pinged twice.
- Failed pings do not spam every tick.
- Tests cover Codex enabled/disabled/reset/not-reset/same-reset cases.

emi-ran:
## Summary

- add provider-generic quota auto-ping scheduler
- add opt-in Codex auto-ping using a tiny `gpt-5.5` request when the 5h Codex window is inactive
- preserve existing Claude auto-ping settings and behavior
- add Codex auto-ping toggles in Quota Tracker and provider connection rows
- simplify Codex reset credits UI by making the reset-credit chip the reset action

Closes #2107

## Notes

Codex auto-ping is disabled by default and only runs for explicitly enabled OAuth connections. It requires the 9Router server process to be running. Failed pings use a cooldown so provider endpoints are not spammed.

Codex reset credits still require confirmation before one credit is consumed. The separate `Reset limit` button was removed to avoid confusion with the auto-ping bolt action on compact card layouts.

Codex session usage is not used to decide whether to ping because the usage API can report values like `session.used = 1` before the 5h window is actually started. The scheduler now uses the simpler reliable signal: `session.resetAt`. If it slides forward between checks, the 5h window is inactive and needs the tiny start-window ping. If it is stable, the 5h window is already running and no extra ping is sent.

Weekly/monthly-style blocking quotas are checked before pinging. If an upper quota is exhausted, Codex auto-ping skips until that quota resets.

## Verification

```powershell
.\tests\node_modules\.bin\vitest.cmd run tests/unit/quota-auto-ping.test.js tests/unit/usage-dispatch.test.js tests/unit/codex-refresh-token.test.js --reporter=verbose
```

```powershell
git diff --check
```

```bash
npm run build
```

Local live verification showed Codex executor success and ping persistence:

```text
[DBG:FETCH] CODEX ← 200
[AutoPing] codex:2afef579-c3cd-46b7-a5ea-c2b58830bfcd: ping sent (reset 2026-06-27T12:17:08.000Z)
```


emi-ran:
It doesn't work right now, I'll fix it ASAP.

emi-ran:
Fixed now.

Root cause was twofold:

- Codex auto-ping initially bypassed the normal Codex executor path, so the request missed Codex-specific request normalization and headers.
- Codex `resetAt` can move by a few seconds, so exact timestamp comparison caused duplicate pings after it started working.

Current behavior:

- Codex auto-ping uses the normal Codex executor path.
- It sends one tiny `gpt-5.5` request per observed session reset window.
- It stores a minute-normalized reset key to avoid duplicate pings when `resetAt` seconds drift.
- I verified it locally: the scheduler sent Codex pings successfully and duplicate pings were fixed.

The PR is now marked ready for review and synced with current `master`.

emi-ran:
@decolua

emi-ran:
Follow-up update after live local testing:

- Codex auto-ping now fires through the normal Codex executor path.
- It sends one tiny `gpt-5.5` request only when the Codex session window has not been used yet.
- It skips auto-ping when the session quota is exhausted or when a blocking quota such as weekly/monthly is exhausted.
- It normalizes `resetAt` to the minute so small second-level drift from the usage API does not cause duplicate pings.

I verified this locally with real Codex quota data: pings are sent when expected, and exhausted weekly/session quota cases are skipped.

emi-ran:
Final follow-up from live testing:

Root cause: Codex usage can report `session.used = 1` even before the 5h window is actually started. That made the scheduler treat inactive sessions as already started.

Fix: the scheduler now watches `session.resetAt`. If it keeps sliding forward between checks, the Codex session is still inactive and auto-ping sends the tiny `gpt-5.5` request. If `resetAt` is stable, the 5h window is already running and no extra ping is sent.

Verified locally with the real Codex executor:

```text
[DBG:FETCH] CODEX ← 200
[AutoPing] codex:2afef579-c3cd-46b7-a5ea-c2b58830bfcd: ping sent (reset 2026-06-27T12:17:08.000Z)
```

emi-ran:
Simplified the Codex decision logic after the live test findings:

- Codex no longer uses `session.used` to decide whether to ping.
- Codex now only uses `session.resetAt` drift: sliding means inactive 5h window, stable means already running.
- Weekly/monthly-style exhausted quotas block auto-ping before any request is sent.
- Added regression coverage for first observation, sliding resetAt, stable resetAt, weekly exhausted, monthly exhausted, session exhausted, and duplicate prevention.

Verification passed:

```text
23 tests passed
npm run build passed
git diff --check: CRLF warnings only
```

emi-ran:
Superseded by the cleaner PR #2157.

Closing this one because the timeline and fix-up commits became noisy while we debugged Codex `resetAt` behavior. The replacement PR keeps the same final behavior with a clean branch and one focused commit.

Issue #1926, opened 2026-06-20
Cloudflare Worker API Returns 400 Bad Input Error: "oneOf at '/' not met"


theseven99:
Hi 9router team,

I'm currently unable to use any models through the API.

### Issue

All requests are returning a 400 error with the following message:

```text
[400]: {
  "errors": [
    {
      "message": "AiError: Bad input: Error: oneOf at '/' not met"
    }
  ]
}
```

Thank you.

PR #2110, opened 2026-06-26
fix(paramSupport): tolerate strip rules without match/drop (#2097)

### LINKED ISSUE #2097: Cloudflare provider crashes when paramSupport rule has no match predicate
## Summary

The built-in `cloudflare-ai` provider can fail every chat request with HTTP 502:

```text
[cloudflare-ai/@cf/meta/llama-3.2-1b-instruct] [502]: Cannot read properties of undefined (reading 'test')
```

This appears to come from `open-sse/translator/concerns/paramSupport.js`.

## Environment

- 9Router: `v0.5.8`
- Runtime: Docker image `decolua/9router:latest`
- Provider: built-in `cloudflare-ai`
- Models tested:
  - `cf/@cf/meta/llama-3.2-1b-instruct`
  - `cf/@cf/zai-org/glm-4.7-flash`

## Reproduction

1. Configure built-in Cloudflare provider with a valid API token and Account ID.
2. Send a local 9Router chat completion request using a Cloudflare model, for example:

```json
{
  "model": "cf/@cf/meta/llama-3.2-1b-instruct",
  "messages": [{ "role": "user", "content": "Reply exactly OK" }],
  "stream": false
}
```

3. The request fails with HTTP 502 and:

```text
Cannot read properties of undefined (reading 'test')
```

## Expected behavior

The request should be forwarded to Cloudflare Workers AI and return a normal OpenAI-compatible chat completion.

## Actual behavior

The 9Router request fails before successfully completing through the built-in Cloudflare provider.

## Why this seems to happen

`paramSupport.js` has a Cloudflare rule without a `match` predicate:

```js
{ provider: "cloudflare-ai", flattenContent: true }
```

But `matches()` assumes every matching rule has `rule.match`:

```js
function matches(rule, model) {
  return typeof rule.match === "function" ? rule.match(model) : rule.match.test(model);
}
```

For the Cloudflare rule, `rule.match` is undefined, so `rule.match.test(model)` throws.

## Local hotfix tested

This local patch fixed the crash:

```js
function matches(rule, model) {
  if (!rule.match) return true;
  return typeof rule.match === "function" ? rule.match(model) : rule.match.test(model);
}
```

Also, `stripUnsupportedParams()` should tolerate rules without `drop`:

```js
for (const key of rule.drop || []) {
  if (body[key] !== undefined) delete body[key];
}
```

After applying the local hotfix to both source and the built Next.js chunk, these requests returned HTTP 200 through 9Router:

- `cf/@cf/meta/llama-3.2-1b-instruct`
- `cf/@cf/zai-org/glm-4.7-flash`

Direct Cloudflare OpenAI-compatible calls with the same account/token/model also returned HTTP 200, so this does not appear to be a credential or model-access issue.

StevanusPangau:
## Problem

The built-in `cloudflare-ai` provider fails every chat request with HTTP 502:

```text
[cloudflare-ai/@cf/meta/llama-3.2-1b-instruct] [502]: Cannot read properties of undefined (reading 'test')
```
Root cause is in `open-sse/translator/concerns/paramSupport.js`. `STRIP_RULES` has a provider-wide rule without a `match` predicate or a `drop` list:

```js
{ provider: "cloudflare-ai", flattenContent: true }
```
But two spots assume every rule has those fields:

1. `matches()` runs `rule.match.test(model)` — throws when `rule.match` is `undefined`.
2. `stripUnsupportedParams()` iterates `for (const key of rule.drop)` — throws when `rule.drop` is `undefined` (the second crash, surfaced once the first is fixed).

So any request routed through `cloudflare-ai` crashes before reaching the upstream.

## Fix

Make both spots tolerate rules that omit `match`/`drop`:

- `matches()`: a rule without `match` applies to every model of that provider (`if (!rule.match) return true;`).
- `stripUnsupportedParams()`: iterate `rule.drop || []`.

This keeps the existing `flattenContent` behavior (#1926) working while no longer crashing.

## Why this is safe

- Existing rules that define `match`/`drop` are unaffected (verified by tests).
- A `match`-less rule is intentionally provider-wide; returning `true` matches that intent.
- A `drop`-less rule simply skips the drop loop.

## Tests

Adds `tests/translator/paramSupport-cloudflare.test.js`:

- `cloudflare-ai` model no longer throws and content is flattened to a plain string.
- Non-matching provider (`openai`) is left untouched.
- A rule that defines a `drop` list (claude-opus-4 `temperature`) still drops the param.

```bash
npm exec -- vitest run tests/translator/paramSupport-cloudflare.test.js --config tests/vitest.config.js
npm run build
```
All pass; the full translator concern/request/coverage suites stay green (no regressions).

Fixes #2097

### DIFF
--- a/open-sse/translator/concerns/paramSupport.js
+++ b/open-sse/translator/concerns/paramSupport.js
@@ -15,7 +15,10 @@ const STRIP_RULES = [
 ];
 
 // Test a rule's match (regex or predicate) against the model id.
+// A rule without a `match` (e.g. a provider-wide rule like flattenContent)
+// applies to every model of that provider.
 function matches(rule, model) {
+  if (!rule.match) return true;
   return typeof rule.match === "function" ? rule.match(model) : rule.match.test(model);
 }
 
@@ -25,7 +28,7 @@ export function stripUnsupportedParams(provider, model, body) {
   for (const rule of STRIP_RULES) {
     if (rule.provider && rule.provider !== provider) continue;
     if (!matches(rule, model)) continue;
-    for (const key of rule.drop) {
+    for (const key of rule.drop || []) {
       if (body[key] !== undefined) delete body[key];
     }
     // CF Workers AI oneOf root schema only accepts content as plain string (#1926)
--- a/tests/translator/paramSupport-cloudflare.test.js
+++ b/tests/translator/paramSupport-cloudflare.test.js
@@ -1,0 +1,38 @@
+// Regression tests for paramSupport rules without a `match` predicate or `drop`
+// list (e.g. the provider-wide `cloudflare-ai` flattenContent rule). Previously
+// `matches()` ran `rule.match.test(model)` on an undefined `rule.match` and the
+// drop loop iterated an undefined `rule.drop`, crashing every Cloudflare request
+// with "Cannot read properties of undefined (reading 'test')". Fixes #2097.
+import { describe, it, expect } from "vitest";
+import { stripUnsupportedParams } from "../../open-sse/translator/concerns/paramSupport.js";
+
+describe("stripUnsupportedParams: rule without match/drop (cloudflare-ai)", () => {
+  it("does not throw for a cloudflare-ai model and flattens content", () => {
+    const body = {
+      messages: [
+        { role: "user", content: [{ type: "text", text: "Reply exactly OK" }] },
+      ],
+    };
+
+    expect(() =>
+      stripUnsupportedParams("cloudflare-ai", "@cf/meta/llama-3.2-1b-instruct", body)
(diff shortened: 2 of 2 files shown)


decolua:
Thanks @StevanusPangau for the careful root-cause writeup and tests 🙏

The fix here (`if (!rule.match) return true` in `matches()` and `rule.drop || []` in the drop loop) has already landed on `master` via an earlier change, so `open-sse/translator/concerns/paramSupport.js` no longer crashes on the `cloudflare-ai` provider-wide rule. The remaining diff is effectively a no-op against current master.

Closing as already-fixed. Really appreciate the clear analysis and the regression coverage idea — much appreciated!

PR #2135, opened 2026-06-26
fix: replace "continue" followup with "..." to prevent unintended agent actions

Vann-Dev:
## Problem

Kiro's tool-calling flow requires a non-empty user response after every tool invocation. To satisfy this constraint, `"continue"` was used as a filler response — however, this turns out to be semantically meaningful to the agent. The agent reads `"continue"` as an instruction to proceed freely, which can lead to destructive or unintended actions.

## Edge Case Example

> **User:** Prompts something that could result in a file deletion  
> **Agent:** Calls the relevant tool, awaits user follow-up  
> **User (auto):** `"continue"`  
> **Agent:** Proceeds to delete the file — **without explicit user permission**

The root issue is that `"continue"` is an affirmative command. It gives the agent a perceived green light to do whatever it deems necessary next.

## Fix

Swap `"continue"` → `"..."`

`"..."` is semantically neutral — it implies the user is waiting or thinking, not actively instructing. This subtle change prevents the agent from treating the placeholder as permission to proceed with potentially destructive operations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,67 @@
-# Unreleased
+# v0.5.50 (2026-08-05)
 
 ## Features
+- **Providers**: add TokenRouter (300+ models via OpenAI-compatible gateway) with
+  exact per-model pricing for 110 models and `reasoning_effort` thinking config
 - **Providers**: add Self-hosted STT / TTS / Embedding — point 9Router at your own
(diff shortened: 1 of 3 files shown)


Vann-Dev:
This edge cases usually happened when in Claude Code's plan mode where the `"continue"` part making the agent thinks they should execute the plan without user permissions

PR #1618, opened 2026-06-02
fix: update codex integration to 1.136-compatible headers and response events

trananhtung:
## Why this change
This is a compatibility-hardening patch for Codex integration paths in 9router, aligned with upstream Codex client conventions.

## External evidence from upstream Codex source (openai/codex)

- Upstream sets the default originator to `codex_cli_rs` in the HTTP client defaults:
  - `codex-rs/login/src/auth/default_client.rs:36` (`DEFAULT_ORIGINATOR: "codex_cli_rs"`)
  - `codex-rs/login/src/auth/default_client.rs:232-247` (default headers include `originator` and default `User-Agent`)
- Upstream `User-Agent` construction is built from originator + version:
  - `codex-rs/login/src/auth/default_client.rs:133-145` (`get_codex_user_agent`), with tests confirming it starts with `{originator}/`:
    - `codex-rs/login/src/auth/default_client_tests.rs:7-12`
- Upstream tests explicitly verify default header behavior:
  - `codex-rs/login/src/auth/default_client_tests.rs:71-83` (created client must send `originator` and `user-agent` headers matching above defaults)
- Upstream protocol includes `response.done` as a first-class event type in protocol mapping/parsing:
  - `codex-rs/codex-api/src/endpoint/realtime_websocket/protocol_v2.rs:70`
  - `codex-rs/codex-api/src/endpoint/realtime_websocket/methods.rs:1312-1323` (test for `parse_realtime_v2_response_done_event`)
- Upstream codex tooling also emits `response.done` in mock response streams:
  - `scripts/mock_responses_websocket_server.py:42-50`
- Upstream /models flow includes `client_version` query handling for model refresh:
  - `codex-rs/core/tests/suite/models_etag_responses.rs:149-156` (assert `client_version` query exists)
  - `codex-rs/model-provider/src/models_endpoint.rs:81-105` (passes `client_version` arg into client request)

## What this PR changes in 9router

- Sync Codex identity headers with upstream shape:
  - `open-sse/config/providers.js:74-76`
  - `open-sse/handlers/imageProviders/codex.js:5-8`
  - `src/app/api/providers/[id]/test/testUtils.js:28`
- Handle both `response.completed` and `response.done` across Codex stream conversion paths:
  - `open-sse/transformer/streamToJsonConverter.js:30`
  - `open-sse/translator/response/openai-responses.js:493`
  - `open-sse/handlers/chatCore/sseToJsonHandler.js:170-171`
- Keep unit test expectation aligned:
[...]

PR #2142, opened 2026-06-27
fix(responses): handle response.done terminal events

rifuki:
## Summary

- treat `response.done` as a terminal OpenAI Responses stream event
- keep emitting the OpenAI `data: [DONE]` sentinel for same-format Responses passthrough streams
- add a regression test for `response.done`

## Why

Codex/OpenAI Responses streams can terminate with:

```text
event: response.done
data: {"type":"response.done", ...}
```

9router already recognizes `response.done` in some older response conversion paths, but the current Responses passthrough terminal-event helper only accepts `response.completed`, `response.failed`, and `error`. When upstream sends `response.done`, the stream can be treated as incomplete and a synthetic `response.failed` may be emitted, which causes clients to surface errors like `stream closed before response.completed` / stream disconnected before completion.

This patch makes the same-format Responses passthrough path accept `response.done` as terminal, and restores the `[DONE]` sentinel for Responses passthrough streams so OpenAI-compatible SSE clients see a clean stream ending.

## Related scan

I checked the current open PRs/issues before opening this:

- #2089 is related to retryable/semantic Responses stream failures, but does not add `response.done` to the terminal event set and is currently conflicting.
- #2082 and #2112 cover Responses usage/text-format issues, not stream termination.
- #1618 previously mentioned `response.done`, but it was closed without merge and the current `master` terminal helper still does not include it.
- #1648 introduced the incomplete Responses stream test coverage, but was also closed without merge.

## Tests

```bash
./node_modules/.bin/vitest run --config tests/vitest.config.js \
  tests/unit/openai-responses-terminal-event.test.js \
  tests/unit/responses-abort-terminal.test.js \
  tests/unit/headroom-responses-format.test.js \
  tests/unit/codex-tool-normalization.test.js
```

Result: 4 files passed, 11 tests passed.

### DIFF
--- a/open-sse/utils/responsesStreamHelpers.js
+++ b/open-sse/utils/responsesStreamHelpers.js
@@ -5,6 +5,7 @@ import { formatSSE } from "./streamHelpers.js";
 // Responses API events that signal the stream has reached a terminal state
 const OPENAI_RESPONSES_TERMINAL_EVENTS = new Set([
   "response.completed",
+  "response.done",
   "response.failed",
   "error"
 ]);
--- a/open-sse/utils/stream.js
+++ b/open-sse/utils/stream.js
@@ -217,9 +217,11 @@ export function createSSEStream(options = {}) {
             sseEmittedCount++;
           }
 
-          // [DONE] not emitted in translate mode — some clients' SSE decoders
-          // fail to parse the OpenAI sentinel on Claude-format translated streams.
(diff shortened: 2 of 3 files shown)


decolua:
Thanks @rifuki! Reviewed and accepted — `response.done` correctly added to the Responses passthrough terminal-event set, and the `data: [DONE]` sentinel is restored with proper dedupe guards. Merging into master locally.

PR #2143, opened 2026-06-27
fix(github): drop trailing assistant prefill for Copilot chat

baslr:
## Problem

The newest Claude Desktop version ends its Anthropic request with an assistant
message (a prefill seed). The Anthropic API supports this, but GitHub Copilot's
`/chat/completions` endpoint does not and responds with a 400: 

> This model does not support assistant message prefill. The conversation must end with a user message.

For `claude-*` models routed through the **github** provider, the request is
translated claude→openai; the trailing assistant message survives translation
and triggers the 400.

## Fix

`sanitizeMessagesForChatCompletions` (the existing Copilot-specific message
adapter in the GitHub executor) now drops trailing assistant message(s) via a
new `dropTrailingAssistantPrefill` helper before dispatching to
`/chat/completions`.

Deliberately scoped to the GitHub executor only — not the shared translator:
real Anthropic/OpenAI endpoints honor prefill, so it must be preserved there.

- No-op when the conversation already ends with a user/tool message
- Never empties the array (keeps at least one message)

## Test plan

- [x] New regression test `tests/unit/github-prefill-sanitize.test.js` (6 cases): passing
- [x] Existing `github-responses-routing` test: passing (no regression)
- [x] End-to-end verified: `claudeToOpenAIRequest` → `sanitizeMessagesForChatCompletions` — conversation ends with`user` after sanitize
- [x] Live smoke test against real GitHub Copilot with Claude Desktop

-----
Human message: his was done with Opus 4.8
The fix runs locally from master (cce47dd809b1b01e23ebedeb78ad26e85bad429b).
Claude Desktop App runs again with this fix 🎊

### DIFF
--- a/open-sse/executors/github.js
+++ b/open-sse/executors/github.js
@@ -102,7 +102,24 @@ export class GithubExecutor extends BaseExecutor {
       return msg;
     });
 
+    // GitHub Copilot's /chat/completions rejects a conversation that ends with an
+    // assistant message ("does not support assistant message prefill. The conversation
+    // must end with a user message."). Anthropic clients such as Claude Desktop send a
+    // trailing assistant turn as a prefill seed, which the Anthropic API honors but
+    // Copilot does not — drop it so the request is accepted.
+    sanitized.messages = this.dropTrailingAssistantPrefill(sanitized.messages);
+
     return sanitized;
+  }
+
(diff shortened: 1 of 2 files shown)


baslr:
@decolua possible to merge this?
Thank you.

PR #2149, opened 2026-06-27
fix(cursor): enable OpenAI/OpenCode tool calling on Cursor route

DevEstacion:
## Summary

Wires Cursor's tool-name translation layer through `cursorToolMapping.js` so OpenAI-format and OpenCode-format clients can drive Cursor's native tool-call flow without getting 400'd by the upstream.

## What it does

- Adds a new `open-sse/utils/cursorToolMapping.js` module that:
  - Maps between OpenAI tool names (`bash`, `read_file`, etc.) and Cursor's expected native names.
  - Translates tool call argument shapes between the two formats.
  - Provides a fallback text-based parser for clients that emit native tool calls as prose.
- Wires the mapper into `open-sse/utils/cursorProtobuf.js` so the protobuf encoder sends the names Cursor expects.
- Adjusts `open-sse/executors/cursor.js` so the Cursor route doesn't drop tool definitions from the request body on the way out.

## Files

- `open-sse/utils/cursorToolMapping.js` (+204, new)
- `open-sse/utils/cursorProtobuf.js` (+56, modified — wires mapper into encodeToolResult)
- `open-sse/executors/cursor.js` (+14, modified)
- `tests/unit/cursor-tool-mapping.test.js` (+42, new — round-trip mapping + native-text parser)

## Why a dedicated module

Cursor's native tool-call protocol is positional and uses names that don't match the OpenAI ecosystem. Inlining the name/arg mapping into the protobuf encoder scattered the logic and made it hard to test. The dedicated module gives a single round-trippable surface with its own unit tests.

## Test plan

- `cd tests && ./node_modules/.bin/vitest run tests/unit/cursor-tool-mapping.test.js`
- Manual: send an OpenAI-format `tools: [...] + tool_calls` payload to a `cursor/*` route and verify the protobuf-encoded request reaches Cursor with the correct tool names.
- Manual: send a native Cursor-format payload (with prose tool calls) and verify the parser extracts them.

🤖 Generated with [opencode](https://opencode.ai)

### DIFF
--- a/open-sse/executors/cursor.js
+++ b/open-sse/executors/cursor.js
@@ -4,8 +4,10 @@ import { HTTP_STATUS } from "../config/runtimeConfig.js";
 import {
   generateCursorBody,
   parseConnectRPCFrame,
-  extractTextFromResponse
+  extractTextFromResponse,
+  parseNativeToolCallsFromText,
 } from "../utils/cursorProtobuf.js";
+import { shouldForceAgentMode } from "../utils/cursorToolMapping.js";
 import { buildCursorHeaders } from "../utils/cursorChecksum.js";
 import { estimateUsage } from "../utils/usageTracking.js";
 import { SSE_DONE, SSE_HEADERS } from "../utils/sseConstants.js";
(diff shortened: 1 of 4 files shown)


AMSeify:
This looks like the same Cursor AgentService empty-reply / hang cluster. Tracking the framing fix in https://github.com/decolua/9router/issues/4016 and PR https://github.com/decolua/9router/pull/4017. cc @decolua @anuragg-saxenaa

PR #2160, opened 2026-06-27
fix(kiro): strip reasoning tags from content stream (#2158)

### LINKED ISSUE #2158: Bug: `reasoning_content` leaks into `content` for `kr/*` Claude models (v0.5.12)
## Summary

On 9Router **0.5.12** (latest, `hasUpdate:false`), responses from `kr/`
Claude models that use thinking surface the model's reasoning **inside the
visible `content`** as a literal `<thinking>...</thinking>` block, instead of
being delivered only through the separate `reasoning_content` channel.

The earlier `content: null` variant (#1836, fixed in v0.5.8) is **not** what
I'm seeing — content is preserved correctly. This is the inverse: reasoning
that should live in `reasoning_content` is *also* emitted as part of the
answer text, so downstream clients render the chain-of-thought as if it were
the final answer.

## Environment

- **9Router version:** 0.5.12 (`GET /api/version` → `{"currentVersion":"0.5.12","latestVersion":"0.5.12","hasUpdate":false}`)
- **Deployment:** local, global install (`~/.bun/install/global/node_modules/9router/app`), Next.js `next-server v16.2.1`, Node v25.9.0, bound to `127.0.0.1:20128`
- **Model:** `kr/claude-opus-4.8` , `kr/claude-sonnet-4.5` 
- **Provider:** `kr`
- **Client:** Pi coding agent 0.80.2 (OpenAI-compatible chat completions, streaming)
- **Thinking:** enabled at the client (`thinking level = medium`)

## What I observed

When the client streams a request with thinking enabled, the assistant's
**final `content`** contains a literal reasoning block:

```
<thinking>
The user said "continue" but I already answered their question...
</thinking>

The AuditValidator class is defined in:
**File path:** src/nested/callers.ts
**Line:** 5
```

The client captured the reasoning with a field signature of
`reasoning_content`, confirming the upstream provider returns reasoning in the
dedicated `reasoning_content` field — yet the same reasoning text also appears
inline in `content`.

Two consequences for clients:

1. **CoT shown as answer.** The `<thinking>` block renders as if it were part
   of the model's reply, because it's embedded in `content` rather than kept
   in `reasoning_content`.
2. **Confabulated conversation history surfaces to the user.** The leaked
   reasoning includes the model inventing a prior turn (e.g. *"the user said
   continue"*) that never happened. That's model behavior, but it's only
   *visible to the end user* because the reasoning is being emitted into
   `content` instead of the hidden reasoning channel.

## Steps to reproduce

1. Use model `kr/claude-opus-4.8` through 9Router (OpenAI-compatible
   `/v1/chat/completions`).
2. Enable thinking on the request (client-side thinking level `medium`, i.e.
   the request asks the model to reason).
3. Send a single self-contained prompt, e.g.:
   `"Which file in this repo defines the AuditValidator class? Give me the file path and line."`
4. Inspect the assistant message: the `content` text includes a
   `<thinking>...</thinking>` block in addition to (or merged with) the final
   answer.

Notes that help isolate it:

- Reproduces even with a brand-new, single-message conversation (no history),
  so it is not the client replaying prior reasoning back into the prompt.
- Reproduces non-interactively (`--print`, streaming JSON output).

## Expected behavior

Reasoning should be delivered **only** in `reasoning_content` (OpenAI shape)
or as a dedicated `thinking` content block (Claude shape). The `content`
field should contain just the final answer, with no `<thinking>` markup.

## Actual behavior

`content` contains the model's reasoning as an inline `<thinking>` block,
duplicating / mixing the chain-of-thought into the user-visible answer.

## Hypothesis (please verify proxy-side)

For `kr/*` Claude models, the reasoning emitted by the upstream appears to be
written into `content` (with `<thinking>` tags) rather than being routed into
the separate `reasoning_content` / Claude `thinking` channel during format
conversion. This is adjacent to the conversion handling fixed in #1836 /
PR #1844 and discussed in #1189, but the symptom is different: content is no
longer dropped, but reasoning is no longer cleanly separated from it either.

I can't see the proxy-internal conversion, so this is a hypothesis from the
client side — happy to provide full streaming captures (request + raw
upstream response + client-visible response) if helpful.

## Related

- #1836 — Empty content when provider returns `reasoning_content` (claude↔openai), fixed in v0.5.8 / PR #1844. Different symptom (content dropped); this report is the inverse (reasoning leaks into content).
- #1189 — DeepSeek `reasoning_content` not preserved across turns (open). Same subsystem, different model family.

hamsa0x7:
## Summary

Fixes the issue where CodeWhisperer (kiro) `<thinking>` literal blocks leak into the SSE `content` stream.

## Problem

CodeWhisperer sometimes outputs literal `<thinking>...</thinking>` tags into the standard `assistantResponseEvent`. Because the actual reasoning payload is already successfully intercepted and routed via `reasoningContentEvent`, this leakage results in duplicate reasoning text in the final chunk stream.

## Changes

- **`open-sse/executors/kiro.js`:** Added state tracking (`inThinking`) to strip literal `<thinking>` tags during the `transformEventStreamToSSE` streaming phase.
- **`tests/unit/kiro-thinking-strip.test.js`:** Added unit tests verifying stripping logic and handling of split chunks across tag boundaries.

## Related Issues

Fixes #2158

### DIFF
--- a/open-sse/executors/kiro.js
+++ b/open-sse/executors/kiro.js
@@ -122,7 +122,8 @@ export class KiroExecutor extends BaseExecutor {
       hasReasoningContent: false,
       reasoningChunkCount: 0,
       toolCallIndex: 0,
-      seenToolIds: new Map()
+      seenToolIds: new Map(),
+      inThinking: false
     };
 
     const transformStream = new TransformStream({
@@ -159,7 +160,36 @@ export class KiroExecutor extends BaseExecutor {
 
           // Handle assistantResponseEvent
           if (eventType === "assistantResponseEvent" && event.payload?.content) {
-            const content = event.payload.content;
+            let content = event.payload.content;
+
+            // Kiro Claude models can leak <thinking> blocks into the content stream.
+            // We strip these literal tags to prevent duplication, as the reasoning 
+            // is already routed correctly via reasoningContentEvent.
+            if (state.inThinking) {
+              if (content.includes("</thinking>")) {
+                state.inThinking = false;
+                const after = content.split("</thinking>").slice(1).join("</thinking>");
+                content = after.startsWith("\n") ? after.substring(1) : after;
+              } else {
+                content = ""; // Drop entirely while inside thinking block
+              }
+            } else if (content.includes("<thinking>")) {
+              state.inThinking = true;
+              if (content.includes("</thinking>")) {
+                state.inThinking = false;
+                const before = content.split("<thinking>")[0];
+                const after = content.split("</thinking>").slice(1).join("</thinking>");
+                content = before + (after.startsWith("\n") ? after.substring(1) : after);
+              } else {
+                content = content.split("<thinking>")[0];
+              }
+            }
+
+            if (!content && state.hasReasoningContent) {
+              // If we stripped everything, skip emitting an empty content chunk
+              continue;
+            }
+
             state.totalContentLength += content.length;
 
             const chunk = {
--- a/tests/unit/kiro-thinking-strip.test.js
+++ b/tests/unit/kiro-thinking-strip.test.js
@@ -1,0 +1,124 @@
+import { describe, it, expect } from "vitest";
+import { KiroExecutor } from "../../open-sse/executors/kiro.js";
+
+function createMockFrame(eventType, payloadObj) {
+  const payloadStr = JSON.stringify(payloadObj);
+  const payloadBytes = new TextEncoder().encode(payloadStr);
+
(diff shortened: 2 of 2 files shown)


decolua:
Thanks @hamsa0x7! Merged into master as eff81b1 (your authorship preserved). The fix correctly strips leaked <thinking> tags from the content stream while keeping real reasoning routed via reasoningContentEvent. Appreciate the tests covering split chunks across tag boundaries.

PR #2169, opened 2026-06-27
fix(alicode): preserve cache_control in content blocks for DashScope providers (#2069)

### LINKED ISSUE #2069: [Bug] cache_control markers are stripped when routing to Alibaba DashScope (alicode) providers, causing extremely low cache hit rates
## Describe the bug

When using 9router as a proxy to access Alibaba DashScope models (e.g., qwen3.7-plus, qwen3-coder-plus) via the `alicode` or `alicode-intl` providers, prompt cache hit rates are extremely low or zero, even though DashScope natively supports explicit caching with `cache_control: {"type": "ephemeral"}`.

## To Reproduce

1. Configure 9router with an `alicode` or `alicode-intl` provider connection
2. Send a request with `cache_control` markers in the message content blocks:
   ```json
   {
     "model": "alicode/qwen3-coder-plus",
     "messages": [
       {
         "role": "system",
         "content": [
           {
             "type": "text",
             "text": "<large system prompt >1024 tokens>",
             "cache_control": {"type": "ephemeral"}
           }
         ]
       },
       {"role": "user", "content": "Hello"}
     ]
   }
   ```
3. Send the same request again (to trigger cache hit)
4. Check the response `usage.prompt_tokens_details.cached_tokens`

**Expected**: Second request should show `cached_tokens > 0`
**Actual**: `cached_tokens` is always 0

## Root Cause Analysis

The issue is in 9router's request translation pipeline. When translating from Claude format (source) to OpenAI format (target for alicode), the `cache_control` markers are systematically stripped at three points:

### Stripping Point 1: `filterToOpenAIFormat()` in `open-sse/translator/formats/openai.js:36`

```javascript
// When converting Claude content blocks to OpenAI format
const { signature, cache_control, ...cleanBlock } = block;
filteredContent.push(cleanBlock);  // ← cache_control is destructured away
```

### Stripping Point 2: `claudeToOpenAIRequest()` in `open-sse/translator/request/claude-to-openai.js:32-43`

System messages are converted from Claude's `body.system` array to OpenAI's `messages[0].role="system"` string, losing all `cache_control` metadata.

### Stripping Point 3: `prepareClaudeRequest()` only injects cache_control for Claude targets

The function `prepareClaudeRequest()` in `open-sse/translator/formats/claude.js:147-251` correctly re-injects `cache_control` markers, but **only when the target format is Claude**. For alicode/DashScope (OpenAI format target), this function is never called.

### Comparison: Why Claude endpoints work correctly

- **Claude target**: `prepareClaudeRequest()` is called → strips old cache_control → re-injects at correct positions → request has cache_control ✓
- **alicode target**: `filterToOpenAIFormat()` is called → strips all cache_control → **no code re-injects** → DashScope receives request without cache_control ✗

## DashScope Caching Capability

According to [Alibaba Cloud documentation](https://www.alibabacloud.com/help/en/model-studio/context-cache):
- **Explicit caching**: Supports `cache_control: {"type": "ephemeral"}` markers
- **Implicit caching**: Automatic caching of identical prefixes (limited effectiveness)
- **Minimum cache length**: 1024 tokens
- **Response format**: `usage.prompt_tokens_details.cached_tokens` reports cache hits
- **Supported models**: qwen3-coder-plus, qwen3.5-plus, and others

## 9router's Usage Tracking Already Handles Cache Reporting

The code in `open-sse/utils/usageTracking.js:199-207` correctly parses DashScope's `prompt_tokens_details.cached_tokens` response, so **if caching were enabled, 9router would report it correctly**. The problem is entirely on the request side where cache_control markers are stripped.

## Proposed Fix

The translation pipeline should preserve (not strip) `cache_control` markers when the target provider supports them. Since DashScope's OpenAI-compatible API natively supports `cache_control: {"type": "ephemeral"}`, the fix is to **pass through** these markers.

### Fix 1: Preserve cache_control in `filterToOpenAIFormat()`

`open-sse/translator/formats/openai.js`

```javascript
// Before (line 36)
const { signature, cache_control, ...cleanBlock } = block;
filteredContent.push(cleanBlock);

// After
const { signature, ...cleanBlock } = block;
if (block.cache_control) cleanBlock.cache_control = block.cache_control;
filteredContent.push(cleanBlock);
```

Apply the same change to line 43 for tool_result handling.

### Fix 2: Inject cache_control for alicode providers

Add cache_control injection logic in `DefaultExecutor.transformRequest()` (file: `open-sse/executors/default.js:84-96`) or create a dedicated `AlicodeExecutor`. Follow the injection strategy from `prepareClaudeRequest()`:
1. System message's last content block → `cache_control: {"type": "ephemeral"}`
2. Last assistant message's last content block → `cache_control: {"type": "ephemeral"}`
3. Last tool in tools array → `cache_control: {"type": "ephemeral"}`

### Fix 3: Keep system messages as content arrays

DashScope's cache_control requires content to be in array format (not plain strings). When the target provider supports cache_control, system messages should remain as content arrays instead of being converted to strings in `claudeToOpenAIRequest()`.

## Verification

After applying the fix:
1. Send two identical large requests (>1024 tokens) through 9router to alicode provider
2. Check that the second request's response includes `usage.prompt_tokens_details.cached_tokens > 0`
3. Verify in 9router's Request Logger (debug mode) that the target request body contains `cache_control` fields

## Environment

- 9router version: 0.5.8
- Provider: alicode / alicode-intl
- Models tested: qwen3-coder-plus, qwen3.5-plus, qwen3.7-plus
- Client: oh-my-pi (OMP) sending Claude-format requests to `/v1/messages`

## Additional Context

This issue affects any client that sends requests with `cache_control` markers to DashScope models via 9router. The fix should be provider-aware: only preserve/inject cache_control for providers that support it (like alicode), to avoid breaking providers that reject unknown fields.

anki1kr:
## Summary

Fixes #2069.

DashScope (`alicode` / `alicode-intl`) supports explicit `cache_control: { type: "ephemeral" }` markers on content blocks, enabling prompt caching. The default `filterToOpenAIFormat` was stripping `cache_control` unconditionally, causing every request to arrive at DashScope without cache markers and resulting in zero cache hits.

## Changes

- **`open-sse/providers/registry/alicode.js`** and **`alicode-intl.js`**: add `quirks: { preserveCacheControl: true }` to the transport config (same pattern as existing `dropOutputConfig`, `cloakToolsOnOAuth`, etc.).
- **`open-sse/translator/formats/openai.js`**: `filterToOpenAIFormat` now accepts an `opts` argument. When `opts.preserveCacheControl` is true, `cache_control` is kept on content blocks; `signature` is always stripped regardless.
- **`open-sse/translator/index.js`**: passes `{ preserveCacheControl: PROVIDERS[provider]?.quirks?.preserveCacheControl }` to `filterToOpenAIFormat`.
- **`tests/unit/alicode-cache-control-2069.test.js`**: 4 unit tests covering default stripping, opt-in preservation, always-strip signature, and no spurious addition.

## What is NOT changed

System-message caching (the `claudeToOpenAIRequest` system→string flattening at stripping point 2) is left for a follow-up — it requires verifying that DashScope accepts array-format system content with `cache_control`, which is outside standard OpenAI spec.

## Test

```
npx vitest run tests/unit/alicode-cache-control-2069.test.js
# 4/4 passed
```

### DIFF
--- a/open-sse/providers/registry/alicode-intl.js
+++ b/open-sse/providers/registry/alicode-intl.js
@@ -16,6 +16,7 @@ export default {
   transport: {
     baseUrl: "https://coding-intl.dashscope.aliyuncs.com/v1/chat/completions",
     headers: {},
+    quirks: { preserveCacheControl: true },
   },
   models: [
     { id: "qwen3.5-plus", name: "Qwen3.5 Plus" },
--- a/open-sse/providers/registry/alicode.js
+++ b/open-sse/providers/registry/alicode.js
@@ -16,6 +16,7 @@ export default {
   transport: {
     baseUrl: "https://coding.dashscope.aliyuncs.com/v1/chat/completions",
     headers: {},
+    quirks: { preserveCacheControl: true },
   },
   models: [
     { id: "qwen3.5-plus", name: "Qwen3.5 Plus" },
--- a/open-sse/translator/formats/openai.js
+++ b/open-sse/translator/formats/openai.js
@@ -6,42 +6,46 @@ export { VALID_OPENAI_CONTENT_TYPES, VALID_OPENAI_MESSAGE_TYPES };
 
 // Filter messages to OpenAI standard format
 // Remove: thinking, redacted_thinking, signature, and other non-OpenAI blocks
-export function filterToOpenAIFormat(body) {
+// opts.preserveCacheControl: keep cache_control on content blocks (e.g. for DashScope/alicode)
+export function filterToOpenAIFormat(body, opts = {}) {
   if (!body.messages || !Array.isArray(body.messages)) return body;
-  
+  const keepCache = !!opts.preserveCacheControl;
+
+  function stripBlock(block) {
+    const { signature, cache_control, ...rest } = block;
+    return keepCache && cache_control ? { ...rest, cache_control } : rest;
+  }
+
   body.messages = body.messages.map(msg => {
     // Normalize developer role to system (many providers don't support developer)
     if (msg.role === ROLE.DEVELOPER) msg = { ...msg, role: ROLE.SYSTEM };
-    
+
     // Keep tool messages as-is (OpenAI format)
     if (msg.role === ROLE.TOOL) return msg;
-    
+
     // Keep assistant messages with tool_calls as-is
     if (msg.role === ROLE.ASSISTANT && msg.tool_calls) return msg;
-    
+
     // Handle string content
     if (typeof msg.content === "string") return msg;
-    
+
     // Handle array content
     if (Array.isArray(msg.content)) {
       const filteredContent = [];
-      
+
       for (const block of msg.content) {
         // Skip thinking blocks
         if (block.type === CLAUDE_BLOCK.THINKING || block.type === CLAUDE_BLOCK.REDACTED_THINKING) continue;
-        
+
         // Only keep valid OpenAI content types
         if (VALID_OPENAI_CONTENT_TYPES.includes(block.type)) {
-          // Remove signature field if exists
-          const { signature, cache_control, ...cleanBlock } = block;
-          filteredContent.push(cleanBlock);
+          filteredContent.push(stripBlock(block));
         } else if (block.type === CLAUDE_BLOCK.TOOL_USE) {
           // Convert tool_use to tool_calls format (handled separately)
           continue;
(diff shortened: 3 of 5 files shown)


decolua:
Thanks @anki1kr! Merged into master. Clean opt-in quirk, signature still always stripped, other providers unaffected. 🙏

PR #2170, opened 2026-06-27
fix(codebuddy-cn): show bonus packs as one-time, not monthly-replenishing

whale9820:
## Problem

CodeBuddy CN ships two credit types behind its billing endpoint:

- **Refill packs** ("基础体验包" / labelled "Monthly", "Weekly", ...): a recurring allowance whose cycle resets monthly — "Reset in Xd" is correct.
- **Bonus packs** ("活动赠送包" / labelled "Bonus Pack N"): one-shot promotional credits that run a single cycle and expire for good (`CycleEndTime == DeductionEndTime`).

The dashboard rendered **both** with "Reset in Xd" / "Reset at ...", which made bonus packs look like they replenish every month. In reality "Bonus Pack 1" is one-time — once depleted it never comes back.

## Fix

Surface a `recurring` flag on each quota so the UI can distinguish a monthly refresh from a hard expiry:

1. **`open-sse/services/usage/codebuddy-cn.js`** — tag refill packs `recurring: true` and bonus packs `recurring: false`.
2. **`ProviderLimits/utils.js`** — add a `codebuddy-cn` case to `parseQuotaData` that forwards `recurring` (defaults to `true` when absent for back-compat).
3. **`QuotaTable.js` / `QuotaProgressBar.js` / `ProviderLimitCard.js`** — render "Expires in" / "Expires at" when `recurring === false`, "Reset in" / "Reset at" otherwise.

## Test

- `tests/unit/codebuddy-cn-bonus-recurring.test.js` — asserts `parseQuotaData("codebuddy-cn", ...)` forwards `recurring: false` for bonus packs and `true` for refills, and defaults to `true` when the flag is absent.

Before / after for a bonus pack row:

| | Before | After |
|---|---|---|
| Countdown | `in 12d 4h 30m` | `expires in 12d 4h 30m` |
| Timestamp | `Reset at Jul 15, ...` | `Expires at Jul 15, ...` |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Support Gemini native TTS generateContent endpoint — nguyenha935
 - Add missing zh-CN endpoint key label (i18n) — weimaozhen
 - CodeBuddy: only send reasoning params when client requests reasoning (#2071) — Rex
+- CodeBuddy CN: show one-shot bonus packs as expiring, not monthly-replenishing
 - Show custom provider models in combo picker — Sapto
 - Docker: add docker-compose.yml with headroom enabled by default — nitsuahlabs
 - Clarify token diagnostics vs provider billing (headroom, #1998) — Sutarto Jordan Chrisfivo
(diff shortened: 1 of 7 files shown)


decolua:
Thanks @whale9820 — clean fix. Bonus packs (CycleEndTime == DeductionEndTime) are genuinely one-shot, so wording them as "Expires in" instead of "Reset in" is correct. The `recurring !== false` default keeps every other provider back-compatible. Merged into master. 🙏

PR #2179, opened 2026-06-28
fix(headroom): skip unsafe responses tool history

### LINKED ISSUE #2132: [BUG] Token saver Headroom in version 0.5.12 ruins plan mode in Codex CLI
Token saver Headroom in version 0.5.12 makes Codex CLI never create plan in plan mode, in the version 0.5.8 still works fine.

Jordannst:
## Description

Fixes #2132 by preventing Headroom compression from running on OpenAI Responses requests that contain structured non-message history, such as tool calls, tool outputs, or reasoning summaries.

Headroom compression only understands OpenAI Chat Completions-style `messages`. For Responses-format requests, 9Router currently converts `body.input` into chat messages before sending it to Headroom. That is safe for plain text/message-only Responses input, but unsafe once the history contains Responses-native items.

## Issue

Codex / OpenAI Responses requests can include history entries like:

- `type: "function_call"`
- `type: "function_call_output"`
- `type: "reasoning"`
- grammar/custom tool definitions, for example `apply_patch`

Those entries are not plain chat messages. If Headroom compression runs on that mixed history, the structured Responses items can be collapsed into normal OpenAI messages. That loses the Responses contract and can break downstream providers/tools because `body.input` no longer preserves the original tool/reasoning structure.

## Findings

- The existing Responses compression path was already fixed to keep `body.input` Responses-shaped for message-only input (#1998).
- The unsafe case is different: `body.input` can be Responses-shaped but still contain non-message item types.
- Headroom cannot preserve those items because its `/v1/compress` API returns Chat Completions-style `messages`, not Responses items.
- Best safe behavior is to skip compression for mixed Responses history instead of attempting a lossy conversion.

## Changes

- Add a guard for OpenAI Responses compression:
  - if `body.input` contains an object item with string `type` other than `message`, skip Headroom compression.
- Record a diagnostic reason:
  - `skipped: openai-responses tool/reasoning input is not safe to compress`
- Leave the original request body untouched when skipped.
- Add regression test coverage for Responses tool/reasoning history.

## Why skip instead of partially compress?

Partial compression would need to split message content from tool/reasoning entries and then reconstruct valid Responses history after Headroom returns chat messages. That is riskier and easy to get wrong, especially around `call_id`, reasoning summaries, and custom/grammar tool calls.

This patch chooses the conservative behavior: preserve correctness over token savings. Message-only Responses requests can still use Headroom compression.

## Verification

```bash
git diff --check
npx vitest run tests/unit/headroom-responses-format.test.js
```

Result:

```text
Test Files  1 passed (1)
Tests       2 passed (2)
```

Closes #2132

### DIFF
--- a/open-sse/rtk/headroom.js
+++ b/open-sse/rtk/headroom.js
@@ -73,6 +73,14 @@ function maskEndpoint(endpoint) {
   }
 }
 
+function hasUnsafeResponsesInputForCompression(body) {
+  if (!Array.isArray(body?.input)) return false;
+  return body.input.some((item) => {
+    if (!item || typeof item !== "object" || Array.isArray(item)) return false;
+    return typeof item.type === "string" && item.type !== "message";
+  });
+}
+
 // POST messages to Headroom /v1/compress; returns compressed messages + stats or null.
 async function callCompress(url, messages, model, timeoutMs, compressUserMessages, diagnostics) {
   const endpoint = buildCompressEndpoint(url);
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @Jordannst! Reviewed and merged into master (conservative skip preserves the Responses contract for tool/reasoning history). Closes #2132.

Issue #2180, opened 2026-06-28
MITM not exists in the 0.5.12 version


may215:
Where is the MITM option?

fajarkurnia0388:
<img width="1881" height="524" alt="Image" src="https://github.com/user-attachments/assets/06c9042e-156a-44cb-8d80-1675fc03e5ae" />

http://localhost:20128/dashboard/cli-tools

http://localhost:20128/dashboard/mitm

Issue #2167, opened 2026-06-27
nvidia/minimaxai/minimax-m3 fails via 9router but works perfectly via direct NVIDIA API


ODMDHJJ:
I am encountering an internal server error when trying to use the nvidia/minimaxai/minimax-m3 model hosted on NVIDIA NIM through 9router.

The issue seems specific to how 9router processes or forwards requests/streams for this particular model. If I bypass 9router and send the exact same request directly to the NVIDIA API, it works flawlessly. Furthermore, other models (like z-ai/glm-5.1 or qwen) work completely fine through 9router with similar payloads.

suryacagur:
resolved at PR #2185

Issue #2151, opened 2026-06-27
No credentials for provider: openai


nicemkbk:
version: [v0.5.12]
[14:07:23] 📥 POST /api/v1/search | openai
[14:07:23] ℹ️ [ROUTING] Provider: openai
[14:07:23] ❌ [AUTH] No credentials for provider: openai

nicemkbk:
unexpected status 404 Not Found: No active credentials for provider: openai, url: xxx

<img width="4244" height="1356" alt="Image" src="https://github.com/user-attachments/assets/89c044d9-2e60-41b0-8d54-8478f1b128a8" />

suryacagur:
im working on it

rognales:
this is a known issue? i got it working in the CLI, but not the macapp

PR #2185, opened 2026-06-28
fix: resolve provider bugs, add combo UX enhancements, improve CLI, and update models (#2151, #2158, #2138, #2114, #2117, #2167, #2077, #2180)

suryacagur:
## Summary

This PR bundles 13 commits covering provider fixes, combo creation UX, CLI improvements, and proxy/SSE fixes.

### Provider fixes

- **OpenAI search (#2151)**: Added missing `endpoint` field to `searchViaChat` config — `chatSearch.js` was returning an empty string for `searchEndpoint('openai')`, causing `fetch('')` to fail even with valid credentials.

- **Kiro reasoning leak (#2158)**: Stripped inline `<think>/<thinking>` tags from assistant content in two places — the Kiro executor (strip before emitting as `delta.content`) and the Claude-to-OpenAI translator (stop emitting `<think>` literals). Kiro returns thinking text in both `reasoningContentEvent` AND wrapped in tags inside `assistantResponseEvent.content`.

- **Qoder model listing (#2138)**: Recovered `userId` from the userinfo API when the token-poll endpoint no longer reliably returns it. Also made the model list parser resilient to multiple response shapes (`chat`, `data`, `models`, `result`, flat array).

### Combo creation UX (#2114)

- **Auto-populate**: First available (non-disabled) model from connected providers is auto-selected when the create-combo modal opens — fewer clicks for users.
- **Disabled model badge**: Disabled models now appear greyed-out with a "Disabled" badge and strikethrough styling instead of being silently hidden. Also skips inactive provider connections.

### CLI & infra

- **Process rename (#2117)**: Next.js standalone server process now appears as "9router next-server (v0.5.12)" instead of generic "next-server (v16.2.1)" — uses a `--require` preload hook. Updated PID-detection logic accordingly.
- **Package.json cleanup**: Removed unused scripts (`install:all`, `build:all`, `clean:*`), unused dependency (`rimraf`), and unnecessary `allowScripts` entries.
- **Anthropic models**: Updated model registry with latest Claude versions (Opus 4.8/4.7/4.6/4.5/4.1, Sonnet 4.6/4.5, Haiku 4.5).

### Proxy/SSE fixes

- **NVIDIA NIM (#2167)**: Set `thinkingFormat: "openai"` for models served through NVIDIA's OpenAI-compatible API. Pattern-based capability resolution was assigning `thinkingFormat: "minimax"`, which injected `body.thinking = { type: "adaptive" }` — a MiniMax-native format unknown to NVIDIA, causing 502 errors.
- **GLM 5.2 deferred tool name (#2077)**: GLM 5.2 streams tool call `id` and `function.name` across separate SSE delta chunks. The response translator now defers `content_block_start` until the tool name arrives, preventing "No such tool: " rejections.
- **MITM sidebar (#2180)**: Restored MITM Proxy navigation link that was accidentally removed when the Token Saver page was added.

- **Reasoning effort clamp**: Clamp `reasoning_effort` value `"max"` to `"xhigh"` for OpenAI-compatible endpoints. `toLevel()` can return `"max"` from budget-based or level-based config, but OpenAI API only accepts: `none`, `minimal`, `low`, `medium`, `high`, `xhigh`. Sending `"max"` caused 400 errors with `"Invalid value: 'max'"`.


asyncDreams:
can confirm this fixes the nvidia - minimax m3 issue https://github.com/decolua/9router/issues/2167

suryacagur:
> can confirm this fixes the nvidia - minimax m3 issue #2167

done

Issue #2047, opened 2026-06-24
[BUG] 0.5.8 is incompatible with Minimax‑M3


torgitor:
After upgrading to 0.5.8, Minimax‑M3 produces no output, while the upstream vendor’s output remains normal. Reverting to 0.5.4 (the web‑UI still shows version 0.5.3) restores normal operation.

When tested with curl, setting stream: true only returns data: [DONE], while stream: false correctly returns the output.
```shell
curl --request POST \
  --url http://localhost:20128/v1/chat/completions \
  --header 'Authorization: Bearer api-key' \
  --header 'content-type: application/json' \
  --data '{"stream":true,"model":"minimax-cn/MiniMax-M3","messages":[{"role":"system","content":""},{"role":"user","content":"ping"}]}'
```
`"stream":true`
<img width="1911" height="188" alt="Image" src="https://github.com/user-attachments/assets/782d9ae6-94a3-4727-b409-768ebe22ed9d" />
`"stream":false`
<img width="2121" height="234" alt="Image" src="https://github.com/user-attachments/assets/2a7323f4-0357-4ec3-a56b-0520274faf93" />

Pieenut:
up , i faced the same issue as well. I suspect it have something to do with the claude to openai translator or viceversa.

amirmufid-eiger:
up

torgitor:
I have verified that this patch is effective. Thank you very much for your help. I look forward to the official merge of the PR.

Issue #1939, opened 2026-06-20
[ERROR] [400] when using minimax token plan


Pieenut:
[15:09:10] [PENDING] END (ERROR) | provider=minimax | model=MiniMax-M3
[ERROR] [400]: {"type":"error","error":{"type":"bad_request_error","message":"invalid params, invalid tool type: (2013)","http_code":"400"},"request_id":"0685e0161ed0c13ca93840e4599e6957"}
❌ minimax [400]: [400]: {"type":"error","error":{"type":"bad_request_error","message":"invalid params, invalid tool t

this keeps happening on my 9router, this doesnt happen before i update the 9router. I have been using this for almost a month. When i try to test the connection manually it says valid.

burak-ayd:
Up. The issue persists in version 0.5.15. I don't experience this issue in version 0.5.4.

Pieenut:
This issue still persist even in the latest version 0.5.18. Which made me have to revert back to 0.5.4. @decolua pls merge the relevant fix

PR #2196, opened 2026-06-28
fix(tools): default type:"custom" on Claude-format tools when missing

### LINKED ISSUE #2195: [Bug] HTTP 400 from strict Anthropic-compatible gateways (MiniMax) when Claude tools omit `type` field
## Bug

Strict Claude-compatible gateways can reject tools whose `type` discriminator is missing or falsy. Upstream v0.5.30 still dispatches final Claude-format tools without a general default, so the schema gap remains.

## Expected fix

For final Claude-format requests, default only missing/falsy tool types to `"custom"` using `{ ...tool, type: "custom" }`. Preserve truthy built-in types and object identity for untouched tools. PR #2196 implements this in `defaultClaudeToolType()` and passes 11/11 cases plus the 126/126 production build on `9845a17`.

## MiniMax scope clarification

#1939 and #2047 reported the normalization effective. Newer reports #2435 and #2509 show that MiniMax-M3 also has a distinct transport/endpoint-selection defect. Competing PRs #2463, #2525, and #2533 address that route mismatch. The general Claude-schema normalization is complementary and must not be described as the sole root-cause fix for every MiniMax-M3 error 2013.

PR #2196 is closed and should not be reopened automatically.

warelik:
Fixes #2195
Related: #1939 (same MiniMax error 2013 `invalid tool type`), #2047 (MiniMax-M3 incompatibility)

## Summary

When proxying Claude-format tool definitions through strict Anthropic-compatible gateways (e.g. the MiniMax Anthropic-compatible endpoint), requests are rejected with `400 Bad Request` (MiniMax error code `2013`) whenever individual tools in the `tools` array arrive without an explicit `type` field.

Anthropic's Claude tool schema documents `type` as a required discriminator (e.g. `"custom"`, `"computer_20241022"`, `"bash_20241022"`). Legacy clients and older SDK versions frequently omit this field, relying on Anthropic's first-party API to infer `"custom"`. While Anthropic's own API tolerates the omission, strict Anthropic-compatible gateways such as MiniMax enforce the documented schema and reject the payload.

This PR normalizes `tools[].type` for every Claude-format request before dispatch, ensuring 9router emits a valid Anthropic Messages payload regardless of the originating client.

## Changes

- **`open-sse/translator/concerns/toolCall.js`**: New exported function `defaultClaudeToolType(tools)` — maps tools without a truthy `type` to `{ ...tool, type: "custom" }`. Uses optional chaining (`tool?.type`) so `null`/`undefined` array entries don't crash. Spread-first ensures falsy `type` values (`null`, `undefined`, `""`) don't survive the override — the original `{ type: "custom", ...tool }` order let `type: null` overwrite the default (confirmed with test cases). Tools that already carry a truthy `type` are returned by reference (no clone).

- **`open-sse/handlers/chatCore.js`**: Calls `defaultClaudeToolType(translatedBody.tools)` when `finalFormat === "claude"`, positioned immediately after the TTS special case and before the RTK compress step. Import line extended to include `defaultClaudeToolType` alongside the existing `salvageOrphanedToolResults` and `fixMissingToolResponses` imports.

- **`open-sse/utils/claudeToolTypeSelfCheck.mjs`**: New 11/11 no-framework assert self-check covering: no-type, type:null, type:undefined, type:"", built-in pass-through (computer_use), already-custom, mixed (built-in + function + web_search), non-array input, empty array, non-mutation by reference, and null-array-entry (optional chaining guard).

## Spread order fix

The original inline implementation `tool.type ? tool : { type: "custom", ...tool }` had a subtle bug: when `tool.type` was a falsy value (`null`, `undefined`, `""`), the spread `{ ...tool }` would overwrite `type: "custom"` with the falsy value, reintroducing the exact condition that causes MiniMax to reject the payload.

```js
// Buggy: type:null survives the spread
{ type: "custom", ...tool }  // → { type: null, name: "foo", ... }

// Fixed: type:"custom" always wins (spread first, override last)
{ ...tool, type: "custom" }  // → { type: "custom", name: "foo", ... }
```

This affects the passthrough path (Claude client → Claude/anthropic-compatible provider) where legacy clients may send `"type": null` in JSON. The translated path (`openai-to-claude.js`) was not affected as it does not generate a `type` property for custom tools.

## Related issues

- **#1939** — `[ERROR] [400] when using minimax token plan` — reports the exact same MiniMax error `invalid tool type: (2013)`. This PR fixes that root cause.
- **#2047** — `[BUG] 0.5.8 is incompatible with Minimax-M3` — may be related (streaming issue, not tool type, but same provider/gateway).

## Notes

This is a general schema-conformance fix, not provider-specific. It benefits any Anthropic-compatible gateway that enforces the documented tool schema. The placement was chosen to run after translation and after the TTS special case (which can delete `translatedBody.tools`), and before the RTK/headroom compression passes that operate on the final outbound body.

## Verification

```bash
node open-sse/utils/claudeToolTypeSelfCheck.mjs   # 11/11
npm run build                                      # 125/125 pages
```

Applies cleanly and independently on a fresh `origin/HEAD` checkout.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -26,6 +26,7 @@ import { compressWithHeadroom, formatHeadroomLog, formatHeadroomSizeLog, isHeadr
 import { getCapabilitiesForModel } from "../providers/capabilities.js";
 import { stripUnsupportedModalities } from "../translator/concerns/modality.js";
 import { prefetchRemoteImages } from "../translator/concerns/prefetch.js";
+import { defaultClaudeToolType } from "../translator/concerns/toolCall.js";
 
 /**
  * Core chat handler - shared between SSE and Worker
@@ -154,6 +155,12 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   if (getModelType(alias, model) === "tts" && translatedBody.messages) {
     translatedBody.messages = translatedBody.messages.filter(msg => msg.role !== "tool");
     delete translatedBody.tools;
+  }
+
+  // Claude tool schema requires `type` to be explicitly set; strict gateways (e.g., MiniMax)
+  // reject legacy payloads that omit it with HTTP 400. Default to "custom" when missing.
+  if (finalFormat === "claude" && Array.isArray(translatedBody.tools)) {
+    translatedBody.tools = defaultClaudeToolType(translatedBody.tools);
   }
 
   // RTK: compress tool_result content
--- a/open-sse/translator/concerns/toolCall.js
+++ b/open-sse/translator/concerns/toolCall.js
@@ -151,3 +151,16 @@ export function fixMissingToolResponses(body) {
   return body;
 }
 
+// Default `type: "custom"` on Claude-format tools that arrive without one.
+// Anthropic's Claude tool schema requires `type` to be explicitly set; strict gateways
+// (e.g., MiniMax Anthropic-compatible endpoint, error 2013) reject legacy payloads that
+// omit it with HTTP 400. Tools that already carry a truthy `type` (e.g., `computer_use`,
+// `bash`, `web_search_20250305`) are passed through untouched.
+//
+// Spread order matters: `{ ...tool, type: "custom" }` (spread first, override last)
+// ensures that falsy `type` values (null, undefined, "") in the original tool don't
+// overwrite the default. `{ type: "custom", ...tool }` would let `type: null` survive.
+export function defaultClaudeToolType(tools) {
+  if (!Array.isArray(tools)) return tools;
+  return tools.map(tool => tool?.type ? tool : { ...tool, type: "custom" });
+}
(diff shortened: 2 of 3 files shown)


warelik:
Rebuilt this branch as a single clean commit on the latest `master` (v0.5.18) to include the full schema normalization logic and self-check:

- Extracted `defaultClaudeToolType(tools)` into `toolCall.js`
- Applied spread-first override (`{ ...tool, type: 'custom' }`) to ensure falsy types (`null`, `undefined`, `""`) don't survive
- Calls normalization for Claude-format outbound requests in `chatCore.js`
- Added 11/11 zero-dependency self-check covering all edge cases (mixed schemas, built-ins, nulls)

Files changed: `chatCore.js`, `toolCall.js`, `claudeToolTypeSelfCheck.mjs`.

decolua:
Thanks @warelik for the contribution! Reviewed and accepted — applying it to master. 🙏

warelik:
## v0.5.30 re-audit

Re-audited this fix against upstream v0.5.30 (`9845a17`). The general Claude-schema gap is still present: final Claude-format tools are dispatched without defaulting a missing/falsy discriminator, and the standalone patch still passes its 11/11 self-check and the 126/126 production build.

Scope clarification: the newer MiniMax-M3 reports #2435 and #2509 also expose a distinct transport/endpoint-selection defect, addressed by competing PRs #2463, #2525, and #2533. This closed PR should therefore be read as a general strict Claude-compatible schema fix, not as the sole root-cause fix for every MiniMax-M3 route.

I am not reopening this PR. A clean one-commit v0.5.30 candidate is available if maintainers want the general normalization as a follow-up.

PR #2200, opened 2026-06-29
fix: strip 'deprecated' from Antigravity tool schemas before Gemini

Yudhistira-Official:
## Summary

Adds `deprecated` to `UNSUPPORTED_SCHEMA_CONSTRAINTS` in `open-sse/translator/formats/gemini.js` so it gets stripped from nested tool schemas before forwarding to Gemini.

This is a follow-up to #1841 (which added `optional`). Without this fix, Gemini rejects requests with `INVALID_ARGUMENT`:

```
Unknown name "deprecated" at 'request.tools[0].function_declarations[25].parameters.properties[1].value': Cannot find field.
```

## Test

```bash
```


decolua:
Thanks @Yudhistira-Official! Merged into master. Good follow-up to #1841 — 'deprecated' now stripped alongside 'optional' before forwarding to Gemini.

PR #2223, opened 2026-06-29
fix(antigravity): preserve Claude tool delta index

### LINKED ISSUE #2218: bug(translator): missing `index` field in Anthropic-format tool_call stream chunks (input_json_delta) for Antigravity/Gemini
### Describe the bug

When using the `ag/gemini-pro-agent` model via 9router with a strict client (like ZCode / Cursor) that expects Anthropic-format streaming tool calls, the stream chunks are missing the required `index` field. 

This causes the client to crash and abort the turn with a type validation error when a tool call starts.

### To Reproduce

1. Connect a strict client (like ZCode) to 9router.
2. Select `ag/gemini-pro-agent` (or potentially other models relying on the same translator path).
3. Ask the agent to perform an action requiring a tool (e.g., executing a bash command or reading a file).
4. Client crashes with the following error parsing the SSE stream:

```json
Turn execution failed
Type validation failed: Value: {"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"{\"command\":\"git restore vite.shared.ts\"}"}}. Error message: [ { "expected": "number", "code": "invalid_type", "path": [ "index" ], "message": "Invalid input: expected number, received undefined" } ]
[ { "expected": "number", "code": "invalid_type", "path": [ "index" ], "message": "Invalid input: expected number, received undefined" } ]
provider=326ee3e9-e026-467b-aead-8e733b434b63 model=ag/gemini-pro-agent request=77013f49-99c9-449a-a631-cc489e031f8b reason=unknown retryable=false
```

### Expected behavior

The translated Anthropic-format `content_block_delta` chunk should include the `index` field to identify which tool call block is being streamed, complying with the Anthropic streaming API specification. 

Expected payload structure should look more like:
```json
{
  "type": "content_block_delta",
  "index": 0,
  "delta": {
    "type": "input_json_delta",
    "partial_json": "..."
  }
}
```

### Environment
- **9router version**: v0.5.15
- **Client**: ZCode
- **Model**: ag/gemini-pro-agent

Jordannst:
## Issue

When Antigravity/Gemini emits a tool call and the client expects Claude/Anthropic streaming format, the response can contain a malformed Claude SSE event.

Observed bad shape:

```json
{"type":"content_block_delta","delta":{"type":"input_json_delta","partial_json":"{...}"}}
```

For Anthropic tool argument deltas, `content_block_delta.index` is required because it points to the `tool_use` content block started earlier in the stream. Without it, Claude-compatible clients cannot associate the JSON argument delta with the correct content block.

This affects the response translation path:

```txt
Antigravity/Gemini response
→ gemini-to-openai.js
→ openai-to-claude.js
→ Claude SSE client
```

## Findings / Root cause

The bug was not caused by `index` being deleted later in the stream layer. It was caused by shared translator state being reused for two different meanings.

`gemini-to-openai.js` stored Gemini/OpenAI-shaped tool-call bookkeeping in the shared `state.toolCalls` map:

```js
state.toolCalls.set(toolCallIndex, toolCall);
```

Those entries look like OpenAI tool calls:

```js
{
  id,
  index,
  type: "function",
  function: { name, arguments }
}
```

But the downstream `openai-to-claude.js` translator also uses `state.toolCalls`. In that translator, the map is expected to contain Claude content block metadata:

```js
{
  id,
  name,
  blockIndex
}
```

Because Gemini had already populated `state.toolCalls`, OpenAI→Claude saw the tool-call index as already initialized:

```js
if (tc.id && !state.toolCalls.has(idx)) {
  // create Claude tool_use block metadata with blockIndex
}
```

That branch was skipped, so no Claude `blockIndex` was created. Later, when emitting the buffered tool arguments:

```js
index: toolInfo.blockIndex
```

`toolInfo.blockIndex` was `undefined`. JSON serialization then omitted `index`, producing the malformed Anthropic event.

## Fix

Keep Gemini-specific bookkeeping out of the shared `state.toolCalls` map.

Changes:

- Removed Gemini response translation writes to `state.toolCalls`.
- Added `state.geminiToolCallCount` to track whether Gemini emitted any function calls.
- Kept existing finish-reason behavior: Gemini `STOP` is still converted to OpenAI `tool_calls` when a function call was emitted.
- Left `state.toolCalls` available for `openai-to-claude.js` to create and store Claude block metadata safely.

Relevant change:

```diff
- state.toolCalls.set(toolCallIndex, toolCall);
+ state.geminiToolCallCount = (state.geminiToolCallCount || 0) + 1;
```

and:

```diff
- if (finishReason === OPENAI_FINISH.STOP && state.toolCalls.size > 0) {
+ if (finishReason === OPENAI_FINISH.STOP && state.geminiToolCallCount > 0) {
    finishReason = OPENAI_FINISH.TOOL_CALLS;
  }
```

## Tests

Added regression coverage for the exact failing translation path:

- `Antigravity → Claude` tool call response emits `content_block_delta` with Anthropic `index`.
- The emitted `input_json_delta.partial_json` still contains the expected tool arguments.

Also added a small guard for the existing direct `Kiro → Claude` path so its `input_json_delta` continues to include `index`.

Verification command:

```sh
npx vitest run tests/translator/bugs-antigravity.test.js tests/translator/claude-kiro-direct.test.js tests/translator/golden-response-stream.test.js
```

Result:

```txt
Test Files  3 passed (3)
Tests       20 passed | 1 expected fail (21)
```

## Impact

This is a narrow state-isolation fix. It does not change Gemini/OpenAI tool-call output shape; it only prevents Gemini-side bookkeeping from corrupting the downstream Claude translator state.

Closes #2218

### DIFF
--- a/open-sse/translator/response/gemini-to-openai.js
+++ b/open-sse/translator/response/gemini-to-openai.js
@@ -25,7 +25,10 @@ function emitFunctionCall(functionCall, state) {
     type: OPENAI_BLOCK.FUNCTION,
     function: { name: fcName, arguments: JSON.stringify(fcArgs) },
   };
-  state.toolCalls.set(toolCallIndex, toolCall);
+  // Keep Gemini bookkeeping separate from the shared translator state.toolCalls map.
+  // The downstream OpenAI→Claude translator uses state.toolCalls for Claude block
+  // metadata; pre-populating it here makes Anthropic tool deltas lose index.
+  state.geminiToolCallCount = (state.geminiToolCallCount || 0) + 1;
   return buildChunk(chunkMeta(state), { tool_calls: [toolCall] }, null);
 }
 
@@ -46,6 +49,7 @@ export function geminiToOpenAIResponse(chunk, state) {
     state.messageId = response.responseId || `msg_${Date.now()}`;
     state.model = response.modelVersion || "gemini";
     state.functionIndex = 0;
+    state.geminiToolCallCount = 0;
     results.push(buildChunk(chunkMeta(state), { role: ROLE.ASSISTANT }, null));
   }
 
@@ -117,7 +121,7 @@ export function geminiToOpenAIResponse(chunk, state) {
   // Finish reason - include usage in final chunk
   if (candidate.finishReason) {
     let finishReason = toOpenAIFinish(candidate.finishReason, "gemini");
-    if (finishReason === OPENAI_FINISH.STOP && state.toolCalls.size > 0) {
+    if (finishReason === OPENAI_FINISH.STOP && state.geminiToolCallCount > 0) {
       finishReason = OPENAI_FINISH.TOOL_CALLS;
     }
     
--- a/tests/translator/bugs-antigravity.test.js
+++ b/tests/translator/bugs-antigravity.test.js
@@ -1,7 +1,7 @@
 // Real Antigravity-MITM requests (Gemini-internal: { request: { contents, ... } }) → OpenAI.
 import { describe, it, expect } from "vitest";
 import "./registerAll.js";
-import { translateRequest } from "../../open-sse/translator/index.js";
+import { translateRequest, translateResponse, initState } from "../../open-sse/translator/index.js";
 import { FORMATS } from "../../open-sse/translator/formats.js";
 import { AntigravityExecutor } from "../../open-sse/executors/antigravity.js";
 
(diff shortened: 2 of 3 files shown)


decolua:
Thanks @Jordannst for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2228, opened 2026-06-29
fix(mitm): generate root ca on server startup

### LINKED ISSUE #2224: MITM cannot initialize on Windows because Root CA is never generated (rootCA.key missing)
# Title

MITM cannot initialize on Windows because Root CA is never generated (rootCA.key missing)

## Environment

* OS: Windows 10 Pro 22H2 (Build 19045.7291)
* Node.js: (fill with `node -v`)
* Package Manager: pnpm 11.7.0
* 9Router: 0.5.15

## Steps to Reproduce

1. Install 9Router globally:

   ```bash
   pnpm add -g 9router@0.5.15
   ```
2. Delete `%APPDATA%\9router`
3. Run:

   ```bash
   9router
   ```
4. Open `http://localhost:20128`
5. Go to **MITM Proxy**

## Expected Behavior

The application should generate:

* rootCA.key
* rootCA.crt

inside:

```
%APPDATA%\9router\mitm
```

and allow the MITM server to start.

## Actual Behavior

The MITM page shows:

> Administrator required

The Start Server button is disabled.

The folder:

```
%APPDATA%\9router\mitm
```

contains only:

```
aliases.json
```

Running the runtime manually:

```bash
node %APPDATA%\9router\runtime\mitm\server.js
```

returns:

```
[MITM] Root CA not found:
...AppData\Roaming\9router\mitm\rootCA.key
```

## Investigation

The bundled `server.js` contains a `generateRootCA()` function.

However, the startup path exits when `rootCA.key` is missing instead of generating it first.

The bundle exports:

* generateRootCA
* loadRootCA

but startup immediately attempts to read `rootCA.key` and exits with:

```
Root CA not found
```

It appears `generateRootCA()` is never invoked during initialization.

## Additional Notes

* Running as Administrator
* `certutil -key` works
* Runtime files are present
* Reinstalling 9Router and deleting `%APPDATA%\9router` does not resolve the issue

Jordannst:
## Description

Fixes #2224.

Direct MITM server startup could fail before the Root CA was ever created because `server.js` read `rootCA.key` / `rootCA.crt` immediately and exited when either file was missing. That bypasses the `manager.js` setup path that normally generates the CA.

## Changes

- Generate the MITM Root CA from `server.js` when `rootCA.key` or `rootCA.crt` is missing.
- Make `generateRootCA()` / `generateCert()` synchronous, matching their filesystem work and avoiding a startup race before `readFileSync`.
- Add a targeted unit test proving Root CA files are created synchronously for direct server startup.

## Verification

```sh
./node_modules/.bin/vitest run tests/unit/mitm-root-ca.test.js
node --check src/mitm/server.js && node --check src/mitm/cert/rootCA.js && node --check src/mitm/cert/generate.js
git diff --check
```

Also ran:

```sh
./node_modules/.bin/vitest run tests/unit/mitm-root-ca.test.js tests/unit/antigravity-mitm.test.js
```

The new MITM Root CA test passed. `tests/unit/antigravity-mitm.test.js` has an unrelated existing failure on the out-of-box Antigravity mandatory model assertion (`expected undefined to be true`).

### DIFF
--- a/src/mitm/cert/generate.js
+++ b/src/mitm/cert/generate.js
@@ -7,8 +7,8 @@ const { generateRootCA, loadRootCA, generateLeafCert } = require("./rootCA");
  * Generate Root CA certificate (one-time setup)
  * This replaces the old static wildcard cert approach
  */
-async function generateCert() {
-  return await generateRootCA();
+function generateCert() {
+  return generateRootCA();
 }
 
 /**
--- a/src/mitm/cert/rootCA.js
+++ b/src/mitm/cert/rootCA.js
@@ -23,7 +23,7 @@ function isCertExpired(certPath) {
  * Generate Root CA certificate (only once, auto-regenerate if expired)
  * This Root CA will sign all dynamic leaf certificates
  */
-async function generateRootCA() {
+function generateRootCA() {
   const exists = fs.existsSync(ROOT_CA_KEY_PATH) && fs.existsSync(ROOT_CA_CERT_PATH);
   if (exists && !isCertExpired(ROOT_CA_CERT_PATH)) {
     console.log("✅ Root CA already exists");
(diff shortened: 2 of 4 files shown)


decolua:
Thanks @Jordannst for the contribution! Reviewed and accepted — applied to master. 🙏

PR #2230, opened 2026-06-30
fix(codebuddy-cn): strip empty tool_calls arrays to preserve reasoning

zmf963:
CodeBuddy CN API includes `"tool_calls": []` (empty array) in every SSE streaming delta. @ai-sdk/openai-compatible checks `delta.tool_calls != null` — an empty array passes this check ([] != null is true in JS), triggering premature `reasoning-end` on every chunk that contains reasoning_content.

This causes tools like OpenCode to show zero or 1ms reasoning durations instead of the actual thinking time.

Fix: in passthrough mode, delete empty tool_calls arrays from delta before hasValuableContent filtering. This is a zero-side-effect change — real tool_calls always have at least one element.

### DIFF
--- a/open-sse/utils/stream.js
+++ b/open-sse/utils/stream.js
@@ -130,6 +130,20 @@ export function createSSEStream(options = {}) {
                 }
               }
 
+              // Strip empty tool_calls arrays that break AI SDK reasoning tracking.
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @zmf963 for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2231, opened 2026-06-30
feat(minimax): extract <think> to reasoning_content for M3

zmf963:
MiniMax M3's OpenAI endpoint embeds thinking content as <think>...</think> XML tags inside the `content` field instead of providing a separate `reasoning_content`. This prevents tools like OpenCode from displaying the reasoning process.

Changes:
- stream.js: extract <think> tags from content into reasoning_content during passthrough streaming, with cross-chunk state tracking
- thinkStripper.js: utility for stripping <think> from non-streaming responses
- nonStreamingHandler.js: integrate thinkStripper for non-streaming
- minimax-cn.js / minimax.js: remove targetFormat:"claude" from MiniMax-M3 to use native OpenAI transport (zero-translation) The Claude endpoint has issues: streaming returns empty [DONE], tool calls return invalid tool type (2013)

This makes MiniMax M3 reasoning visible in OpenCode and other tools using @ai-sdk/openai-compatible.

#2047


Pieenut:
up

Issue #2182, opened 2026-06-28
Bug: Kiro translator injects a fake "continue" user message on tool-result turns


jetmiky:
## Summary

On 9Router 0.5.12, requests to `kr/*` (Kiro / CodeWhisperer) models in an **agentic tool loop** get a fabricated **`continue`** user message injected into the conversation. After the assistant emits a `tool_use`, the next request turn carries only tool results with no user text — and the Kiro request translator substitutes the literal word `"continue"` for the empty `content`. The model reads this as a real user instruction, and it surfaces in the visible conversation as if the user typed "continue" on every tool turn.

This is on the **request** path (client → Kiro). It is unrelated to #2158, which is a response-path bug (`reasoning_content` leaking into `content`).

## Environment

- **9Router version:** 0.5.12 (latest)
- **Deployment:** local global install (`~/.bun/install/global/node_modules/9router/app`), `next-server v16.2.1`, Node v25.9.0, bound to `127.0.0.1:20128`
- **Models:** `kr/claude-opus-4.8`, `kr/claude-sonnet-4.5` (any `kr/*`)
- **Provider:** `kr` (kiro)
- **Client:** any OpenAI/Claude tool-calling agent (reproduced with an agentic client that loops tool calls)

## Observed Behavior

In an agentic loop, every turn that consists solely of tool results (no user text) arrives at the model prefixed as:

```
[Context: Current time is 2026-06-28T08:14:30.741Z]

continue
```

The `continue` is not sent by the user or the client — it is injected by the translator. The model then treats it as a user instruction.

### Root cause

Two identical spots in the Kiro request translator:

- `open-sse/translator/request/openai-to-kiro.js` (`flushPending`):
  ```js
  const content = pendingUserContent.join("\n\n").trim() || "continue";
  ```
- `open-sse/translator/request/claude-to-kiro.js` (`flushPending`):
  ```js
  const content = pendingUserContent.join("\n\n").trim() || "continue";
  ```

Kiro's CodeWhisperer API rejects a **fully-empty** `content`, so a placeholder was added. But on a tool-result turn the payload already carries `userInputMessageContext.toolResults`, and **Kiro accepts an empty `content` in that case** — so the placeholder is unnecessary there, and `"continue"` is actively harmful because the model reads it as a prompt.

Note: this is **not** caused by RTK (the "request token-killer"). RTK only compresses tool-result text and never adds `continue`. Disabling RTK in the dashboard does not affect this.

## Reproduction (live Kiro probe)

[...]

Issue #2122, opened 2026-06-26
Image-only tool_result is stringified into the OpenAI tool message -> Codex "input exceeds the context window"


alican532:
### Summary
In `open-sse/translator/request/claude-to-openai.js`, the `TOOL_RESULT` branch keeps only `TEXT`
blocks and falls back to `JSON.stringify(block.content)` for anything else. When a `tool_result`
contains an **image** (and no text), the base64 image gets serialized as text into the `tool`
message — often hundreds of KB — so OpenAI-protocol clients (Codex especially) fail with
**"Your input exceeds the context window of this model."**

This triggers with any client that returns images inside a `tool_result` (e.g. Claude Code's
image-returning Read/screenshot tools) routed through 9router to an OpenAI-format provider.

### Current code (`claude-to-openai.js`, `TOOL_RESULT` case)
```js
} else if (Array.isArray(block.content)) {
  resultContent = block.content
    .filter(c => c.type === CLAUDE_BLOCK.TEXT)
    .map(c => c.text)
    .join("\n") || JSON.stringify(block.content);   // image-only -> base64 as text
}
```
For image-only content the `filter(TEXT)` is empty, so `|| JSON.stringify(block.content)` embeds the
whole array (including base64) as the tool-message text.

### Note
The same file already converts **standalone** image blocks correctly a few lines above —
`CLAUDE_BLOCK.IMAGE` → `{ type: OPENAI_BLOCK.IMAGE_URL, image_url: { url: encodeDataUri(...) } }`
pushed into `parts`. The `tool_result` path just doesn't reuse it.

### Reproduction
Send (via a Codex/OpenAI client through 9router → an OpenAI-format provider) a message with:
```json
{ "role": "user", "content": [
  { "type": "tool_result", "tool_use_id": "toolu_x",
    "content": [ { "type": "image",
      "source": { "type": "base64", "media_type": "image/png", "data": "<~200KB base64>" } } ] } ] }
```
- **Observed:** the outbound `tool` message `content` is the stringified image (~280k tokens) → context-window error.
- **Expected:** the image is forwarded as a native `image_url`, at normal token cost.

I have a small fix ready (reuses the existing `encodeDataUri`/`parts` path) and will open a PR.

warelik:
[...]

PR #2192, opened 2026-06-28
fix(translator): normalize Gemini contents to prevent 400 invalid_argument

warelik:
Resolves #2191

## Summary

When translating OpenAI/Claude request payloads to the Gemini API format, the resulting request payload can contain consecutive same-role messages (e.g. adjacent `user` messages from system/user prompts or tool history conversion) or empty `parts`. This causes the Gemini API to reject the request with `400 INVALID_ARGUMENT`.

This PR introduces a `normalizeGeminiContents` helper to safely merge adjacent same-role content blocks and strip empty parts, ensuring fully normalized payloads.

## Changes

- **`open-sse/translator/request/openai-to-gemini.js`**:
  - Implemented `normalizeGeminiContents` to concatenate parts of adjacent same-role blocks and filter out empty roles/parts.
  - Normalized `result.contents` in `openaiToGeminiBase` (affects standard OpenAI to Gemini and Gemini CLI requests).
  - Normalized `envelope.request.contents` in `wrapInCloudCodeEnvelopeForClaude` (affects Anthropic/Claude to Gemini conversions).

PR #2237, opened 2026-06-30
fix(translator): salvage orphaned tool results across request formats

### LINKED ISSUE #2236: [Bug] Orphaned tool results are not stripped consistently across request formats
## Bug

History truncation or compression can leave a tool result without its call, or a call without its result. Duplicate results are also invalid. v0.5.30 still lacks one general, non-lossy invariant repair across Responses `input[]` and OpenAI/Claude/Gemini intermediate shapes.

## Expected behavior

- Responses `function_call_output`: strip when no matching `function_call` exists.
- Intermediate OpenAI `role:"tool"`, Claude `tool_result`, and Gemini `functionResponse`: salvage orphaned or duplicate text to `[Tool result: ...]`; keep only the first real result.
- Drop image-only orphans; matched image-only results remain the separate #2122 problem.
- Merge adjacent user turns created by salvage.
- Re-run orphan and missing-result repair after RTK/Headroom compression (#2338).
- Preserve Kiro text semantics (#2182) rather than deleting content.

PR #2237 implements this contract. Competing #2298 remains delete-based and retains the reviewed call-role guard issue.

## Additional translator fix

The same touched Responses translator leaked Chat-only `max_tokens` / `max_completion_tokens`; PR #2237 maps them to `max_output_tokens` on both construction and passthrough paths.

## Verification on v0.5.30

- Responses pairing self-check: 9/9
- intermediate pairing self-check: 22/22
- production build: 126/126

warelik:
Fixes #2236

## Summary

Client-side history truncation/summarization can leave a **tool result** in the request after the matching **tool call** was removed from the active history. Strict upstream APIs reject the whole request before model execution.

This is not specific to one protocol. It is the same pairing invariant expressed by different wire formats:

| Wire format | Tool call carrier | Tool result carrier |
|---|---|---|
| OpenAI Chat Completions | `assistant.tool_calls[].id` | `role:"tool"`, `tool_call_id` |
| OpenAI Responses API | `input[].function_call.call_id` | `input[].function_call_output.call_id` |
| Anthropic Messages | `content[].tool_use.id` | `content[].tool_result.tool_use_id` |
| Gemini / Antigravity | `parts[].functionCall.id\|name` | `parts[].functionResponse.id\|name` |

The invariant: every tool result in the outbound request must reference a tool call still present in the same request.

9router already had request-side helpers for half of this surface in `open-sse/translator/concerns/toolCall.js`:

- `ensureToolCallIds` — make call ids well-formed;
- `fixMissingToolResponses` — repair the inverse case where a call exists but its result is missing.

This PR adds the missing half — on both request envelopes 9router validates:

1. **Responses API `input[]`** — strip orphaned `function_call_output` items with no matching `function_call`, including the zero-call truncation case where every call was removed but outputs remain.
2. **Canonical `messages[]` / intermediate format** — **salvage** orphaned OpenAI `role:"tool"` messages, Claude-shaped `tool_result` blocks, and Gemini `functionResponse` parts with no matching tool call — fold orphan text content into `[Tool result: ...]` user text instead of deleting.
3. **Post-compression repair** — re-run both `salvageOrphanedToolResults` and `fixMissingToolResponses` after RTK/Headroom compression in `chatCore.js` to restore the pairing invariant when compression drops assistant turns.

## Salvage-to-text (not strip)

Based on @bloodf's review, the original strip approach preempted format-specific salvage paths (notably Kiro's `reconcileOrphanedToolResults`). This PR **salvages** orphaned tool results instead of deleting them:

- Orphaned `role:"tool"` message → `{ role: "user", content: "[Tool result: <text>]" }`
- Orphaned `tool_result` block in user content → `{ type: "text", text: "[Tool result: <text>]" }`
- Orphaned Gemini `functionResponse` part → `{ text: "[Tool result: <text>]" }`
- Image-only `tool_result` (no text) → still dropped (no text representation; tracked in #2122)
- Matched tool results → unchanged
- No-op fast path (same body reference) when nothing needs salvaging

The `[Tool result: ...]` format matches the pattern already used in the merged combo fix (#1859).

## Consecutive user message merge

After salvage, consecutive same-role user messages are merged so downstream translators that don't merge (notably `openai-to-gemini.js`, which pushes each `role:"user"` as a separate `contents[]` entry) don't emit adjacent user turns that trigger Gemini 400 INVALID_ARGUMENT. Only string-content user messages are merged; array-content messages (which may carry structured blocks like `tool_result`) are left as-is.

## Post-compression tool pairing repair

RTK/Headroom compression can drop assistant turns containing `tool_calls`, breaking the pairing invariant. This PR re-runs both `salvageOrphanedToolResults` and `fixMissingToolResponses` on the compressed body in `chatCore.js`, restoring the invariant after compression. This addresses #2338 (tool_use without tool_result after headroom).

The resulting pipeline enforces the full tool-pairing tripod before dispatch:

```js
ensureToolCallIds(result);              // ids are well-formed
fixMissingToolResponses(result);        // every call has a result
salvageOrphanedToolResults(result);     // every result has a call (salvaged to text)
// ... RTK/Headroom compression ...
salvageOrphanedToolResults(translatedBody);  // re-check after compression
fixMissingToolResponses(translatedBody);     // re-check after compression
```

### `open-sse/translator/concerns/toolCall.js`

- `extractToolResultText(content)` — extracts text from string / array of content blocks (text blocks only), returns `""` for image-only.
- `salvageOrphanedToolResults(body)` — handles three wire formats:
  - OpenAI/Claude `messages[]`: orphaned `role:"tool"` / `tool_result` blocks → user text
  - Gemini/Antigravity `contents[]`: orphaned `functionResponse` parts → text part (with name-based fallback pairing)
  - Merges consecutive same-role user messages after salvage
  - Fail-open (returns body unchanged on any error)
  - Assistant-only `knownCallIds` collection (fixes false-positive bug from #2298 review)
- `fixMissingToolResponses` — fast-path with `changed` flag (no array allocation on no-op)
- `mergeConsecutiveUserMessages(messages)` — merge adjacent string-content user messages

### `open-sse/handlers/chatCore.js`

- Imports `salvageOrphanedToolResults` + `fixMissingToolResponses`.
- Post-compression call site: re-runs both helpers after RTK/Headroom, before Caveman/Ponytail.

### `open-sse/translator/index.js`

- Wired `salvageOrphanedToolResults(result)` in `translateRequest` immediately after `fixMissingToolResponses`.

### `open-sse/translator/request/openai-responses.js`

- `collectKnownCallIds(input)` — collect `function_call.call_id` values from a Responses API `input[]`.
- `stripOrphanedToolOutputs(input)` — remove `function_call_output` items with no matching `function_call`. (Responses API envelope — no text representation to salvage, so strip is correct here.)
- Applied in Responses→OpenAI translation, OpenAI→Responses translation, and Responses passthrough.

### `open-sse/utils/openaiResponsesToolPairingSelfCheck.mjs` (new)

5/5 checks — Responses envelope invariant.

### `open-sse/utils/toolPairingSelfCheck.mjs` (new)

15/15 checks — OpenAI/Claude/Gemini intermediate invariant, including:
- Matched preserved (OpenAI + Claude + Gemini)
- Orphan salvaged to user text (OpenAI + Claude + Gemini)
- Zero-call truncation salvages all
- Image-only orphan dropped (#2122)
- Claude mixed: image-only orphan dropped, text block preserved
- Gemini name-based fallback pairing
- Consecutive user messages merged after salvage
- Array-content messages not merged
- No-op same body reference
- Mixed matched + orphan

## Related

- #2298 — competing PR by @anki1kr on the same #2236. Coordinating to merge efforts — this PR takes the salvage-to-text approach, adds Gemini `functionResponse` coverage and post-compression call site from #2298, and fixes the assistant-only id collection bug (Copilot review).
- #2338 — reverse problem (tool_use without tool_result), same root cause (headroom compression). Addressed by post-compression repair.
- #2182 — Kiro fake "continue" on tool-result turns. Salvage-to-text should reduce the empty-content case that triggers it.
- #2122 — image-only tool_result. Salvage skips non-text content; remains a separate issue for `claude-to-openai.js`.

## Verification

```bash
node --check open-sse/translator/request/openai-responses.js
node --check open-sse/translator/concerns/toolCall.js
node --check open-sse/translator/index.js
node --check open-sse/handlers/chatCore.js
node --check open-sse/utils/openaiResponsesToolPairingSelfCheck.mjs
node --check open-sse/utils/toolPairingSelfCheck.mjs
node open-sse/utils/openaiResponsesToolPairingSelfCheck.mjs
node open-sse/utils/toolPairingSelfCheck.mjs
npm run build
```

Results:

- `openaiResponsesToolPairingSelfCheck.mjs`: 5/5 passed
- `toolPairingSelfCheck.mjs`: 15/15 passed
- `npm run build`: compiled successfully, 125/125 pages

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -22,6 +22,7 @@ import { detectClientTool, isNativePassthrough } from "../utils/clientDetector.j
 import { dedupeTools } from "../utils/toolDeduper.js";
 import { injectCaveman } from "../rtk/caveman.js";
 import { injectPonytail } from "../rtk/ponytail.js";
+import { salvageOrphanedToolResults, fixMissingToolResponses } from "../translator/concerns/toolCall.js";
 import { compressMessages, formatRtkLog } from "../rtk/index.js";
 import { compressWithHeadroom, formatHeadroomLog, formatHeadroomSizeLog, isHeadroomPhantomSavings } from "../rtk/headroom.js";
 import { compressWithPxpipe } from "../rtk/pxpipe.js";
@@ -257,6 +258,10 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
 
   // Token-saver flags accumulator for the single "⚙" log line below.
   const xf = [];
+
+  // Compression may remove a tool call while retaining its result. Restore the invariant before dispatch.
+  salvageOrphanedToolResults(translatedBody);
+  fixMissingToolResponses(translatedBody);
 
   // Caveman: inject terse-style system prompt
   if (tokenSaverEnabled && cavemanEnabled && cavemanLevel) {
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -21,6 +21,7 @@ import { detectClientTool, isNativePassthrough } from "../utils/clientDetector.j
 import { dedupeTools } from "../utils/toolDeduper.js";
 import { injectCaveman } from "../rtk/caveman.js";
 import { injectPonytail } from "../rtk/ponytail.js";
+import { salvageOrphanedToolResults, fixMissingToolResponses } from "../translator/concerns/toolCall.js";
 import { compressMessages } from "../rtk/index.js";
 import { compressWithHeadroom, formatHeadroomSizeLog, isHeadroomPhantomSavings } from "../rtk/headroom.js";
 import { compressWithPxpipe } from "../rtk/pxpipe.js";
@@ -213,6 +214,10 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   } else if (headroomEnabled) {
     log?.warn?.("HEADROOM", `skipped: ${headroomDiagnostics.reason || "compression unavailable"}${headroomDiagnostics.endpoint ? ` (${headroomDiagnostics.endpoint})` : ""}`);
   }
+
+  // Compression may remove a tool call while retaining its result. Restore the invariant before dispatch.
+  salvageOrphanedToolResults(translatedBody);
+  fixMissingToolResponses(translatedBody);
 
   // Caveman: inject terse-style system prompt
   if (cavemanEnabled && cavemanLevel) {
--- a/open-sse/translator/concerns/toolCall.js
+++ b/open-sse/translator/concerns/toolCall.js
@@ -118,10 +118,232 @@ export function hasToolResults(msg, toolCallIds) {
 }
 
 // Fix missing tool responses - insert empty tool_result if assistant has tool_use but next message has no tool_result
+// Extract text content from a tool_result block (Claude-shaped) or tool message (OpenAI-shaped).
+// Returns "" when there is no text (e.g. image-only tool_result — tracked separately in #2122).
+function extractToolResultText(content) {
+  if (typeof content === "string") return content.trim();
+  if (!Array.isArray(content)) return "";
+  return content
+    .filter(block => block?.type === "text" && typeof block.text === "string")
+    .map(block => block.text)
+    .join("\n")
+    .trim();
+}
+
+// Salvage orphaned tool results — results that reference a tool call no longer
+// present in the same request. Mirrors fixMissingToolResponses on the result side:
+// that helper ensures every call has a result; this one ensures every result has a call.
+//
+// Instead of deleting orphaned results (which preempts format-specific salvage paths
+// like Kiro's reconcileOrphanedToolResults and loses tool output context), this folds
+// the orphan's text content into a user-text block: `[Tool result: <text>]`.
+// Image-only results (no text) are still dropped — no text representation exists (#2122).
+//
+// Handles two wire envelopes (three schemas):
+//   - OpenAI/Claude messages[]: orphaned role:"tool" / tool_result blocks → user text
+//   - Gemini/Antigravity contents[]: orphaned functionResponse parts → user text part
+//
+// Responses API function_call_output is handled separately in openai-responses.js
+// (stripOrphanedToolOutputs) because Responses items have no text representation
+// to salvage — they are structural call/output pairs, not content blocks.
+//
+// After salvaging messages[], consecutive same-role user messages are merged so
+// downstream translators that don't merge (notably openai-to-gemini, which pushes
+// each role:"user" as a separate contents[] entry) don't emit adjacent user turns
+// that trigger Gemini 400 INVALID_ARGUMENT.
+//
+// Fail-open: any error returns the body unchanged.
+export function salvageOrphanedToolResults(body) {
+  if (!body || typeof body !== "object") return body;
+
+  try {
+    let changed = false;
+
+    // ── OpenAI/Claude messages[] ───────────────────────────────────────────
+    if (Array.isArray(body.messages)) {
+      const knownCallIds = new Set();
+      for (const msg of body.messages) {
+        // Only assistant turns carry tool calls — collecting from other roles
+        // causes false positives (a user message with tool_use would mask orphans).
+        if (msg.role !== "assistant") continue;
+
+        if (Array.isArray(msg.tool_calls)) {
+          for (const tc of msg.tool_calls) {
+            if (typeof tc?.id === "string") knownCallIds.add(tc.id);
+          }
+        }
+
+        if (Array.isArray(msg.content)) {
+          for (const block of msg.content) {
+            if (block?.type === "tool_use" && typeof block.id === "string") {
+              knownCallIds.add(block.id);
+            }
+          }
+        }
+      }
+
+      const salvagedMessages = [];
+      // Track IDs already matched to a tool result — duplicates are salvaged too,
+      // since strict providers reject a second result for the same call_id.
+      const consumedIds = new Set();
+
+      for (const msg of body.messages) {
+        // OpenAI shape: orphaned role:"tool" message → salvage to user text
+        if (msg.role === "tool" && msg.tool_call_id) {
+          if (knownCallIds.has(msg.tool_call_id) && !consumedIds.has(msg.tool_call_id)) {
+            consumedIds.add(msg.tool_call_id);
+            salvagedMessages.push(msg);
+          } else {
+            const text = extractToolResultText(msg.content);
+            if (text) {
+              salvagedMessages.push({ role: "user", content: `[Tool result: ${text}]` });
+            }
+            changed = true;
+          }
+          continue;
+        }
+
+        // Claude shape: orphaned tool_result block in user content → salvage to text block
+        if (Array.isArray(msg.content)) {
+          let orphanCount = 0;
+          const rebuiltContent = [];
+
+          for (const block of msg.content) {
+            if (block?.type !== "tool_result") {
+              rebuiltContent.push(block);
+              continue;
+            }
+            if (typeof block.tool_use_id === "string" && knownCallIds.has(block.tool_use_id) && !consumedIds.has(block.tool_use_id)) {
+              consumedIds.add(block.tool_use_id);
+              rebuiltContent.push(block);
+              continue;
+            }
+            // Orphaned or duplicate tool_result — salvage text content
+            orphanCount++;
+            const text = extractToolResultText(block.content);
+            if (text) {
+              rebuiltContent.push({ type: "text", text: `[Tool result: ${text}]` });
+            }
+          }
+
+          if (orphanCount > 0) {
+            changed = true;
+            if (rebuiltContent.length === 0) continue; // all blocks were image-only orphans
+            msg.content = rebuiltContent;
+          }
+        }
+
+        salvagedMessages.push(msg);
+      }
+
+      if (changed) {
+        body.messages = mergeConsecutiveUserMessages(salvagedMessages);
+      }
+    }
+
+    // ── Gemini/Antigravity contents[] ──────────────────────────────────────
+    if (Array.isArray(body.contents)) {
+      const knownFnIds = new Set();
+      for (const turn of body.contents) {
+        // Only model turns carry functionCall — collecting from other roles
+        // causes false positives (a malformed user turn with functionCall would mask orphans).
+        if (turn.role !== "model") continue;
+        if (!Array.isArray(turn.parts)) continue;
+        for (const part of turn.parts) {
+          if (part?.functionCall) {
+            // Prefer explicit id; fall back to name (older Gemini shapes pair by name).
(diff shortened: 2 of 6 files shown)


bloodf:
This global strip preempts format-specific orphan handling that already exists on `master`, causing a regression for the Kiro direct route.

`master`'s `claude-to-kiro.js` already handles orphaned tool_results by **salvaging** them: `reconcileOrphanedToolResults` folds an orphan's content into the user text (and `flattenClaudeToolInteractions` covers the no-tools case). This PR calls `stripOrphanedToolResults(result)` in `translateRequest` *before* the target translator runs, which **deletes** the orphaned `tool_result` block — and drops the whole message when it becomes empty. So for `claude:kiro`, the orphan content is gone before Kiro can salvage it.

Repro — `translateRequest(CLAUDE, KIRO, …)`:

```js
messages: [
  { role: "user", content: "go" },
  { role: "user", content: [{ type: "tool_result", tool_use_id: "ghost", content: "salvage me" }] },
]
```

On `master` the resulting current-message content contains `salvage me` (folded in); with this PR it is dropped.

Two possible directions:
1. Skip the global strip for targets that salvage orphans themselves, e.g. `if (targetFormat !== FORMATS.KIRO) stripOrphanedToolResults(result);`
2. Make `stripOrphanedToolResults` salvage-to-text (like the Kiro path) instead of deleting, so it is non-lossy across all formats.

warelik:
Thanks @bloodf — great catch. You're right: the global `stripOrphanedToolResults(result)` in `translateRequest` runs **before** the target translator, so it preempts Kiro's `reconcileOrphanedToolResults` salvage path. Going with your **option 2** (salvage-to-text) — it's non-lossy for every format, not just Kiro.

## What changed

Renamed `stripOrphanedToolResults` → `salvageOrphanedToolResults`. Instead of deleting orphaned tool results, fold their text content into a user-text block:

- Orphaned `role:"tool"` message → `{ role: "user", content: "[Tool result: <text>]" }`
- Orphaned `tool_result` block in user content → `{ type: "text", text: "[Tool result: <text>]" }`
- Orphaned Gemini `functionResponse` part → `{ text: "[Tool result: <text>]" }`
- Image-only `tool_result` (no text) → still dropped (no text representation; tracked separately in #2122)
- Matched tool results → unchanged
- No-op fast path (same body reference) when nothing needs salvaging

The `[Tool result: ...]` format matches the pattern already used in the merged combo fix (#1859).

## Consecutive user message merge

After salvage, consecutive same-role **string-content** user messages are merged so downstream translators that don't merge (notably `openai-to-gemini.js`, which pushes each `role:"user"` as a separate `contents[]` entry — verified in source) don't emit adjacent user turns that trigger Gemini 400 INVALID_ARGUMENT. Array-content messages (which may carry structured blocks like `tool_result`) are left as-is.

The merge uses **clone-on-merge**: when concatenating, the last entry is replaced with a new object instead of mutating it in place, so original message references held by callers (continuity canonicalization, logging) are not silently corrupted. Non-merged messages are pushed by reference (no unnecessary clones on the common no-merge path).

*(Earlier draft of this comment claimed "handled by existing Gemini normalization (merged in #2192)" — that was wrong: `openai-to-gemini.js` does not merge consecutive user turns. The merge is now done in `salvageOrphanedToolResults` itself.)*

## Gemini `contents[]` coverage

Added `functionResponse` salvage for Gemini/Antigravity `contents[]`:
- Collects `functionCall` ids **only from `role:"model"` turns** (symmetric with the assistant-only guard for `messages[]` — collecting from other roles causes false positives).
- Pairs by `id` with name-based fallback (older Gemini shapes pair by name).
- Salvages `response.result` (or `response` if no `result` key) to a text part.
- Drops the turn entirely if all parts were image-only/empty orphans — Gemini rejects turns with empty `parts:[]` (400 INVALID_ARGUMENT).

## Post-compression tool pairing repair

RTK/Headroom compression can drop assistant turns containing `tool_calls`, breaking the pairing invariant. The PR now re-runs both `salvageOrphanedToolResults` and `fixMissingToolResponses` on the compressed body in `chatCore.js`, restoring the invariant after compression. This addresses #2338 (tool_use without tool_result after headroom).

`fixMissingToolResponses` also got a fast-path (`changed` flag) — no array allocation on no-op — since it's now called twice per request.

## Related issues this addresses

- **#2338** — reverse problem (`tool_use` without `tool_result`), same root cause: headroom/RTK compression drops assistant turns. Addressed by post-compression repair.
- **#2182** — Kiro fake "continue" on tool-result turns. Salvage-to-text preserves the tool result content as user text instead of dropping it, which should make the empty-content case that triggers the fake "continue" less likely.
- **#2122** — image-only `tool_result`. Salvage skips non-text content (can't represent an image as `[Tool result: ...]`), so this stays a separate issue for the `claude-to-openai.js` image-handling path.

## Related PR

#2298 by @anki1kr is a competing approach on the same #2236. It's broader (covers Gemini `functionResponse`, adds post-compression strip) but has the same delete-not-salvage issue you flagged here, plus a Copilot-noted bug: it collects `tool_use` ids from any message role, not just assistant, which can cause false positives. This PR takes the salvage-to-text approach, adds the Gemini `functionResponse` coverage and post-compression call site from #2298, and fixes the assistant-only id collection (for `messages[]`) and the model-only id collection (for `contents[]`).

## Self-checks

- `toolPairingSelfCheck.mjs`: **19/19** — OpenAI/Claude/Gemini matched preserved, orphan salvaged to text, image-only dropped, Gemini role guard (non-model `functionCall` doesn't mask orphans), Gemini empty-orphan turn dropped, consecutive user merge, array-content merge skip, clone-on-merge (original messages not mutated by reference), no-op same body ref, null body guard.
- `openaiResponsesToolPairingSelfCheck.mjs`: 5/5 (unchanged).
- `npm run build`: 125/125 pages compiled successfully.

warelik:
Rebuilt this branch as a single clean commit on the latest `master` (v0.5.18). The patch is now fully autonomous (no baseline dependencies) and includes:

- **Salvage-to-text** for OpenAI `role:tool`, Claude `tool_result`, and Gemini `functionResponse` (per @bloodf's review feedback)
- **Post-compression repair** in `chatCore.js` — re-runs `salvageOrphanedToolResults` + `fixMissingToolResponses` after RTK/Headroom (#2338)
- **Consecutive user merge** after salvage (prevents Gemini 400 on adjacent same-role turns, #2191)
- **Responses API** orphaned `function_call_output` stripping
- Self-checks: `openaiResponsesToolPairingSelfCheck.mjs` (5/5) + `toolPairingSelfCheck.mjs` (19/19)

Files changed: `chatCore.js`, `toolCall.js`, `translator/index.js`, `openai-responses.js`, 2 self-checks.

warelik:
## Follow-up fix: Chat→Responses `max_tokens` leak (pre-existing upstream bug in the same function)

While auditing this patch's touched file (`translator/request/openai-responses.js`) for an unrelated production `400 Unsupported parameter: max_tokens` report against an `openai-compatible-responses-*` target, found a pre-existing bug in **stock upstream** (present before this PR's changes, in the same `openaiToOpenAIResponsesRequest` function this PR already touches):

- The Responses→Chat direction (`openaiResponsesToOpenAIRequest`) already correctly maps `max_output_tokens → max_tokens` (added earlier, commit `aba4c45`).
- The reverse Chat→Responses direction only had `if (body.max_tokens !== undefined) result.max_tokens = body.max_tokens;` — this leaks the Chat-only field name straight through. The Responses API only accepts `max_output_tokens` and rejects `max_tokens`/`max_completion_tokens` outright with HTTP 400.

Any client resolving to `openai-responses` as its target format (both the built-input path and the already-`input[]` passthrough path, e.g. Cursor CLI/agents calling `/chat/completions` with Responses-shaped `input[]`) could hit this whenever `max_tokens` was present in the request.

**Fix** (folded into this PR since it's the same function/file this PR already modifies): both branches of `openaiToOpenAIResponsesRequest` now map `max_tokens`/`max_completion_tokens` → `max_output_tokens` and delete the Chat-only names, without clobbering an already-present `max_output_tokens`.

**Verification:**
- Extended `open-sse/utils/openaiResponsesToolPairingSelfCheck.mjs` from 5/5 → 9/9 (added: built-input `max_tokens` mapping, `max_completion_tokens` mapping, passthrough-branch mapping, no-clobber-of-explicit-max_output_tokens).
- `open-sse/utils/toolPairingSelfCheck.mjs` still 19/19 (unaffected, different file).
- Full 5-patch stack (`0001`→`0002`→`0004`→`0006`→`0007`) applies cleanly, `npm run build` succeeds.

Not provider-specific: this is a root-cause fix in the shared translator direction, not a special case for any one `openai-compatible-responses-*` credential.

warelik:
## Update: duplicate tool-result salvaging + verified counts

Extended `salvageOrphanedToolResults` to also handle **duplicate** tool results — a second `role:"tool"` message, `tool_result` block, or `functionResponse` part carrying the same id as one already matched earlier in the same request. Strict providers reject a second result for the same call, so duplicates are now salvaged to `[Tool result: ...]` text the same way orphans are (only the first result for a given id is kept as a real tool result).

Updated counts (self-checks extended, no behavior change to the existing matched/orphan/merge cases):

- `toolPairingSelfCheck.mjs`: **22/22** (was 19/19 — +3 for OpenAI/Claude/Gemini duplicate-result salvaging)
- `openaiResponsesToolPairingSelfCheck.mjs`: **9/9** (unchanged from the `max_tokens` mapping update above)
- `npm run build`: 125/125 pages, 0 errors

Re-verified independently on a fresh `origin/HEAD` checkout (no other changes applied).

warelik:
## v0.5.30 standalone rebuild

Re-audited the full behavior against upstream v0.5.30 (`9845a17`). The problem is still present in current master, and the competing #2298 remains delete-based (loses orphan text/Kiro salvage) and retains the reviewed role-guard issue.

The rebuilt standalone commit now contains the complete non-lossy contract:

- Responses `input[]`: strip orphaned `function_call_output` items.
- OpenAI/Claude/Gemini intermediate shapes: salvage orphaned **and duplicate** results to `[Tool result: ...]` user text; keep only the first real result for a call.
- Drop image-only orphans, merge consecutive user turns after salvage, and re-run both directions of pairing repair after RTK/Headroom compression.
- Chat→Responses: map `max_tokens` / `max_completion_tokens` to `max_output_tokens` on both construction and passthrough paths without clobbering an explicit value.

Files remain the six standalone files documented by this PR. Verification on clean v0.5.30:

- `openaiResponsesToolPairingSelfCheck.mjs`: 9/9
- `toolPairingSelfCheck.mjs`: 22/22
- production build: 126/126
- `git apply --check`, `git diff --check`: clean

Because @bloodf provided external review, I am adding this summary instead of rewriting the discussion history.

PR #2242, opened 2026-06-30
feat(capabilities): model capability metadata on `/v1/models`, combo aggregation, pattern fixes

mihado:
## Summary

`/v1/models` returns bare model IDs with no capability metadata, so downstream tools (OpenCode) default vision/tools/reasoning to false for every model. This PR wires capability metadata to every entry and fixes several incorrect pattern matches.

### `open-sse/providers/capabilities.js`

- Moves the `aggregateComboCapabilities` out of `route.js` so it can be imported by the dashboard and tested. Union for vision/pdf/audio/search, intersection for tools, primary model drives reasoning fields, min `contextWindow`, max `maxOutput`.
- **Nested combo resolution:** accepts an optional `comboLookup` map; when a combo member is itself a combo, recurses into its members instead of pattern-matching the name. Depth guard at 6. Backwards-compatible.
- **Pattern fixes:**
  - add MiMo v2.5/Omni (vision + deepseek reasoning, always-on `<think>`)
  - add `vision: true` to `*qwen*max*` / `*qwen*plus*` (previously swallowed by a later pattern)
  - add `*minimax-m2.5*` with vision
  - add `glm-5.2` with 1M context
  - remove `vision` from `*deepseek-v4*`

### `src/app/api/v1/models/route.js`

- Every `/v1/models` entry now carries a `capabilities` object.
- Static models use `getCapabilitiesForModel`, live-resolved models fall through `capabilitiesFromServiceKind`
- Combos use `aggregateComboCapabilities(combo.models, comboByName)` with a name→models lookup so nested combos resolve correctly.

### `src/app/(dashboard)/dashboard/combos/page.js`

- Switch combo list to the `useModelCaps` hook so custom and combo-within-combo entries showed no icons.
- Add a `ctx Xk · max Yk` metadata line per combo card from aggregated capabilities.
- Fix nested-combo chip icons to resolve through `comboByName`.

### Provider registry

| Provider | Change |
|---|---|
| commandcode | Add Kimi K2.7 Code/HighSpeed, GLM 5.2/5.2-Fast, MiniMax M3, MiMo v2.5/v2.5-Pro, Qwen 3.7 Max/Plus, StepFun 3.7 Flash, Nemotron 3 Ultra |
| openai | Add `gpt-5.5` |
| opencode-go | Fix typo in subscription notice |

## Tests

39 unit tests across `capabilities.test.js` and `combo-capabilities.test.js` covering the 4-tier resolver, pattern order, combo aggregation, nested lookup, cycle safety, and backwards-compat.


mihado:
@decolua what do you think about this? I have been building & running this locally without issues in the last few days. This is quite useful for downstream tool to work out the model capabilities automatically.

zierocode:
I added conservative nested-Combo and opaque-validator acceptance tests in draft companion #2656. They cover intersection/minimum aggregation, cycle/missing/unknown omission, byte-stable canonical output, membership-sensitive HMAC ETags, and If-None-Match semantics. This is intended to help #2242 fail closed for fallback routing rather than replace its implementation.

Fresh cross-check on 2026-07-23:

- this PR head: `83bc0af95f6c5d7dd2a82e6c4e974c25a902473a`
- focused capability suites: 40/40 passed
- the complete #2656 contract applied to vanilla: 2 baseline passes and 13 expected failures
- the same contract applied to this PR exposes one safety mismatch: when a Combo member cannot be resolved, `/v1/models` still publishes a capability object for the logical Combo instead of omitting aggregate metadata

Observed value for the unresolved case included `contextWindow: 200000` and `tools: true`. For fallback routing, that is stronger than the available evidence and can route a request to a leaf that cannot satisfy it.

Recommended convergence: keep this PR as capability-registry owner, but make unresolved/cyclic/unknown nested membership return no public aggregate. Leave strong ETag/private-revision behavior in #2656 so the two PRs stay independently reviewable.

eddoww:
Supporting this — it addresses a real interop gap. Clients that consume `GET /v1/models` currently have no way to learn `vision` / `tools` / `reasoning` for combo (alias) entries, because those entries are emitted as bare stubs. The clients therefore default every capability to false and drop image parts before the request is ever sent, which then reads to the user as "the model can't see images" even when every member can.

Emitting conservative, member-derived metadata (union where all members agree, min context / max output, omit when unresolved) is exactly what downstream tooling needs.

One caveat from testing against OpenCode: a generic OpenAI-compatible provider that is configured manually may not consume this field today — its capability resolution goes through models.dev or explicit per-model config, not the live `/v1/models` payload — so this PR may need a companion client-side change for that specific path. It is still clearly worth landing for the clients that do read it, and it fixes the "combo looks capability-less" problem at the source.

Would love to see this reviewed/merged; it has been open a while and overlaps with #3909 / #4034.

decolua:
Thanks @mihado for the contribution! Reviewed and merged into master. 🙏

PR #2248, opened 2026-06-30
fix(antigravity): preserve Claude tool delta index

### LINKED ISSUE #2225: Antigravity Provider fails with Google AI Pro account in Claude Code
When using the **Antigravity** provider with **Claude Code**, requests consistently fail when authenticated with a **Google Pro** account.

The exact same setup works correctly after switching to a **Google Free** account.

This suggests the issue may be related to differences in the API response or streaming format returned for Pro accounts.

## Environment

- Claude Code: 2.1.177
- Provider: Antigravity
- 9Router: 0.5.15
- OS: Ubuntu 24.04.4 LTS

## Steps to Reproduce

1. Configure Claude Code to use the Antigravity provider.
2. Sign in with a **Google Pro** account.
3. Start a new conversation.
4. Send any prompt.

## Expected Behavior

The model should stream and complete the response successfully.

## Actual Behavior

The request fails with one of the following errors:

```text
● API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request

● API Error: Content block not found
```

## Comparison

| Google Account | Result |
| -------------- | ------ |
| Pro            | ❌ Fails |
| Free           | ✅ Works |

The following remain exactly the same:

- Claude Code version
- 9Router version
- Machine
- Network
- Configuration
- Prompt

Only the Google account is changed.

Jordannst:
## Summary

Fixes a Claude Code streaming failure when using the Antigravity provider.

Issue #2225 reports `API Error: Content block not found` after upgrading to 9Router `0.5.15`. The report first mentioned Google Pro, but follow-up comments indicate the same error can happen with Free accounts too, so this fix treats it as an Antigravity → Claude Code stream translation issue rather than an account-tier or OAuth issue.

## Findings

The Antigravity/Gemini response path is translated through OpenAI before being translated to Claude:

```text
Antigravity/Gemini response → OpenAI response → Claude response
```

`geminiToOpenAIResponse()` was using the shared `state.toolCalls` map for Gemini-side bookkeeping. The downstream `openaiToClaudeResponse()` also uses `state.toolCalls`, but for Claude content-block metadata.

When a Gemini `functionCall` arrived, the upstream translator pre-populated `state.toolCalls`. Then the Claude translator saw the tool-call index as already present and did not emit the matching `content_block_start` for the Claude `tool_use` block. Later tool argument/stop events could still be emitted for that block index, which can make Claude Code reject the stream with:

```text
API Error: Content block not found
```

## Fix

Keep Gemini bookkeeping separate from Claude tool-block state:

- replace Gemini-side `state.toolCalls` mutation with `state.geminiToolCallCount`
- use `state.geminiToolCallCount` when normalizing Gemini `STOP` to OpenAI `tool_calls`
- add a regression test for `Antigravity → Claude` tool-call streaming that verifies `input_json_delta` keeps a valid Claude block index
- keep the existing Kiro/Claude direct-route coverage aligned with the same invariant

## Verification

```bash
npx vitest run tests/translator/bugs-antigravity.test.js tests/translator/claude-kiro-direct.test.js tests/unit/openai-to-claude-response-tools.test.js
```

Result:

```text
Test Files  3 passed (3)
Tests       15 passed | 1 expected fail (16)
```

Fixes #2225

### DIFF
--- a/open-sse/translator/response/gemini-to-openai.js
+++ b/open-sse/translator/response/gemini-to-openai.js
@@ -25,7 +25,10 @@ function emitFunctionCall(functionCall, state) {
     type: OPENAI_BLOCK.FUNCTION,
     function: { name: fcName, arguments: JSON.stringify(fcArgs) },
   };
-  state.toolCalls.set(toolCallIndex, toolCall);
+  // Keep Gemini bookkeeping separate from the shared translator state.toolCalls map.
+  // The downstream OpenAI→Claude translator uses state.toolCalls for Claude block
+  // metadata; pre-populating it here makes Anthropic tool deltas lose index.
+  state.geminiToolCallCount = (state.geminiToolCallCount || 0) + 1;
   return buildChunk(chunkMeta(state), { tool_calls: [toolCall] }, null);
 }
 
@@ -46,6 +49,7 @@ export function geminiToOpenAIResponse(chunk, state) {
     state.messageId = response.responseId || `msg_${Date.now()}`;
     state.model = response.modelVersion || "gemini";
     state.functionIndex = 0;
+    state.geminiToolCallCount = 0;
     results.push(buildChunk(chunkMeta(state), { role: ROLE.ASSISTANT }, null));
   }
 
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @Jordannst for the contribution! 🙏 This fix (separating Gemini bookkeeping into `state.geminiToolCallCount` so the downstream OpenAI→Claude translator keeps a valid Claude `tool_use` block index) is already present on `master`, along with the `Antigravity → Claude` and Kiro `input_json_delta` index regression tests. Closing as already applied.

PR #2251, opened 2026-06-30
fix: xiaomi mimo token-plan regions and multi-connection for compatible providers

MiQieR:
### Summary

This PR fixes three bugs related to the Xiaomi MiMo Token Plan provider and one bug affecting all Custom Compatible providers.

---

#### 1. Missing region selector when adding a Token Plan key

**Problem**: The `xiaomi-tokenplan` registry entry defined its cluster URLs inside `transport.regions`, but omitted a top-level `regions` array. The `AddApiKeyModal` reads `AI_PROVIDERS[provider].regions` to decide whether to render a region `<Select>` — because that field was absent, the dropdown was never shown. Every connection silently fell back to the default `sgp` (Singapore) cluster, causing API key validation to fail for users whose subscription belongs to the `cn` or `ams` cluster.

**Fix**: Added a top-level `regions` array to the `xiaomi-tokenplan` registry entry (`open-sse/providers/registry/xiaomi-tokenplan.js`), making the region dropdown visible in `AddApiKeyModal`.

---

#### 2. Token Plan key validation always fails / times out

**Problem**: The validation handler for `xiaomi-tokenplan` used `isValid = res.ok` (requires HTTP 200). The Token Plan `/v1/models` endpoint returns `403` for valid keys that lack model-listing permission, causing those keys to be rejected as invalid. Additionally, the `fetch` call had no timeout, so connections to unreachable clusters would hang indefinitely and eventually throw a `fetch failed` error.

**Fix**: Changed the validity check for `xiaomi-tokenplan` to `res.status !== 401 && res.status !== 403` (consistent with other providers that use the same pattern), and added `AbortSignal.timeout(8000)` to the `fetch` call (`src/app/api/providers/validate/route.js`).

---

#### 3. Region cannot be changed after a Token Plan connection is saved

**Problem**: `EditConnectionModal` had no awareness of provider regions. It neither read the saved region from `providerSpecificData` nor rendered a `<Select>` for it, making it impossible to fix a connection that was saved with the wrong cluster.

**Fix**: Updated `EditConnectionModal` (`src/shared/components/EditConnectionModal.js`) to:
- Read and initialize the `region` state from `connection.providerSpecificData.region` on open.
- Render a `<Select>` dropdown when `AI_PROVIDERS[provider].regions` is defined (generic — works for any future region-aware provider, not just Token Plan).
- Pass the selected region in `providerSpecificData` during both key validation (`Check`) and save.

---

#### 4. Cannot add more than one API key to a Custom Compatible node

**Problem**: The `POST /api/providers` handler contained guards that rejected any second connection for OpenAI-compatible, Anthropic-compatible, and Custom Embedding nodes with `"Only one connection is allowed…"`. This broke Bulk Add (always `✓ 1 added, ✗ n-1 failed`) and prevented users from configuring multiple keys for rotation or load balancing.

The comment in the code stated these guards were *"restored to honor the contract locked in by tests"*, but the underlying router and DB layer already support multiple active connections per provider ID natively.

**Fix**: Removed the duplicate-connection guards from `src/app/api/providers/route.js` and updated the corresponding unit test in `tests/unit/compatible-provider-connections.test.js` to assert that multiple connections are successfully created on the same node.

---

### Files Changed

| File | Change |
|---|---|
| `open-sse/providers/registry/xiaomi-tokenplan.js` | Added top-level `regions` array |
| `src/app/api/providers/validate/route.js` | Fixed `xiaomi-tokenplan` validity check; added fetch timeout |
| `src/shared/components/EditConnectionModal.js` | Added region selector support for region-aware providers |
| `src/app/api/providers/route.js` | Removed single-connection restriction for compatible nodes |
| `tests/unit/compatible-provider-connections.test.js` | Updated test to assert multiple connections are allowed |

---

### Testing

**Manual verification performed:**
- Opened the Add Connection modal for Xiaomi MiMo (Token Plan) — the **Region** dropdown now appears with Singapore, China, and Amsterdam options.
- Selected **China (中国大陆)**, entered a `cn`-cluster Token Plan key — validation returns **Valid**.
- Opened the Edit Connection modal for an existing Token Plan connection — the current region is pre-selected and can be changed.
- Added two API keys to the same OpenAI-compatible node — both succeed; Bulk Add correctly reports `✓ 2 added`.

**Automated tests:**
- Updated unit test `allows multiple connections on the same compatible node` passes as expected.

### DIFF
--- a/open-sse/providers/registry/xiaomi-tokenplan.js
+++ b/open-sse/providers/registry/xiaomi-tokenplan.js
@@ -21,6 +21,11 @@ export default {
   },
   category: "apikey",
   hasProviderSpecificData: true,
+  regions: [
+    { id: "sgp", label: "Singapore (新加坡)" },
+    { id: "cn", label: "China (中国大陆)" },
+    { id: "ams", label: "Amsterdam (阿姆斯特丹)" },
+  ],
   defaultRegion: "sgp",
   transport: {
     baseUrl: "https://token-plan-sgp.xiaomimimo.com/v1/chat/completions",
--- a/src/app/api/providers/route.js
+++ b/src/app/api/providers/route.js
@@ -126,17 +126,10 @@ export async function POST(request) {
 
     let providerSpecificData = normalizeProviderSpecificData(provider, body, body.providerSpecificData);
 
-    // Compatible/embedding nodes allow exactly one connection each. These guards were
-    // dropped accidentally during the bun:sqlite refactor (v0.4.28); restored to honor
-    // the contract locked in by tests/unit/compatible-provider-connections.test.js (#925).
     if (isOpenAICompatibleProvider(provider)) {
       const node = await getProviderNodeById(provider);
       if (!node) {
         return NextResponse.json({ error: "OpenAI Compatible node not found" }, { status: 404 });
-      }
-      const existingConnections = await getProviderConnections({ provider });
-      if (existingConnections.length > 0) {
-        return NextResponse.json({ error: "Only one connection is allowed for this OpenAI Compatible node" }, { status: 400 });
       }
       providerSpecificData = {
         prefix: node.prefix,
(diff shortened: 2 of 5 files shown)


decolua:
Thanks @MiQieR for the contribution! Reviewed and accepted — applying it to master. 🙏

Two small logic tweaks applied on top:
- Kept xai's validation as `status === 200 || 403` (its 403 = valid-but-no-credit) instead of folding it into the shared non-401/403 check, which would have started rejecting valid xai keys.
- For xiaomi-tokenplan, used `status !== 401` so a 403 (valid key without list permission) is treated as valid, matching the intent described in the PR.

PR #2279, opened 2026-07-01
fix(translator): prevent doubled tool args in OpenAI-to-Claude response

vishalrajv:
## Problem

When using OpenAI-compatible models (e.g. mimo) with Claude Code clients, tool calls frequently fail with `InputValidationError` because the model's arguments arrive as doubled JSON (e.g. `{query:x}{query:x}`).

## Root Cause

The OpenAI-compatible model sends **duplicate SSE chunks** with `finish_reason` set. The translator's finish block had no guard against this, so it ran **twice** — emitting tool arguments as two separate `input_json_delta` events. Claude Code concatenated both halves, producing invalid doubled JSON.

## Changes

- **`openai-to-claude.js`**: Add `!state.finishReason` guard to prevent double finish emission
- **`openai-to-claude.js`**: Add `deduplicateDoubledJson()` as defense-in-depth in `sanitizeToolArgs`
- **`claude-to-openai.js`**: Guard string tool input on request side to prevent double-encoding
- **`openai-to-claude.test.js`**: Fix test state (add missing `nextBlockIndex`, `finish_reason`)

## Testing

- All 12 translator unit tests pass
- Verified with live Claude Code + mimo-v2.5-free — tool calls now work correctly


bloodf:
Heads-up: this guard regresses the multi-format pivot paths.

The `!state.finishReason` guard added here keys off `state.finishReason`, which is a **shared** translator-state field that upstream pivot stages also write. In `Antigravity/Gemini → OpenAI → Claude` (and any `X → OpenAI → Claude` pivot whose first stage sets `state.finishReason`), `gemini-to-openai.js` sets `state.finishReason` — for `stream.js` usage injection — *before* `openaiToClaudeResponse` runs. So by the time the finish chunk reaches openai-to-claude, `state.finishReason` is already truthy → the finish block is skipped → the buffered tool-call arguments are never flushed as `input_json_delta`, and `content_block_stop`/`message_stop` are never emitted. Tool calls then reach Claude clients with empty `input`.

Minimal repro — `translateResponse(ANTIGRAVITY, CLAUDE, …)` with a single `functionCall` emits only `message_start` + `content_block_start` (no `input_json_delta`, no stops).

A fix that preserves the dedupe intent without the cross-stage collision is to use a Claude-specific dedupe flag instead of the shared field:

```js
// before
if (choice.finish_reason && !state.finishReason) {
// after
if (choice.finish_reason && !state.claudeFinishHandled) {
  state.claudeFinishHandled = true;
```

Keep `state.finishReason = choice.finish_reason;` inside the block for `stream.js`. This still fixes the mimo double-finish case while not being suppressed by an upstream pivot stage.

vishalrajv:
Addressed the regression flagged in the review.

**Fix:** Replaced the shared `!state.finishReason` guard with a Claude-specific `!state.claudeFinishHandled` flag. This prevents the finish block from being suppressed when an upstream pivot stage (e.g. Antigravity/Gemini → OpenAI → Claude) sets `state.finishReason` before `openaiToClaudeResponse` runs.

- `state.claudeFinishHandled = true` is set inside the finish block (prevents double emission)
- `state.finishReason = choice.finish_reason` is kept inside the block for `stream.js` usage injection

All translator tests pass (golden-response-stream, format-roundtrip, bugs-antigravity, bugs-openai-bridge).

vishalrajv:
Hi @decolua, just a gentle reminder about this PR when you get a chance. It fixes doubled tool arguments that cause InputValidationError when using OpenAI-compatible models with Claude Code clients. All translator tests pass. Happy to address any feedback or make adjustments. Thank you for your time!

PR #2296, opened 2026-07-02
fix(translator): strict Anthropic content block compliance (#2225)

### LINKED ISSUE #2225: Antigravity Provider fails with Google AI Pro account in Claude Code

SahrulRamadhanHardiansyah:
## Summary
Fixes #2225 — Anthropic Claude API rejects requests routed via the Antigravity provider due to malformed content blocks in the messages array.

## Root Cause
The `antigravity-to-openai.js` request translator produced payloads that violate Anthropic's strict Messages API requirements:

1. **Empty text blocks from `thoughtSignature`-only parts** — A part with `{ thoughtSignature: "sig", text: "" }` emitted `{ type: "text", text: "" }` which Anthropic rejects as an invalid content block.
2. **Dropped tool calls when `functionResponse` + `functionCall` coexist** — When a single Gemini content block contained both, the early `return toolResults` discarded the `functionCall` entries entirely, causing downstream "Content block not found" errors.
3. **Empty regular text parts leaking through** — Parts with `text: ""` were not filtered, producing empty text blocks in the final Claude payload.

### `open-sse/translator/request/antigravity-to-openai.js`
- Filter `thoughtSignature` parts where `text` is empty (only emit when non-empty)
- When `functionResponse` and `functionCall` coexist in the same content block, emit both: tool result messages first, then an assistant message with the tool calls
- Guard regular text parts to skip `text: ""` before they reach the Claude API

### `tests/translator/bugs-antigravity.test.js`
- Changed `it.fails(...)` to `it(...)` for the previously known-failing mixed content test

## Test Results
```text
 ✓ tests/unit/openai-to-claude-response-tools.test.js (2 tests)
 ✓ tests/translator/bugs-antigravity.test.js (5 tests)
 ✓ tests/translator/claude-kiro-direct.test.js (9 tests)

 Test Files  3 passed (3)
      Tests  16 passed (16)
```
Previously: 15 passed + 1 expected fail → Now: **16 passed** (known-failing test now passes).

### DIFF
--- a/open-sse/translator/request/antigravity-to-openai.js
+++ b/open-sse/translator/request/antigravity-to-openai.js
@@ -138,12 +138,14 @@ function convertContent(content) {
 
     // Text with thoughtSignature = regular text after thinking
     if (part.thoughtSignature && part.text !== undefined) {
-      textParts.push({ type: OPENAI_BLOCK.TEXT, text: part.text });
+      if (part.text) {
+        textParts.push({ type: OPENAI_BLOCK.TEXT, text: part.text });
+      }
       continue;
     }
 
     // Regular text
-    if (part.text !== undefined) {
+    if (part.text !== undefined && part.text !== "") {
       textParts.push({ type: OPENAI_BLOCK.TEXT, text: part.text });
     }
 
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @SahrulRamadhanHardiansyah for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2297, opened 2026-07-02
Fix Kiro IdC auth: regional CodeWhisperer routing + drop placeholder ARN

lossless1:
AWS IAM Identity Center (authMethod=idc) connections failed every request
with 403 "bearer token invalid". Three causes, all now handled the same way
api_key/external_idp already were:

- executors/kiro.js: idc tokens were routed to runtime.us-east-1.kiro.dev
  first, which rejects AWS SSO tokens (403 halts before the CodeWhisperer
  host is tried). Now idc uses the *.amazonaws.com surface, region-aware from
  credentials.region instead of hardcoded us-east-1.
- openai-to-kiro.js / claude-to-kiro.js: idc fell back to the shared builder-id
  placeholder profileArn (another account's ARN) -> 403. Now idc/external_idp
  send the resolved ARN or empty, like api_key.

### DIFF
--- a/open-sse/executors/kiro.js
+++ b/open-sse/executors/kiro.js
@@ -64,9 +64,22 @@ export class KiroExecutor extends BaseExecutor {
   getOrderedBaseUrls(credentials) {
     const baseUrls = this.getBaseUrls();
     const authMethod = credentials?.providerSpecificData?.authMethod;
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @lossless1 for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2298, opened 2026-07-02
fix(chat): strip orphaned tool results before dispatch (#2236)

### LINKED ISSUE #2236: [Bug] Orphaned tool results are not stripped consistently across request formats

anki1kr:
## Problem

Closes #2236.

Client-side history truncation (summarisation, context-window management) can remove an **assistant turn that contained `tool_calls` / `tool_use`** while leaving the corresponding tool result sitting in the history. Strict upstream APIs then reject the whole request with a **400 before the model executes**:

```
No tool call found for function call output with call_id <id>
```

The same invariant appears in all four wire formats 9router handles:

| Format | Call carrier | Result carrier | Pairing key |
|--------|-------------|----------------|-------------|
| OpenAI Chat Completions | `assistant.tool_calls[].id` | `role:"tool"`, `tool_call_id` | `tool_call_id` ↔ `id` |
| OpenAI Responses API | `input[type=function_call].call_id` | `input[type=function_call_output].call_id` | `call_id` ↔ `call_id` |
| Anthropic Messages | `content[type=tool_use].id` | `content[type=tool_result].tool_use_id` | `tool_use_id` ↔ `id` |
| Gemini / Antigravity | `parts[].functionCall.id\|name` | `parts[].functionResponse.id\|name` | `id` or `name` |

## Fix

**`open-sse/translator/concerns/toolCall.js`** — adds `stripOrphanedToolResults(body)`:
- Collects every live tool-call id present in the request
- Removes any result item whose call id has no matching call — across all four formats
- Mutates body in-place; returns count of removed items (0 = nothing changed)

**`open-sse/handlers/chatCore.js`** — calls it twice:
1. **Pre-translation** — removes client-sent orphans before format conversion
2. **Post-RTK/Headroom** — removes orphans that compression creates by dropping assistant turns mid-history

Both call-sites log at `debug` level (`TOOLCLEAN`) so operators can observe when stripping fires.

## Tests

`tests/unit/strip-orphaned-tool-results.test.js` — **15 tests, all passing**:
- OpenAI: orphan removed, paired kept, mixed, no-tool-messages
- Anthropic: `tool_result` block orphan removed, paired kept
- Responses API: `function_call_output` orphan removed, paired kept, mixed
- Gemini: `functionResponse` orphan removed, paired kept, name-based fallback
- Edge cases: empty body, null messages, result-without-id (pass-through)

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -19,6 +19,7 @@ import { handleNonStreamingResponse } from "./chatCore/nonStreamingHandler.js";
 import { handleStreamingResponse, buildOnStreamComplete } from "./chatCore/streamingHandler.js";
 import { detectClientTool, isNativePassthrough } from "../utils/clientDetector.js";
 import { dedupeTools } from "../utils/toolDeduper.js";
+import { stripOrphanedToolResults } from "../translator/concerns/toolCall.js";
 import { injectCaveman } from "../rtk/caveman.js";
 import { injectPonytail } from "../rtk/ponytail.js";
 import { compressMessages, formatRtkLog } from "../rtk/index.js";
@@ -119,6 +120,13 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     } catch (e) { log?.warn?.("MODALITY", `image prefetch failed: ${e.message}`); }
   }
 
+  // Strip orphaned tool results before translation so the translator never sees
+  // stale call_id references that client-side history truncation left behind.
+  const preStripped = stripOrphanedToolResults(body);
+  if (preStripped > 0) {
+    log?.debug?.("TOOLCLEAN", `pre-translation: stripped ${preStripped} orphaned tool result(s)`);
+  }
+
   let translatedBody;
   let toolNameMap;
   if (passthrough) {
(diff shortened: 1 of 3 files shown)


warelik:
Nice work on the Gemini `functionResponse` coverage and the post-compression call site — both are gaps in my approach on #2237.

A heads-up: the same Kiro regression that @bloodf flagged on #2237 applies here too. Both PRs delete orphaned tool results, which preempts `claude-to-kiro.js`'s `reconcileOrphanedToolResults` salvage path. I'm updating #2237 to salvage-to-text (fold orphan content into `[Tool result: ...]` user text instead of deleting) per bloodf's option 2.

Also worth noting from the Copilot review: the live-id collection gathers `tool_use` ids from any message role, not just assistant turns. A user/system message containing `{type:"tool_use"}` would incorrectly mark that id as live and could prevent orphaned `tool_result` blocks from being stripped. The fix is to gate on `msg.role === "assistant"` for the `tool_use` scan.

Would be good to coordinate — happy to fold your Gemini coverage and post-compression call site into #2237's salvage approach, or vice versa. The end goal is one merged solution, not two competing PRs.

PR #2313, opened 2026-07-03
feat(kenari): add Kenari OpenAI-compatible gateway provider

doedja:
### What

Adds [Kenari](https://kenari.id) as an API-key provider. Kenari is an Indonesian OpenAI-compatible AI gateway billed in Rupiah (IDR): one `kn-` API key covers the whole catalog (Claude, GPT, DeepSeek, GLM, Kimi and more).

### How

- New registry entry `open-sse/providers/registry/kenari.js`, wired into `registry/index.js`.
- Standard `Authorization: Bearer` auth and OpenAI format (registry defaults), so no executor or translator changes.
- Live model list via `modelsFetcher` on the public `https://kenari.id/v1/models` catalog (no key needed) plus `passthroughModels: true`, mirroring the vercel-ai-gateway pattern. `validateUrl` powers the dashboard key test.

### Verified

- `https://kenari.id/v1/models` returns the catalog with no auth (24 models).
- `POST https://kenari.id/v1/chat/completions` returns 401 without a key, 200 with a `kn-` key.
- Registry loads: `PROVIDERS.kenari` resolves with `format: "openai"`.
- Docs: https://kenari.id/docs, OpenAPI: https://kenari.id/openapi.json.

Disclosure: I run kenari.id.

### DIFF
--- a/open-sse/providers/registry/index.js
+++ b/open-sse/providers/registry/index.js
@@ -119,6 +119,7 @@ import p116 from "./tokenrouter.js";
 import p117 from "./selfhosted-stt.js";
 import p118 from "./selfhosted-tts.js";
 import p119 from "./selfhosted-embedding.js";
+import p120 from "./kenari.js";
 
 export default [
   p0,
@@ -239,4 +240,5 @@ export default [
   p117,
   p118,
   p119,
+  p120,
 ];
--- a/open-sse/providers/registry/kenari.js
+++ b/open-sse/providers/registry/kenari.js
@@ -1,0 +1,24 @@
+export default {
+  id: "kenari",
+  priority: 60,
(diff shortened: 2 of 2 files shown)


doedja:
Rebased onto current master (v0.5.50), so this is mergeable again.

Kenari now registers as p120. p100 was free when this was opened but belongs to codebuddy-intl now, and the slots through p119 filled up with poolside, tokenrouter and the selfhosted STT/TTS/embedding providers, so I took the next free integer.

Diff is still just the two registry files, 26 lines added. Locally: node --check clean on both, the registry import resolves 116 entries, and the PROVIDERS build produces kenari with the expected baseUrl and the openai format default.

Issue #2257, opened 2026-06-30
Kiro API Key: Claude models return INVALID_MODEL_ID while GLM works (IDE/Web works with same account)


mosabbir-maruf:
Hi,

I'm running the latest 9Router source (also tested with my latest Docker image), and I'm having an issue with the Kiro provider.

GLM models work perfectly, but every Claude model returns:

```json
{
  "message": "Invalid model ID. Please select a different model to continue.",
  "reason": "INVALID_MODEL_ID"
}

HTTP 400: [kiro/claude-sonnet-4.5] [400]: {"message":"Invalid model ID. Please select a different model to continue.","reason":"INVALID_MODEL_ID"} (reset after 30s)

VitzS7:
its because need "-": kr/claude-sonnet-4-5

mosabbir-maruf:
I just tested kr/claude-sonnet-4-5 as well, but it returns the exact same INVALID_MODEL_ID error. So it doesn’t seem to be related to the model alias format.

VitzS7:
this is a problem, because the model in kiro is using "-".

mosabbir-maruf:
Thanks for looking into this.

I tested both:

kr/claude-sonnet-4.5
kr/claude-sonnet-4-5

Unfortunately both return the same INVALID_MODEL_ID error.

I also checked the request logs generated by 9Router. The request sent upstream contains the exact model ID that I configured, so it doesn't look like a client-side alias formatting issue on my setup.

I'll inspect the generated request (4_req_target.json) again to confirm exactly which model and modelId are being sent for the dash variant, and I'll report back with the results.

VitzS7:
i using a proxy to fix this, are my creation if u want i put in github

Issue #2306, opened 2026-07-02
fix: system prompt injected as raw user message when routing Claude Code to Kiro


VitzS7:
## Problem

When Claude Code CLI sends a request through 9router to the Kiro/CodeWhisperer backend, the `system` field from the Anthropic API format gets converted to a plain `role: user` message **without any wrapper**, causing it to appear as raw text inside the conversation context indistinguishable from actual user input.

This results in the full Claude Code system prompt (including environment info, tool definitions, memory system instructions, billing headers, etc.) being visible as if the user had typed it — polluting the model context and causing confusion.

## Root Cause

In `open-sse/translator/request/openai-to-kiro.js`, the `convertMessages` function:

```js
// Normalize: system/tool -> user
if (role === ROLE.SYSTEM || role === ROLE.TOOL) {
  role = ROLE.USER;
}
```

This simply reassigns the role without wrapping the content, so the system prompt ends up as plain user text in the Kiro/CodeWhisperer payload.

## Expected Behavior

The system prompt should be wrapped in `<system-reminder>` tags before being merged into the user message — exactly as the project already does for mid-conversation system messages in `claude-to-openai.js`:

```js
function systemReminderText(content) {
  return `<system-reminder>
${text}
</system-reminder>`;
}
```

## Proposed Fix

In `open-sse/translator/request/openai-to-kiro.js`, wrap system message content in `<system-reminder>` tags before converting to user role:

```js
if (role === ROLE.SYSTEM) {
  role = ROLE.USER;
  if (typeof msg.content === 'string') {
    msg = { ...msg, content: `<system-reminder>
${msg.content}
</system-reminder>` };
  } else if (Array.isArray(msg.content)) {
    const wrappedText = msg.content
      .filter(c => c.type === OPENAI_BLOCK.TEXT || c.text)
      .map(c => c.text || '')
      .filter(Boolean)
      .join('
');
    if (wrappedText) {
      msg = { ...msg, content: `<system-reminder>
${wrappedText}
</system-reminder>` };
    }
  }
}
if (role === ROLE.TOOL) {
  role = ROLE.USER;
}
```

## Environment

- 9router version: 0.5.15
- Client: Claude Code CLI (`@anthropic-ai/claude-code`)
- Provider: `kiro` (AWS CodeWhisperer)

## Impact

[...]

PR #2315, opened 2026-07-03
fix(kiro): convert Claude model ID dots to dashes before upstream dispatch

### LINKED ISSUE #2308: bug: Kiro provider rejects Claude models with INVALID_MODEL_ID due to dot vs dash model ID format
## Problem

When routing Claude Code CLI requests through 9router to the Kiro provider, all Claude models return:

```json
{
  "message": "Invalid model ID. Please select a different model to continue.",
  "reason": "INVALID_MODEL_ID"
}
```

`HTTP 400: [kiro/claude-sonnet-4.5] Invalid model ID`

Non-Claude models (e.g. GLM, DeepSeek) work fine through the same provider.

## Root Cause

The 9router expects model IDs in **dot** notation (e.g. `claude-sonnet-4.5`), but the Kiro/CodeWhisperer backend requires **dash** notation (e.g. `claude-sonnet-4-5`). When the dot format reaches the Kiro API, it rejects it with `INVALID_MODEL_ID`.

Claude Code CLI sends models with **dashes** (e.g. `claude-sonnet-4-5-thinking`) because its capability-gating logic uses `.includes()` on the model string to detect features. The 9router then converts dashes to dots for its internal routing, but that dot format is what gets sent to Kiro — which rejects it.

The model ID translation chain is:
```
Claude Code CLI: claude-sonnet-4-5-thinking  (dashes)
9router internal: claude-sonnet-4.5          (dots — for routing)
Kiro API expects: claude-sonnet-4-5          (dashes — original format)
```

## Suggested Fix

When the target provider is `kiro`, the 9router should send the **dash** format to the upstream Kiro API instead of the dot format. The dot format can remain for internal routing/aliasing, but the final model ID sent to the Kiro backend should use dashes.

The current workaround is running a separate proxy (kr-cc-proxy) that translates dots back to dashes before forwarding to 9router — but this should be handled by 9router itself.

## Environment

- 9router version: 0.5.15
- Provider: kiro (AWS CodeWhisperer)
- OS: Windows 11

## Related

- #2257 (same INVALID_MODEL_ID error reported by another user)
- Workaround proxy: kr-cc-proxy translates `claude-opus-4.8` back to `claude-opus-4-8` before sending to Kiro

anki1kr:
## Problem

When routing requests to the Kiro / CodeWhisperer backend, every Claude model returns:

```json
{ "message": "Invalid model ID. Please select a different model to continue.", "reason": "INVALID_MODEL_ID" }
```

HTTP 400: `[kiro/claude-sonnet-4.5] Invalid model ID`

Non-Claude models (DeepSeek, GLM, Qwen) work correctly through the same provider.

## Root Cause

9router stores model IDs with dot notation for the version number (`claude-sonnet-4.5`), but the Kiro/CodeWhisperer API only accepts dash notation (`claude-sonnet-4-5`). Both translators (`openai-to-kiro.js` and `claude-to-kiro.js`) forwarded the dot-format ID as `modelId` directly to the upstream API.

## Fix

Added `toKiroModelId(upstream)` to `kiroConstants.js` that replaces dots with dashes in Claude model IDs only:

```js
// claude-sonnet-4.5  →  claude-sonnet-4-5
// deepseek-3.2       →  deepseek-3.2 (unchanged)
```

The function is applied in both translators right after `resolveKiroModel()` strips synthetic suffixes (`-agentic`, `-thinking`), so the Kiro wire format always receives the correct dash-notation ID.

## Tests

8 new unit tests in `tests/unit/kiro-model-id-format.test.js` — all pass:
- Dot→dash conversion for Claude models
- Non-Claude models unchanged
- Idempotency on already-correct IDs
- Full pipeline: `resolveKiroModel` + `toKiroModelId`

Fixes #2308. Related: #2257.

### DIFF
--- a/open-sse/config/kiroConstants.js
+++ b/open-sse/config/kiroConstants.js
@@ -195,6 +195,24 @@ export function stripThinkingSuffix(model) {
 }
 
 /**
+ * Normalise a 9router model id to the wire format the Kiro API expects.
+ * Kiro rejects Claude model IDs that use dot notation for the version number
+ * (e.g. "claude-sonnet-4.5") — it only accepts dash notation ("claude-sonnet-4-5").
+ * Non-Claude models (deepseek-3.2, MiniMax-M2.5) are passed through unchanged
+ * because their dots are semantically part of the stable upstream model name and
+ * those models work correctly with dots.
+ *
+ * @param {string} upstream Model id with synthetic suffixes already stripped
+ * @returns {string} Kiro-wire model id
+ */
+export function toKiroModelId(upstream) {
+  if (typeof upstream === "string" && upstream.startsWith("claude-")) {
+    return upstream.replace(/\./g, "-");
+  }
+  return upstream;
+}
+
+/**
  * Resolve a 9router model id to the real upstream Kiro model id, plus flags
  * describing which behaviours the suffixes implied.
  *
--- a/open-sse/translator/request/claude-to-kiro.js
+++ b/open-sse/translator/request/claude-to-kiro.js
@@ -27,6 +27,7 @@ import { FORMATS } from "../formats.js";
 import { v4 as uuidv4 } from "uuid";
 import {
   resolveKiroModel,
+  toKiroModelId,
   resolveKiroThinkingBudget,
   buildThinkingSystemPrefix,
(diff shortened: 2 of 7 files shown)


VitzS7:
this fix the #2306? @anki1kr

PR #2325, opened 2026-07-03
fix(github): use goldeneye-free-auto as default model for Copilot free tier (#2276)

### LINKED ISSUE #2276: No model available for free acc GithubCopilot
Image

Image

Image

Image

Image

anki1kr:
## Problem

Free-tier GitHub Copilot users see "No model available" after connecting (#2276).

The root cause: `goldeneye-free-auto` is GitHub's auto-routing model that works on ALL Copilot tiers (free and paid alike), but it was listed **last** in the model registry. The default model — used for connection tests and the model selector — was `gpt-5.2`, which free accounts can't access (returns 403). So the connection test fails, the connection is marked unavailable, and the dashboard shows "No model available."

## Fix

Move `goldeneye-free-auto` to **first** in the model list so it becomes:
- The default for connection tests (free users now pass)
- The default in the model selector (free users can send messages immediately)

Paid users still see and can select all premium models below it.

Display label changed to `"GoldenEye (Auto)"` to communicate that this model auto-selects the best model for the user's current plan.

## Impact

- Free tier: connection test now uses `goldeneye-free-auto` → passes
- Paid tier: `goldeneye-free-auto` auto-routes to the best model anyway; no regression
- Model list order: `goldeneye-free-auto` first, then premium models as before

Fixes #2276

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -214,6 +214,9 @@ function killAllAppProcesses(appPort) {
     try {
       // Kill MIT first (privileged process, needs special handling)
       killProxyByPidFile();
+      // Kill Headroom proxy by PID file — detached process that outlives the main server.
+      // Must stop before npm rename; it holds a handle on the app/ directory on Windows (#2265).
+      killByPidFile(path.join(getAppDataDir(), "headroom", "proxy.pid"));
       // Kill cloudflared/tailscale by PID file (precise, only this app's tunnel)
       killTunnelByPidFile();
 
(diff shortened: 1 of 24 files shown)


floydous:
My GitHub account with student plan cannot use the model as you mentioned. Did I do something wrong?

image

phamhuyhien335:
> My GitHub account with student plan cannot use the model as you mentioned. Did I do something wrong?
> 
> image

same issues, so cannot using github copilot free/student with 9router

PR #2344, opened 2026-07-03
fix(codex): fallback on SSE capacity errors

ryanngit:
## Summary
- detects Codex capacity errors embedded in HTTP 200 SSE bodies
- converts capacity/model_at_capacity to a 503 response so existing account fallback rotates credentials
- preserves normal SSE streams after peeking

## Checks
- node assertion covering capacity detection and normal SSE reassembly
- git diff --check


ryanngit:

PR #2366, opened 2026-07-04
fix(kiro): system prompt delivery, Opus 4.8 models, model id normalization

### LINKED ISSUE #2306: fix: system prompt injected as raw user message when routing Claude Code to Kiro

VitzS7:
## What this fixes

This PR addresses three issues with the Kiro provider:

### 1. System prompt not reaching the model (closes #2306)

When Claude Code CLI sends requests through 9router to Kiro, the system prompt was being wrapped in `<system-reminder>` tags before merging into the user turn. Claude models are trained to treat `<system-reminder>` as informational-only context that must NOT override behavior, so the system prompt was effectively ignored.

**Fix:** Use the native `systemInstruction` field on `userInputMessage` that the Kiro/Q API supports. The model now receives the system prompt as a real system prompt, not user-authored text. Also adds `<instructions>` fallback in user content for upstreams that do not support the native field.

**Verified:** Tested with a directive to prefix every response with "PIZZA:" and confirmed the model follows it:
```
PIZZA: 2+2 = 4.
```

Also replaced `<system-reminder>` with `<instructions>` in `claude-to-openai.js` and `openai-to-kiro.js` translators for consistency.

### 2. Missing Kiro models

The registry was missing several models now available in Kiro (per kiro.dev/changelog/models and kiro.dev/docs/models):

- Claude Opus 4.8 (base, thinking, agentic, thinking+agentic)
- Claude Opus 4.7 (base, thinking, agentic, thinking+agentic)
- Claude Opus 4.5 (base, thinking, agentic, thinking+agentic)

### 3. Model id dash/dot normalization (Kiro only)

Clients often send Kiro model ids with dashes (`claude-sonnet-4-5`, `claude-opus-4-8-thinking`) while the registry uses dots (`claude-sonnet-4.5`, `claude-opus-4.8-thinking`). Added `normalizeModelId()` that converts hyphens between digits to dots while preserving word/suffix hyphens (`-thinking`, `-agentic`, `qwen3-coder-next`).

This normalization is scoped to the Kiro provider only (`DOT_VERSION_PROVIDERS = new Set(["kr", "kiro"])`). All other providers keep exact-match lookup, so ids that legitimately use hyphens elsewhere are never rewritten. Applied via a shared `findModel(models, modelId, aliasOrId)` helper across the registry lookups in `providerModels.js`.

## Files changed

| File | Change |
|------|--------|
| `providers/registry/kiro.js` | Add Opus 4.8, 4.7, 4.5 models |
| `providers/models/schema.js` | Add `normalizeModelId()` |
| `config/providerModels.js` | Add `findModel()` with Kiro-scoped dash/dot tolerance |
| `translator/request/claude-to-kiro.js` | Native `systemInstruction` field |
| `translator/request/claude-to-openai.js` | `<instructions>` instead of `<system-reminder>` |
| `translator/request/openai-to-kiro.js` | `<instructions>` for system messages |
| `tests/unit/model-name-regex.test.js` | New test cases for normalization |

### DIFF
--- a/open-sse/config/providerModels.js
+++ b/open-sse/config/providerModels.js
@@ -18,11 +18,18 @@ export function getDefaultModel(aliasOrId) {
   return models?.[0]?.id || null;
 }
 
-// Find a registry entry by id, tolerating dash/dot version separators ("claude-sonnet-4-5" ~= "claude-sonnet-4.5").
-function findModel(models, modelId) {
+// Providers whose registry uses dots in version numbers (e.g. "claude-sonnet-4.5").
+// For these, we tolerate clients sending dashes ("claude-sonnet-4-5") by normalizing
+// digit-hyphen-digit to digit-dot-digit before lookup. Other providers are left untouched.
+const DOT_VERSION_PROVIDERS = new Set(["kr", "kiro"]);
+
+// Find a registry entry by id. For Kiro models, tolerates dash/dot version separators
+// ("claude-sonnet-4-5" ~= "claude-sonnet-4.5"). Other providers use exact match only.
+function findModel(models, modelId, aliasOrId) {
   if (!models) return undefined;
   const found = models.find(m => m.id === modelId);
   if (found) return found;
+  if (!DOT_VERSION_PROVIDERS.has(aliasOrId)) return undefined;
   const normalized = normalizeModelId(modelId);
   if (normalized === modelId) return undefined;
   return models.find(m => m.id === normalized);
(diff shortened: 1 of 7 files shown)


decolua:
Thanks @VitzS7 for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2372, opened 2026-07-04
fix(headroom): proxy dashboard through app

### LINKED ISSUE #2330: [Bug] Headroom dashboard cant access token saver data
I run headroom on tokensaver:

Image

But when I tried to access its dashboard I got status of error:

Image

And seems like the headroom is working fine when it is being start via Token Saver:

Image

Jordannst:
## Issue

Starting Headroom from the Token Saver modal can successfully launch the local Headroom proxy, but opening the Headroom dashboard from a browser that is not running on the same host can still fail to load token-saver data.

The dashboard page itself may load, but its data widgets cannot reach Headroom, so the dashboard looks broken even though Headroom is running.

## Findings

Headroom serves the dashboard HTML from its own proxy and the page uses root-relative fetch calls for dashboard data:

- `/stats?cached=1`
- `/health`
- `/stats-history`
- `/transformations/feed`

That works when the browser opens Headroom directly at `http://localhost:8787/dashboard` on the same machine.

It does not work reliably when 9Router starts Headroom on the server and the user opens the dashboard from another device/browser through the 9Router app. In that case, those root-relative requests resolve against the 9Router origin instead of the Headroom origin.

## Root cause

9Router managed the Headroom process server-side, but did not provide a browser-safe path for the Headroom dashboard and its data endpoints.

So the browser could end up requesting dashboard data from 9Router paths such as `/stats` and `/health`, while the actual endpoints live on the Headroom proxy.

## Fix

This PR adds a 9Router-side Headroom dashboard proxy:

- proxy Headroom requests through `/api/headroom/proxy/[...path]`
- add an `Open Headroom Dashboard` link in the Token Saver modal
- open the dashboard at `/api/headroom/proxy/dashboard`
- rewrite Headroom dashboard HTML fetch calls so dashboard data stays on the proxy path:
  - `/stats` → `/api/headroom/proxy/stats`
  - `/health` → `/api/headroom/proxy/health`
  - `/stats-history` → `/api/headroom/proxy/stats-history`
  - `/transformations/feed` → `/api/headroom/proxy/transformations/feed`

## Why this approach

This keeps Headroom private to the 9Router host while still giving the user's browser a same-origin URL that can reach both the dashboard page and its backing data endpoints.

It also avoids requiring users to manually expose `localhost:8787` or know whether Headroom is running locally, in Docker, or on the 9Router server.

## Verification

- `git diff --check`
- `npm run build`

Closes #2330

### DIFF
--- a/src/app/(dashboard)/dashboard/token-saver/TokenSaverClient.js
+++ b/src/app/(dashboard)/dashboard/token-saver/TokenSaverClient.js
@@ -374,6 +374,16 @@ export default function TokenSaverClient() {
               {headroomStatusLabel}
             </span>
           </div>
+          {headroomRunning && (
+            <a
+              href="/api/headroom/proxy/dashboard"
+              target="_blank"
+              rel="noreferrer"
+              className="w-full rounded border border-border px-4 py-2 text-center text-sm hover:bg-surface-2"
+            >
+              Open Headroom Dashboard
+            </a>
+          )}
           <div className="flex flex-col gap-1">
             <p className="text-sm font-medium">Proxy URL</p>
             <Input
--- a/src/app/api/headroom/proxy/[...path]/route.js
+++ b/src/app/api/headroom/proxy/[...path]/route.js
@@ -1,0 +1,98 @@
+import { NextResponse } from "next/server";
+import { getSettings } from "@/lib/localDb";
+import { DEFAULT_HEADROOM_URL } from "@/lib/headroom/detect";
+
+export const dynamic = "force-dynamic";
+
+const HOP_BY_HOP_HEADERS = new Set([
+  "connection",
+  "keep-alive",
+  "proxy-authenticate",
+  "proxy-authorization",
+  "te",
+  "trailer",
+  "transfer-encoding",
+  "upgrade",
+]);
+
+const DASHBOARD_PREFIX = "/api/headroom/proxy";
+
+async function getTargetBase() {
+  const settings = await getSettings();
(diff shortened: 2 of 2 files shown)


decolua:
Thanks @Jordannst for the contribution! Reviewed and accepted — applying it to master. During review I added two small hardening tweaks: gating `/api/headroom/proxy` as LOCAL_ONLY (matching `start`/`stop`) and stripping `cookie`/`authorization` when the Headroom target is non-loopback, so viewer credentials aren't forwarded to an external host. 🙏

tcno4:
> Thanks @Jordannst for the contribution! Reviewed and accepted — applying it to master. During review I added two small hardening tweaks: gating `/api/headroom/proxy` as LOCAL_ONLY (matching `start`/`stop`) and stripping `cookie`/`authorization` when the Headroom target is non-loopback, so viewer credentials aren't forwarded to an external host. 🙏感谢您的贡献！已经审核并通过了您的建议，将会应用到我们的主代码中。在审核过程中，我添加了两个小的改进：将 gating `/api/headroom/proxy` 设置为 LOCAL_ONLY（与 `start` / `stop` 一致）；当 Headroom 目标不是循环回环时，会移除 `cookie` / `authorization` ，这样观众的身份凭证就不会被转发到外部主机了。🙏

In which version will this feature be available?

Jordannst:
@tcno4 Looks like it’s already in v0.5.20. The applied commit `481e7e4` is included in that release.

PR #2381, opened 2026-07-05
fix(claude): reconcile max_tokens vs thinking budget and lift per-model ceiling

thienpvt:
## Summary

Fixes two related `max_tokens` defects on the **translated** (non-passthrough) request path to Claude-format providers. Native Claude Code → Anthropic passthrough is unaffected — it bypasses this pipeline entirely.

### 1. Budget collision → API 400 at `max` effort

`adjustMaxTokens` caps `max_tokens` **before** `applyThinking` sets `thinking.budget_tokens`, and the `claude-budget` thinking format maps `max` effort to `budget_tokens: 128000`. On a 64k-ceiling model this produced:

```
max_tokens: 64000, thinking.budget_tokens: 128000
```

Anthropic requires `max_tokens` **strictly greater** than `budget_tokens`, so the request 400s. The existing guard in `adjustMaxTokens` runs a step too early to catch it (no budget exists yet).

**Fix:** `prepareClaudeRequest` now reconciles the two *after* the budget is known — prefer raising `max_tokens` to preserve the requested thinking depth; only shrink the budget when it alone meets/exceeds the model ceiling.

### 2. Global 64000 cap ignored real model limits

The ceiling was `min(maxOutput, 64000)`, so high-output models were clamped to 64000 regardless of their real capability, and `max`-effort thinking could never use its full budget:
- `claude-budget` 128k models (`fable`/`mythos`)
- adaptive 128k models (`claude-opus-4.8`, `claude-sonnet-4.6`, `claude-sonnet-5`)

**Fix:** the ceiling is now the model's real `maxOutput`. `adjustMaxTokens` takes an optional `ceiling` arg (**default unchanged**, so gemini / antigravity / claude-to-openai callers are byte-identical), and `openai-to-claude` passes the model's `maxOutput` so a high-output model isn't pre-clamped before the model-aware step runs.

## End-to-end pipeline results

`translateRequest` (OpenAI→Claude) → `applyThinking` → `prepareClaudeRequest`:

| model | effort | max_tokens | budget_tokens | |
|---|---|---|---|---|
| claude-opus-4-20250514 (64k) | max | 64000 | 62976 | valid |
| claude-fable-5 (128k, budget) | max | **128000** | 126976 | valid |
| claude-opus-4-20250514 (64k) | xhigh, client max=16k | 33792 | 32768 | valid |
| claude-opus-4-20250514 (64k) | high | 64000 | 24576 | valid |
| claude-opus-4-8 (128k, adaptive) | max | 64000 | n/a (output_config) | valid |

Default 64k models still clamp over-large requests to 64000 — the lift is per-model, not global.

## Changes

- **`open-sse/translator/formats/maxTokens.js`** — `adjustMaxTokens(body, ceiling = DEFAULT_MAX_TOKENS)`; cap against the passed ceiling. Default preserves existing behavior for all current callers.
- **`open-sse/translator/request/openai-to-claude.js`** — pass the model's `maxOutput` as the ceiling.
- **`open-sse/translator/formats/claude.js`** — `prepareClaudeRequest` ceiling is now `maxOutput || DEFAULT_MAX_TOKENS` (was `min(maxOutput, 64000)`), plus the budget↔max_tokens reconciliation.

## Test plan

- **`tests/translator/bugs-toClaude-context.test.js`** — 8 new cases calling `prepareClaudeRequest` directly: max-effort budget shrink on a 64k model; xhigh with low client max (raise, preserve); high unchanged; adaptive/non-budget untouched; 128k model uses full budget at max; per-model clamp regression (64k still clamps, 128k allows through).
- Full pipeline spot-checked via `translateRequest` (table above) — every case yields `budget_tokens < max_tokens`.
- Verified **zero regressions**: the pre-existing failures in the broad suite (7 credential-gated `real/`/`e2e` tests, 3 `golden-url-header` snapshots, 4 `translator-request-normalization` cases) reproduce identically with this branch's source reverted to baseline.

### DIFF
--- a/open-sse/translator/formats/claude.js
+++ b/open-sse/translator/formats/claude.js
@@ -192,10 +192,27 @@ export function prepareClaudeRequest(body, provider = null, apiKey = null, conne
     delete body.output_config;
   }
 
-  // Clamp max_tokens to the model output ceiling (never above DEFAULT_MAX_TOKENS)
+  // Clamp max_tokens to the model's real output ceiling. Models whose caps
+  // declare a higher maxOutput (e.g. Opus 4.8 / Sonnet 4.6 = 128000) are allowed
+  // up to it, so max-effort thinking gets full budget; others fall back to the
+  // conservative 64000 default.
   if (body.max_tokens) {
-    const ceiling = Math.min(getCapabilitiesForModel(provider, body.model).maxOutput, DEFAULT_MAX_TOKENS);
+    const ceiling = getCapabilitiesForModel(provider, body.model).maxOutput || DEFAULT_MAX_TOKENS;
     if (body.max_tokens > ceiling) body.max_tokens = ceiling;
+
+    // Reconcile against thinking budget. applyThinking (thinkingUnified.js) runs
+    // AFTER adjustMaxTokens capped max_tokens, and the claude-budget format maps
+    // max effort → budget_tokens 128000 — larger than the clamped max_tokens.
+    // Anthropic requires max_tokens strictly greater than budget_tokens (else 400).
+    // Prefer raising max_tokens to preserve the requested thinking depth; if the
+    // budget alone meets/exceeds the ceiling, cap output and shrink the budget so
+    // some tokens remain for the answer.
+    if (body.thinking?.type === "enabled" && body.thinking.budget_tokens && body.thinking.budget_tokens >= body.max_tokens) {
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @thienpvt for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2409, opened 2026-07-06
feat(proxy-pools): auto-rotate strategy for no-auth providers

fdjrr:
## Summary

Add a rotation strategy selector to **NoAuthProxyCard** (provider detail page, e.g. OpenCode Free) that allows no-auth providers to automatically rotate requests across **all active proxy pools** — round-robin or random.

## Problem

Free-tier providers like OpenCode Free share public upstream IPs, making them prone to per-IP rate limits. Previously, users could only assign a **single** static proxy pool. When that proxy IP got rate-limited, requests would fail until the user manually switched pools. No fallback, no load distribution.

### UI — `NoAuthProxyCard.js`

- **Proxy Pool dropdown** preserved as-is — default mode "None (single pool)"
- **Rotation Strategy dropdown** added (None / Round-robin / Random)
- Pool dropdown is disabled with an explanatory hint when rotation is active
- Round-robin/Random options are disabled if fewer than 2 active pools exist
- Both values are persisted to `settings.providerStrategies[providerId]` and auto-save on change

### Backend — `auth.js`

```
getProviderCredentials("opencode")
  → strategy = override.rotateStrategy
  → if "none":         use override.proxyPoolId (existing behavior)
  → if "round-robin":  getProxyPools({isActive:true}) → pickProxyPoolId()
  → if "random":       getProxyPools({isActive:true}) → pickProxyPoolId()
  → resolveConnectionProxyConfig() → proxyAwareFetch()
```

### Shared utility — `connectionProxy.js`

New function `pickProxyPoolId(poolIds, strategy, providerId)`:
- **round-robin**: in-memory Map keyed by `providerId → { index }`, cycles on each call
- **random**: `Math.random()` per request
- **none**: returns poolIds[0] (fallback)

## Why in-memory?

The round-robin state is deliberately **not persisted** — resetting the index on server restart is fairer than resuming from a potentially biased last index. If a pool is deleted, the cycle automatically adjusts without stale DB state.

## Testing

This change only affects pool ID selection logic in the no-auth credential path. All API-key provider paths are untouched. The downstream proxy resolution (`resolveConnectionProxyConfig` → `proxyAwareFetch`) is completely unchanged.

## Files Changed

| File | Changes |
|------|---------|
| `src/shared/components/NoAuthProxyCard.js` | + rotation strategy dropdown, auto-save, pool selector disabled hint |
| `src/sse/services/auth.js` | + import `getProxyPools`, `pickProxyPoolId`; rotation vs single pool resolution |
| `src/lib/network/connectionProxy.js` | + `pickProxyPoolId()` with in-memory round-robin + random |

### DIFF
--- a/src/lib/network/connectionProxy.js
+++ b/src/lib/network/connectionProxy.js
@@ -6,6 +6,33 @@ function normalizeString(value) {
   return String(value).trim();
 }
 
+// ─── Proxy pool rotation state (in-memory) ─────────────────────────
+const rotateState = new Map(); // providerId → { index }
+
+/**
+ * Pick one proxy pool ID from a list based on strategy.
+ * round-robin: cycle sequentially (in-memory, resets on restart)
+ * random:      uniform random pick
+ * none/single: return first entry
+ */
+export function pickProxyPoolId(poolIds, strategy, providerId) {
+  if (!poolIds || poolIds.length === 0) return null;
+  if (poolIds.length === 1) return poolIds[0];
+
+  if (strategy === "round-robin") {
+    const state = rotateState.get(providerId) || { index: -1 };
+    state.index = (state.index + 1) % poolIds.length;
+    rotateState.set(providerId, state);
+    return poolIds[state.index];
+  }
+
+  if (strategy === "random") {
+    return poolIds[Math.floor(Math.random() * poolIds.length)];
+  }
+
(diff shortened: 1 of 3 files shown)


decolua:
Thanks for the contribution! 🙏 Reviewed and accepted — applying to master.

PR #2418, opened 2026-07-06
feat(vinext): migrate from Next.js to vinext (Vite) build pipeline IVT-0517

nhattran998:
## Summary

- Migrates the web app from Next.js to **vinext** (Cloudflare's Vite-based reimplementation of the Next.js App Router). `bun run dev` (Next) still works unchanged; the vinext paths are additive.
- All 27 pages, 131 API route handlers, and 8 rewrites serve correctly in both **dev** and **prod** (verified end-to-end).
- Ports the security-critical trusted-client-IP middleware (the old `custom-server.js` hook) to the vinext runtime, since it doesn't run under vinext's own server.

## Changes

- **`build` — vite/vinext pipeline**: add `vite.config.ts` (`appDir: "src"` — vinext appends `/app`, so this points at the dir containing `app/`; plus `@`/`open-sse` aliases, watch-ignore, and the trusted-IP dev middleware). Add vinext/vite/plugin deps + `chalk` (transitive under Next, missing under Vite). Drop the `webpack` fn + `experimental.proxyClientMaxBodySize`/`serverComponentsHmrCache` from `next.config.mjs` (vinext streams bodies unbounded; Vite replaces webpack). New `dev:vinext` / `build:vinext` / `start:vinext` / `start:vinext:custom` scripts.
- **`refactor(mitm)` + `refactor(api)`** — convert 15 CJS files in the web app's import graph to real ESM (`require`→`import`, `module.exports`→`export`, hoist inline dynamic requires in `manager.js`, add an ESM `__dirname` shim). Next auto-transpiled these; Vite/Rolldown parses them as ESM since the root `package.json` has `"type": "module"`. The two route handlers' inline `require("better-sqlite3")` / `require("child_process").execSync` were hoisted to static imports.
- **`fix(layout)`** — `@next/third-parties/google`'s `<GoogleAnalytics>` depends on `next/script`, which does not render under vinext SSR (caused a 500 on every page). Replaced with an equivalent inline `gtag` snippet.
- **`feat(security)`** — `custom-server.js` injected `x-9r-real-ip` / `x-9r-via-proxy` (consumed by `loginLimiter` + `dashboardGuard`) by monkey-patching Next's standalone `http.createServer`. That hook does not run under vinext. Extracted the logic to `src/lib/clientIp.js` and wired it into both modes: `server.vinext.js` (prod) uses `server.prependListener("request", ...)` on the `http.Server` returned by `startProdServer`; the vite plugin `9router:trusted-client-ip` does the dev equivalent as connect middleware. The original `custom-server.js` is left intact for the Next.js path.
- **`fix(mitm)`** — `node-machine-id` only exports a default; switched its named import to default-import + destructure for bare-Node ESM compatibility.

## Quality Gates

- [x] `deslop` skill run — nothing to remove (all comments are WHY/contract)
- [x] `lint` — **no `lint` script** in `package.json`; ad-hoc eslint run hits a pre-existing `typescript-eslint`/typescript version-resolution error unrelated to this change. Skipped per the gate's "no lint script" allowance.
- [x] Review Army + `simplify` run (complex change). **Two findings fixed before push:**
  - **P1 (BLOCK):** the initial `http.createServer` monkey-patch was a **no-op in prod** — vinext imports `createServer` as an immutable ESM named binding, so the old CJS trick doesn't apply and injection silently bypassed → `loginLimiter` degraded to a single `"unknown"` bucket (rate-limit bypass). Fixed by capturing the `http.Server` returned from `startProdServer` and using `prependListener("request", ...)` so the sanitizer runs before vinext's handler. **Verified empirically** that injection now fires on every prod request.
  - **P2:** `node-machine-id` named import fails under bare-Node ESM (only worked via vinext's `vite-plugin-commonjs` rewrite). Fixed with default-import + destructure.

## Testing

- [x] `bun run build:vinext` succeeds (5 build phases, ~1k modules)
- [x] **Prod** (`node server.vinext.js`): `/api/health` 200, `/login` 200 (renders full HTML), `/landing` 200, `/` 307→dashboard (guard), `/api/auth/status` 200, `/v1/models` (rewrite) 200, `/manifest.webmanifest` 200
- [x] **Dev** (`bun run dev:vinext`): same routes all 200, no SSR errors
- [x] `next/font/google` self-hosts Inter correctly under vinext (`/_next/static/_vinext_fonts/`)
- [x] Trusted-IP injection confirmed running in prod via diagnostic
- [x] `node --check` passes on all converted files

## Notes

- The Dockerfile still runs `node custom-server.js` (Next path). To run vinext in Docker, change the `CMD` to `node server.vinext.js`.
- Two CJS files (`src/lib/updater/updater.js`, `src/lib/crashLogger.js`) are intentionally **not** converted — they're outside the vinext build graph (one is a dual CJS/ESM hybrid, the other loads lazily) and the build succeeds without touching them.
- Pre-existing runtime warning (`usageRepo pruneOldUsageHistory failed: db.transaction is not a function`) is a sql.js/better-sqlite3 API mismatch unrelated to this migration.

## Related Issues

Ref: IVT-0517


rizkhal:
🚀

PR #2235, opened 2026-06-30
feat: OpenAI and Anthropic API parity

bloodf:
# PR: OpenAI/Anthropic API parity improvements

## Summary
Ports the fork's OpenAI/Anthropic API parity improvements to the upstream worktree as plain JavaScript (ESM, `.js` extensions).

### Translator parity
- `open-sse/translator/validate.js` — outbound payload validation gate. Runs before `executor.execute()` in `chatCore`, catching missing required fields, shape violations per target format, and internal-key leaks (`_toolNameMap`, `_clientSessionId`). Disabled via `VALIDATE_OUTBOUND=false`.
- `open-sse/translator/index.js` now imports `validateOutboundPayload`/`stripInternalKeys` and gates dispatch through them.
- `open-sse/config/runtimeConfig.js` adds `VALIDATE_OUTBOUND` (default `true`).
- `open-sse/utils/earlyStreamKeepalive.js` — SSE keepalive wrapper for slow-to-first-token streaming endpoints (Codex CLI / Anthropic SDK compatibility). Branding stripped.

### Capability cores
- `open-sse/handlers/imageEditCore.js` — OpenAI-compatible `/v1/images/edits` multipart passthrough.
- `open-sse/handlers/moderationsCore.js` — generic OpenAI-compatible `/v1/moderations` passthrough.
- `open-sse/handlers/rerankCore.js` — generic `/v1/rerank` passthrough for Cohere/Jina/Voyage-style providers.
- `open-sse/handlers/countTokensCore.js` — native `/messages/count_tokens` for Claude-compatible providers, heuristic fallback (~4 chars/token) otherwise.
- `open-sse/services/modelFallback.js` — per-model fallback runner with deterministic-payload-error guard and roundrobin/random/ordered strategies.
- `src/sse/handlers/typeHelpers.js` — credential/result mapping between SSE auth layer and open-sse core handlers.

### Routes
- `src/app/api/v1/responses/route.js` now routes through `handleChat` + translator with early SSE keepalive for `text/event-stream` requests.
- `src/app/api/v1/images/edits/route.js` — new image-edit endpoint.
- `src/app/api/v1/audio/translations/route.js` — new Whisper translation endpoint.
- `src/app/api/v1/moderations/route.js` — new moderation endpoint.
- `src/app/api/v1/rerank/route.js` — new rerank endpoint.
- `src/app/api/v1/messages/count_tokens/route.js` — new count_tokens endpoint.
[...]

PR #2419, opened 2026-07-06
fix(count_tokens): include structured Anthropic blocks

### LINKED ISSUE #2337: count_tokens only counts text blocks (tool_use/tool_result = 0), silently breaking Claude Code auto-compaction
## Summary

The Anthropic-compatible `/v1/messages/count_tokens` endpoint only counts `text` content blocks. All other block types — `tool_use`, `tool_result`, `thinking` — are counted as **0 tokens**. Since ~95% of tokens in an agentic coding session live in tool results, this makes the endpoint return wildly underestimated (often near-zero) counts for real Claude Code conversations.

This **silently breaks Claude Code's auto-compaction**: conversations grow past the model's real context limit without ever being compacted, until the upstream API rejects requests — at which point even manual `/compact` fails (the compaction request itself exceeds the input limit), deadlocking the session.

This is very likely the root cause behind #39.

## Reproduction

```bash
# 1. Plain text message — works (chars/4 estimate)
curl -s http://127.0.0.1:20128/v1/messages/count_tokens \
  -H "x-api-key: $KEY" -H "anthropic-version: 2023-06-01" -H "content-type: application/json" \
  -d '{"model":"cx/gpt-5.5","messages":[{"role":"user","content":"hello world"}]}'
# => {"input_tokens":3}

# 2. tool_use + tool_result message (the shape of real agent conversations) — counted as ZERO
curl -s http://127.0.0.1:20128/v1/messages/count_tokens \
  -H "x-api-key: $KEY" -H "anthropic-version: 2023-06-01" -H "content-type: application/json" \
  -d '{"model":"cx/gpt-5.5","messages":[{"role":"assistant","content":[{"type":"tool_use","id":"toolu_01","name":"Read","input":{"file_path":"/tmp/a.txt"}}]},{"role":"user","content":[{"type":"tool_result","tool_use_id":"toolu_01","content":"line1 line2 line3 some file content here"}]}]}'
# => {"input_tokens":0}
```

## Source location

`app/api/v1/messages/count_tokens/route.js` (deobfuscated):

```js
let total = 0;
for (const msg of body.messages || []) {
  if (typeof msg.content === "string") total += msg.content.length;
  else if (Array.isArray(msg.content))
    for (const block of msg.content)
      if (block.type === "text" && block.text) total += block.text.length; // only text blocks!
}
return { input_tokens: Math.ceil(total / 4) };
```

## Impact on Claude Code (verified with `claude --debug`)

Claude Code's auto-compaction **trigger** works fine (it uses response `usage`, which 9router converts correctly — thumbs up for that). But the **grouping/apply stage** calls `count_tokens` per message to decide which message groups to summarize. With tool messages counted as 0, grouping collapses:

```
[DEBUG] precomputed compact: started (main, 14 msgs, ~31554 tok, attempt 1, trigger api_response)
[INFO]  Reactive compact: no assistant messages in summarize set, bailing
[DEBUG] precomputed compact: failed (too_few_groups) (main, 1ms)
...
[DEBUG] precomputed compact: ready (main, 66286ms)   <- summary computed but never applied
```

Result: context grows to 200%+ of the window with zero compactions (reproduced 6 times in controlled sessions; also hit in production with `cx/gpt-5.5`, where a session died at 272k input tokens with no way to compact).

Note: Claude Code connected to providers with a *real* `count_tokens` implementation (e.g. Zhipu's Anthropic-compatible endpoint) auto-compacts normally through long sessions — which isolates the failure to this endpoint rather than anything else in 9router's translation layer.

## Suggested fix

Count every block type, not just `text`. Even the crude one-liner

```js
for (const msg of body.messages || []) total += JSON.stringify(msg).length;
```

would be a strict improvement (tool_use inputs and tool_result contents get counted). Ideally, use a real tokenizer (e.g. `gpt-tokenizer` for OpenAI-backed models) and also account for `system` / `tools` fields when present.

Happy to provide more debug logs or test a patched build. Thanks for the great project!

spacesky-cell:
## Summary
- Fix `/v1/messages/count_tokens` fallback estimates so structured Anthropic content blocks contribute tokens.
- Count `tool_use.input`, `tool_result.content`, `thinking`, `system`, and `tools` while preserving the existing plain-text estimate.
- Add focused route tests for plain text, structured tool/thinking blocks, and system/tool definitions.

## Verification
- `C:\tmp\9router-vitest\node_modules\.bin\vitest.cmd run E:\ClaudeCode\9router\repo\tests\unit\count-tokens.test.js --reporter=verbose --config C:\tmp\9router-vitest\vitest.empty.config.mjs --root E:\ClaudeCode\9router\repo\tests`
- `node --check src\app\api\v1\messages\count_tokens\route.js`

## Notes
- Checked open PR #2235 because it also touches `count_tokens`; its fallback estimator still only counts text blocks, so it does not resolve this issue.

Closes #2337

### DIFF
--- a/src/app/api/v1/messages/count_tokens/route.js
+++ b/src/app/api/v1/messages/count_tokens/route.js
@@ -11,6 +11,64 @@ export async function OPTIONS() {
   return new Response(null, { headers: CORS_HEADERS });
 }
 
+function countValueChars(value) {
+  if (value == null) return 0;
+  if (typeof value === "string") return value.length;
+  if (typeof value === "number" || typeof value === "boolean") {
+    return String(value).length;
+  }
+  if (Array.isArray(value)) {
+    return value.reduce((total, item) => total + countValueChars(item), 0);
+  }
+  if (typeof value === "object") {
+    return Object.entries(value).reduce((total, [key, item]) => {
+      return total + key.length + countValueChars(item);
+    }, 0);
+  }
+  return 0;
+}
+
+function countContentBlockChars(block) {
+  if (block == null) return 0;
+  if (typeof block === "string") return block.length;
+  if (typeof block !== "object") return countValueChars(block);
+
+  switch (block.type) {
+    case "text":
+      return countValueChars(block.text);
+    case "tool_use":
+      return countValueChars(block.name) + countValueChars(block.input);
+    case "tool_result":
+      return countValueChars(block.content);
+    case "thinking":
+      return countValueChars(block.thinking);
+    default:
+      return countValueChars(block);
+  }
+}
+
+function countMessageChars(message) {
+  if (!message || typeof message !== "object") return 0;
+  const content = message.content;
+
+  if (typeof content === "string") return content.length;
+  if (Array.isArray(content)) {
+    return content.reduce((total, block) => total + countContentBlockChars(block), 0);
+  }
+  return countValueChars(content);
+}
+
+export function estimateAnthropicInputTokens(body = {}) {
+  const messages = Array.isArray(body.messages) ? body.messages : [];
+  let totalChars = countValueChars(body.system) + countValueChars(body.tools);
+
+  for (const msg of messages) {
+    totalChars += countMessageChars(msg);
+  }
+
+  return Math.ceil(totalChars / 4);
+}
+
 /**
  * POST /v1/messages/count_tokens - Mock token count response
  */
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @spacesky-cell for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2448, opened 2026-07-07
fix(rtk/find): detect and group Windows backslash-style find output

Diwak4r:
## Summary

The RTK token saver's `autodetect` treats any line containing a colon as non-path. That excludes Windows absolute paths (e.g. `C:\Users\me\a.js`), so Windows `find`/path dumps returned `null` and were **never compacted**. Even when detected, `find.js` split only on `/`, mis-grouping backslash paths under a single `./` bucket.

Fixes:
- `autodetect.js` -> `isPathLike`: allow the drive-letter colon (`X:\` / `X:/`) before the general colon rejection, so Windows paths are recognized as paths.
- `filters/find.js`: split on the last `/` **or** `\` separator and normalize emitted directory labels to forward slashes.

Fail-open behavior and all non-Windows detection paths are unchanged.

## Test plan
- Added `tests/unit/rtkFindWindows.test.js` covering: Windows path detection (find), Windows `file:line` routing to a compacting filter, grouping of 3 Windows files under one normalized dir, and forward-slash normalization (no backslashes leak).
- `node --check` passes on all touched files.
- Regression verified: git-diff, git-status, grep, and Unix `find` detection + grouping all intact.

## Note on verification scope
This change lives entirely in `open-sse/rtk/` (a Node.js token-saver module) with no browser/Dashboard surface, so there is no UI for a browser-session (Reticle) check to exercise. Verification was done via Node assertions and `node --check` instead.

### DIFF
--- a/open-sse/rtk/autodetect.js
+++ b/open-sse/rtk/autodetect.js
@@ -84,6 +84,9 @@ function isGrepLine(line) {
 function isPathLike(line) {
   const t = line.trim();
   if (t.length === 0) return false;
+  // Windows absolute paths carry a drive letter (e.g. "C:\Users\me"),
+  // so only reject colons that are NOT the drive-letter separator.
+  if (/^[A-Za-z]:[\\/]/.test(t)) return true;
   if (t.includes(":")) return false;
   return t.startsWith(".") || t.startsWith("/") || t.includes("/");
 }
--- a/open-sse/rtk/autodetect.js
+++ b/open-sse/rtk/autodetect.js
@@ -84,8 +84,10 @@ function isGrepLine(line) {
 function isPathLike(line) {
   const t = line.trim();
   if (t.length === 0) return false;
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @Diwak4r for the contribution! Reviewed and accepted — applying it to master. 🙏

Diwak4r:
Hi @decolua — thanks for the review! Quick check-in on this one: your comment said you were applying it to master, but the compare API still shows the head commit (`ca80f968c36613a5a3ab36aff26dd9e8a097f9fa`) as not on master — `compare/master...ca80f968c36613a5a3ab36aff26dd9e8a097f9fa` returns `ahead_by: 3`. Same situation as #3030, where the commit also never landed on master. If it went in under a different commit or branch, could you point me at it and I'll close this out? Otherwise I'm happy to rebase onto current master and reopen — the fix for Windows backslash-style find output is still useful.

PR #2452, opened 2026-07-08
fix(codex): preserve GPT-5.6 max reasoning

### LINKED ISSUE #2501: Cannot configure gpt-5.6 nor Luna, Terra, Sol max reasoning with Pro subscription
Image

I can configure xhigh reasoning in those newly released GPT 5.6 models, but max reasoning does not work.

ryanngit:
## Summary

- expose `max` for Codex GPT-5.6 Sol, Terra, and Luna through provider/model capability metadata
- preserve `reasoning_effort: max` on their Codex wire requests
- retain existing `max -> xhigh` compatibility for OpenAI, Kiro, and older Codex models
- cover provider isolation and prefixed `cx/gpt-5.6-sol` model IDs

## Root cause

9Router v0.5.45 exposes `max` for Sol through a model-name pattern, but `thinkingUnified.js` unconditionally clamps every OpenAI-format `max` request to `xhigh`. Terra and Luna do not expose `max` at all. This makes Console Log show `THINK:xhigh` and sends lower effort than requested.

This change moves supported levels into existing provider-specific Codex capability records. Translator checks that metadata instead of hardcoding GPT model names. Review aliases inherit the same capability records.

Fixes #2501. Reproduces the exact logged symptom from #2785.

## Verification

- focused capability/translator/provider-isolation suite: 7 files, 75 tests passed
- changed-file ESLint: passed
- `git diff --check`: passed
- commit-range Gitleaks: no leaks
- full single-worker run has no candidate-only failure; remaining 111 failures match pristine v0.5.45 baseline

Base: pristine `v0.5.45` (`6fcd27337a7893642c7fe630840d0a641743f28f`). Five files, one commit.

## Scope

Fast/Priority and long-context behavior are intentionally excluded from this PR.


ryanngit:
## v0.5.40 refresh

- Base: `v0.5.40` (`79918c7`)
- Branch head: `bf03ebcc0193c8eb198bca13070e14fa784b6624` (was `dfcd956d5b7a5baa6817123d530589c22ff31437`)
- Merge: merged (detached worktree push)
- Tests: see log summary
- ` Test Files  3 passed (3);      Tests  24 passed (24);`

pixelbrow720:
## Independent runtime reproduction (v0.5.40)

I independently reproduced the GPT-5.6 `max` regression with:

- 9Router `0.5.40`
- Codex Desktop `0.144.6`
- route `cx/gpt-5.6-sol -> codex/gpt-5.6-sol`
- client request `reasoning.effort: "max"`

### Observed before the fix

The generic OpenAI thinking normalization clamps semantic `max` to `xhigh` before the request-line metadata is emitted, so the dashboard reports:

```text
THINK:xhigh
```

The released source explains the behavior: `applyFormat("openai", ...)` does not receive provider/model context and therefore applies the generic OpenAI clamp unconditionally. A later Codex-specific transform can restore the outbound effort, but the console metadata has already observed `xhigh`.

### Independent validation of this PR's scope

I applied a narrow local equivalent of this PR's translator change for provider `codex` and the GPT-5.6 Sol/Terra/Luna models, restarted 9Router, and drove normal Codex traffic through the running server. The dashboard then consistently reported:

```text
THINK:max
```

Requests completed successfully. Negative checks also passed:

- other Codex/OpenAI models retain the existing `max -> xhigh` compatibility behavior;
- non-Codex providers are unchanged;
- non-`max` efforts remain unchanged;
- prefixed model ids such as `cx/gpt-5.6-sol` resolve correctly.

This independently confirms the provider/model-aware change in `thinkingUnified.js` and the Sol/Terra/Luna scope in this PR. A focused regression around the post-translation value used by Console Log would also protect the user-visible `THINK:max` signal.

No credentials, account identifiers, prompts, or raw request bodies are included here.

pixelbrow720:
## Independent runtime reproduction (v0.5.40)

I independently reproduced the GPT-5.6 `max` regression with:

- 9Router `0.5.40`
- Codex Desktop `0.144.6`
- route `cx/gpt-5.6-sol -> codex/gpt-5.6-sol`
- client request `reasoning.effort: "max"`

### Observed before the fix

The generic OpenAI thinking normalization clamps semantic `max` to `xhigh` before the request-line metadata is emitted, so the dashboard reports:

```text
THINK:xhigh
```

The released source explains the behavior: `applyFormat("openai", ...)` does not receive provider/model context and therefore applies the generic OpenAI clamp unconditionally. A later Codex-specific transform can restore the outbound effort, but the console metadata has already observed `xhigh`.

### Independent validation of this PR's scope

I applied a narrow local equivalent of this PR's translator change for provider `codex` and the GPT-5.6 Sol/Terra/Luna models, restarted 9Router, and drove normal Codex traffic through the running server. The dashboard then consistently reported:

```text
THINK:max
```

Requests completed successfully. Negative checks also passed:

- other Codex/OpenAI models retain the existing `max -> xhigh` compatibility behavior;
- non-Codex providers are unchanged;
- non-`max` efforts remain unchanged;
- prefixed model ids such as `cx/gpt-5.6-sol` resolve correctly.

This independently confirms the provider/model-aware change in `thinkingUnified.js` and the Sol/Terra/Luna scope in this PR. A focused regression around the post-translation value used by Console Log would also protect the user-visible `THINK:max` signal.

No credentials, account identifiers, prompts, or raw request bodies are included here.

ryanngit:
Final `v0.5.45` audit for maintainer review: head `a31070b1` is one commit on pristine `6fcd2733`, merge state `CLEAN`. Fresh 7-file provider-isolation matrix passed 75/75; changed-file ESLint, diff check, and commit-range Gitleaks passed. Scope directly fixes #2501 and the `THINK:xhigh` symptom in #2785 without changing older/OpenAI/Kiro compatibility.

Issue #2431, opened 2026-07-07
[BUG] Antigravity to Antigravity MITM ResponseAborted


Nautilaceae:
I dont really know how it happens. but it happens when the agents try to write big thing such as long implementation plan. 
LOG :

[07:59:15] ℹ️ [AUTH] Using antigravity account: xxx@xxx.xxx
[07:59:15] ℹ️ [TOKEN_REFRESH] Refreshing provider credentials proactively {"provider":"antigravity","expiresIn":-54580,"refreshLeadMs":300000,"lastRefreshAt":"2026-07-06T08:49:36.429Z"}
[07:59:15] ℹ️ [TOKEN_REFRESH] Successfully refreshed Google token {"hasNewAccessToken":true,"expiresIn":3599}
[07:59:15] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"ebff9875-9b9b-4d54-812b-600465286054","success":true}
[07:59:15] [PENDING] START | provider=antigravity | model=claude-opus-4-6-thinking
[07:59:16] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"ebff9875-9b9b-4d54-812b-600465286054","success":true}
[08:00:12] 📥 POST /v1/chat/completions | ag/claude-opus-4-6-thinking | 0 msgs
[08:00:12] ℹ️ [ROUTING] ag/claude-opus-4-6-thinking → antigravity/claude-opus-4-6-thinking
[08:00:12] ℹ️ [AUTH] Using antigravity account: xxx@xxx.xxx
[08:00:12] ℹ️ [TOKEN_REFRESH] Refreshing provider credentials proactively {"provider":"antigravity","expiresIn":-54614,"refreshLeadMs":300000,"lastRefreshAt":"2026-07-06T08:49:59.335Z"}
[08:00:12] ℹ️ [TOKEN_REFRESH] Successfully refreshed Google token {"hasNewAccessToken":true,"expiresIn":3599}
[08:00:12] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"9d9baf58-ea74-42fe-b7c9-4a1a1416ab8b","success":true}
[08:00:12] [PENDING] START | provider=antigravity | model=claude-opus-4-6-thinking
[...]

PR #2370, opened 2026-07-04
fix: handle Antigravity capacity and stream usage

nguyenha935:
## Summary
- Treat Antigravity `MODEL_CAPACITY_EXHAUSTED` as request-local fallback, not account/model cooldown.
- Preserve normal cooldown behavior for other providers and real quota/rate-limit errors.
- Track Antigravity/Gemini passthrough stream content so successful streams without `usageMetadata` can estimate usage and appear in Recent Requests without showing 0/0 junk rows.

## Test Plan
- `node tests/node_modules/vitest/vitest.mjs run tests/unit/antigravity-retry-hook.test.js tests/unit/base-executor-retry.test.js tests/unit/antigravity-capacity-fallback.test.js tests/unit/antigravity-recent-requests.test.js --config tests/vitest.config.js --configLoader runner`

Issue #2259, opened 2026-07-01
● API Error: API returned an empty or malformed response (HTTP 200) — check for a proxy or gateway intercepting the request


Ranshiv:
[ClaudeHeaders] Cached 14 identity headers from Claude Code client
[20:56:03] 📥 POST /v1/v1/messages | ag/gemini-pro-agent | 280 msgs | 19 tools
[20:56:03] ℹ️ [ROUTING] ag/gemini-pro-agent → antigravity/gemini-pro-agent
[20:56:03] ℹ️ [AUTH] Using antigravity account: email@email.com
[20:56:03] [PENDING] START | provider=antigravity | model=gemini-pro-agent
[20:56:18] 🌊 [STREAM] ANTIGRAVITY | gemini-pro-agent | 15479ms | disconnect: ResponseAborted
[20:56:18] [PENDING] END | provider=antigravity | model=gemini-pro-agent
[ClaudeHeaders] Cached 14 identity headers from Claude Code client
[20:56:18] 📥 POST /v1/v1/messages | ag/gemini-pro-agent | 280 msgs | 19 tools
[20:56:18] ℹ️ [ROUTING] ag/gemini-pro-agent → antigravity/gemini-pro-agent
[20:56:18] ℹ️ [AUTH] Using antigravity account: email@email.com
[20:56:18] ℹ️ [TOKEN_REFRESH] Refreshing provider credentials proactively {"provider":"antigravity","expiresIn":295,"refreshLeadMs":300000,"lastRefreshAt":"2026-06-30T20:18:18.636Z"}
[20:56:18] ℹ️ [TOKEN_REFRESH] Successfully refreshed Google token {"hasNewAccessToken":true,"expiresIn":3599}
[20:56:18] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"34ca4378-0368-40c4-a58b-9dc16ad9a0b4","success":true}
[20:56:18] [PENDING] START | provider=antigravity | model=gemini-pro-agent
[20:56:19] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"34ca4378-0368-40c4-a58b-9dc16ad9a0b4","success":true}
[20:56:39] [PENDING] END | provider=antigravity | model=gemini-pro-agent
[20:56:39] 🌊 [STREAM] ANTIGRAVITY | gemini-pro-agent | 21128ms | complete

[...]

Issue #2229, opened 2026-06-29
Empty Output Response Error.


Fuad7061:
For some custom models it shows 200 code with success but acutally gives null or empty response. As example:

{
 "text": ""
}

So, it will be good that if the system can detect any null or empty response and can mark them as error so that it can retry with other models.

lucaspeyrin:
I get the same thing with Antigravity's gemini-3.6-flash-high model. Would love if empty response could be seen as an error, as this breaks / stops the agents that are running

PR #2462, opened 2026-07-08
fix(antigravity): retry empty streams, surface aborted turns, finalize truncated Claude streams

### LINKED ISSUE #2250: Please clarify the task and I will continue
if i prompt hi, it works. but if i type a long prompt, it always error Please clarify the task and I will continue.

Image

anhdiepmmk:
## Problem

Claude Code → 9router → Antigravity (Gemini) is unstable in ways the other providers are not:

1. **Empty responses** — upstream answers HTTP 200 with a stream carrying no usable output (no candidates, thought-only parts, or a bare `STOP`). The stream "succeeds", so no retry fires and no combo fallback engages; Claude Code hangs on an empty turn (#2188, #2229).
2. **Model says it will do something but does nothing** — Gemini aborts the tool call with `finishReason: MALFORMED_FUNCTION_CALL`, which was forwarded as a clean `end_turn`. The user sees "I'll run the command now." and must type *continue* (#2250).
3. **Streams truncated mid-turn** (dropped connection, no `finishReason`) never emit `message_delta`/`message_stop` — dangling message, client hangs (#2259, #2431).

## Root Cause

- `translator/response/gemini-to-openai.js` dropped every candidate-less chunk — including `promptFeedback.blockReason`-only chunks and `{error:{...}}` objects embedded mid-stream — and discarded upstream `functionCall.id`s.
- `translator/concerns/finishReason.js` defaulted all unmapped Gemini finish reasons (`MALFORMED_FUNCTION_CALL`, `UNEXPECTED_TOOL_CALL`, `OTHER`, …) to `stop` → `end_turn`.
- `translator/response/openai-to-claude.js` only finalized inside `if (choice.finish_reason)` — no flush-time close; the pivot in `translator/index.js` also never forwarded the flush sentinel past the gemini stage.
- Retries (`executors/base.js`, `executors/antigravity.js`) only fire on non-2xx/network errors — a 200-but-empty stream never reaches any retry or fallback path.

## Fix

Five focused changes (one commit each, scoped to the Gemini-family path), plus three review-fix commits:

1. **Gemini error finish reasons → internal hub finish `"error"`** (`GEMINI_ERROR_FINISH_REASONS`); safety family extended with `SPII`/`IMAGE_SAFETY` (now a shared `GEMINI_CONTENT_FILTER_FINISH_REASONS` set); truly unknown values still map to `stop` so a future benign reason can't regress gemini/gemini-cli/vertex. The `STOP→tool_calls` upgrade no longer fires for aborted turns.
2. **gemini-to-openai hardening**: candidate-less chunks now harvest usage (keep-alives), close as `content_filter` (prompt blocked), or close with `"error"` + `state.upstreamError` (embedded error objects). Upstream `functionCall.id` is preserved (dedup via `state.seenToolCallIds`); generated fallback ids use `name_<ts>_<idx>` and the Anthropic id charset.
3. **Claude-side finalization**: flush-time `finalizeOnFlush()` closes open blocks, flushes buffered tool args, and emits `message_delta` + `message_stop` for truncated streams (usage converted to the Claude shape via `toClaudeUsage()` when the gemini stage left OpenAI-shaped counts in the shared state); `state.claudeFinishHandled` guards duplicate finishes (deliberately NOT keyed on the shared `state.finishReason`, which the gemini stage already sets); hub finish `"error"` emits an Anthropic `error` SSE event (retryable by Claude Code) instead of a fake clean `end_turn`.
4. **Empty-stream guard** (`handlers/chatCore/emptyStreamGuard.js`, gated `provider === "antigravity" && stream`) — oh-my-pi parity: **everything, thinking included, streams to the client live**; emptiness is judged per upstream attempt, after the fact. An attempt that ends with a bare `STOP` (or an aborted/error finish, or dies) having produced no meaningful content (tool call / non-whitespace text; thought-only = empty, same definition as oh-my-pi) has its terminal event withheld and is retried with the identical request up to 2× (`500ms·2^n`); the retried attempt **splices into the same client stream** (the translator inits its message once, so the splice continues the same Claude message). On exhaustion an `{error:{...}}` event is emitted in-stream — the translator turns it into the Anthropic `error` SSE event (auto-retried by Claude Code); a held real upstream error (e.g. `RESOURCE_EXHAUSTED`) is re-emitted verbatim so the true message reaches the client. `onUpstreamEmptyExhausted` benches the account (30s via `markAccountUnavailable`) so the client's automatic retry rotates to the next account — the practical replacement for pre-stream fallback, which is impossible once headers are sent. **Deterministic content blocks are never retried**: `promptFeedback.blockReason` chunks and SAFETY-family/`MAX_TOKENS` finishes forward untouched so the translator closes them as `content_filter`/`max_tokens`. Client aborts are classified as aborts (never retried or misreported), and upstream readers are cancelled on every discard path.
5. **functionResponse name recovery** tolerates both generated-id shapes and degrades to the id itself for foreign ids.

**Not touched** (no overlap with open PRs): #2317 (multipleOf), #2222 (ensureObjectType), #2321 (assistant prefill), #2370 (capacity exhausted), #2443 (system prompt injection), #2135 (fake continue turn).

## Behavior notes / trade-offs (from review vs oh-my-pi & antigravity-claude-proxy)

- **Thinking streams live; a retried empty attempt can duplicate it.** Same trade-off oh-my-pi makes: a thought-only attempt has already streamed its thinking before it can be judged empty, so the client may see the discarded attempt's thinking followed by the retry's thinking inside one message. In exchange the UI is never frozen during long thinking phases, and thought-only empties (#2229) stay retryable.
- **Empty-stream exhaustion is an in-stream error, not a pre-stream 502.** Headers are already sent when exhaustion is known, so account/combo fallback cannot run inside the same request. Instead: the account is benched (awaited **before** the error event is emitted, so the client's automatic retry always lands on the next account; quota-style exhaustion parses "reset after XhYmZs" from the upstream message via `parseRetryFromErrorMessage` and benches until the actual reset instead of the generic 30s). Combo **model**-fallback does not trigger for this case (it still triggers for all pre-stream failures).
- **Truncated streams close cleanly** (`end_turn`, or `tool_use` with the buffered args flushed) rather than surfacing a retryable error like oh-my-pi does. Partial content is preserved and the tool loop continues; the cost is that a mid-answer truncation looks complete. Deliberate choice — strictly better than the current hang either way.
- **MALFORMED-family finishes before any content are retried server-side** (oh-my-pi surfaces them immediately and relies on the client retrying; retrying in place saves a full client round-trip with a large prompt). After content they forward as `error` events, exactly like oh-my-pi.
- **OpenAI-format clients** now receive `finish_reason: "error"` (non-standard enum) for aborted gemini turns instead of a silent empty stream; the upstream error message itself is only attached for Claude clients (possible follow-up).

## Tests

- **17 regression cases** in `tests/translator/bugs-antigravity.test.js` (finish-reason mapping, no-upgrade-on-error, promptFeedback close, keep-alive usage, mid-stream error objects, id preservation, flush finalization, flush usage shape, duplicate-finish dedup, MALFORMED end-to-end, id→name round-trip both separators).
- **18 unit cases** in `tests/unit/empty-stream-guard.test.js` (live thought-streaming byte-fidelity, empty-attempt suppression + splice, split reads, duplicate-thinking wart locked, exhaustion synthetic vs real-error re-emit, bench-awaited-before-error ordering, upstream-error passthrough for quota reset parsing, MALFORMED before/after content, blockReason/SAFETY/MAX_TOKENS never retried, truncation-after-content no-retry, downstream cancel propagation, abort classification, reexecute failure, splice→Claude single-message sanity).
- `tests/unit/finish-reason-concern.test.js`: `OTHER` moves to the aborted-turn set (`error`); `SPII` → `content_filter`; `UNKNOWN_XYZ` still locks the default-to-`stop`.
- **Full suite vs clean master: zero new failures** — verified twice, including after the review fixes, by running the JSON reporter on both trees and diffing the failing-test sets (identical 45-test set, all pre-existing).

## Evidence (before/after)

Pathological Gemini streams pushed through `translateResponse(ANTIGRAVITY → CLAUDE)` on master vs this branch — what the Claude Code client receives:

```
=== 1. Model narrates tool call, Gemini aborts with MALFORMED_FUNCTION_CALL (#2250) ===
master: … content_block_stop → message_delta stop_reason=end_turn → message_stop   ← fake "finished answer"
branch: … content_block_stop → error api_error (retryable)

=== 2. Stream truncated mid-answer, no finishReason (#2259, #2431) ===
master: message_start → content_block_start → content_block_delta                  ← dangling, client hangs
branch: … → content_block_stop → message_delta stop_reason=end_turn → message_stop

=== 3. Stream truncated with an unfinished tool call ===
master: message_start → content_block_start                                        ← tool args lost, hangs
branch: … → input_json_delta flushed → message_delta stop_reason=tool_use → message_stop

=== 4. Prompt blocked: promptFeedback-only chunk (#2188) ===
master: (no events — empty 200 stream)                                             ← hangs
branch: message_start → message_delta → message_stop  (guard releases, no retry)

=== 5. {error:{RESOURCE_EXHAUSTED}} embedded in 200 stream ===
master: message_start → content_block_start → content_block_delta                  ← hangs
branch: … → error api_error: Quota exceeded
```

Fixes #2250. Addresses the antigravity/gemini portion of #2188, #2229, #2259, #2431.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -333,7 +333,16 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
         reexecute,
         signal: streamController.signal,
         log,
-        onExhausted: (reason) => onUpstreamEmptyExhausted?.(formatProviderError(new Error(reason), provider, model, HTTP_STATUS.BAD_GATEWAY)),
+        onExhausted: (reason, { upstreamError } = {}) => {
+          if (!onUpstreamEmptyExhausted) return;
+          // Quota-style exhaustion carries the reset time only inside the error
+          // message ("Your quota will reset after 2h7m23s") — bench precisely.
+          const resetMs = executor.parseRetryFromErrorMessage?.(upstreamError?.message || reason);
+          return onUpstreamEmptyExhausted(
+            formatProviderError(new Error(reason), provider, model, HTTP_STATUS.BAD_GATEWAY),
+            resetMs ? Date.now() + resetMs : undefined
+          );
+        },
       }),
       { status: providerResponse.status, headers: providerResponse.headers }
     );
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -18,6 +18,7 @@ import { buildRequestDetail, extractRequestConfig } from "./chatCore/requestDeta
 import { handleForcedSSEToJson } from "./chatCore/sseToJsonHandler.js";
 import { handleNonStreamingResponse } from "./chatCore/nonStreamingHandler.js";
 import { handleStreamingResponse, buildOnStreamComplete } from "./chatCore/streamingHandler.js";
+import { probeSSEStream, replayStream, EMPTY_STREAM_MAX_RETRIES, EMPTY_STREAM_BASE_DELAY_MS } from "./chatCore/emptyStreamGuard.js";
 import { detectClientTool, isNativePassthrough } from "../utils/clientDetector.js";
 import { dedupeTools } from "../utils/toolDeduper.js";
 import { injectCaveman } from "../rtk/caveman.js";
@@ -307,6 +308,59 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     console.log(`${COLORS.red}[ERROR] ${errMsg}${COLORS.reset}`);
     reqLogger.logError(new Error(message), finalBody || translatedBody);
     return createErrorResult(statusCode, errMsg, resetsAtMs);
+  }
+
+  // Antigravity empty-stream guard: probe the upstream SSE before anything is
+  // sent to the client. Gemini occasionally answers 200 with no usable output
+  // (empty candidates / thought-only / bare STOP) or aborts the turn with
+  // MALFORMED_FUNCTION_CALL before emitting content — both usually succeed on a
+  // plain retry with the identical request. On exhaustion return 502 so the
+  // account/combo fallback chain engages instead of the client hanging on an
+  // empty stream (#2188, #2229, #2259).
+  if (provider === "antigravity" && stream && providerResponse.body) {
+    for (let attempt = 0; ; attempt++) {
+      const probe = await probeSSEStream(providerResponse.body, { signal: streamController.signal });
+      if (probe.verdict === "aborted") {
+        trackPendingRequest(model, provider, connectionId, false, true);
+        return createErrorResult(499, "Request aborted");
+      }
+      if (probe.verdict === "ok") {
+        providerResponse = new Response(replayStream(probe.buffered, probe.reader), {
+          status: providerResponse.status,
+          headers: providerResponse.headers,
+        });
+        break;
+      }
+
+      log?.warn?.("STREAM", `${provider.toUpperCase()} | ${probe.verdict}${probe.reason ? ` (${probe.reason})` : ""} | attempt ${attempt + 1}/${EMPTY_STREAM_MAX_RETRIES + 1}`);
+      if (attempt >= EMPTY_STREAM_MAX_RETRIES) {
+        trackPendingRequest(model, provider, connectionId, false, true);
+        appendRequestLog({ model, provider, connectionId, status: `FAILED ${HTTP_STATUS.BAD_GATEWAY}` }).catch(() => { });
+        const errMsg = formatProviderError(
+          new Error(`empty response from upstream (${probe.verdict}${probe.reason ? `: ${probe.reason}` : ""}) after ${attempt + 1} attempts`),
+          provider, model, HTTP_STATUS.BAD_GATEWAY
+        );
+        return createErrorResult(HTTP_STATUS.BAD_GATEWAY, errMsg);
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, EMPTY_STREAM_BASE_DELAY_MS * 2 ** attempt));
+      let retryResult;
+      try {
+        retryResult = await executor.execute({ model, body: translatedBody, stream, credentials, signal: streamController.signal, log, proxyOptions });
+      } catch (error) {
+        trackPendingRequest(model, provider, connectionId, false, true);
+        if (error.name === "AbortError") return createErrorResult(499, "Request aborted");
+        return createErrorResult(HTTP_STATUS.BAD_GATEWAY, formatProviderError(error, provider, model, HTTP_STATUS.BAD_GATEWAY));
+      }
+      if (!retryResult.response.ok) {
+        trackPendingRequest(model, provider, connectionId, false, true);
+        const { statusCode, message, resetsAtMs } = await parseUpstreamError(retryResult.response, executor);
+        return createErrorResult(statusCode, formatProviderError(new Error(message), provider, model, statusCode), resetsAtMs);
+      }
+      providerResponse = retryResult.response;
+      providerUrl = retryResult.url;
+      finalBody = retryResult.transformedBody;
+    }
   }
 
   const sharedCtx = { provider, model, body, stream, translatedBody, finalBody, requestStartTime, connectionId, apiKey, clientRawRequest, onRequestSuccess };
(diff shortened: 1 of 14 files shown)


anhdiepmmk:
cc @decolua @babymum7 — could you review this when you have a chance? It targets the Antigravity/Gemini empty-response and "narrates-but-does-nothing" instability with Claude Code (#2188, #2229, #2250, #2259, #2431). Before/after evidence and full-suite regression comparison are in the PR body. Happy to adjust anything to fit the codebase conventions.

anhdiepmmk:
@decolua — @babymum7 is on my team and will take responsibility for validating this PR: he'll run it locally against a real Antigravity account and post the stability evidence here, so you don't have to spend your own quota verifying it.

@babymum7 — please run the checklist below and report results in this thread.

## Setup

```bash
gh repo clone anhdiepmmk/9router && cd 9router
git checkout fix/antigravity-empty-stream-stability
npm install && (cd tests && npm install)
PORT=20128 NEXT_PUBLIC_BASE_URL=http://localhost:20128 npm run dev
```

- Connect your Antigravity (Google) account in the dashboard (`http://localhost:20128/dashboard`).
- Point Claude Code at it: `ANTHROPIC_BASE_URL=http://localhost:20128 ANTHROPIC_AUTH_TOKEN=<your 9router api key> claude`, select an `ag/gemini-*` model.
- Keep the 9router console visible — the new guard logs lines like `STREAM | ANTIGRAVITY | empty (…) | attempt 1/3`.

## Test checklist

**T1 — Unit/regression suite (5 min)**
```bash
cd tests && npx vitest run --config vitest.config.js translator/bugs-antigravity.test.js unit/empty-stream-guard.test.js unit/finish-reason-concern.test.js
```
✅ Expect: all pass (16 + 9 + 26 tests).

**T2 — Long agentic session, tool-heavy (the main scenario)**
In a real repo, ask Claude Code for a multi-step task that forces many tool calls, e.g. *"read the project structure, find all TODO comments, fix the three easiest ones, and run the tests"*. Let it run 15–30 minutes.
✅ Expect: no empty responses; no turn where the model says "I'll do X" and then stops without a tool call; you never need to type `continue`.
📋 Evidence: session duration, number of turns, count of `STREAM | ANTIGRAVITY` retry lines in the console (recoveries are expected and fine — silent hangs are not).

**T3 — Empty-response recovery (was #2188/#2229)**
Trigger the known empty-response pattern: long conversations near context limits, or rapid consecutive requests. On master this shows an empty reply and Claude Code just stops.
✅ Expect: either a normal answer (silent retry succeeded) or an explicit error + fallback to your next account/combo — never a silent empty turn.
📋 Evidence: console log lines showing `empty … attempt n/3` followed by recovery or `FAILED 502`.

**T4 — Aborted tool call (was #2250)**
Use gemini-flash models with complex/nested tool schemas (MCP tools are good) — these trigger MALFORMED_FUNCTION_CALL most often.
✅ Expect: when Gemini aborts a tool call, Claude Code shows a retryable API error (and retries) instead of ending the turn as a finished answer.

**T5 — Interrupted stream**
Mid-generation, briefly cut your network (or suspend/resume the machine).
✅ Expect: the message closes cleanly (no permanently-hanging spinner); Claude Code can continue the conversation.

**T6 — Multi-turn tool loop with hyphenated tool names**
Use an MCP server whose tool names contain `-`, run a session with several tool round-trips.
✅ Expect: tool results are accepted by the model across turns (no "tool loop resets" / model re-asking for data it already received).

**T7 — Regression: non-Antigravity Gemini path**
Repeat a short version of T2 on `gemini-cli` (or vertex) provider.
✅ Expect: unchanged behavior — the shared translator changes are format-generic; the stream guard is antigravity-only.

**T8 — Regression: your daily drivers**
One short session each on claude / kiro / any OpenAI-compatible provider you use.
✅ Expect: zero behavior change (their code paths are untouched).

## Evidence format for this thread

Per test: `T# — PASS/FAIL`, model used, rough session length, and for T2/T3 the retry-log excerpt from the console. If anything fails, please attach the request detail from the dashboard (Requests → the failing entry) with the raw upstream chunks.

lucaspeyrin:
This would help me tremendously. Thank you for the work.

PR #2463, opened 2026-07-08
fix(minimax): resolve runtimeTransport/modelTargetFormat conflict and extract <think> for M3

### LINKED ISSUE #2435: invalid tool type
Title: [Bug] Combo model gọi MiniMax qua OpenAI-completions endpoint bị lỗi "invalid tool type" (400) trên toàn bộ connections

Version: 9Router Proxy v0.5.18

Mô tả:
Khi dùng combo (VD "openclaw") có model minimax/MiniMax-M3, mọi request có kèm tools đều fail 400 trên cả 4 connections (round-robin), lỗi giống hệt nhau ở tất cả key → không phải do quota/auth.

Log lỗi:
[ERROR] [400]: {"type":"error","error":{"type":"bad_request_error","message":"invalid params, invalid tool type: (2013)","http_code":"400"}}

Đối chứng:
- Gọi thẳng MiniMax /anthropic endpoint bằng tool schema chuẩn Anthropic (name/description/input_schema) → thành công.
- "Test Connection One-by-One" trong Providers → success (không có tools).
- Chỉ fail khi agent (OpenCLAW) gửi request thật có kèm tools qua combo, tức route qua lớp openai-completions → convert sang anthropic-messages.

zmf963:
### 1. `invalid tool type: (2013)` errors (#2435)

When an OpenAI client requests `minimax-cn/MiniMax-M3`, the `modelTargetFormat: "claude"` overrides the matching `runtimeTransport` (openai), causing the request body to be translated to Claude format before being sent to the OpenAI endpoint — resulting in 400 errors.

**Root cause**: `chatCore.js` used `modelTargetFormat || runtimeTransport?.format` — `modelTargetFormat` always wins.

**Fix**: When `runtimeTransport.format === sourceFormat`, prefer the transport for zero-translation passthrough instead of the model's declared targetFormat. This aligns with `buildUrl()` and `buildHeaders()` which already prioritize `runtimeTransport`.

### 2. MiniMax M3 embeds reasoning as `<think>` tags

M3's OpenAI endpoint returns reasoning as `<think>...</think>` XML tags inside the `content` field rather than using `reasoning_content`. This leaks raw XML to downstream clients and prevents tools like OpenCode from displaying the reasoning process.

**Fix**: New `thinkExtractor.js` module extracts `<think>` tags for both streaming (cross-chunk state tracking via `createThinkExtractor()`) and non-streaming (single-pass regex via `extractThinkTags()`) responses.

## Changes

| File | Description |
|------|------------|
| `chatCore.js` | Prefer `runtimeTransport.format` when it matches `sourceFormat` |
| `stream.js` | Extract `<think>` to `reasoning_content` during passthrough streaming |
| `nonStreamingHandler.js` | Extract `<think>` for non-streaming; guard extracted content from being stripped |
| `thinkExtractor.js` | New module: `extractThinkTags()` (non-streaming) + `createThinkExtractor()` (streaming) |

## Testing

- `npm run build` ✅
- End-to-end verified on arm64 deployment: streaming + non-streaming both correctly extract reasoning
- 48/48 related unit tests pass

Fixes #2435
Closes #2231

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -47,10 +47,16 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
 
   const alias = PROVIDER_ID_TO_ALIAS[provider] || provider;
   const modelTargetFormat = getModelTargetFormat(alias, model);
-  // Multi-endpoint providers: pick transport matching sourceFormat → zero translation
+  // Multi-endpoint providers: pick transport matching sourceFormat → zero translation.
+  // If found, force targetFormat=sourceFormat so we skip translation entirely —
+  // otherwise the body could be translated to modelTargetFormat and sent to a
+  // transport that doesn't understand it.
   const runtimeTransport = resolveTransport(provider, sourceFormat);
-  const targetFormat = modelTargetFormat || runtimeTransport?.format || getTargetFormat(provider);
-  if (runtimeTransport && credentials) credentials.runtimeTransport = runtimeTransport;
+  const skipTranslation = runtimeTransport?.format === sourceFormat;
+  if (skipTranslation && credentials) credentials.runtimeTransport = runtimeTransport;
+  const targetFormat = skipTranslation
+    ? sourceFormat
+    : (modelTargetFormat || runtimeTransport?.format || getTargetFormat(provider));
   const stripList = getModelStrip(alias, model);
   const upstreamModel = getModelUpstreamId(alias, model);
 
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -9,6 +9,7 @@ import { parseSSEToOpenAIResponse } from "./sseToJsonHandler.js";
 import { buildRequestDetail, extractRequestConfig, extractUsageFromResponse, saveUsageStats } from "./requestDetail.js";
 import { appendRequestLog, saveRequestDetail } from "@/lib/usageDb.js";
 import { decloakToolNames } from "../../utils/claudeCloaking.js";
+import { extractThinkTags } from "../../utils/thinkExtractor.js";
 
 function parseToolArguments(value) {
   if (!value) return {};
@@ -242,6 +243,25 @@ export async function handleNonStreamingResponse({ providerResponse, provider, m
     : responseBody;
   const isClaudeMessageResponse = sourceFormat === FORMATS.CLAUDE && translatedResponse?.type === "message";
 
+  // Extract <think>...</think> from content into reasoning_content for providers
+  // that inline thinking as XML tags in the content field (e.g. MiniMax M3 on
+  // OpenAI-format tiers). Non-streaming responses are a single JSON blob — a
+  // simple regex pass suffices.
+  let extractedThink = false;
+  if (translatedResponse?.choices) {
+    for (const choice of translatedResponse.choices) {
+      const msg = choice?.message;
+      if (msg?.content && typeof msg.content === "string") {
+        const { content, reasoning } = extractThinkTags(msg.content);
+        if (reasoning) {
+          msg.reasoning_content = reasoning;
+          msg.content = content;
+          extractedThink = true;
+        }
+      }
+    }
+  }
+
   // Fix finish_reason for tool_calls: some providers return non-standard values (e.g. "other")
   if (translatedResponse?.choices?.[0]) {
     const choice = translatedResponse.choices[0];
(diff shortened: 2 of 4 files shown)


samir-abis:
Thanks @zmf963 — verifying on v0.5.20, single-connection (not a combo). Tracing the live SQLite log at `~/.9router/db/data.sqlite` shows the (2013) response always paired with provider label `minimax`, model `MiniMax-M3`, connectionId `e3029bca-ca66-45d7-ad94-0f0fcac20e30`. Three identical failures over the last hour of Cursor usage, same request_id prefix family (`069d88…` / `069d8c…`), same error body:

```json
{"type":"error","error":{"type":"bad_request_error","message":"invalid params, invalid tool type:  (2013)","http_code":"400"},"request_id":"069d888dc80b5c177a8d8c1e38cea1a4"}
```

```json
{"type":"error","error":{"type":"bad_request_error","message":"invalid params, invalid tool type:  (2013)","http_code":"400"},"request_id":"069d886bf70b97ac9ecd0208fe28046c"}
```

```json
{"type":"error","error":{"type":"bad_request_error","message":"invalid params, invalid tool type:  (2013)","http_code":"400"},"request_id":"069d8c97e5bcdc961d10668b92f93e60"}
```

The upstream request body 9router actually sent (from `data.providerRequest` on row `2026-07-08T14:07:24.626Z-bozwqm-MiniMax-M3`, which was small enough not to be truncated) shows the partial translation path your fix addresses — system prompt and user message have already been lifted into Anthropic-shape `system:[…]` and `messages:[{content:[{type:"text",…}]}]`, but the `tools` array from the original `request` is **not** present in `providerRequest` for that row (it was dropped by the runtime before this point — interesting, see note below). For the larger rows the body is `_truncated` at 446400 / 447333 bytes; here is the literal first-2000-byte `_preview` from row `2026-07-08T13:50:08.167Z-j9asg2-MiniMax-M3`, which contains the system-prompt boundary the truncation cuts at:

```text
{"messages":[{"role":"system","content":"You are an AI coding assistant, powered by main-combo.\n\nYou operate in Cursor.\n\nYou are a coding agent in the Cursor IDE that helps the USER with software 
```

```text
{"model":"MiniMax-M3","max_tokens":32000,"stream":false,"messages":[{"role":"user","content":[{"type":"text","text":"weather in SF"}]}],"system":[{"type":"text","text":"You are Claude Code, Anthropic's official CLI for Claude."},{"type":"text","text":"Respond like terse caveman. All technical substance stay exact, only fluff die. Drop: articles (a/an/the), filler (just/really/basically/actually/simply), pleasantries, hedging. Fragments OK. Short synonyms (big not extensive, fix not implement a solution for). Pattern: [thing] [action] [reason]. [next step]. Not: \"Sure! I'd be happy to help you with that. The issue you're experiencing is likely caused by...\" Yes: \"Bug in auth middleware. Token expiry check use `<` not `<=`. Fix:\" Code blocks, file paths, commands, errors, URLs: keep exac
```

One observation that may explain a residual case PR #2463 doesn't cover: the body 9router sends upstream still has the full OpenAI-shape `tools:[{type:"function", function:{…}}]` block intact on the larger rows (the `request` blob in row `2026-07-08T14:07:24.626Z-bozwqm-MiniMax-M3` clearly carries it; the upstream `providerRequest` for the small row dropped `tools` entirely, which is why its body is small — but on the larger Cursor-driven rows the body 446 KB+ is being forwarded with `tools` left in OpenAI shape), and MiniMax's Anthropic-compatible gateway rejects with `invalid tool type:` (and never the analogous `function arguments` wording #12167 hit). That smells like a second-layer schema-validation issue on the OpenAI→Anthropic translation that your fix's `skipTranslation` branch should bypass — but if the body still gets translated upstream by something the runtime path doesn't see (e.g. a proxy in front of MiniMax, or a different transport in the chain), the OpenAI shape reaches a strict Anthropic validator and 2013s.

Happy to test your branch on top of a local hotfix if useful; otherwise I'll open a small follow-up PR with a regression test that exercises the `MiniMax-M3` OpenAI-shape tools path against a mock that returns the upstream (2013) verbatim — flagged as complementary not duplicate. Maintainers decide.

samir-abis:
Heads-up: #2525 addresses the same transport-selection root cause (#2435) plus `reasoning_split` defaults, cross-chunk `<think>` sanitization, and stripping reasoning fields from the client SSE stream (OpenCode-tested). Happy to consolidate or rebase if you prefer one PR — the transport fix is identical in spirit.

zmf963:
Closing in favor of #2525, which addresses the same runtimeTransport/modelTargetFormat root cause (#2435) with a more complete scope (reasoning_split defaults, cross-chunk <think> sanitization, stricter stream handling).

For reference, this PR's thinkExtractor module (open-sse/utils/thinkExtractor.js) was designed as a standalone closure-based extractor to avoid the shared-regex lastIndex race condition in Node.js async I/O contexts — if #2525's inline approach doesn't already handle that, the pattern may be worth adopting.

Thanks for the review pointer @samir-abis.

PR #2470, opened 2026-07-08
fix(kiro): anchor Claude sessions from cache control

SemonCat:
## Problem

The direct Claude -> Kiro translator could generate a fresh Kiro `conversationState.conversationId` for requests that were semantically part of the same cached Claude prompt prefix. Hermes/Slack requests also include volatile request ids and changing trailing user text, so Kiro cache/session affinity was unstable and Opus traffic missed cache reuse.

## Fix

Use `resolveSessionId` for Claude -> Kiro conversation ids, matching the OpenAI -> Kiro path.

Session resolution now preserves explicit `prompt_cache_key`, `session_id`, and `conversation_id` fields first. If those are absent, Anthropic `cache_control` markers on system blocks, tools, or message content blocks are treated as an internal affinity hint: the translator hashes the prompt prefix through the last cache breakpoint and feeds that as the internal `prompt_cache_key`. This keeps same cached-prefix requests on the same Kiro conversation while still ignoring later uncached user text.

The translator does not forward `cache_control` as a made-up Kiro payload field. Requests without `cache_control` keep the Slack/thread anchor and deterministic fallback behavior.

## Tests

- `NODE_PATH=tests/node_modules tests/node_modules/.bin/vitest run --reporter=verbose --config tests/vitest.config.js tests/translator/claude-kiro-direct.test.js` (22/22)
- `git diff --check`

### DIFF
--- a/open-sse/executors/kiro.js
+++ b/open-sse/executors/kiro.js
@@ -384,33 +384,12 @@ export class KiroExecutor extends BaseExecutor {
             state.hasMeteringEvent = true;
           }
 
-          // Handle metricsEvent for token usage
-          if (eventType === "metricsEvent") {
-            // Extract usage data from metricsEvent payload
-            const metrics = event.payload?.metricsEvent || event.payload;
-            if (metrics && typeof metrics === 'object') {
-              const inputTokens = metrics.inputTokens || 0;
-              const outputTokens = metrics.outputTokens || 0;
-              // ponytail: Amazon Q upstream does not expose cache fields today,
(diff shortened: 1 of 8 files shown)


SemonCat:
Closing this for now. Follow-up testing showed Kiro native cachePoint does not implement the Anthropic cache_control prefix + dynamic user tail semantics Hermes needs; the live 9Router instance has also been reverted to the original image/mounts. This PR improves affinity/request shape, but it does not achieve the actual Hermes -> 9Router -> Kiro prompt-cache goal.

PR #1819, opened 2026-06-15
fix(codex): bind usage and quota calls to ChatGPT account

ryanngit:
## Summary

- resolves one trimmed ChatGPT account ID in `workspaceId`, `chatgptAccountId`, `accountId`, then ID-token order
- uses that identity consistently for Codex usage, reset-credit GET/POST, image requests, and quota auto-ping
- preserves per-connection proxy options
- rejects irreversible reset-credit consumption before network I/O when no account ID exists

## Scope

v0.5.45 already absorbed inference-header binding and bare-email dedup through #2477. This recut excludes those changes. Email is never used as `ChatGPT-Account-ID`; PAT/FedRAMP and organization-level dedup remain separate work (#2985/#941).

## Verification

- TDD baseline: 15 focused failures on pristine v0.5.45
- focused suite: 5 files, 51/51 passed
- changed-file ESLint: 0 errors; one unchanged anonymous-default-export warning also exists on pristine v0.5.45
- `git diff --check`: passed
- Gitleaks: no leaks

Base: `v0.5.45` (`6fcd2733`).

PR #2477, opened 2026-07-08
fix(codex): avoid bare-email OAuth dedup

lucasjustinudin:
## Summary

Avoid merging Codex OAuth connections by email alone when the ChatGPT account id is missing.

## Problem

Codex OAuth tokens are account-scoped and refresh tokens are rotated. The generic OAuth dedup path can merge a new Codex login into an existing row when both rows share the same email but do not both expose `providerSpecificData.chatgptAccountId`.

That can overwrite the first connection's token pair and make the existing account appear invalid after another Codex account is added.

## Changes

- For `provider === "codex"`, only update an existing OAuth row when both existing and incoming rows have the same `providerSpecificData.chatgptAccountId`.
- Use `workspaceId || chatgptAccountId || accountId` when setting the Codex `chatgpt-account-id` request header.

## Related

Related to the existing Codex workspace/account binding work in #941 and #1819. This PR keeps the scope narrow and covers the missing-account-id fallback case.

## Verification

```bash
node --check src/lib/db/repos/connectionsRepo.js
node --check open-sse/executors/codex.js
git diff --check
```

### DIFF
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -135,10 +135,17 @@ export class CodexExecutor extends BaseExecutor {
     headers["session_id"] = this._currentSessionId || credentials?.connectionId || "default";
     // Identify client type to Codex backend (matches official codex CLI)
     if (!headers["originator"]) headers["originator"] = "codex_cli_rs";
-    // Workspace binding header — improves account scope + cache affinity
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @lucasjustinudin for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2479, opened 2026-07-09
fix(cli): allow staged app bundle builds

ryanngit:
## Summary
- lets build-cli write the standalone app bundle to NINEROUTER_CLI_APP_DIR when set
- lets the MITM bundle follow the same staged app destination
- keeps the default output path unchanged at cli/app

## Verification
- node --check cli/scripts/build-cli.js
- node --check cli/scripts/buildMitm.js
- live deploy used the staged build flow successfully before swapping the bundle and restarting PM2 once


decolua:
Thanks! Reviewed and accepted — applying to master. 🙏

PR #2480, opened 2026-07-09
fix(gemini): handle Gemma 4 thinking replay

chy1211:
## Summary

- Map Gemma 4 models on the Gemini API to `gemini-level` thinking instead of budget-based Gemini thinking.
- Avoid replaying synthetic Gemini thought parts and `thoughtSignature` fields for Gemma 4 multi-turn tool history.
- Add regression coverage for Gemma 4 thinking config and OpenAI-to-Gemini tool-history conversion.

## Why

Gemma 4 on the Gemini API rejects some request shapes that are valid for other Gemini-family routes:

1. `generationConfig.thinkingConfig.thinkingBudget` / `includeThoughts` can produce `400 INVALID_ARGUMENT`; Gemma 4 expects level-based thinking (`thinkingLevel`) when thinking is explicitly requested.
2. Multi-turn tool history converted from OpenAI format can include synthetic replay artifacts such as assistant `reasoning_content` converted to Gemini `thought` parts and `thoughtSignature` attached to replayed `functionCall` parts. Gemma 4 rejects these with the same generic `400 INVALID_ARGUMENT`.

This keeps existing behavior for regular Gemini routes while special-casing Gemma 4.

## Test Plan

```bash
NODE_PATH=./node_modules ./tests/node_modules/.bin/vitest run \
  tests/translator/gemma4-gemini.test.js \
  tests/translator/thinking-unified.test.js \
  tests/unit/capabilities.test.js \
  --reporter=verbose
```

Result: 3 files passed, 38 tests passed.

Also ran:

```bash
git diff --check
```

### DIFF
--- a/open-sse/providers/capabilities.js
+++ b/open-sse/providers/capabilities.js
@@ -163,6 +163,8 @@ export const PATTERN_CAPABILITIES = [
   { pattern: "*gemini-2.5*",    caps: { vision: true, audioInput: true, videoInput: true, reasoning: true, search: true, thinkingFormat: "gemini-budget", thinkingRange: { min: 0, max: 24576 }, contextWindow: 1048576, maxOutput: 65536 } },
   { pattern: "*gemini-2*",      caps: { vision: true, audioInput: true, videoInput: true, search: true, contextWindow: 1048576, maxOutput: 65536 } },
   { pattern: "*gemini*",        caps: { vision: true, search: true, contextWindow: 1048576 } },
+  // Gemma 4 on Gemini API accepts thinkingLevel, not Gemini 2.5 thinkingBudget.
+  { pattern: "*gemma-4*",       caps: { vision: true, reasoning: true, thinkingFormat: "gemini-level", contextWindow: 128000 } },
   { pattern: "*gemma*",         caps: { vision: true, contextWindow: 128000 } },
   { pattern: "*nanobanana*",    caps: { vision: true, imageOutput: true } },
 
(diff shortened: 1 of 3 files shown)


chy1211:
Additional context / reproduction evidence:

We hit this through a Hermes -> 9Router -> Gemini/Gemma route using the combo `Hermes-Gemini-Gemma-4-31B` -> `gemini/gemma-4-31b-it`.

Observed 9Router log shape:

```text
POST /v1/chat/completions | Hermes-Gemini-Gemma-4-31B | 4 msgs | 48 tools | effort=high
[CHAT] Combo "Hermes-Gemini-Gemma-4-31B"
[COMBO] Trying model 1/1: gemini/gemma-4-31b-it
[ROUTING] Provider: gemini, Model: gemma-4-31b-it
[PENDING] END (ERROR) | provider=gemini | model=gemma-4-31b-it
[400]: { "error": { "code": 400, "message": "Request contains an invalid argument.", "status": "INVALID_ARGUMENT" } }
```

Two separate request-shape issues reproduced:

1. Thinking format mismatch
   - `reasoning_effort: "high"` on Gemma 4 can be normalized to Gemini budget-style `thinkingBudget`.
   - Gemma 4 on Gemini API expects level-style thinking config (`thinkingLevel`) instead.

2. Multi-turn tool history replay
   - First/simple requests can succeed.
   - A later agent/tool-use turn can fail when converted history contains synthetic Gemini replay artifacts, e.g. model turns with `thoughtSignature` attached to replayed `functionCall`, or previous assistant `reasoning_content` replayed as Gemini `thought` parts.
   - Removing those synthetic thought artifacts for `gemma-4-*` made the same multi-turn/tool-history request return HTTP 200 locally, while preserving the existing behavior for regular Gemini models.

The added regression test covers both cases and also checks that `gemini-2.5` still keeps the previous `thoughtSignature` behavior.

PR #2486, opened 2026-07-09
Fix Gemini CLI thinking and tool calls

dmdfami:
## Summary

- Raise Gemini `maxOutputTokens` floors when applying Gemini thinking config so explicit thinking modes have enough room for reasoning plus answer tokens.
- Add `request.toolConfig.functionCallingConfig.mode = "VALIDATED"` for Gemini CLI tool requests, matching the existing Antigravity behavior.
- Cover both behaviors with translator tests.

## Root Cause

Gemini CLI (`gc/*`) requests can arrive with small OpenAI `max_tokens` values. After translating `reasoning_effort` to Gemini `thinkingConfig`, Gemini can spend that output budget on thoughts and return blank content or `length`. Tool requests also lacked Cloud Code `toolConfig`, which can produce `MALFORMED_FUNCTION_CALL` for Gemini CLI even when `functionDeclarations` are valid.

## Validation

- `npx vitest run --config tests/vitest.config.js tests/translator/thinking-unified.test.js`
- `npx vitest run --config tests/vitest.config.js tests/translator/golden-request.test.js -t "Gemini CLI tool requests"`

Note: running the full `tests/translator/golden-request.test.js` currently fails on an unrelated Kiro snapshot mismatch in the existing suite; the new Gemini CLI targeted test passes.

## Runtime Verification

The same behavior was verified locally against the installed 9router runtime on `gc/gemini-3.1-pro-preview` high thinking for text, long-context, non-stream tool call, and stream tool call scenarios.

### DIFF
--- a/open-sse/translator/concerns/thinkingUnified.js
+++ b/open-sse/translator/concerns/thinkingUnified.js
@@ -148,16 +148,55 @@ function toKimiReasoningEffort(cfg) {
   return null;
 }
 
+const GEMINI_LEVEL_OUTPUT_FLOOR = {
+  minimal: 4096,
+  low: 8192,
+  medium: 16384,
+  high: 65535,
+};
+
+function geminiBudgetOutputFloor(budget) {
+  if (budget === -1) return 32768;
+  if (!Number.isFinite(budget)) return 32768;
+  if (budget <= 1024) return 8192;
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @dmdfami for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2498, opened 2026-07-09
fix(github): label Copilot profiles by account identity

ryanngit:
Adds stable labels for GitHub Copilot OAuth profiles.\n\n- Maps GitHub login/name/email from OAuth user info onto top-level connection fields\n- Uses stored GitHub login/email/name as fallback instead of generic Account N\n- Keeps existing connection dedupe behavior unchanged\n\nTested:\n- ./node_modules/.bin/vitest run --config tests/vitest.config.js --testTimeout 20000 tests/unit/db-sqlite-vs-lowdb.test.js

### DIFF
--- a/src/lib/db/repos/connectionsRepo.js
+++ b/src/lib/db/repos/connectionsRepo.js
@@ -56,6 +56,17 @@ function upsert(db, c) {
   );
 }
 
+function deriveConnectionName(data, fallbackName) {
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @ryanngit for the contribution! Reviewed and accepted — applying it to master. 🙏

PR #2503, opened 2026-07-10
fix: Codex Responses API 400 — input[0].content unknown parameter

datj9:
## Problem

Codex CLI requests through 9router intermittently fail with:
```
400 Unknown parameter: '"'input[0].content'"'
```

This happens when **caveman** or **ponytail** is enabled and the Codex CLI sends a request **without** `body.instructions` (some conversation turns). The system prompt injector falls through to the `input[]` unshift path and creates:

```js
{ role: "system", content: "prompt string" }  // no type:"message", content is bare string
```

The ChatGPT Codex Responses backend requires every `input[]` item to have a `type` field — without `type: "message"`, it rejects `content` as an unknown parameter.

## Root cause

`injectMessagesSystem()` in `open-sse/rtk/systemInject.js` did not distinguish between:
- **Chat Completions** (`messages[]`): `{ role, content: "string" }` — correct
- **Responses API** (`input[]`): needs `{ type: "message", role, content: [{ type: "input_text", text }] }`

### 1. Root cause — `systemInject.js`
- Pass `format` through to `injectMessagesSystem` so it knows when it's targeting the Responses API
- For Responses format: unshift proper `{ type: "message", role: "system", content: [{ type: "input_text", text }] }`
- `appendToOpenAIMessage` converts string content → typed array for Responses, and adds `type: "message"` if missing

### 2. Defense-in-depth — `codex.js`
- New `sanitizeCodexInput()` in `CodexExecutor.transformRequest()`:
  - Adds `type: "message"` to any input item with `role` but no `type`
  - Converts bare-string `content` to `[{ type: "input_text"|"output_text", text }]`
- Runs after `convertSystemToDeveloperRole`, before upstream dispatch
- Catches both the caveman/ponytail bug AND any client-side anomalies

## Tests

- **6 new tests** in `codex-request-transform.test.js`: malformed input items, caveman simulation, well-formed passthrough, non-message items untouched
- **9 new tests** in `system-inject-responses.test.js`: Responses-format injection, append to existing system, double injection (caveman+ponytail), Chat Completions backward compatibility
- All 25 new tests pass; existing tests unaffected

## Files changed
- `open-sse/rtk/systemInject.js` — format-aware injection (+25 lines)
- `open-sse/executors/codex.js` — `sanitizeCodexInput()` defense-in-depth (+20 lines)
- `tests/unit/codex-request-transform.test.js` — new (fork-only test file, now upstreamed)
- `tests/unit/system-inject-responses.test.js` — new

PR #2511, opened 2026-07-10
fix(codex): support Responses Lite transport

ryanngit:
## Summary

- forward explicit Codex Responses Lite opt-in plus bounded native metadata
- preserve `additional_tools`, force `parallel_tool_calls: false`, and set `reasoning.context: all_turns`
- keep `/responses/compact` unary and route compact requests before URL resolution
- avoid mutating `additional_tools` during system-prompt injection
- deep-clone request bodies before passthrough and account fallback
- recognize current `codex_exec` and `codex_cli_rs` user agents

## Problem

Responses Lite uses a native contract that differs from ordinary Responses requests. Dropping its internal header or injecting `content` into the leading `additional_tools` item causes upstream 400 errors. Lite also requires all-turn reasoning context and serial tool execution. Compact requests must reach `/backend-api/codex/responses/compact` as unary JSON, including retries.

## Trust boundary

Lite activates only when the client explicitly sends `X-OpenAI-Internal-Codex-Responses-Lite: true`. The router forwards a fixed metadata allowlist, validates `originator` and native Codex user agents, bounds metadata values to 16 KiB, and never copies client `authorization`, `cookie`, `proxy-authorization`, `x-api-key`, or `x-forwarded-for`.

## Verification

Tested on pristine `v0.5.45` (`6fcd2733`):
- focused executor/client/system matrix: 22/22 passed
- adjacent 66-test matrix: 62 passed; four failures reproduce identically on pristine v0.5.45, so candidate-only failures are zero
- concurrent compact and standard requests on shared executor route independently
- production `npm run build` passed
- Gitleaks found no leaks

#2518 overlaps the Lite transport goal but is not present in v0.5.45. #2503 overlaps only Responses system injection.


ryanngit:
## v0.5.45 recut

- Base: `upstream/master` = `v0.5.45` (`6fcd2733`)
- New head: `03bb0d83ccb421e384516152efbefae692c6177e`
- Merge state: `CLEAN`
- Removed old v0.5.40 merge commit; branch now contains one feature commit
- Focused matrix: 4 files, 22/22 tests passed
- Adjacent matrix: zero candidate-only failure identities
- Production build: passed
- Changed-file ESLint and `git diff --check`: clean
- Gitleaks: no leaks

Recut adds explicit forbidden-header coverage and a concurrent compact/standard routing regression for the shared Codex executor.

PR #2515, opened 2026-07-10
fix(codex): retry rejected input namespaces once

nguyenha935:
## Problem

Codex GPT-5.6 Responses Lite sessions can fail after a tool call with:

```
400 Unknown parameter: 'input[150].namespace'
```

The rejected field is attached to a historical `input[]` function call. The namespace tool declaration itself is still valid and must be preserved.

## Root cause

The upstream Codex endpoint can reject `namespace` on historical input items even though the current client contract uses namespaced tools. Retrying accounts or combo models cannot change that request-schema incompatibility.

## Behavior before

- The exact schema error is returned immediately.
- Higher layers may rotate accounts or combo models with the same invalid history.

## Behavior after

- Retry the same request once after the exact `input[n].namespace` error.
- Remove only `namespace` fields from `input[]` history.
- Preserve namespace tool declarations, call IDs, ordering, auth, and all other input fields.
- Unrelated 400 responses are not retried.

## Backward compatibility

The compatibility retry is gated by HTTP 400 plus the exact upstream parameter path. Standard Responses requests and other Codex errors are unchanged.

## Tests

- `cd tests && npx vitest run unit/codex-namespace-retry.test.js unit/codex-tool-normalization.test.js`
- 7 tests passed.
- `git diff --check` passed.

## Live verification

Verified with Codex CLI and Codex Desktop 0.144.0-alpha.4 against GPT-5.6 Sol through 9Router. Tool execution and the following resumed turn completed without the namespace error.

## Related OpenAI issues

- OpenAI Codex #31760
- OpenAI Codex #31875
- OpenAI Codex #31882

## Related 9Router PRs

- #2511 handles the broader Responses Lite transport contract; this PR is a narrow backend-compatibility retry for the observed historical namespace rejection.
- #1193 covers broader namespace round-trip behavior and is not copied here.

No header or credential behavior changes. The retry stays on the selected account and mutates only the request body copy already owned by the executor.

### DIFF
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -44,6 +44,21 @@ const RESPONSES_API_ALLOWLIST = new Set([
   "text"
 ]);
 
+export function stripRejectedCodexInputNamespaces(body) {
+  if (!Array.isArray(body?.input)) return false;
+  let stripped = false;
+  for (const item of body.input) {
+    if (!item || typeof item !== "object" || Array.isArray(item) || !("namespace" in item)) continue;
+    delete item.namespace;
+    stripped = true;
+  }
+  return stripped;
+}
+
+function isRejectedCodexInputNamespace(status, errorText) {
+  return Number(status) === 400
+    && /unknown parameter:\s*['"]input\[\d+\]\.namespace['"]/i.test(String(errorText || ""));
+}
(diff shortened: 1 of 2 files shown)


nguyenha935:

PR #2516, opened 2026-07-10
fix(codex): avoid fallback on request schema errors

nguyenha935:
## Problem

A Codex request-schema 400 currently enters the generic fallback path. That can mark a healthy account unavailable, rotate through other accounts, and fan the same malformed request across every model in a combo.

## Root cause

`checkFallbackError()` deliberately treats unmatched errors as transient. Request-schema incompatibilities are deterministic and model/account independent, so the generic policy is incorrect for this narrow Codex case.

## Behavior before

- Codex schema 400 errors can lock or rotate accounts.
- Combo routing can send the same invalid request to subsequent models.

## Behavior after

- Classify only Codex HTTP 400 schema incompatibilities with recognized metadata and schema-specific messages.
- Return the original error without calling `markAccountUnavailable()`.
- Stop combo fan-out when a `cx/*` model returns the classified error.
- Keep normal fallback for 401, 403, 429, usage limits, capacity, overload, and unsupported-account/model errors.

## Backward compatibility

The classifier requires provider `codex`, status 400, recognized error metadata, and a schema-specific message. Other providers and existing retry/fallback rules are unchanged.

## Tests

- `cd tests && npx vitest run unit/codex-schema-fallback.test.js unit/combo-routing.test.js`
- 12 tests passed.
- `git diff --check` passed.

## Live verification

The original `input[150].namespace` failure was reproduced through a GPT-5.6 Sol combo. After applying the classifier with the namespace compatibility fix, the request completed without locking the account or falling through the combo.

## Related OpenAI issues

- OpenAI Codex #31760
- OpenAI Codex #31875
- OpenAI Codex #31882

## Related 9Router PRs

- #2511 Responses Lite transport
- Companion namespace-retry PR from this fork

No auth or header changes. Client error bodies are inspected only for classification and are not logged with credentials.

### DIFF
--- a/open-sse/config/errorConfig.js
+++ b/open-sse/config/errorConfig.js
@@ -28,6 +28,14 @@ export const DEFAULT_ERROR_MESSAGES = {
   504: "Gateway timeout"
 };
 
+export const CODEX_REQUEST_SCHEMA_ERROR_CODES = new Set([
+  "invalid_request_error",
+  "unknown_parameter",
+  "unsupported_value",
+]);
+
+export const CODEX_REQUEST_SCHEMA_ERROR_PATTERN =
+  /unknown[_ ]parameter|unsupported[_ ]value|(?:invalid|unsupported)[_ ]?(?:parameter|schema)|(?:parameter|schema).*(?:invalid|unsupported)/i;
 // Exponential backoff config for rate limits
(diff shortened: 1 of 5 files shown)


nguyenha935:

PR #2517, opened 2026-07-10
fix(api): support Codex model catalog responses

nguyenha935:
## Problem

Codex Desktop requests `GET /v1/models?client_version=...` and decodes a Codex catalog envelope containing `models`. 9Router correctly returns the OpenAI-compatible `{ object, data }` envelope, which causes Codex Desktop to log `missing field models` even though `/v1/responses` still works.

## Root cause

The same `/models` path is used by two compatible clients with different response envelopes.

## Behavior before

- OpenAI-compatible clients receive `{ object: list, data: [...] }`.
- Codex Desktop model refresh fails to decode that response.

## Behavior after

- Standard clients still receive the exact existing OpenAI envelope.
- Requests with both a Codex user agent and `client_version` receive the same OpenAI fields plus `models: []`.
- The empty Codex catalog intentionally lets Codex use its bundled, version-matched model capabilities instead of 9Router fabricating private `ModelInfo` fields.

## Backward compatibility

Content negotiation is narrow. OpenAI SDKs, OpenCode, and clients without the Codex version query receive byte-shape-compatible JSON with no additional field.

## Tests

- `cd tests && npx vitest run unit/models-response-compatibility.test.js`
- 7 tests passed.
- Covers Codex Desktop, `codex_cli_rs`, `codex_exec`, standard OpenAI user agents, and missing version queries.
- `git diff --check` passed.

## Live verification

Verified with Codex Desktop 0.144.0-alpha.4 through the production HTTPS reverse proxy. Model refresh decoded successfully and GPT-5.6 Sol returned a streamed response.

## Related 9Router PRs

- #2511 Responses Lite transport (independent request-path concern)

No auth, credential, proxy, or header-forwarding behavior changes. The user agent is used only to select the response envelope.

### DIFF
--- a/src/app/api/v1/models/route.js
+++ b/src/app/api/v1/models/route.js
@@ -475,10 +475,24 @@ export async function OPTIONS() {
  * GET /v1/models - OpenAI compatible models list (LLM/chat models only by default).
  * For other capabilities use /v1/models/{kind} (image, tts, stt, embedding, image-to-text, web).
  */
-export async function GET() {
+export function isCodexModelsRequest(request) {
+  if (!request?.url || typeof request.headers?.get !== "function") return false;
+  const clientVersion = new URL(request.url).searchParams.get("client_version");
+  const userAgent = request.headers.get("user-agent") || "";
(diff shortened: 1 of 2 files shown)


nguyenha935:

PR #2500, opened 2026-07-10
feat(providers): add max thinking level for gpt-5.6-sol

raflyazf:
## Summary
- Adds `max` to the thinking-level picker for `gpt-5.6-sol` only (Codex / OpenAI GPT-5.6 Sol).
- Other Codex models keep existing levels (`low` / `medium` / `high` / `xhigh`).
- Unit tests cover inclusion for `gpt-5.6-sol` and non-regression for other models.

## Context
- OpenAI docs list GPT-5.6 Sol reasoning efforts including `max`.
- Live probes via local 9Router against ChatGPT/Codex: `max` returns 200; `ultra` is rejected as invalid `reasoning.effort` (Ultra is a Codex subagent mode, not a wire effort).
- Wire path already maps internal `max` appropriately in the thinking layer.

## Test plan
- [x] `cd tests && npx vitest run unit/thinking-levels-gpt56-sol.test.js --config vitest.config.js`
- [ ] Open dashboard → Providers → OpenAI Codex with `gpt-5.6-sol` available
- [ ] Confirm Thinking dropdown includes **Max**
- [ ] Select Max and copy model name → suffix `(max)` only on `gpt-5.6-sol`
- [ ] Request `cx/gpt-5.6-sol` with `reasoning_effort: "max"` succeeds

PR #2518, opened 2026-07-10
fix(codex): support GPT-5.6 Responses Lite clients

nguyenha935:
## Problem

Codex GPT-5.6 Responses Lite clients expose several related compatibility failures through 9Router:

- the Lite opt-in header and native payload contract can be lost or normalized incorrectly;
- resumed tool history can fail with `Unknown parameter: 'input[n].namespace'`;
- deterministic schema 400 responses can rotate/lock healthy accounts and fan out through combos;
- Codex Desktop model refresh expects a `models` catalog envelope while the shared endpoint returns the OpenAI-compatible `{ object, data }` envelope.

These failures belong to the same Codex client compatibility path and are fixed together here.

## Root cause

9Router previously treated Codex requests as standard Responses requests at several boundaries:

1. The Codex executor did not preserve the Responses Lite header, native input items, `parallel_tool_calls`, or Lite reasoning context.
2. The current upstream Codex backend can reject `namespace` on historical `input[]` function calls even though namespace tool declarations remain valid.
3. Generic account/combo fallback treats unmatched errors as transient, but request-schema incompatibilities are deterministic.
4. `/v1/models` is shared by standard OpenAI clients and Codex clients with different response-envelope expectations.

## Behavior before

- Responses Lite metadata or native items could be dropped or rewritten.
- `input[n].namespace` caused a terminal 400 or repeated fallback attempts.
- Healthy Codex accounts could be marked unavailable for a malformed request.
- Codex Desktop logged `missing field models` during catalog refresh.

## Behavior after

- Forward only the explicit Responses Lite opt-in header; client authorization, cookies, and API keys are never forwarded.
- Preserve `additional_tools`, namespace declarations/calls, call IDs, encrypted reasoning, input order, and `parallel_tool_calls`.
- Keep GPT-5.6 `max` reasoning on the Lite wire contract while preserving existing model behavior.
- Retry the same selected account once after the exact `input[n].namespace` rejection, removing only historical input namespace fields and preserving tool declarations.
- Return classified Codex schema 400 errors without locking/rotating accounts or fanning through combo models.
- Content-negotiate `/v1/models`: standard clients retain the exact `{ object, data }` envelope; Codex clients with `client_version` receive the same data plus `models: []` so Codex uses its bundled version-matched capabilities.

## Backward compatibility

- Standard Responses requests and GPT-5.5/GPT-5.4 behavior remain unchanged.
- Existing 401, 403, 429, quota, capacity, overload, and token-refresh fallback behavior remains unchanged.
- Unsupported-account/model 400 errors are not classified as request-schema errors.
- OpenAI SDKs, OpenCode, and clients without the Codex version query receive the unchanged model-list envelope.

## Tests

```bash
cd tests
npx vitest run \
  unit/codex-responses-lite.test.js \
  unit/codex-namespace-retry.test.js \
  unit/codex-schema-fallback.test.js \
  unit/combo-routing.test.js \
  unit/models-response-compatibility.test.js \
  unit/codex-tool-normalization.test.js
```

- 6 test files passed.
- 32 tests passed.
- `git diff --check` passed.
- Production Next.js build passed on Linux through the deployment updater.

## Live verification

Verified against GPT-5.6 Sol using both Codex CLI and Codex Desktop 0.144.0-alpha.4 through a production HTTPS reverse proxy:

- text response;
- shell tool call;
- resumed turn containing tool history;
- Responses Lite namespace history;
- combo routing/account fallback;
- Desktop `/v1/models?client_version=0.144.0` refresh;
- streamed `/v1/responses` completion.

The original `input[150].namespace` failure no longer occurs, and Desktop model refresh no longer reports `missing field models`.

## Related OpenAI issues

- OpenAI Codex #31760
- OpenAI Codex #31875
- OpenAI Codex #31882

## Related 9Router PRs

- #2511 independently covers Responses Lite transport and compact handling. Credit to @ryanngit for that work; this PR includes the production-verified GPT-5.6 compatibility path together with namespace retry, fallback policy, and Desktop catalog negotiation.
- #1193 covers broader namespace round-trip behavior; no unrelated changes from that PR are copied.
- #2503 addressed malformed Responses message injection; its normalization direction was reviewed but not copied wholesale.
- #2500 covered GPT-5.6 Sol reasoning-level UI; this PR does not claim that work.

- Forwarded Codex headers are allowlisted in `open-sse/config`.
- Client `authorization`, `cookie`, `proxy-authorization`, and `x-api-key` are not forwarded.
- 9Router continues to use the OAuth account selected by its own account router.
- No tokens, account IDs, emails, captures, runtime scripts, deployment patches, or personal configuration are included.

### DIFF
--- a/open-sse/config/codexConstants.js
+++ b/open-sse/config/codexConstants.js
@@ -1,0 +1,7 @@
+export const CODEX_RESPONSES_LITE_HEADER = "x-openai-internal-codex-responses-lite";
+export const CODEX_RESPONSES_LITE_HEADER_VALUE = "true";
+export const CODEX_FORWARD_HEADER_ALLOWLIST = Object.freeze([CODEX_RESPONSES_LITE_HEADER]);
+export const CODEX_RESPONSES_LITE_INPUT_TYPES = Object.freeze([
+  "additional_tools", "function_call", "function_call_output", "custom_tool_call",
+  "custom_tool_call_output", "tool_search_call", "tool_search_output", "reasoning", "message",
+]);
--- a/open-sse/config/errorConfig.js
+++ b/open-sse/config/errorConfig.js
@@ -28,6 +28,14 @@ export const DEFAULT_ERROR_MESSAGES = {
   504: "Gateway timeout"
 };
 
+export const CODEX_REQUEST_SCHEMA_ERROR_CODES = new Set([
+  "invalid_request_error",
+  "unknown_parameter",
+  "unsupported_value",
+]);
+
+export const CODEX_REQUEST_SCHEMA_ERROR_PATTERN =
+  /unknown[_ ]parameter|unsupported[_ ]value|(?:invalid|unsupported)[_ ]?(?:parameter|schema)|(?:parameter|schema).*(?:invalid|unsupported)/i;
 // Exponential backoff config for rate limits
 export const BACKOFF_CONFIG = {
   base: 2000,
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -11,6 +11,12 @@ import { getModelUpstreamId } from "../config/providerModels.js";
 import { DEFAULT_RETRY_CONFIG, HTTP_STATUS, resolveRetryEntry } from "../config/runtimeConfig.js";
 import { dbg } from "../utils/debugLog.js";
 import { resolveSessionId } from "../utils/sessionManager.js";
+import {
+  CODEX_FORWARD_HEADER_ALLOWLIST,
+  CODEX_RESPONSES_LITE_HEADER,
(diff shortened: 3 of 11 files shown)


decolua:

nguyenha935:
Thanks again for reviewing this. I re-audited the current upstream `master` at `9845a170` (v0.5.30). The GPT-5.6 registry/pricing/thinking entries are present, but I could not find the Responses Lite header/payload handling, one-time retry for rejected `input[].namespace`, Codex-specific `/v1/models` catalog envelope, or request-schema fallback guard from this PR. Could you point me to the commit or equivalent implementation that absorbed those pieces? I am keeping the downstream compatibility patch for now and will remove it once the upstream equivalent is confirmed.

PR #2530, opened 2026-07-11
fix: resolve better-sqlite3 parameter array binding crash

### LINKED ISSUE #2529: TypeError: SQLite3 can only bind... (Internal Server Error) when updating settings using better-sqlite3 driver
I encountered an "Internal Server Error" when trying to set my new password after logging in for the first time.

What’s interesting is that this works perfectly on my local machine, but fails when I deploy it to my server environment.

After some debugging, I discovered that the issue depends entirely on which SQLite driver 9router dynamically loads:

On my local machine, it falls back to node:sqlite or sql.js, both of which handle parameter arrays correctly.
On my server, better-sqlite3 is installed globally. 9router prioritizes better-sqlite3 and initializes the betterSqliteAdapter.
The betterSqliteAdapter has a parameter binding bug where it passes query parameters as a raw JavaScript array instead of spreading them, which crashes better-sqlite3 on any parameterized query (like updating the password in the settings).

Error Details
When submitting a new password, the server crashes with:

TypeError: SQLite3 can only bind numbers, strings, bigints, buffers, and null
This happens because better-sqlite3 expects parameters to be unpacked/spread (e.g., stmt.run(val1, val2)) but the adapter passes the raw params array object.

Steps to Reproduce
Run 9router in an environment where the better-sqlite3 module is installed and successfully loaded by the app.
Log in to the dashboard using the default password (123456).
Attempt to set a new password.
The UI returns an "Internal Server Error" (HTTP 500).

Root Cause & Proposed Fix
The issue is located in app/src/lib/db/adapters/betterSqliteAdapter.js (and naturally propagated into the .next-cli-build compiled chunks).

The parameters are passed without the spread operator (...), unlike the nodeSqliteAdapter and bunSqliteAdapter which do it correctly.

Current buggy code:

run(sql, params = []) { return prepare(sql).run(params); },
get(sql, params = []) { return prepare(sql).get(params); },
all(sql, params = []) { return prepare(sql).all(params); },

cubewin07:
This PR fixes the "Internal Server Error" caused by the `betterSqliteAdapter` passing query parameters as a raw JavaScript array instead of spreading them. 

By using the spread operator (`...params`) on the `run`, `get`, and `all` statements, we prevent `better-sqlite3` from crashing when executing parameterized queries (like updating a user password).

Fixes #2529

### DIFF
--- a/src/lib/db/adapters/betterSqliteAdapter.js
+++ b/src/lib/db/adapters/betterSqliteAdapter.js
@@ -40,9 +40,9 @@ export function createBetterSqliteAdapter(filePath) {
 
   return {
     driver: "better-sqlite3",
-    run(sql, params = []) { return prepare(sql).run(params); },
-    get(sql, params = []) { return prepare(sql).get(params); },
-    all(sql, params = []) { return prepare(sql).all(params); },
+    run(sql, params = []) { return prepare(sql).run(...params); },
+    get(sql, params = []) { return prepare(sql).get(...params); },
+    all(sql, params = []) { return prepare(sql).all(...params); },
     exec(sql) { return db.exec(sql); },
     transaction(fn) { return db.transaction(fn)(); },
     checkpoint() { try { db.pragma("wal_checkpoint(TRUNCATE)"); } catch {} },


decolua:
Thanks @cubewin07 for the contribution! Reviewed and merged into master. 🙏

PR #2534, opened 2026-07-11
fix(xai): strip reasoning_effort for grok-composer; dashboard thinking None

gitcommit90:
## Problem

`grok-composer-2.5-fast` (xAI) returns HTTP 400 when the chat request includes `reasoningEffort` / `reasoning_effort`, including when clients send `none`. Hermes and other OpenAI-compatible clients that set `agent.reasoning_effort` therefore fail through 9router while bare requests succeed.

The provider dashboard thinking picker also hid **None** (filtered out of dropdown options), so there was no way to tell the gateway to omit reasoning params for a provider.

## Fix

- **paramSupport**: strip `thinking`, `reasoning_effort`, and `reasoning` for `xai` + `grok-composer` models before upstream call.
- **capabilities**: mark `*grok-composer*` as non-reasoning so unified thinking normalization does not re-inject effort.
- **chatCore**: when `providerThinking.mode === "none"`, delete thinking/reasoning fields instead of injecting effort.
- **Dashboard**: include **None** in the thinking dropdown; `none` does not append a model suffix.
- **Test**: unit coverage in `param-support.test.js`.
- **CHANGELOG**: v0.5.30 fixes entry.

## Verification

`npx vitest run unit/param-support.test.js -t grok-composer` passes.

Fixes Mattermost/Hermes failures when global model is `grok-composer-2.5-fast` via rerouted with `reasoning_effort: medium`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 ## Fixes
 - **xAI**: strip `reasoning_effort` for `grok-composer` models (upstream rejects any reasoningEffort); honor provider thinking mode `none` — Joseph Yaksich
+- **grok-cli**: guard `transformRequest()` with capabilities (no reasoning on composer/build); cap tools at 200 for cli-chat-proxy — folded from #2539 / @hameduc
 - **Cloudflare-AI**: support accountId in bulk key import (#2449)
 - **DB**: backup on schema change, MCP child cleanup, codex models, usage providers OOM
 - **Codex**: avoid bare-email OAuth dedup (#2477)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - **Proxy-Pools**: auto-rotate strategy for no-auth providers (#2409)
 
 ## Fixes
+- **xAI**: strip `reasoning_effort` for `grok-composer` models (upstream rejects any reasoningEffort); honor provider thinking mode `none` — Joseph Yaksich
 - **Cloudflare-AI**: support accountId in bulk key import (#2449)
 - **DB**: backup on schema change, MCP child cleanup, codex models, usage providers OOM
 - **Codex**: avoid bare-email OAuth dedup (#2477)
--- a/open-sse/executors/grok-cli.js
+++ b/open-sse/executors/grok-cli.js
@@ -9,6 +9,7 @@ import { normalizeResponsesInput } from "../translator/formats/responsesApi.js";
 import { getModelUpstreamId } from "../config/providerModels.js";
 import { resolveSessionId } from "../utils/sessionManager.js";
 import { getConsistentMachineId } from "../shared/machineId.js";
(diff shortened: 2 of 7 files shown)


hameduc:
Nice PR — the `paramSupport.js` strip rule and `chatCore.js` mode "none" handling are the right architectural approach, and the dashboard None option + tests are great.

One gap: the `grok-cli` executor (`cli-chat-proxy.grok.com`) has its own `transformRequest()` that hardcodes `body.reasoning = { effort: "high", summary: "concise" }` **after** the translator layer runs. The `paramSupport.js` strip rule (scoped to `provider: "xai"`) doesn't catch it because:

1. It's keyed to provider `xai`, not `grok-cli`
2. The executor rebuilds `body.reasoning` unconditionally in `transformRequest()`, overriding whatever the translator stripped

I opened #2539 which adds the `getCapabilitiesForModel()` guard inside the grok-cli executor's `transformRequest()` itself, plus covers `grok-build` and caps tools at 200 (the cli-chat-proxy maximum). These two PRs are complementary — #2534 handles the `xai` provider and dashboard UX, #2539 handles the `grok-cli` executor and tool limits.

Happy to close #2539 and fold its executor-side fixes into this PR if you prefer a single combined PR.

gitcommit90:
Thanks @hameduc — good catch on the executor ordering.

You're right that `paramSupport` + `chatCore` `none` only cover the **`xai`** / DefaultExecutor path. Anything routed through **`grok-cli`** still gets `body.reasoning` re-injected in `transformRequest()` after translation, so the strip rule never wins there.

Happy to treat **#2539 as complementary** (grok-cli guard + tool cap + build/composer caps). I'm fine either way for upstream:

- **Merge both** as-is with cross-links, or  
- **Fold #2539 into #2534** if maintainers prefer one PR — I can cherry-pick your executor/capability bits onto this branch if that's easier to review.

Either way the dashboard **None** option and provider-level `none` handling in `chatCore` still help clients that send `reasoning_effort` into the `xai` provider without wanting effort forwarded.

Appreciate the review.

hameduc:
Sounds good — go ahead and cherry-pick the executor guard, capability entries for grok-build/grok-composer, and the 200-tool cap from #2539 onto this branch. Those changes are isolated to `open-sse/executors/grok-cli.js` and `open-sse/providers/capabilities.js` so they should apply cleanly on top of your work here.

gitcommit90:
@hameduc Done — cherry-picked the isolated bits from #2539 onto this branch:

- `open-sse/executors/grok-cli.js` (capabilities guard + 200-tool cap)
- `open-sse/providers/capabilities.js` (`grok-composer` / `grok-build` MODEL_CAPABILITIES + patterns)

Kept the rest of #2534 (xai `paramSupport`, dashboard **None**, `chatCore` `none`, tests). CHANGELOG updated; pushed to `fix/xai-grok-composer-reasoning-effort`.

Happy for you to close #2539 if this single PR is enough for review.

PR #2541, opened 2026-07-11
fix: allow application/x-ndjson streaming from ollama-local (closes #2386)

### LINKED ISSUE #2386: Bug: Ollama 0.31.1 streaming fails on OpenAI endpoint (/v1/chat/completions)
### Environment

- 9router: v0.5.18
- Ollama: 0.31.1
- Windows 11
- Cursor latest

### Reproduce

Ollama works normally.

```
ollama run qwen2.5:14b
```

works.

```
curl http://127.0.0.1:11434/api/generate
```

works.

9router detects all Ollama models.

```
GET /v1/models
```

returns:

- ollama-local/qwen2.5:14b
- ollama-local/qwen2.5:7b
- ollama-local/deepseek-coder:6.7b
- ollama-local/llama3

### stream=false

This works:

```bash
curl -X POST http://127.0.0.1:20128/v1/chat/completions ^
-H "Content-Type: application/json" ^
-H "Authorization: Bearer xxx" ^
-d "{\"model\":\"ollama-local/qwen2.5:7b\",\"stream\":false,\"messages\":[{\"role\":\"user\",\"content\":\"hello\"}]}"
```

returns a normal response.

### stream=true

Cursor always uses stream=true.

9router log:

```
blocked pipe:
Upstream returned non-SSE response (application/x-ndjson) [200]

STREAM OLLAMA-LOCAL
error: upstream non-SSE: 200
```

Looks like the OpenAI compatible endpoint does not convert Ollama's application/x-ndjson stream into SSE correctly.

qveys:
Closes #2386.

## Root cause

Ollama's native /api/chat streams as `application/x-ndjson`, never SSE. The non-SSE guard in `handleStreamingResponse` (open-sse/handlers/chatCore/streamingHandler.js) only special-cased `text/event-stream` and `application/json`, so any ndjson response from Ollama was treated as an upstream error page (e.g. a Cloudflare 5xx page) and blocked - before it ever reached the existing Ollama ndjson->SSE transcoder (parseSSELine + ollamaToOpenAIResponse). This matches the log captured in #2386 exactly: `blocked pipe: ... / upstream non-SSE: 200`.

`stream:false` always worked because it skips this streaming code path entirely - which is why the bug only shows up with clients that default to streaming (Cursor, Claude Code, etc).

While tracing this I found a second, related bug: when this (or any non-HTTP) error occurs, `markAccountUnavailable` (src/sse/services/auth.js) collapses the real error into a generic "Provider error" string whenever errorText isn't a plain string, and only logs to console when a numeric HTTP status is present. So a streaming-parse failure like this one left zero usable trace server-side - just the generic error surfaced to the client.

## Fix

- `streamingHandler.js`: let `application/x-ndjson` through the guard specifically when `targetFormat === FORMATS.OLLAMA`; reuse the shared `createErrorResult` helper for the remaining blocked-response path instead of a hand-built `{ success, response }` object, so status/error survive downstream.
- `auth.js`: extract `errorText.message` (or a JSON-stringified fallback) instead of defaulting straight to "Provider error", and log the real cause even when there's no numeric HTTP status.

## Testing

Repro'd locally against a real `ollama-local` connection (v0.5.18 and v0.5.30): `stream:true` requests to `ollama-local/*` models failed 100% of the time with the generic "Provider error" before this change, and succeeded consistently after, producing well-formed `chat.completion.chunk` SSE events ending in the usual `finish_reason`. Verified `stream:false` and non-Ollama streaming providers (tested against groq and mistral) are unaffected.

The repo's Vitest suite fails to run independent of this change (pre-existing circular import error, reproducible on an unmodified checkout), so I validated the transcode/guard logic with a standalone script driving the real production modules directly instead.

### DIFF
--- a/open-sse/handlers/chatCore/streamingHandler.js
+++ b/open-sse/handlers/chatCore/streamingHandler.js
@@ -8,6 +8,7 @@ import { buildAbortedResponsesTerminalBytes } from "../../utils/responsesStreamH
 import { buildRequestDetail, extractRequestConfig, saveUsageStats, formatDoneLine } from "./requestDetail.js";
 import { saveRequestDetail } from "@/lib/usageDb.js";
 import { SSE_HEADERS_CORS as SSE_HEADERS } from "../../utils/sseConstants.js";
+import { createErrorResult } from "../../utils/error.js";
 
 // Codex returns Responses API SSE → which client format to translate INTO, by request sourceFormat.
 // Gemini-family all map to ANTIGRAVITY decoder; unknown sources fall back to OPENAI.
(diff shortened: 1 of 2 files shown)


qveys:
Addressed all 3 review comments in the latest commit:
- reason is now clamped to 100 chars regardless of which branch (string vs Error/object) produced it
- safeStringifyError guarantees a string return even when JSON.stringify returns undefined
- the log.warn line uses the same status ?? "no-status" fallback as console.error, so it no longer prints a literal [undefined]

PR #2542, opened 2026-07-11
fix(headroom): allow larger prompt compression to finish

KunN-21:
## Problem

9router aborted Headroom `/v1/compress` after 3 seconds. Large local compression can take longer, so affected requests continued without compression.

## Change

Raise fail-open bridge timeout from 3s to 15s. Add regression test for default timeout.

## Scope

No routing, provider, OAuth, dashboard, or retry behavior changed. If compression exceeds 15s, 9router keeps existing fail-open behavior and sends original request.

## Test

`npx vitest run tests/unit/headroom.test.js`

### DIFF
--- a/open-sse/rtk/headroom.js
+++ b/open-sse/rtk/headroom.js
@@ -5,7 +5,7 @@ import {
   openaiToOpenAIResponsesRequest,
 } from "../translator/request/openai-responses.js";
 
-const DEFAULT_TIMEOUT_MS = 3000;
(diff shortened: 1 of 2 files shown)


KunN-21:
Maintenance note: #3493 is the intended broader replacement for the Headroom timeout/lifecycle path. It is still open and undergoing validation, so this PR remains open until that replacement lands and the original timeout regression is confirmed covered. This is not a claim that the replacement has merged.

PR #2563, opened 2026-07-12
fix(cursor): bump client version and speak HTTP/2 to AgentService

### LINKED ISSUE #2487: Problem with Cursor IDE, shows up as an outdated version
HTTP 429: [cursor/default] [429]: {"error":{"message":"Update Required","type":"rate_limit_error","code":"ERROR_GPT_4_VISION_PREVIEW_RATE_LIMIT"}} (reset after 2s)

This is the error I get when I try to test a model after it imports my access token and Machine ID automatically.

long2ice:
## Summary
- Bump `x-cursor-client-version` from `3.1.0` → `3.11.13` so Cursor stops rejecting requests with `ERROR_GPT_4_VISION_PREVIEW_RATE_LIMIT` / "Update Required" (fixes #2487).
- Route Cursor **AgentService** (`agent.api5.cursor.sh`, HTTP/2-only) through Node `http2` duplex instead of `fetch`/undici, which only speaks HTTP/1.1 and fails with `fetch failed`.
- Teach `chatCore` to honor executor `responseFormat` when AgentService already decodes to OpenAI-compatible SSE.
- Fetch the live `GetUsableModels` catalog (also over HTTP/2) for account-specific model lists in the dashboard / `/v1/models`.

## Test plan
- [ ] Import Cursor OAuth credentials; provider test / chat with `cu/default` no longer returns 429 `Update Required`
- [ ] Claude Code (or any client that always sends tool schemas) against `cu/default` streams a normal reply instead of hanging / `AgentService request failed: fetch failed`
- [ ] Dashboard Cursor model picker loads live models (or falls back to static catalog on failure)
- [ ] `tests/__baseline__/providers-baseline.json` clientVersion matches registry (`3.11.13`)

### DIFF
--- a/open-sse/executors/cursor.js
+++ b/open-sse/executors/cursor.js
@@ -1,8 +1,11 @@
 import { BaseExecutor } from "./base.js";
-import { PROVIDERS } from "../config/providers.js";
+import { PROVIDERS, PROVIDER_OAUTH } from "../config/providers.js";
 import { HTTP_STATUS } from "../config/runtimeConfig.js";
 import {
   generateCursorBody,
+  encodeField,
+  wrapConnectRPCFrame,
+  decodeMessage,
   parseConnectRPCFrame,
   extractTextFromResponse
 } from "../utils/cursorProtobuf.js";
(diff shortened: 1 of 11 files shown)


long2ice:
Closing — not pursuing this Cursor AgentService / ChatService path further for now.

PR #2574, opened 2026-07-13
fix(stream): emit client keepalive on silent upstream IVT-0666

nhattran998:
## Summary

- Prevents client-side `Response stalled mid-stream` errors when sub-agents return large output to the main agent (huge prompt → slow upstream TTFT → client watchdog fires while 9router is silent).
- Adds a client-facing SSE keepalive heartbeat (`: ka` comment, spec-ignored by every SSE parser) emitted every 15s while the translate-transform output is silent.

## Root cause

`pipeWithDisconnect` measured stall on raw upstream bytes (360s) but emitted **nothing** to the client during upstream silence. Only `kiro` had a reactive keepalive — and only mid-stream, not during the TTFT gap. The client's stream-read watchdog (~60s) fired before 9router's own stall timer, producing `Response stalled mid-stream`.

## Changes

- `fix(stream)`: `createKeepaliveStream(ms)` TransformStream in `open-sse/utils/streamHandler.js`, wired into `pipeWithDisconnect` after the translate transform. New `STREAM_CLIENT_KEEPALIVE_MS` env (default 15000) in `open-sse/config/runtimeConfig.js`. Default-on for all providers; `clientKeepaliveMs: 0` disables per-call.
- `test(stream)`: 4 regression tests in `tests/unit/stream-keepalive.test.js` — emits keepalive when silent, silent while chunks flow, forwards chunks + keepalive in gaps, disabled when ms=0.

## Quality Gates

- [x] `deslop` skill run (0 slop — all comments are WHY/spec)
- [~] lint skipped (no `lint` script; root `eslint.config.mjs` crashes with pre-existing `defaultMeta` error, unrelated to this change)
- [~] Review Army skipped — simple change (3 files, ~136 lines, 1 module, no auth/security/migration); TDD-covered, zero regressions

## Testing

- [x] 4/4 new keepalive tests pass
- [x] `responses-abort-terminal` 4/4 pass (no regression)
- [x] `force-stream-config` (2/4) and `executor-const-guard` (14/2) fail identically on `master` — pre-existing, unrelated

## Related Issues

Ref: IVT-0666


nhattran998:
Retargeting to investtal/9router (our fork). New PR opening there.

PR #2580, opened 2026-07-13
Improve Kiro direct session cache reuse

SemonCat:
## Summary

This improves Kiro direct request shaping so resumed client sessions reuse Kiro's cache-affinity fields instead of starting unrelated CodeWhisperer conversations.

- keep `conversationState.conversationId` stable when the client sends an explicit session id (`x-session-id`, `session_id`, `conversation_id`, Claude Code session metadata, etc.)
- add a stable `conversationState.agentContinuationId` per Kiro session
- send `conversationState.agentTaskType: "vibe"` and `agentMode: "vibe"`, matching the normal Kiro CLI/KAS chat path
- move Kiro thinking instructions into Kiro-compatible `systemPrompt` / `additionalModelRequestFields` instead of generic top-level `thinking`
- keep volatile timestamp context out of the top-level `systemPrompt`; it remains only in user content fallback
- suppress `additionalModelRequestFields` for legacy 4.5-era Claude/Kiro models that reject it, while defaulting future Claude/Kiro model ids to supported
- preserve Kiro `meteringEvent` credit usage internally for accounting without leaking provider-specific fields into OpenAI-compatible client usage
- prevent unrelated headerless Kiro requests from sharing one connection-wide continuation
- cap/evict continuation sessions so long-running processes do not grow the continuation map unbounded
- treat generated headerless Kiro sessions as one-shot so they do not evict real explicit-session continuations
- keep credit-only Kiro metering valid for internal persistence when token metrics are unavailable

Hermes integration note: this PR no longer parses Hermes prompt payloads. Hermes should provide an explicit session id header (for example via its existing `llm_request` plugin that sets `x-session-id`). That keeps 9Router's behavior deterministic and avoids guessing from untrusted chat metadata.

## Evidence

Static Kiro CLI/KAS comparison:

- inspected local Kiro KAS bundle: `@kiro/agent/dist/server/acp-server.js`
- `GenerateAssistantResponseCommand` request includes `conversationState`, `profileArn`, `additionalModelRequestFields`, `agentMode`, and `systemPrompt`
- CLI/KAS `conversationState` includes `conversationId`, `agentContinuationId`, `agentTaskType`, `currentMessage`, `history`, and `chatTriggerType: "MANUAL"`
- CLI creates and reuses a per-session continuation id; normal chat task type is `vibe`
- Kiro credit billing is reported through `meteringEvent.usage`

Live direct Kiro credit A/B:

- without stable `agentContinuationId` / `agentTaskType`:
  - turn 1: `0.0052469782421227205` credits
  - turn 2: `0.00525832152570481` credits
  - total: `0.01050529976782753` credits
- with stable `agentContinuationId` + `agentTaskType: "vibe"`:
  - turn 1: `0.0052469782421227205` credits
  - turn 2: `0.0030093663018242118` credits
  - total: `0.008256344543946932` credits
- observed improvement:
  - second turn: about 43% fewer credits
  - two-turn total: about 21.4% fewer credits

Hermes chain smoke:

- AWS Hermes was tested with an external plugin that sends `x-session-id` to 9Router for Kiro requests
- two replies in the same Slack/Hermes thread used the same Kiro `conversationId` and `agentContinuationId`
- this proves Hermes → 9Router reaches the same Kiro continuation; exact Hermes credit delta was not claimed because the live quota window had concurrent traffic

Compatibility / safety checks:

- sending `additionalModelRequestFields` to a legacy 4.5-era Kiro model produced a real Kiro 400 (`additionalModelRequestFields is not supported for this model`), so this PR suppresses that field for legacy 4.5-era model ids
- pure top-level `systemPrompt` was not consistently enforced on the direct CodeWhisperer surface, so this PR keeps both top-level `systemPrompt` and the user-content fallback
- formal debug logging used during AWS validation was removed before finalizing this PR
- unrelated headroom/RTK changes are not part of this PR diff
- Codex review findings fixed: raw `metadata.user_id` is not used as a Kiro session key, credit-only Kiro metering is preserved as valid usage, and one-shot headerless sessions do not pollute the continuation LRU

## Tests

```bash
npm --prefix tests exec vitest -- \
  unit/session-manager.test.js \
  translator/claude-kiro-direct.test.js \
  unit/kiro-thinking-strip.test.js \
  unit/openai-to-kiro.test.js \
  unit/cached-token-usage.test.js
```

Result: 5 files passed, 90 tests passed.


decolua:

PR #2583, opened 2026-07-13
fix(v1/messages): reverse-translate OpenAI→Claude for all non-Claude providers + SSE-to-JSON

mahdiwafy:
## Summary

Same fix as https://github.com/Vanszs/VansRouter/pull/49 but adapted for 9Router v0.5.30 codebase (uses existing `openAICompletionToClaudeMessage` function).

## Changes

**nonStreamingHandler.js** (4 lines net):
- Widen the reverse-translation gate: instead of `targetFormat === FORMATS.OPENAI && sourceFormat === FORMATS.CLAUDE`, use `sourceFormat === FORMATS.CLAUDE && responseBody?.choices`. This catches combo routing where the selected provider is Gemini/Antigravity (not just OpenAI) — `translateNonStreamingResponse` always converts to OpenAI shape first, so `choices[]` is the signal.

**sseToJsonHandler.js** (31 lines):
- Import `openAICompletionToClaudeMessage` from nonStreamingHandler
- Claude format reverse translation in:
  - **Responses API path** (codex provider): builds an intermediate `chat.completion` shape, converts to Claude message format
  - **Standard Chat Completions path** (force-stream providers): SSE aggregated to OpenAI JSON → Claude format

## Motivation

Claude Code sends `/v1/messages` with Claude-format requests. When routed through a combo that selects a non-OpenAI provider (Gemini, Antigravity, etc.), the upstream response is translated to OpenAI `chat.completion` shape, but was never converted back to Claude format. The Claude Code client receives a malformed response.

Also, when a provider forces streaming (SSE) despite `stream: false`, the aggregated OpenAI JSON was returned directly even for Claude-format requests — this path now also converts back to Claude format.

## Existing foundation

9Router v0.5.30 already has `openAICompletionToClaudeMessage` in `nonStreamingHandler.js` and uses it when `targetFormat === FORMATS.OPENAI && sourceFormat === FORMATS.CLAUDE`. This PR:
1. Fixes the gate to not require FORMATS.OPENAI specifically
2. Adds the same reverse translation to `sseToJsonHandler.js` for streaming providers

### DIFF
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -65,7 +65,11 @@ function openAICompletionToClaudeMessage(responseBody) {
  */
 export function translateNonStreamingResponse(responseBody, targetFormat, sourceFormat) {
   if (targetFormat === sourceFormat) return responseBody;
-  if (targetFormat === FORMATS.OPENAI && sourceFormat === FORMATS.CLAUDE) {
+  // After upstream translation, if client sent Claude format and the response
+  // is in OpenAI shape (choices[] array), convert back to Claude message format.
+  // This must catch all non-Claude target formats (combo routing picks Gemini,
+  // Antigravity, etc.) — not just FORMATS.OPENAI.
+  if (sourceFormat === FORMATS.CLAUDE && responseBody?.choices) {
     return openAICompletionToClaudeMessage(responseBody);
   }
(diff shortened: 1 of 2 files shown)


mahdiwafy:
Fork repo salah — head nunjuk ke mahdiwafy/VansRouter, harusnya mahdiwafy/9router. Branch udah diperbaiki, buka PR baru.

PR #2587, opened 2026-07-13
fix(providers): bulk-add API keys no longer overwrite existing keys

asynx6:
Bulk-add named auto-generated keys by paste-line index, blind to existing connection names. The backend upserts apikey connections by exact name (connectionsRepo), so a colliding generated name silently replaced an existing key instead of inserting a new one.

Add a collision-aware planner (src/shared/utils/bulkAdd.js) that gap-fills the smallest free "<base> <n>" against both existing connection names and names assigned earlier in the same batch, so a generated name is never reused and the backend always inserts. Applies to auto-named lines, custom name|apiKey lines, and Cloudflare name|apiKey|accountId lines.

Wire the planner into AddApiKeyModal and pass existing connection names from the provider detail page.

Add unit tests covering gap-fill, custom names, Cloudflare 3-part format, and robustness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+# v0.5.31 (2026-07-13)
+
+## Fixes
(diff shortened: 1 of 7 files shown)


decolua:
Thanks @asynx6 for the contribution! Reviewed and merged into master. 🙏

PR #2588, opened 2026-07-13
feat(reliability): add provider skip-error rules and retry policy

nguyenha935:
## Summary

- Supersedes #2370 while preserving its Antigravity capacity fallback, stream usage, and Recent Requests fixes.
- Adds request-scoped transport retry policy and configurable provider skip-error rules.
- Adds a compact Settings UI, strict API validation, and an idempotent migration for the legacy Antigravity capacity behavior.

## Behavior

- `maxTransportAttempts` caps in-place transport attempts per request without mutating cached executor config.
- `providerSkipRules` supports `kind`, `status`, and `contains` conditions with AND semantics and first-match ordering.
- Rules can `retry` the current transport or `skip` to the next account/model; `headerTimeoutMs` customizes connect/header timeout and `sweep` opts a skip rule into a full account-pool resweep.
- Connect timeout classification uses an attempt-local closure flag, so caller aborts, network failures, and real HTTP errors remain distinct.
- Combo fail-fast metadata is kept in an in-process `WeakMap`; it is never serialized into response headers or bodies.
- Transport retry policy does not change Codex SSE-overload retry, namespace compatibility retry, token refresh, or other higher-level retry loops.

## Migration and UI

- Schema migration v2 seeds the former Antigravity `503 + capacity` behavior as an editable `skip + sweep` rule when it is not already covered, while respecting existing user rules.
- Settings validation limits attempts to `1..5`, rules to 100, and validates match/action/timeout/sweep combinations before writing.
- The responsive Skip-error rules modal uses compact, clearly separated fields in light and dark themes and merges static providers with dynamic compatible provider nodes.

## Test plan

- 41 targeted skip-rule, fail-fast, routing metadata, no-auth, concurrency, modal, and Antigravity capacity tests passed with `--unhandled-rejections=warn`.
- 32 inherited Antigravity Recent Requests, retry-hook, request-detail usage, migration-chain, and seed tests passed.
- `git diff --check` passed.
- `npm run build` passed.

## Notes

- No production patch, database, screenshot helper, or generated snapshot is included in this PR.

### DIFF
--- a/open-sse/executors/antigravity.js
+++ b/open-sse/executors/antigravity.js
@@ -397,6 +397,11 @@ export class AntigravityExecutor extends BaseExecutor {
 
     const errorMessage = this.extractErrorMessage(errorJson, bodyText);
 
+    // NOTE: capacity 503 fail-fast is no longer hardcoded here. It is handled by
+    // the seeded skip-rule ({ provider:"antigravity", status:503, contains:"capacity",
+    // action:"skip" }): a skip-rule resolves attempts=0, so BaseExecutor.tryRetry
+    // short-circuits before ever calling this hook. If the user deletes that rule,
+    // capacity 503 falls through to the transient-retry path below — their choice.
     if (!retryMs) {
       retryMs = this.parseRetryFromErrorMessage(errorMessage);
     }
--- a/open-sse/executors/base.js
+++ b/open-sse/executors/base.js
@@ -1,5 +1,6 @@
 import { HTTP_STATUS, RETRY_CONFIG, DEFAULT_RETRY_CONFIG, resolveRetryEntry, FETCH_CONNECT_TIMEOUT_MS } from "../config/runtimeConfig.js";
 import { shouldRefreshCredentials } from "../services/oauthCredentialManager.js";
+import { matchSkipRule } from "../services/accountFallback.js";
 import { proxyAwareFetch } from "../utils/proxyFetch.js";
 import { dbg } from "../utils/debugLog.js";
 import { ANTHROPIC_API_VERSION, OPENAI_COMPAT_BASE, ANTHROPIC_COMPAT_BASE } from "../providers/shared.js";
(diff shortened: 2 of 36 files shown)


nguyenha935:
Pushed `c4c5f26d`, which corrects the semantics of `action: "retry"` in this PR.

**What was wrong.** `BaseExecutor.resolveAttempts()` served a matched retry rule by returning `maxTransportAttempts - 1` in-place transport attempts:

```js
if (rule?.action === "retry") {
  return { attempts: cap != null ? cap : base.attempts, delayMs: base.delayMs };
}
```

That produced two behaviours the modal's own description ("gọi lại account") does not describe. The retry count was really a global transport setting the user never associated with the rule — at the shipped `maxTransportAttempts = 2`, every retry rule retried exactly once with no way to ask for more. And once those attempts were spent the failure fell through to the ordinary account-fallback path: `markAccountUnavailable()` has an early return for skip rules but had none for retry, so `checkFallbackError()` ran and `updateProviderConnection()` wrote `testStatus: "unavailable"`, a model lock, `errorCode`, `lastError` and a bumped `backoffLevel`. The account was cooled down for a failure the user had explicitly marked as worth retrying.

**What changed.** The two budgets are now separate. Transport retry keeps `maxTransportAttempts` and the provider retry registry for network failures and base-URL fallback; a matched retry rule resolves to `0` transport attempts and, like a skip rule, stops cycling the remaining base URLs, because the account layer is about to re-call the same account.

Account retry moved to the account-selection loop in `src/sse/handlers/chat.js`, the only layer holding the current `connectionId` and able to both re-select it and move past it. `markAccountUnavailable()` returns `{ accountRetry: true, retryAttempts, ruleKey }` before any DB read or write, and the loop keeps a request-scoped `Map` of consumed attempts keyed by `(connectionId, rule)` — never persisted, since a durable counter would leak between unrelated requests. The account is pinned through `getProviderCredentials(..., { preferredConnectionId })` while budget remains. When it runs out the loop advances to the next account with no DB write at all.

**Schema.** Rules gain an optional `retryAttempts` (integer 1–10) meaning extra calls to the same account after the first. Rejected on skip rules by the settings API and dropped from the row when the modal switches Retry → Skip. Rules saved before the field existed default to 1, owned by `resolveRetryAttempts()` in `accountFallback.js` and deliberately not derived from `maxTransportAttempts` — that coupling is the thing being removed. An out-of-range or non-integer stored value falls back to the default rather than silently disabling a rule the user created.

Client aborts end the request instead of consuming budget. A no-auth provider retries its single virtual connection the configured number of times and then returns a terminal response for combo to route on, rather than re-selecting the same virtual account forever (its `connectionId` is `undefined`, so excluding it would filter nothing). Errors matching no rule keep their existing cooldown behaviour.

**Tests.** The test that locked the old semantics (`action:retry ... retries maxTransportAttempts-1 times`) is replaced; two `connect_timeout` tests asserting the same transport behaviour were updated, and the concurrency test that used a retry rule to prove per-request policy isolation now proves it without depending on retry counts. `tests/unit/account-retry-semantics.test.js` drives the real `chat.js` loop and the real `auth.js` against mocked DB deps: exact call sequence for `retryAttempts: 3`, not advancing early, the legacy default, success on a later retry, absence of any `updateProviderConnection()` write, independence from `maxTransportAttempts` 1 and 5, unchanged skip and unmatched-error behaviour, abort, and the no-auth path. Plus API validation and modal round-trip coverage.

Run against this branch: the 8 skip/retry/fallback suites pass 91/91, full suite unchanged at 46 failing before and after with 36 newly passing and zero regressions, and `npx next build` exits 0.

Issue #2546, opened 2026-07-11
GrokCLI Provider does not use refresh token in Auth.json and token expire after around 40-45 mins


Knguyen-data:
## Description

The GrokCLI provider does not appear to use the `refresh_token` stored in `auth.json` when the access token expires.

After approximately **40–45 minutes**, the access token expires and all subsequent requests return **401 Unauthorized**. Instead of automatically refreshing the access token using the stored `refresh_token`, the provider requires the user to manually log in again.

## Steps to Reproduce

1. Authenticate with the GrokCLI provider.
2. Confirm that `auth.json` contains both `access_token` and `refresh_token`.
3. Wait for approximately 40–45 minutes until the access token expires.
4. Make another API request.

## Expected Behavior

- The provider should detect that the access token has expired.
- It should automatically use the stored `refresh_token` to obtain a new access token.
- The request should succeed without requiring user intervention.

## Actual Behavior

- The request fails with **401 Unauthorized** after the access token expires.
- No automatic token refresh is attempted.
- The user must manually log in again to continue using the provider.

## Impact

- Long-running sessions fail unexpectedly.
- Automation workflows are interrupted.
- Users are forced to repeatedly re-authenticate.

## Environment

- Provider: GrokCLI
- Token expiry: ~40–45 minutes
- Error: `401 Unauthorized`

fdjrr:
## Update — Root cause found and fixed

The Grok CLI (xAI device-code OAuth) token refresh issue has two layers:

### Layer 1: Refresh token not readable from PSD (fixed in 608ff20)

Commit 608ff20 added:
- `_resolveRefreshToken()` — reads refresh token from both `credentials.refreshToken` and `credentials.providerSpecificData.refreshToken`
- `mapTokens` mirrors refresh token into PSD during OAuth flow
- `updateProviderCredentials` mirrors refresh token into PSD on persist

### Layer 2: Proactive refresh path bypassed generic infrastructure (fixed in 895c567)

The refactoring in 608ff20 routed `checkAndRefreshToken` through `executor.refreshCredentials()`, which introduced **3 regressions**:

| Provider | Problem |
|----------|---------|
| iflow, codebuddy-cn | Proactive refresh silently skipped — `DefaultExecutor.refreshers` had no entries |
| claude, gemini, cline, clinepass, kimi-coding | Lost `withCredentialRefreshLock` (concurrent refresh dedup) and `mergeRefreshedCredentials` (field preservation) |
[...]

PR #2604, opened 2026-07-14
fix(grok-cli): surface expiresAt so proactive token refresh fires (#2546)

dsitmilis:
## Bug (#2546)
Grok CLI (xAI) sessions die 40-45 minutes after login with token-expired errors.

## Root cause
`mapTokens` for grok-cli stored `expiresIn` but never `expiresAt`. `shouldRefreshCredentials()` only reads `expiresAt`/`tokenExpiresAt`, so the proactive refresh path (`checkAndRefreshToken`) never fired. Only the reactive 401 handler in chatCore could refresh, causing intermittent failures once the ~40-45 min xAI token lapsed.

## Fix
Compute `expiresAt` from `expiresIn` in grok-cli `mapTokens` so proactive refresh triggers before expiry. Verified with a regression test asserting `shouldRefreshCredentials('grok-cli', ...)` returns true for a near-expiry token and false for a far-future one.

## Test plan
- `npx vitest run tests/unit/grok-cli-expiresat-2546.test.js` — 2 passing

### DIFF
--- a/src/lib/oauth/providers.js
+++ b/src/lib/oauth/providers.js
@@ -350,10 +350,20 @@ const PROVIDERS = {
         .join(" ")
         .trim() || null;
 
+      const expiresAt = tokens.expires_in
+        ? new Date(Date.now() + tokens.expires_in * 1000).toISOString()
+        : null;
+
       return {
         accessToken: tokens.access_token,
         refreshToken: tokens.refresh_token || null,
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @dsitmilis for the contribution! Reviewed and merged into master. 🙏

PR #2608, opened 2026-07-14
feat(github): route Claude models through Copilot's native /v1/messages

yidecode:
## Summary
- GitHub Copilot's `/chat/completions` and `/responses` endpoints never surface prompt-cache token counts for Claude models. Copilot also exposes an Anthropic-native `/v1/messages` shim that does report `cached_tokens`.
- Claude models (detected by name pattern — Copilot's live model catalog regularly exposes `claude-*` variants ahead of the static registry, so this is intentionally not a registry `targetFormat` field) are now routed through a new `executeWithMessagesEndpoint()`, translating OpenAI-shape requests to Claude natively (reusing `translateRequest`/`translateResponse`) instead of going through `/chat/completions`.
- Fixes a real bug found along the way: `translateRequest()`'s internal bookkeeping field `_toolNameMap` was being sent upstream as-is, which made Anthropic's strict schema reject tool-call requests with a 400. It's now stripped and threaded through response state instead, same as the rest of the codebase does.
- Removes the now-dead `response_format` Claude JSON-mode workaround in `sanitizeMessagesForChatCompletions()`, since Claude models no longer take that code path.

## Test plan
- [x] `node --check open-sse/executors/github.js` / `open-sse/providers/registry/github.js`
- [x] `npx eslint` on both files (only a pre-existing, unrelated warning)
- [x] `npx vitest run unit/github-responses-routing.test.js` — 5/5 pass
- [x] Manually verified against a real Copilot account: streaming and non-streaming Claude requests both correctly hit `cached_tokens` on the second turn; `reasoning_effort` maps to Claude `thinking.budget_tokens`; streaming `reasoning_content` carries real thinking text; full tool_use → tool_result → final answer round-trip works
- [x] Audited GPT-5.x models — no changes needed, cache/reasoning already worked via the existing endpoints

### DIFF
--- a/open-sse/executors/github.js
+++ b/open-sse/executors/github.js
@@ -4,17 +4,29 @@ import { OAUTH_ENDPOINTS, GITHUB_COPILOT } from "../config/appConstants.js";
 import { HTTP_STATUS } from "../config/runtimeConfig.js";
 import { openaiToOpenAIResponsesRequest } from "../translator/request/openai-responses.js";
 import { openaiResponsesToOpenAIResponse } from "../translator/response/openai-responses.js";
-import { initState } from "../translator/index.js";
+import { initState, translateRequest, translateResponse } from "../translator/index.js";
+import { FORMATS } from "../translator/formats.js";
 import { parseSSELine, formatSSE } from "../utils/streamHelpers.js";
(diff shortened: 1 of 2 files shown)


decolua:

PR #2609, opened 2026-07-14
feat(rtk): add X-9Router-Token-Saver header to bypass token savers per request

joachimBrindeau:
## Summary

Adds a per-request opt-out header so a single chat request can bypass **all** token savers without changing global settings.

Send `X-9Router-Token-Saver: off` and RTK, Headroom, Caveman, Ponytail and PXPIPE are all skipped for that request only. Any other value (or no header) keeps current behavior.

## Motivation

Token savers are global on/off settings. Caveman/Ponytail inject terse-style system prompts into **every** request, which degrades output quality for non-coding work such as prose/writing. Today the only workaround is toggling the global setting back and forth (and it needs dashboard auth). A per-request header lets a client mix coding requests (savers on) and writing requests (savers off) against the same endpoint.

## Changes

- `open-sse/handlers/chatCore.js`: compute `tokenSaverEnabled` from the client header and gate RTK, Headroom, Caveman, Ponytail and PXPIPE on it. Fail-open behavior of each saver is unchanged.
- `open-sse/config/runtimeConfig.js`: add `TOKEN_SAVER_HEADER` constant (no hardcoded header string in the handler, per repo conventions).
- `tests/unit/headroom-chat-core.test.js`: add a test asserting all savers are skipped when the header is `off` (no Headroom fetch, no PXPIPE transform, original messages passed through).
- `README.md`: document the header.

## Behavior

- Header absent / any value ≠ `off` → unchanged (savers run per global settings).
- `X-9Router-Token-Saver: off` → all savers skipped for that request.
- Case-insensitive value; header lookup uses the lowercased key already present on `clientRawRequest.headers`.

## Testing

- `npx vitest run unit/headroom-chat-core.test.js` — new bypass test passes; existing Headroom tests unaffected by this change.
- Manual end-to-end against a local 9Router (Caveman enabled globally): a normal request injected the caveman system prompt into the outbound target body; the same request with `X-9Router-Token-Saver: off` contained zero caveman injections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,6 +448,8 @@ Default URLs:
 | 📊 **Usage Analytics**                                                            | Track tokens, cost, trends over time                                                     | Optimize spending                                 |
 | 🌐 **Deploy Anywhere**                                                            | Localhost, VPS, Docker, Cloudflare Workers                                               | Flexible deployment options                       |
 
+Set `X-9Router-Token-Saver: off` to bypass all token savers for one chat request.
+
 <details>
 <summary><b>📖 Feature Details</b></summary>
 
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @joachimBrindeau for the contribution! Reviewed and merged into master. 🙏

PR #2612, opened 2026-07-14
feat: drop temperature for all Claude models

mhshahmoradi:
Dropped `temperature` from all Claude requests to fix Anthropic 400 errors:

- paramSupport.js — Broadened the strip rule from `/claude-opus-4/i` to `/claude/i`, so `temperature` is removed for every Claude model (not just opus-4). Applies on all OpenAI-compatible routes via default.js and the GitHub executor.


decolua:
Thanks @mhshahmoradi for the contribution! Reviewed and merged into master. 🙏

PR #2627, opened 2026-07-15
fix: list compatible provider models in /v1/models

### LINKED ISSUE #2626: OpenAI-compatible provider models missing from /v1/models due to overly broad UPSTREAM_CONNECTION_RE regex
## Bug Description

Models from OpenAI-compatible providers (e.g. local MLX server, Ollama, vLLM) do not appear in the `/v1/models` endpoint, even though the provider node is successfully created and validated.

## Root Cause

Commit `838d9a7` (#541) introduced `UPSTREAM_CONNECTION_RE = /[-_][0-9a-f]{8,}$/i` to skip dynamic model fetching for upstream/cross-instance connections, preventing recursive `/models` fetches between 9router instances.

However, **all** OpenAI-compatible provider node IDs contain a UUID v4 suffix (format: `openai-compatible-chat-{uuidv4}`), e.g. `openai-compatible-chat-a7135bd4-4bbc-4951-9b85-39e838ae9773`. The last UUID segment (`39e838ae9773`, 12 hex chars) matches the regex `[-_][0-9a-f]{8,}$`, so `fetchCompatibleModelIds` is **never** called for any compatible provider.

This means models are only discoverable via manual "Import from /models" — they never appear automatically in `/v1/models`.

## Reproduction

1. Start a local OpenAI-compatible server (e.g. MLX server on `http://localhost:8080/v1`)
2. Add it as an "OpenAI Compatible" provider in the dashboard
3. Add an API key connection with a default model
4. Call `GET /v1/models` — the provider's models are **not listed**

## Impact

- Local LLM servers (MLX, Ollama, vLLM, LM Studio, etc.) are effectively invisible in the model list
- Users must manually import models, which is unintuitive
- Tools that auto-discover models via `/v1/models` cannot see compatible provider models

## Environment

- 9router v0.5.30
- Next.js 16.2.10

evlon:
## Summary

Fixes #2626

Replaces the overly broad `UPSTREAM_CONNECTION_RE` regex with a custom HTTP header to detect and break cross-instance recursive `/models` fetches.

## Problem

The regex `/[-_][0-9a-f]{8,}$/i` was intended to match upstream/cross-instance connection IDs, but it also matches **all** compatible provider IDs (which contain UUID v4 suffixes from `generateId()`). This prevented `fetchCompatibleModelIds` from ever being called, making compatible provider models invisible in `/v1/models`.

## Solution

Instead of pattern-matching provider IDs, use an HTTP header to detect recursive cross-instance fetches:

1. `fetchCompatibleModelIds()` sends `x-9r-internal-models-fetch: 1` header when fetching `/models` from a compatible provider
2. The `/v1/models` GET handler checks for this header — if present, it skips dynamic model fetching, breaking the recursion loop
3. The existing 5-second `AbortController` timeout remains as a safety net

This allows regular compatible providers (MLX, Ollama, vLLM, etc.) to have their models listed automatically, while still preventing infinite loops between cross-instance connections.

## Changes

- **Removed** `UPSTREAM_CONNECTION_RE` regex constant
- **Added** `INTERNAL_MODELS_FETCH_HEADER` constant
- **Modified** `fetchCompatibleModelIds()` to send the header with its fetch request
- **Modified** `buildModelsList()` to accept `skipDynamicFetch` option (replaces regex check)
- **Modified** `GET()` handler to detect the header and pass `skipDynamicFetch` to `buildModelsList()`

## Testing

- Verified MLX server models now appear in `/v1/models` after adding an OpenAI-compatible provider
- Verified chat completions work with the dynamically-listed model
- Verified cross-instance loop prevention: a 9router instance fetching `/v1/models` from another 9router instance sends the header, which causes the receiving instance to skip dynamic fetch (breaking the loop)

### DIFF
--- a/src/app/api/v1/models/route.js
+++ b/src/app/api/v1/models/route.js
@@ -79,8 +79,9 @@ const parseOpenAIStyleModels = (data) => {
   return data?.data || data?.models || data?.results || [];
 };
 
-// Matches provider IDs that are upstream/cross-instance connections (contain a UUID suffix)
-const UPSTREAM_CONNECTION_RE = /[-_][0-9a-f]{8,}$/i;
+// Header sent by fetchCompatibleModelIds to detect cross-instance /models fetches
+// and break recursive loops between 9router instances connected to each other.
+const INTERNAL_MODELS_FETCH_HEADER = "x-9r-internal-models-fetch";
 
 // LLM kind sentinel — combos/models with no explicit kind default to LLM
 const LLM_KIND = "llm";
@@ -145,7 +146,7 @@ async function fetchCompatibleModelIds(connection) {
     const timeoutId = setTimeout(() => controller.abort(), 5000);
     const response = await fetch(url, {
       method: "GET",
-      headers,
+      headers: { ...headers, [INTERNAL_MODELS_FETCH_HEADER]: "1" },
       cache: "no-store",
       signal: controller.signal,
     });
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @evlon for the contribution! Reviewed and merged into master. 🙏

PR #2641, opened 2026-07-16
perf(startup): skip inactive background services

hungtrinhh:
## Summary

This PR reduces unnecessary startup and background activity by only initializing optional services when their related features are enabled.

Previously, several background services were started even when unused. This included tunnel network monitoring, periodic watchdog checks, Cloudflared installation checks, MITM cache synchronization, and the quota auto-ping scheduler.

## Changes

- Only check or install `cloudflared` when Cloudflare Tunnel is enabled.
- Only run the tunnel watchdog and network monitor when Cloudflare Tunnel or Tailscale is enabled.
- Stop tunnel monitoring when the last active tunnel service is disabled.
- Start tunnel monitoring immediately when Tunnel or Tailscale is enabled at runtime.
- Only synchronize the MITM alias cache and attempt MITM auto-start when MITM is enabled.
- Lazy-load the quota auto-ping service when at least one account has opted in.
- Stop the quota auto-ping scheduler when the last account opts out.
- Avoid loading the quota provider/executor dependency graph from the settings route unless needed.
- Add tests for quota scheduler startup and shutdown behavior.

## Motivation

When optional services were disabled, the application could still:

- Open a TCP connection to `1.1.1.1:443` every few seconds.
- Run tunnel and Tailscale watchdog checks every minute.
- Check or download the Cloudflared binary.
- Read and write the MITM alias cache.
- Keep the quota auto-ping scheduler active.
- Eagerly load provider and executor modules through the quota scheduler.

These operations added unnecessary network, disk, timer, memory, and event-loop activity. They could also introduce latency jitter while chat streams were active.

## Behavior After This Change

When Tunnel, Tailscale, MITM, and quota auto-ping are disabled:

- No tunnel network reachability probes are scheduled.
- No tunnel watchdog timer is active.
- Cloudflared is not checked or downloaded.
- The MITM alias cache is not synchronized at startup.
- The quota auto-ping scheduler is not loaded or started.

Runtime enable and disable operations continue to work without requiring a server restart.

## Testing

- `npx vitest run unit/quota-auto-ping.test.js`
  - 15 tests passed.
- ESLint passed for all changed files.
- `git diff --check` passed.

A production build was also attempted, but the local Windows environment prevented Next.js/Webpack from scanning legacy user-directory junctions:

```text
EPERM: operation not permitted, scandir 'C:\Users\hungt\Application Data'
This failure occurs outside the repository and is unrelated to the changed source files.

### DIFF
--- a/src/app/api/settings/route.js
+++ b/src/app/api/settings/route.js
@@ -2,7 +2,6 @@ import { NextResponse } from "next/server";
 import { getSettings, updateSettings } from "@/lib/localDb";
 import { applyOutboundProxyEnv } from "@/lib/network/outboundProxy";
 import { resetComboRotation } from "open-sse/services/combo.js";
-import { runQuotaAutoPingTick } from "@/shared/services/quotaAutoPing";
 import bcrypt from "bcryptjs";
 
 export const dynamic = "force-dynamic";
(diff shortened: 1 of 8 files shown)


decolua:
Thanks @hungtrinhh for the contribution! Reviewed and merged into master. 🙏

Issue #2582, opened 2026-07-13
Support Anthropic Advisor server-tool passthrough


tmaarcxs:
## Summary

Please add transparent passthrough support for Anthropic's experimental Advisor server tool on 9router's Anthropic-compatible Messages endpoint.

Claude Code documents gateway support as dependent on the gateway forwarding the request intact to Anthropic. With 9router, Claude Code can persist `advisorModel: "fable"` and recognize an official main-model alias, but an attempted consultation fails with:

```text
No such tool available: advisor
```

## Environment

- Claude Code: `2.1.207`
- 9router: current Docker `latest` deployment
- Client API: Anthropic Messages-compatible endpoint via `ANTHROPIC_BASE_URL`
- Main model presented to Claude Code: `claude-sonnet-5`
- Advisor model: `claude-fable-5`

No credentials or private endpoint details are included here.

## Reproduction

1. Point Claude Code's `ANTHROPIC_BASE_URL` at 9router's Anthropic-compatible endpoint.
2. Configure official model aliases that route to the corresponding real Anthropic models:

   ```text
   claude-sonnet-5
   claude-fable-5
   ```

3. In Claude Code, run:

   ```text
   /advisor fable
   ```

4. Confirm Claude Code reports:

   ```text
   Advisor set to Fable 5
   ```

5. Start a session whose main model is reported as Sonnet 5 and ask it to consult the advisor.

## Actual behavior

```text
No such tool available: advisor
```

No successful `Advising` event or advisor result is produced.

## Expected behavior

When an official Anthropic model route is used, 9router should preserve and forward the Advisor request to Anthropic, then preserve the server-tool response blocks back to Claude Code.

Relevant protocol elements include:

- Beta header: `anthropic-beta: advisor-tool-2026-03-01`
- Tool definition: `type: "advisor_20260301"`
- Server-generated `server_tool_use`
- Advisor result/error content blocks
- Streaming behavior while the server-side advisor runs
- Advisor usage data, including advisor iterations/tokens
- Full multi-turn content-block fidelity for subsequent requests

Unsupported upstream providers should return a clear capability error rather than silently exposing an unusable Advisor configuration.

## Suggested scope

- Preserve the Advisor beta header on Anthropic routes.
- Pass `advisor_20260301` tool definitions through without translating them into client-executed tools.
- Preserve all Advisor server-tool result and usage blocks in streaming and non-streaming responses.
- Advertise or document whether a configured route supports Anthropic server-side tools.
[...]

PR #2649, opened 2026-07-16
fix: normalize nested Claude server tool models

### LINKED ISSUE #2642: ANTHROPIC_DEFAULT_*_MODEL with cc/ prefix causes 400 errors on Task tool / subagent dispatch (tools.N.model invalid_request_error)
**Summary**
Setting the cc/-prefixed model aliases via ANTHROPIC_DEFAULT_OPUS_MODEL, ANTHROPIC_DEFAULT_SONNET_MODEL, ANTHROPIC_DEFAULT_HAIKU_MODEL, ANTHROPIC_DEFAULT_FABLE_MODEL (as generated by the "Claude Code" CLI Tools dashboard page) causes Claude Code to embed those raw cc/... strings into internal tool-definition schemas (used for the Task tool / subagent dispatch). These schemas get forwarded to Anthropic's real API, which rejects them with a 400 since cc/claude-opus-4-8 isn't a valid Anthropic model identifier in that context.
This looks related to #1224, which shows the same cc/-prefixed pattern in ANTHROPIC_DEFAULT_*_MODEL, but that issue is about the config UI hiding model fields on remote/undetected installs — this is a distinct downstream failure once that config shape is actually applied.
Config that triggers it
Generated automatically via Dashboard → CLI Tools → Claude Code → Apply:

Image
Image

```
json{
  "hasCompletedOnboarding": true,
  "env": {
    "ANTHROPIC_BASE_URL": "http://127.0.0.1:20128/v1",
    "ANTHROPIC_AUTH_TOKEN": "sk-xxxx",
    "ANTHROPIC_DEFAULT_FABLE_MODEL": "cc/claude-fable-5",
    "ANTHROPIC_DEFAULT_OPUS_MODEL": "cc/claude-opus-4-8",
    "ANTHROPIC_DEFAULT_SONNET_MODEL": "cc/claude-sonnet-5",
    "ANTHROPIC_DEFAULT_HAIKU_MODEL": "cc/claude-haiku-4-5-20251001"
  }
}
```

**Error**
API Error: 400 [claude/claude-opus-4-8] [400]: {"type":"error","error":{"type":"invalid_request_error","message":"tools.36.model: cc/claude-opus-4-8"},"request_id":"req_011Cd5AYVWc2d8azgyyqgY5Y"} (reset after 30s)
Console log (screenshot attached) shows the request correctly reaching api.anthropic.com/v1/messages, then Anthropic rejecting it, then 9Router locking the account for 30s and failing over.
Working workaround
Replacing the four ANTHROPIC_DEFAULT_*_MODEL vars with a single ANTHROPIC_MODEL and switching manually via /model cc/... avoids the issue entirely, since it never populates the internal tool-schema model field:
```
json{
  "env": {
    "ANTHROPIC_BASE_URL": "http://localhost:20128/v1",
    "ANTHROPIC_AUTH_TOKEN": "sk-xxxx",
    "ANTHROPIC_MODEL": "cc/claude-opus-4-8"
  }
}
```

AlanSyue:
## Problem

Claude Code can expand ANTHROPIC_DEFAULT_*_MODEL values such as cc/claude-opus-4-8 into Advisor or Task server tools. 9router normalizes the top-level model for routing, but previously forwarded the nested tools[].model unchanged, causing Anthropic to reject the 9router-only prefix with HTTP 400.

## Fix

Normalize only the known cc/ and claude/ prefixes in tools[].model during native Claude passthrough outbound processing. The top-level model remains unchanged so 9router can continue selecting the Claude OAuth provider.

## Why proxy-layer normalization is safer

A settings-only change that writes a bare top-level claude-* model can make 9router infer the anthropic API-key provider instead of the Claude OAuth provider. Normalizing the actual outbound nested values avoids that routing regression and also covers existing settings plus payloads from other clients.

## Regression coverage

- Advisor advisor_20260301 nested model normalization
- Task/subagent nested model normalization
- Idempotency for already-bare models
- Preservation of missing, non-string, unrelated model values, and other tool fields

## Verification

- npx vitest run tests/unit/translator-helpers-edge.test.js — 10/10 passed
- npx eslint open-sse/translator/formats/claude.js tests/unit/translator-helpers-edge.test.js — passed
- git diff --check — passed

Fixes #2642
Related to #2582

This is a proxy-layer alternative to #2645.

### DIFF
--- a/open-sse/translator/formats/claude.js
+++ b/open-sse/translator/formats/claude.js
@@ -84,6 +84,21 @@ export function fixToolUseOrdering(messages) {
 // Models that reject thinking.type "adaptive" + output_config.effort (Opus 4.5+/Sonnet 4.6+ only)
 const ADAPTIVE_THINKING_UNSUPPORTED = /haiku/i;
 
+// 9router provider prefixes are valid for routing the top-level model, but
+// Anthropic server tools expect an upstream model ID in their nested `model`.
+const CLAUDE_PROVIDER_MODEL_PREFIXES = ["cc/", "claude/"];
+
+function normalizeClaudeServerToolModels(tools) {
+  if (!Array.isArray(tools)) return;
+
+  for (const tool of tools) {
+    if (!tool || typeof tool !== "object" || typeof tool.model !== "string") continue;
+
+    const prefix = CLAUDE_PROVIDER_MODEL_PREFIXES.find(candidate => tool.model.startsWith(candidate));
+    if (prefix) tool.model = tool.model.slice(prefix.length);
+  }
+}
+
 function handlesThinkingBlocks(provider) {
   return provider === "claude" || provider?.startsWith("anthropic-compatible") || provider === "deepseek";
 }
@@ -108,6 +123,7 @@ function buildThinkingPlaceholder(provider) {
 // 1. thinking.type "adaptive" → unsupported on Haiku
 // 2. output_config.effort → unsupported on Haiku
 // 3. role "system" messages (mid-conversation-system beta) → only top-level system is allowed
+// 4. server tool model IDs must not include 9router's Claude provider prefix
 export function normalizeClaudePassthrough(body, model = "") {
   if (!body || typeof body !== "object") return body;
 
@@ -122,7 +138,7 @@ export function normalizeClaudePassthrough(body, model = "") {
     if (Object.keys(body.output_config).length === 0) delete body.output_config;
   }
 
-  // 2. Hoist mid-conversation system messages into the top-level system field
+  // 3. Hoist mid-conversation system messages into the top-level system field
   if (Array.isArray(body.messages)) {
     const systemBlocks = [];
     const messages = [];
@@ -150,7 +166,10 @@ export function normalizeClaudePassthrough(body, model = "") {
     }
   }
 
-  // 3. Drop thinking blocks whose signature is not Claude's (combo mixes models,
+  // 4. Normalize nested server tool model IDs without changing the routing model
+  normalizeClaudeServerToolModels(body.tools);
+
+  // 5. Drop thinking blocks whose signature is not Claude's (combo mixes models,
   // so foreign signatures leak into history and Anthropic rejects them).
   const thinkingEnabled = body.thinking?.type === "enabled";
   if (Array.isArray(body.messages)) {
--- a/tests/unit/claude-server-tool-passthrough-integration.test.js
+++ b/tests/unit/claude-server-tool-passthrough-integration.test.js
@@ -1,0 +1,154 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+const { executeMock } = vi.hoisted(() => ({
+  executeMock: vi.fn(),
+}));
+
+vi.mock("../../open-sse/executors/index.js", () => ({
+  getExecutor: vi.fn(() => ({
+    execute: executeMock,
+    refreshCredentials: vi.fn().mockResolvedValue(null),
+  })),
+}));
+
+vi.mock("../../open-sse/utils/requestLogger.js", () => ({
+  createRequestLogger: vi.fn(async () => ({
+    logClientRawRequest: vi.fn(),
+    logRawRequest: vi.fn(),
+    logTargetRequest: vi.fn(),
+    logError: vi.fn(),
+  })),
+}));
+
+vi.mock("../../open-sse/utils/bypassHandler.js", () => ({
+  handleBypassRequest: vi.fn(() => null),
+}));
+
+vi.mock("../../open-sse/utils/streamHandler.js", () => ({
+  createStreamController: vi.fn(() => ({
+    signal: undefined,
+    handleComplete: vi.fn(),
+    handleError: vi.fn(),
+  })),
+}));
+
+vi.mock("../../open-sse/services/tokenRefresh.js", () => ({
+  refreshWithRetry: vi.fn(),
+}));
+
+vi.mock("../../open-sse/rtk/caveman.js", () => ({
+  injectCaveman: vi.fn(),
+}));
+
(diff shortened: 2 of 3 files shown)


AlanSyue:
## Verification results

Re-verified commit `3cf03072d96d923b197515b1a0438e052488a625` locally.

Environment:
- Node.js `v26.0.0`
- npm `11.12.1`
- Vitest explicitly pinned to `4.1.10` because this repository does not declare/pin Vitest in `package.json`

Results:
- `npx vitest@4.1.10 run tests/unit/translator-helpers-edge.test.js` — 1 file passed, 10/10 tests passed
- `npx vitest@4.1.10 run tests/unit/headroom-chat-core.test.js` — 1 file passed, 6/6 tests passed
- `npx eslint open-sse/translator/formats/claude.js tests/unit/translator-helpers-edge.test.js` — passed
- `git diff --check origin/master...HEAD` — passed

Version note:
- An explicit run with `vitest@4.0.0` failed before test collection with `Unknown method: getBuiltins. Expected "fetchModule"`; no assertions ran. `vitest@4.1.10` collected and passed all tests above. This appears to be runner/runtime compatibility rather than a code assertion failure.

AlanSyue:
Additional verification after adding the dispatch-level integration regression test:

- The new test enters `handleChatCore` through the native Claude passthrough path and captures the final body passed to the executor.
- On baseline `bc252ea`, the same test **FAILS exactly as expected**: the top-level model is correctly normalized to `claude-sonnet-5`, while the nested Advisor model is actually `cc/claude-opus-4-8` instead of the expected `claude-opus-4-8`.
- On this fixed branch, the same test **PASSES**.
- Combined targeted regression results: **11/11 passed**. ESLint and `git diff --check` also passed.

Live validation was also attempted with the exact Claude Code **2.1.211** release, an isolated HOME and DATA_DIR, `advisorModel: "opus"`, main model `cc/claude-sonnet-5`, and default Opus model `cc/claude-opus-4-8`. The real request used native Claude passthrough with 24 tools, and the 129,525-byte request body reached `api.anthropic.com`. However, the Anthropic account returned HTTP 429 before schema validation, with a total cost of $0. Per the retry limit, no live fixed-branch request was attempted.

Therefore, this is **not a successful live before/after E2E**. The deterministic before/after evidence comes from the dispatch-level integration regression test described above.

AlanSyue:
## Live E2E follow-up and limitation

I ran a follow-up live E2E check with exact Claude Code 2.1.211. The baseline and fix runs used isolated HOME and DATA_DIR environments with the same settings, environment variables, prompt, and CLI flags; only the local 9router ports differed.

After account availability and token refresh were restored safely through the existing 20128 instance, both revisions were actually exercised:

- Baseline `bc252ea` on port 20228: CLI exit 0, Anthropic HTTP 200, native Claude passthrough, 24 tools, and no `tools.N.model` error.
- Fix `3875c2a` on port 20229: CLI exit 0, Anthropic HTTP 200, native Claude passthrough, 24 tools, and no `tools.N.model` error.

Critical limitation: neither request contained or used an observable Advisor server tool/result. Claude Code reported that its current toolset had no advisor. A separate offline TTY capability probe with exact 2.1.211 returned `Unknown command: /advisor` and did not write `advisorModel` to the isolated settings. Therefore, these successful HTTP 200 runs are **not** evidence that the Advisor nested-model path passed live.

The deterministic regression evidence remains:

- The identical dispatch-level integration test fails on baseline `bc252ea`, where the actual nested model delivered to the executor remains `cc/claude-opus-4-8`.
- The same test passes on fix `3875c2a`, where the nested model delivered to the executor is `claude-opus-4-8`.
- Combined targeted tests pass 11/11; ESLint and diff-check also pass.

This follow-up supersedes the previous comment's statement that "no live fixed-branch request was attempted."

ryguritno:
Confirming this fix approach works in the wild.

We hit the exact bug described in #2642 with CC LTS 2.1.197 through 9router 0.5.35 on `cc/*` models — every `Task` / `Advisor` dispatch failed with `tools.<N>.model: cc/claude-opus-4-7` 400 from Anthropic, and the secondary effect locked all rotation accounts for 30s per malformed request.

Since #2645/#2649 hadn't landed yet, I mirrored this PR's approach at runtime via a Node fetch-interceptor injected through `NODE_OPTIONS=--require` on launchctl. The whole intercepted-and-mutated payload is exactly what this PR does at the serializer level:

```js
// runtime shim, ~50 LOC, wraps globalThis.fetch for api.anthropic.com only
for (const t of body.tools) {
  if (t && typeof t.model === 'string') {
    if (t.model.startsWith('cc/'))         t.model = t.model.slice(3);
    else if (t.model.startsWith('claude/')) t.model = t.model.slice(7);
  }
}
```

**Result:** 400 → 200 immediately, no other changes required. Verified with `cc/claude-opus-4-7`, `cc/claude-sonnet-4-6`, `cc/claude-haiku-4-5-20251001` — all pong clean via 9router.

**One thing worth noting from our debugging** (not blocking this PR, just diagnostic):

- We initially tried DELETING `tools[].model` entirely — Anthropic still returned 400. The field appears to be required in Task/Advisor tool schemas, only its **value** is wrong when it carries the 9router-internal prefix. Your PR's normalize-in-place approach is correct; a naive strip-the-field workaround does not work.
- Before finding this PR I spent time patching sanitizer destructures in `chunks/318.js` and `chunks/6805.js` (both `{cache_control:X,...rest}=a` patterns) — neither is on the request path that emits `tools[].model` for `cc/*` routes. The serializer layer this PR touches is the actual culprit. Would save future debuggers time if there's a way to add a comment/reference to the offending serialize call linking back to #2642.

`mergeable: true` / `mergeable_state: clean` on this PR — any blocker to merging? Happy to help test if there are edge cases you want validated first. This is affecting real Claude Code users with `ANTHROPIC_DEFAULT_*_MODEL=cc/...` set, and the account-lock cascade turns one malformed request into 30s of unavailability across the whole rotation.

PR #2667, opened 2026-07-17
fix(codex): recover invalid encrypted history

ryanngit:
## Summary

- retry the same Codex account once when upstream rejects stale `reasoning.encrypted_content`
- clone the request, remove ciphertext only from top-level reasoning items, and drop items left without summary/content
- preserve valid encrypted reasoning on ordinary requests
- classify repeated invalid-ciphertext failures as deterministic so outer fallback does not rotate accounts

## Why

Cross-model or cross-account history can retain encrypted reasoning that a later Codex account cannot decrypt. Upstream returns HTTP 400. Account rotation cannot repair that payload; one sanitized retry on the already selected account can.

## Safety

- gated to HTTP 400 plus exact `invalid_encrypted_content` code or observed verification/decryption messages
- reuses model, credentials, workspace binding, proxy, and abort signal
- runs at most once
- logs removed-item count only, never ciphertext
- leaves accepted ciphertext and unrelated errors unchanged

## Verification

Tested on pristine `v0.5.45` (`6fcd2733`):
- recovery suite: 8/8 passed
- adjacent Codex/base matrix: 35/35 passed
- Gitleaks found no leaks

This PR remains independent of #2511. If Responses Lite merges first, this narrow executor hunk can be rebased without changing recovery behavior.


ryanngit:
## v0.5.45 recut

- Base: `upstream/master` = `v0.5.45` (`6fcd2733`)
- New head: `eeac78993c15b36b4a4c79626082061ff1ce7cd3`
- Merge state: `CLEAN`
- Removed old v0.5.40 merge commit; branch now contains one feature commit
- Recovery suite: 8/8 passed
- Adjacent Codex/base matrix: 35/35 passed
- Changed-file ESLint and `git diff --check`: clean
- Gitleaks: no leaks

Added explicit coverage for message-only detection, repeated rejection ceiling, non-400 exclusion, missing ciphertext, original-request ciphertext preservation, and secret-free logging.

PR #2691, opened 2026-07-18
fix(azure): send max_completion_tokens for gpt-5/o-series reasoning deployments

chy1211:
## Summary

- `AzureExecutor.transformRequest` was a pure passthrough (`return body`), so it never translated `max_tokens` for reasoning-model deployments.
- Add `requiresMaxCompletionTokens(model)` + rename `max_tokens` -> `max_completion_tokens` for `gpt-5*` / `o1-` / `o3-` / `o4-` deployments, mirroring the pattern already used by the `github` executor for Copilot-hosted reasoning models.
- Non-reasoning deployments (e.g. `gpt-4o`) are untouched — `transformRequest` still returns the original body reference when no rename is needed.

## Why

Azure AI Foundry reasoning-model deployments reject the legacy `max_tokens` parameter:

```json
{"error":{"message":"Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.","type":"invalid_request_error","param":"max_tokens","code":"unsupported_parameter"}}
```

Any client that sends `max_tokens` (the common/default param name in most OpenAI-compatible SDKs) gets a 400 on every request to a `gpt-5*`/`o`-series Azure deployment, with no way to work around it from the client side through 9Router. The `github` executor already special-cases this for Copilot; `azure` was missing the same handling.

Reproduced end-to-end against a live Azure AI Foundry `gpt-5.6-luna` (GlobalStandard) deployment:

- `POST .../openai/deployments/gpt-5.6-luna/chat/completions?api-version=2025-04-01-preview` with `max_tokens` in the body -> `400 unsupported_parameter`
- Same request with `max_tokens` renamed to `max_completion_tokens` -> `200 OK`

## Test Plan

```bash
NODE_PATH=./node_modules ./tests/node_modules/.bin/vitest run tests/unit/azure-executor.test.js --reporter=verbose
```

Result: 1 file passed, 7 tests passed (new file, covers registry wiring, the `gpt-5`/`o[134]-` regex, and `transformRequest` rename/no-op/passthrough behavior including non-mutation of the original body object).

Also ran the full `tests/unit` suite before and after this change to confirm no regressions:

- Before (upstream/master, stashed): 794 passed / 84 failed / 76 skipped
- After (this branch): 798 passed / 80 failed / 76 skipped

The 4-passed / 4-failed delta is exactly this PR's 7 new tests (net vs. pre-existing flake in that run); all pre-existing failures are unrelated `open-sse` workspace-alias resolution errors in `tests/unit/xai-oauth-service.test.js` and friends, present identically on unmodified `upstream/master` in this environment (Node 20 portable build without the maintainer's workspace linking) — not caused by this change.

### DIFF
--- a/open-sse/executors/azure.js
+++ b/open-sse/executors/azure.js
@@ -65,7 +65,12 @@ export class AzureExecutor extends DefaultExecutor {
     }
 
     const transformed = { ...body };
-    transformed.max_completion_tokens = transformed.max_tokens;
+    // An explicit max_completion_tokens from the caller wins; max_tokens is
+    // only used as a fallback. Either way max_tokens must be stripped, or
+    // Azure 400s on it regardless of max_completion_tokens being present.
+    if (transformed.max_completion_tokens === undefined) {
+      transformed.max_completion_tokens = transformed.max_tokens;
+    }
     delete transformed.max_tokens;
     return transformed;
   }
--- a/open-sse/executors/azure.js
+++ b/open-sse/executors/azure.js
@@ -51,7 +51,22 @@ export class AzureExecutor extends DefaultExecutor {
     return headers;
   }
 
+  // Azure OpenAI reasoning-model deployments (gpt-5*, o1-/o3-/o4- series) reject
+  // the legacy `max_tokens` param with a 400 "Unsupported parameter" error and
+  // require `max_completion_tokens` instead. Same rule the "github" executor
+  // already applies for Copilot-hosted reasoning models.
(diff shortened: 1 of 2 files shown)


chy1211:
Good catch — fixed in 6764532 : `transformRequest` now only falls back to `max_tokens` when `max_completion_tokens` is absent, and always strips `max_tokens` either way. Added a regression test covering the both-fields-present case.

PR #2701, opened 2026-07-18
fix(kiro): map GPT reasoning effort fields

SemonCat:
## Summary

- map GPT-5.6 effort to `additionalModelRequestFields.reasoning.effort`
- preserve Claude adaptive-effort mapping through `output_config.effort`
- keep the existing prompt tags for backward compatibility

## Why

Kiro CLI selects the effort field path from each model schema. GPT models use `reasoning.effort`, while Claude models use `output_config.effort`. 9Router previously only emitted the Claude-shaped request fields, so GPT-5.6 did not receive the same structured thinking level as Kiro CLI.

## Verification

- production Docker build (`npm run build`) passes on Node 22
- targeted mapping smoke confirms GPT `reasoning.effort` and existing Claude `output_config.effort`
- live Kiro GPT-5.6 request with `reasoning_effort: high` returned HTTP 200
- added OpenAI- and Claude-format translator coverage for GPT-5.6 effort mapping and invalid values

### DIFF
--- a/open-sse/config/kiroConstants.js
+++ b/open-sse/config/kiroConstants.js
@@ -144,9 +144,13 @@ export function extractKiroEffortLevel(body) {
   return null;
 }
 
-export function buildKiroAdditionalModelRequestFields(body) {
+export function buildKiroAdditionalModelRequestFields(body, effortPath = "output_config") {
   const effort = extractKiroEffortLevel(body);
   if (!effort) return undefined;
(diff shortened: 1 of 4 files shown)


SemonCat:
Superseded by #2707, which also preserves Responses API `reasoning.effort` through the OpenAI bridge and covers the actual Hermes `codex_responses` path.

decolua:

PR #2708, opened 2026-07-19
feat(i18n): add Khmer language support

seakleangnhak:
## Summary

- register Khmer (`km`) in the supported locale configuration, display-name map, and locale normalization
- add a complete Khmer dashboard translation with 1,394 entries
- show the Cambodia flag in the language selector and header
- keep the selected locale synchronized with the header after switching languages
- place Khmer immediately before Thai in the language selector

## Translation key set

The Khmer file includes the 1,389-key set shared by the complete maintained locales (`fa`, `th`, and `zh-CN`), plus five older keys that are still rendered by the current application: `Failed`, `OK`, `Time`, `Type`, and `Version`.

Technical identifiers, placeholders, commands, paths, URLs, environment variables, provider names, model names, and product names are preserved where appropriate.

## Validation

- parsed `public/i18n/literals/km.json` successfully with Node.js
- confirmed 1,394 Khmer entries with no duplicate JSON keys
- confirmed no missing keys against the maintained 1,389-key locale set
- confirmed locale registration, normalization, API validation, switcher mapping, flag mapping, and filename consistently use `km`
- focused ESLint passes for `src/i18n/config.js`; the existing React hook diagnostics in the switcher/header are unchanged from `upstream/master`
- `npm run build` passes and generates 130/130 pages
- `git diff --check` passes

## Notes

No unrelated locale entries were changed.


decolua:
Thanks @seakleangnhak for the contribution! Reviewed and merged into master. 🙏

PR #2717, opened 2026-07-19
fix(kiro): validate terminal streams before emitting output

SemonCat:
## Summary

- validate AWS EventStream framing, header bounds, CRCs, error frames, and terminal stop metadata before exposing Kiro output
- preserve clean-EOF compatibility while retrying malformed transport or incomplete tool-call structure once
- propagate streaming failures through non-streaming callers instead of converting them into successful stop responses
- keep authoritative refusal and upstream error outcomes terminal

## Root cause

Kiro responses without a trustworthy terminal boundary could be normalized into `finish_reason=stop`. The fix buffers the bounded upstream response, validates its transport and terminal structure, and only then emits it downstream.

## Scope

This PR is intentionally transport-only. It does not inspect answer wording, ellipses, progress language, task completeness, or any other semantic policy. Semantic completion belongs in the caller/agent layer.

## Verification

- Kiro terminal/non-stream/thinking suites: 75 passed
- `node --check open-sse/executors/kiro.js`: passed
- production Next.js build: passed
- `git diff --check`: passed


decolua:

SemonCat:
Follow-up pushed in 966c6dc: all response-semantic heuristics were removed. The branch is now strictly EventStream / terminal integrity; answer completeness stays in the caller layer. Focused Kiro suites (75 tests) and the production build pass.

PR #2753, opened 2026-07-21
feat(antigravity): add gemini-3.6 (low/medium/high) models

AbdoKnbGit:
## Summary

- Add Gemini 3.6 Flash High, Medium, and Low models to Antigravity.
- Expose all three models through `/v1/models`, the CLI, and the MITM mapping UI.
- Route every variant through `gemini-3.6-flash-tiered` while preserving its distinct thinking level.
- Keep existing Gemini 3.5 models and fallback behavior unchanged.

## Implementation

- Register the following public model IDs:
  - `gemini-3.6-flash-high`
  - `gemini-3.6-flash-medium`
  - `gemini-3.6-flash-low`
- Reuse the existing thinking-suffix normalization to emit `high`, `medium`, or `low` in Antigravity's `thinkingConfig`.
- Preserve the resolved upstream model when the Antigravity executor builds the final request.
- Detect the selected thinking level in native Antigravity MITM requests so each model remains independently mappable.
- Default tiered requests without a valid thinking level to Medium.

## Validation

- Verified all three public models resolve to `gemini-3.6-flash-tiered`.
- Verified High, Medium, and Low produce distinct `thinkingLevel` values.
- Passed 136 focused translator and Antigravity tests.
- Successfully built the Next.js application, CLI package, and bundled MITM server.

### DIFF
--- a/cli/src/cli/menus/providers.js
+++ b/cli/src/cli/menus/providers.js
@@ -53,6 +53,9 @@ const PROVIDER_MODELS = {
     { id: "glm-4.7" },
   ],
   ag: [
+    { id: "gemini-3.6-flash-high" },
+    { id: "gemini-3.6-flash-medium" },
+    { id: "gemini-3.6-flash-low" },
     { id: "gemini-3-flash-agent" },
     { id: "gemini-3.5-flash-low" },
     { id: "gemini-3.5-flash-extra-low" },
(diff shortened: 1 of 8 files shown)


ramabondanp:
got 
HTTP 404: [antigravity/gemini-3.6-flash-medium] [404]: { "error": { "code": 404, "message": "Requested entity was not found.", "status": "NOT_FOUND" } }

AbdoKnbGit:
@ramabondanp Google has only one Gemini 3.6 model: gemini-3.6-flash-tiered. High/medium/low aren't models  they're a separate "how hard to think" setting. We were sending gemini-3.6-flash-medium as the model name, so Google said "no such model" → 404.It happened because the 3 models were added on top of the model list instead of inside the provider registry, so most of the code never saw them  and unknown models get forwarded as-is instead of erroring.
 Fix: put them in the registry with the real name gemini-3.6-flash-tiered plus a thinkingLevel field for the tier.

PR #2764, opened 2026-07-22
fix(server): initialize runtime services at process startup

ryanngit:
Runtime background services were imported only from the root dashboard layout. API-only serving could therefore accept requests without starting tunnel recovery, watchdogs, network monitoring, MITM startup, or opted-in quota auto-ping.

This change uses Next.js' native Node instrumentation boundary:
- extends `src/instrumentation.js` to load the existing singleton runtime bootstrap
- keeps Edge instrumentation inert
- relies on existing build-phase suppression and global singleton protection
- avoids loopback HTTP probes and custom-server coupling

Tested on pristine `v0.5.45` base:
- focused Vitest matrix: 39/39
- production `npm run build`
- focused ESLint
- `git diff --check`
- Gitleaks on `v0.5.45..HEAD`


ryanngit:
## v0.5.45 recut

Replaced old loopback `/api/init` implementation because v0.5.45 now provides `src/instrumentation.js`, the native process-start boundary.

- Base: `v0.5.45` (`6fcd2733`)
- New head: `82c9ae89`
- Scope: `src/instrumentation.js` plus one focused test
- Focused tests: 39/39 passed
- Production build: passed
- Focused ESLint and `git diff --check`: clean
- Gitleaks: no leaks

Old `custom-server.js` and `/api/init` changes were removed.

PR #2765, opened 2026-07-22
Prevent stale cloudflared exits from clearing successor PID

ryanngit:
Concurrent or closely spaced tunnel restarts can spawn a replacement cloudflared process before the previous child emits `exit`. Both exit handlers currently clear global process state and unlink the shared PID file unconditionally. The stale child can therefore erase the replacement PID, causing the watchdog to treat a healthy tunnel as dead and rotate the Quick Tunnel URL repeatedly.

This change:
- makes PID cleanup conditional on the exiting child still owning the PID file
- clears in-memory process ownership only when the exiting child is current
- preserves unconditional cleanup for explicit tunnel disable
- adds regression coverage for stale-child and current-child cleanup

Tested:
- `./tests/node_modules/.bin/vitest run --config tests/vitest.config.js --testTimeout 20000 tests/unit/tunnel-pid-ownership.test.js` (2 passed)
- focused ESLint and `git diff --check`

### DIFF
--- a/src/lib/tunnel/cloudflare/cloudflared.js
+++ b/src/lib/tunnel/cloudflare/cloudflared.js
@@ -239,8 +239,8 @@ export async function spawnCloudflared(tunnelToken) {
     });
 
     child.on("exit", (code, signal) => {
-      cloudflaredProcess = null;
-      clearPid();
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @ryanngit for the contribution! Reviewed and merged into master. 🙏

PR #2784, opened 2026-07-22
feat(providers): add configurable error cooldown policies

zyizyiz:
## Summary

This PR improves reliability for custom OpenAI Responses-compatible providers and adds opt-in, per-connection error cooldown policies.

It addresses two related problems:

1. Requests that worked when sent directly could fail through 9Router because custom Responses providers received a generic client identity, Responses-native fields were rewritten into Chat Completions shapes, and network exceptions retried the same dead URL before falling over.
2. Provider errors lost useful metadata before account fallback, so users could not reliably apply different cooldowns by HTTP status, provider error code, or message.

### OpenAI Responses and network forwarding

- Preserve Responses-native `instructions` and `reasoning.effort` instead of converting them to Chat Completions-only fields.
- Avoid mutating non-message Responses input items such as `additional_tools` during system prompt injection.
- Send `codex_cli_rs/0.136.0` as the User-Agent for custom Responses-compatible providers whose upstream filters reject the generic OpenAI/Python client identity.
- Move immediately to the next configured base URL after a network exception instead of consuming status-based retries against the failed connection.

### Configurable error cooldown policies

Each provider connection can now enable an error cooldown policy with:

- Ordered rules; the first matching rule wins.
- AND semantics across all populated match fields.
- HTTP status matching.
- Case-insensitive `error.code` matching.
- Case-insensitive message substring matching.
- Key-wide or current-model scope.
- Fixed durations of 30 minutes, 1 hour, 5 hours, or 24 hours.
- Custom durations from 1 minute through 30 days.
- A required default duration that catches every unmatched failure, including gateway and network errors without structured upstream metadata.

Configured rules intentionally override provider-reported reset times. When the policy is disabled, the existing reset-time and exponential-backoff behavior remains unchanged.

### Error metadata propagation

- Preserve upstream HTTP status, provider error code, `Retry-After`, and `reset_at` / `resets_at` timestamps.
- Normalize JSON, text, and HTML error responses without exposing raw provider pages to the policy matcher.
- Carry metadata through chat, embeddings, image generation, fetch, search, STT, TTS, video, Gemini-native forwarding, and connection validation.
- Preserve MiniMax TTS application-level `base_resp.status_code` errors even when the HTTP response is 200.

### Persistence and concurrency

- Store policies in the existing connection JSON data, so no schema migration is required.
- Never shorten an already active cooldown.
- Compute success-side error clearing atomically from the latest connection state, preventing an older successful request from clearing a newer concurrent failure lock.
- Clear locks only when an API key actually changes or when the user explicitly selects **Clear error cooldown**.
- Keep historical `end-of-day` policy values readable; the editor normalizes them to the current 24-hour option when saved.

### Dashboard and localization

- Add reusable cooldown policy form and status summary components to the connection editor and connection rows.
- Show the active rule/default source, HTTP/provider code, expiry, and countdown.
- Support manual cooldown clearing.
- Add all required strings to all 34 non-English locale files, using native `Intl` formatting for duration units and timezone-aware timestamps.

## Backward compatibility

- The feature is opt-in per connection.
- Connections without an enabled policy keep the legacy fallback behavior.
- Existing connection records require no migration.
- Existing model locks and provider reset timestamps remain supported.

## Verification

- `cd tests && ./node_modules/.bin/vitest run --config vitest.config.js unit/error-cooldown-policy.test.js unit/minimax-tts.test.js unit/base-executor-retry.test.js unit/system-inject.test.js unit/thinking-effort-openai-max-clamp.test.js`
  - 5 files, 26 tests passed.
- `cd tests && ./node_modules/.bin/vitest run --config vitest.config.js unit/db-concurrent.test.js -t 'stale success does not clear a newer cooldown lock' --testTimeout=15000 --hookTimeout=15000`
  - Focused concurrency regression passed.
- All 34 locale JSON files parse successfully and contain the 17 strings used by the new UI.
- `git diff --check origin/master...HEAD` passes.

The full `db-concurrent.test.js` file still reports three unrelated usage-aggregation failures in this environment; the new cooldown race regression passes independently.


GuoZhuang-cn:
This is exactly what we're missing — per-status configurable cooldown policies instead of the hardcoded `30000 / 120000 / 1800000` table in the error-rules chunk.

Anything blocking review or merge? Is there a way to help move this forward (running builds, testing against my configs)?

Context from my deployment that this PR would fix: a provider whose upstream API key got disabled (`401 → 503`, "API key is disabled") stays in the probe rotation on the 30 s default lock, so every combo request pays one wasted upstream round-trip before falling through to the next member. With a configurable 120 s policy for service-unavailable class errors, that member would be skipped silently until the upstream actually recovers — without patching minified chunks (my container rebuilds from `decolua/9router:latest` on every release, ~daily).

I can also confirm the need for #3352-style env knobs for the 429 backoff schedule: the 2 s → 5 min ladder is too fine-grained for quota windows that reset hourly — my MiniMax plan-limit 429s lock 300 s but get re-probed every few minutes all the same.

PR #2789, opened 2026-07-22
fix(kiro): normalize dashboard thinking intensity models

### LINKED ISSUE #2716: Bug: Kiro models fail with REQUEST_BODY_INVALID after setting thinking intensity in dashboard
## Description

After updating 9router and configuring the thinking intensity option in the dashboard, Kiro models that previously worked correctly start returning HTTP 400 errors.

Affected models include:

* `kiro/claude-sonnet-4.5-thinking-agentic`
* `kiro/glm-5-thinking-agentic`
* ...

Error message:

```
HTTP 400:
[kiro/claude-sonnet-4.5-thinking-agentic]
[400]: {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}
```

## Expected behavior

For models that do not support explicit thinking intensity parameters, 9router should either:

1. Not send any thinking/reasoning parameter by default, or
2. Explicitly disable thinking when no value is configured.

Previously, these models worked normally before the thinking intensity setting was introduced.

## Actual behavior

After updating the dashboard and setting thinking intensity support, it seems that all requests are being assigned:

```
thinking: AUTO
```

or an equivalent automatic reasoning parameter.

This causes Kiro upstream requests to contain invalid reasoning configuration, resulting in:

```
REQUEST_BODY_INVALID
```

Models such as `claude-sonnet-4.5-thinking-agentic` and `glm-5-thinking-agentic` can no longer be used.

## Steps to reproduce

1. Update 9router to the latest version.
2. Open dashboard settings.
3. Configure thinking intensity.
4. Use Kiro provider models:

   * `kiro/claude-sonnet-4.5-thinking-agentic`
   * `kiro/glm-5-thinking-agentic`
5. Send a request.

The request fails with HTTP 400.

## Possible cause

The new thinking intensity option may be globally injected into all provider requests.

However, different providers/models have different reasoning parameter formats. Kiro's agentic models may not accept:

```
thinking: AUTO
```

or similar automatic reasoning values.

## Suggested fix

* Default thinking intensity should be empty/disabled instead of `AUTO`.
* Only inject reasoning parameters when the selected provider/model explicitly supports them.
* Add provider-specific handling for Kiro reasoning models.

## Additional information

I have screenshots comparing behavior before and after enabling/configuring thinking intensity, which show that the failure starts only after this change.

Image
Image
Image

Thanks for maintaining 9router. Hopefully this compatibility issue can be fixed soon.

ankit1324:
## Summary

Fixes #2716

Kiro models such as `claude-sonnet-4.5-thinking-agentic` and `glm-5-thinking-agentic` return `REQUEST_BODY_INVALID` after a dashboard thinking level is appended (for example `(high)` or `(medium)`).

## Root cause

The dashboard's generic thinking selector appends a trailing `model(level)` suffix. The Kiro translators preserved that suffix while resolving the synthetic `-thinking` / `-agentic` variants, so the upstream request could contain an invalid model id such as:

`claude-sonnet-4.5-thinking-agentic(high)`

## Fix

- Normalize the generic Kiro `model(level)` suffix before resolving synthetic Kiro variants.
- Keep the clean upstream model id in the Kiro envelope.
- Preserve agentic behavior for unsupported legacy models.
- Map explicit levels to native Kiro effort fields only for supported Claude/GPT model families.
- Stop the dashboard thinking-level helper from advertising native levels for unsupported Kiro models.
- Apply the same normalization to both OpenAI→Kiro and direct Claude→Kiro routes.

## Verification

- Focused Kiro/thinking tests: **102 passed**
- Root ESLint on changed files: passed
- Production build: passed (`npm run build`)
- Full independent test suite: contains the repository's documented baseline/missing-fixture failures; no Kiro-focused failures
- `git diff --check`: passed

The reported models keep their agentic chunked-write behavior, but no longer send the invalid parenthesized model id or unsupported native effort fields.

### DIFF
--- a/open-sse/config/kiroConstants.js
+++ b/open-sse/config/kiroConstants.js
@@ -15,7 +15,7 @@
  * fiction. The suffix is stripped before the request leaves this process.
  */
 
-import { extractThinking } from "../translator/concerns/thinkingUnified.js";
+import { extractThinking, parseSuffix } from "../translator/concerns/thinkingUnified.js";
 import { effortToBudget } from "../translator/concerns/thinking.js";
 
 export const KIRO_AGENTIC_SUFFIX = "-agentic";
@@ -39,6 +39,39 @@ export function resolveDefaultProfileArn(authMethod) {
 }
 
 export const KIRO_THINKING_BUDGET_DEFAULT = 16000;
+
+/**
+ * Resolve a Kiro model after consuming the generic model(level) suffix.
+ * The suffix is a 9router request override, not part of Kiro's upstream model id.
+ */
+export function resolveKiroModelIntent(model) {
+  const { cleanModel, override } = parseSuffix(model);
+  return {
+    model: cleanModel,
+    ...resolveKiroModel(cleanModel),
+    thinkingOverride: override,
+  };
+}
+
+/** Apply a parsed model(level) override without mutating the caller's body. */
+export function applyKiroThinkingOverride(body, override) {
+  if (!override) return body;
+
+  const next = { ...body };
+  if (override.mode === "budget") {
+    delete next.output_config;
+    delete next.reasoning_effort;
+    delete next.reasoning;
+    next.thinking = { type: "enabled", budget_tokens: override.budget };
+    return next;
+  }
+
+  next.output_config = {
+    ...(body.output_config || {}),
+    effort: override.mode === "level" ? override.level : override.mode,
+  };
+  return next;
+}
 
 export const KIRO_AGENTIC_SYSTEM_PROMPT = `
 # CRITICAL: CHUNKED WRITE PROTOCOL (MANDATORY)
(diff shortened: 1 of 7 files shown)


ankit1324:
I verified the issue against the current `master` checkout before implementing. The focused Kiro/thinking suite passes (102 tests), the production build and changed-file ESLint pass, and the reported models preserve agentic behavior while emitting clean upstream model IDs.

ankit1324:
image

image

decolua:
Thanks @ankit1324 for the contribution! Reviewed and merged into master. 🙏

ankit1324:
@decolua Thank you so much! It's truly an honor to contribute to this repository. 🙏

A few months ago, I actually had a very similar idea, but I never had enough time to build it from scratch. Then one day I came across this repo, started exploring the codebase and features, and I was genuinely amazed by the overall architecture and approach. It was exactly the kind of project I had imagined.

I do have one question though. I've been struggling to use Claude Code with this repository. It seems to refuse or fail almost every time I ask it to work on the codebase. If you have any recommended setup, workflow, or tips to make Claude Code work better here, I'd really appreciate it.

I'd also love to hear about the future roadmap for the project. I'm excited to keep contributing, and if there are features or improvements you'd like help with, I'd be more than happy to work on them.

Thanks again for building such an awesome project! 🚀

PR #2792, opened 2026-07-22
fix: recover Jina Reader after transient errors

jacardl:
## Summary

- call Jina Reader through its JSON `POST https://r.jina.ai/` API
- parse the `Title:` metadata returned by Jina before falling back to a Markdown heading
- clear stale provider error codes and account locks after a successful web fetch
- add regression coverage for the Jina request shape, upstream errors, and successful account recovery

## Root cause

A transient Jina Reader error such as HTTP 402 marks the provider connection unavailable and creates an account-wide cooldown. The web fetch handler passed an `onRequestSuccess` callback to the core fetch handler, but the core handler never consumed that callback. A later successful request therefore did not clear the stale provider state, so the UI could continue to show `[jina-reader] Unavailable` even while direct Jina requests worked.

The Jina adapter also embedded an encoded target URL into a GET path instead of using Jina’s documented JSON POST request, and its title parser did not recognize Jina’s `Title:` metadata line.

## Impact

Jina Reader connections now recover automatically after a transient upstream failure. The adapter uses the documented request format and returns the page title from Jina metadata when available.

## Validation

- `npx vitest run --config tests/vitest.config.js tests/unit/jina-reader-fetch.test.js tests/unit/fetch-success-clears-account.test.js`
- 2 test files passed, 3 tests passed
- built the 9Router 0.5.40 CLI package successfully
- verified a local `/v1/web/fetch` request through 9Router returned HTTP 200 and cleared the stale error code and global lock

### DIFF
--- a/open-sse/handlers/fetch/index.js
+++ b/open-sse/handlers/fetch/index.js
@@ -49,7 +49,10 @@ function truncate(text, max) {
 }
 
 function parseJinaTitle(text) {
-  const m = String(text || "").match(/^\s*#\s+(.+)$/m);
+  const source = String(text || "");
+  const metadataTitle = source.match(/^\s*Title:\s*(.+)$/mi);
+  if (metadataTitle) return metadataTitle[1].trim();
+  const m = source.match(/^\s*#\s+(.+)$/m);
   return m ? m[1].trim() : null;
 }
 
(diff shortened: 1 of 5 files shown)


decolua:
Thanks @jacardl for the contribution! Reviewed and merged into master. 🙏

PR #2453, opened 2026-07-08
fix(usage): preserve nonnegative provider-reported exact cost

ryanngit:
## Summary

Preserve trusted provider-reported exact request cost without replacing current pricing tables or usage UI.

- carry `cost_usd`, `cost_in_usd`, and `cost_in_usd_ticks` through usage extraction, normalization, format filtering, and canonical storage
- prefer `cost_usd`, then `cost_in_usd`, then xAI ticks
- accept finite nonnegative totals, including exact zero
- reject negative, empty, NaN, and infinite totals
- trust exact provider total even when local model pricing is missing
- convert xAI ticks using `1 USD = 10^10 ticks`
- fall back to current local token pricing when exact value is invalid or absent

## Scope

This refresh intentionally excludes:
- model pricing table changes
- cache-write/effective-tier transport
- stored/UI component breakdowns
- xAI local quota/dashboard rows
- historical usage rewrites
- API-key grouping

Streaming cache/tier work remains separate and overlaps #2348, #2713, #2747, #2873, and #2880. Exact embedding usage from #2794 is already upstream.

Base: `v0.5.45` at `6fcd27337a7893642c7fe630840d0a641743f28f`

## Verification

- focused/adjacent Vitest: 2 files, 25/25 passed
- tests cover missing pricing, exact zero, precedence, negative rejection, xAI tick scale, streaming extraction, non-streaming extraction, and existing cache accounting
- production `npm run build`: passed
- changed-file ESLint: passed
- `git diff --check`: passed
- Gitleaks: 4 changed files, no findings

Full upstream suite is not used as binary gate because pristine `v0.5.45` currently has 111 unrelated baseline failures.

PR #2794, opened 2026-07-23
fix(usage): record exact embedding tokens

zierocode:
## Problem

Successful embedding responses already return provider usage to clients, but the local usage history does not record it. Embedding traffic is therefore absent from usage and cost accounting.

## Fix

- return normalized provider usage alongside the core success response
- persist usage after a successful embedding request
- accept only exact positive safe-integer totals where input equals total and completion is zero
- reject estimated, malformed, zero, or internally inconsistent usage instead of fabricating counts
- keep persistence non-blocking so a DB failure cannot fail a successful inference

## Scope

This is provider-neutral embedding accounting. It does not change response bodies, provider routing, credentials, or the xAI/cache-cost work in #2453.

## TDD evidence

RED on `79918c7830695bbca4a45c9fea4a42c3e9fd73d1`:

```text
records exact provider usage for successful embedding requests
Number of calls: 0
```

GREEN on `660552583fca0df6c51cbd23312b016ac5e05c32`:

- focused tests: 2 files, 47 tests passed
- ESLint: passed on all changed files
- production build: passed with isolated `DATA_DIR`

### DIFF
--- a/open-sse/handlers/embeddingsCore.js
+++ b/open-sse/handlers/embeddingsCore.js
@@ -116,6 +116,7 @@ export async function handleEmbeddingsCore({
 
   return {
     success: true,
+    usage: normalized.usage || null,
     response: new Response(JSON.stringify(normalized), {
       headers: {
         "Content-Type": "application/json",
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @zierocode for the contribution! Reviewed and merged into master. 🙏

PR #2811, opened 2026-07-24
Include commandcode cache reads in usage statistics

Vaults1:
Small change to include cache reads in the usage statistics from the commandcode provider.

As far as I can tell, there is currently no way to retrieve cache write stats.

### DIFF
--- a/open-sse/translator/concerns/usage.js
+++ b/open-sse/translator/concerns/usage.js
@@ -57,7 +57,10 @@ const USAGE_EXTRACTORS = {
   commandcode(raw) {
     const input = n(raw.inputTokens), output = n(raw.outputTokens);
     const total = typeof raw.totalTokens === "number" ? raw.totalTokens : input + output;
-    return { promptTokens: input, completionTokens: output, totalTokens: total };
+    const cached = n(raw.cachedInputTokens);
+    const out = { promptTokens: input, completionTokens: output, totalTokens: total };
+    if (cached > 0) out.cachedTokens = cached;
+    return out;
   },
 };
 


kushaldotdev:
Real-world confirmation from a live setup (augmenting the fix with field-shape evidence):

I verified against the commandcode **provider endpoint** (`https://api.commandcode.ai/provider/v1/chat/completions`, the OpenAI-compatible one) with a valid `user_*` key and `deepseek/deepseek-v4-flash`:

```json
{
  "prompt_tokens": 107,
  "completion_tokens": 10,
  "total_tokens": 117,
  "prompt_tokens_details": { "cached_tokens": 0, "audio_tokens": 0, "video_tokens": 0 },
  "completion_tokens_details": { "reasoning_tokens": 10, "image_tokens": 0 },
  "cache_creation_input_tokens": 0
}
```

Observations:
1. The provider endpoint returns cache data as **`prompt_tokens_details.cached_tokens`** (OpenAI chat-completions shape), **not** `cachedInputTokens`. The PR reads `raw.cachedInputTokens` — worth confirming which field the `/alpha/generate` (AI SDK v5) stream emits on a cache hit, since the two endpoints differ.
2. `cached_tokens` was `0` across 3 identical repeat prompts (121 tokens) — so cache reads via this endpoint may be rare/none for deepseek, but the **field is present** and the fix direction is correct.

This is exactly the gap I hit: routed through 9Router, the client never sees `cached_tokens`, so cache usage and hit-rate stats are blank. Would be great to see this merged.

Vaults1:
@kushaldotdev this repo only implements the `/alpha/generate` endpoint, so maybe worth running your test again using that?

PR #2818, opened 2026-07-25
fix(models): sync discovered catalogs with configured models

DedSecer:
## Summary
- Load account-specific live catalogs into provider controls for Cursor, GitHub Copilot, Kiro, Qoder, Kimchi, ClinePass, and Grok CLI.
- Ensure **Disable All** can manage every dynamically discoverable model.
- Treat configured custom models on OpenAI/Anthropic Compatible nodes as an explicit whitelist: when such a list exists, `/v1/models` no longer auto-adds every model returned by the upstream `/models` endpoint.
- Add regression coverage for live-only catalogs and compatible-node whitelist discovery.

## Problems fixed
1. The gateway could expose dynamic provider models that were absent from provider-page controls.
2. A compatible node with four manually configured models could still expose the full upstream catalog through `/v1/models`.

## Test plan
- `npx vitest run unit/compatible-model-discovery.test.js unit/merge-model-catalogs.test.js`
- `npm run build`
- Live verification: disabling the full Copilot catalog yields zero `gh/` models; a dex compatible node configured with four custom models now returns exactly those four `dex/` models.

### DIFF
--- a/src/app/(dashboard)/dashboard/providers/[id]/page.js
+++ b/src/app/(dashboard)/dashboard/providers/[id]/page.js
@@ -27,6 +27,18 @@ import BulkImportGrokCliModal from "./BulkImportGrokCliModal";
 
 const ONE_BY_ONE_DELAY_MS = 1000;
 
+// These providers use an account-specific catalog in /v1/models. Keep their
+// dashboard model controls on the same catalog so Disable All remains complete.
+const LIVE_CATALOG_PROVIDER_IDS = new Set([
+  "cursor",
+  "github",
+  "kiro",
+  "qoder",
+  "kimchi",
+  "clinepass",
+  "grok-cli",
+]);
+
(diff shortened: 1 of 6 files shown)


DedSecer:
Hi @decolua, gentle follow-up on PR #2818.

I rebased the PR onto the latest master (v0.5.59) and updated the remote branch successfully. The PR is currently clean and mergeable.

The two issues addressed by this PR still appear to be relevant:
Compatible providers with explicitly configured custom models should treat those models as a whitelist instead of automatically exposing the entire upstream /models catalog.
Account-specific live catalogs should be merged into the dashboard model controls so that “Disable All” covers every model exposed through /v1/models.

The focused regression tests pass:
2 test files passed
6 tests passed

Would you prefer that I:

keep this as one PR, split it into a smaller compatible-provider whitelist PR and a separate live-catalog dashboard PR, or revise it according to the current project direction?
Thanks.

PR #2824, opened 2026-07-25
feat(ollama): add embedding provider adapter and endpoint support for Ollama Local

HaoNgo232:
## Summary
Adds native embedding provider adapter and endpoint support for **Ollama Local** (`ollama-local`) via `/v1/embeddings`.

## Key Changes
- Created `ollamaLocal.js` embedding provider adapter connecting to `http://localhost:11434/v1/embeddings` (or configured custom host with trailing slash normalization).
- Registered `ollama-local` adapter in `open-sse/handlers/embeddingProviders/index.js`.
- Updated `ollama-local` provider registry to include `"embedding"` serviceKind, `embeddingConfig`, and static embedding models catalog (`embeddinggemma`, `nomic-embed-text`, `bge-m3`).
- Added unit tests covering default and custom host routing (including trailing slash stripping) for Ollama Local.

## Testing
- Unit tests added in `tests/unit/embeddingsCore.test.js` (42/42 passing).

### DIFF
--- a/tests/unit/embeddingsCore.test.js
+++ b/tests/unit/embeddingsCore.test.js
@@ -267,6 +267,44 @@ describe("buildEmbeddingsUrl", () => {
     expect(url).toBe("https://myhost.ai/v1/embeddings");
   });
 
+  it("ollama-local default host → http://localhost:11434/v1/embeddings", async () => {
+    vi.mocked(fetch).mockResolvedValueOnce(makeProviderResponse(VALID_EMBEDDING_RESPONSE));
+
+    await handleEmbeddingsCore(makeOptions({
+      modelInfo: { provider: "ollama-local", model: "nomic-embed-text" },
+      credentials: {},
+    }));
+
+    const [url, init] = vi.mocked(fetch).mock.calls[0];
+    const sent = JSON.parse(init.body);
+    expect(url).toBe("http://localhost:11434/v1/embeddings");
+    expect(sent.model).toBe("nomic-embed-text");
+  });
+
+  it("ollama-local custom host → respects baseUrl from providerSpecificData", async () => {
+    vi.mocked(fetch).mockResolvedValueOnce(makeProviderResponse(VALID_EMBEDDING_RESPONSE));
+
+    await handleEmbeddingsCore(makeOptions({
+      modelInfo: { provider: "ollama-local", model: "bge-m3" },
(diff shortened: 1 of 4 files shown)


tests/unit/embeddingsCore.test.js
```diff
@@ -267,6 +267,44 @@ describe("buildEmbeddingsUrl", () => {
...
+
+  it("ollama-local custom host → respects baseUrl from providerSpecificData", async () => {
+    vi.mocked(fetch).mockResolvedValueOnce(makeProviderResponse(VALID_EMBEDDING_RESPONSE));
+
+    await handleEmbeddingsCore(makeOptions({
+      modelInfo: { provider: "ollama-local", model: "bge-m3" },
+      credentials: { providerSpecificData: { baseUrl: "http://192.168.1.100:11434" } },
+    }));
+
+    const [url] = vi.mocked(fetch).mock.calls[0];
```
sunba91-su: If your test fails because fetch was never called, `mock.calls[0]` will be `undefined`. Trying to destructure `[url]` from undefined will throw an ugly JavaScript TypeError: `undefined is not iterable`, which hides the actual reason your test failed.

Using `.toHaveBeenCalledWith()` guarantees you get a clean, readable Vitest error message saying exactly what went wrong.

tests/unit/embeddingsCore.test.js
```diff
@@ -267,6 +267,44 @@ describe("buildEmbeddingsUrl", () => {
...
+    expect(url).toBe("http://192.168.1.100:11434/v1/embeddings");
```
sunba91-su: The `.toBe()` assertion checks for exact string equality. If your application code calls fetch with `http://localchat.example.com/v1/embeddings`, but your test expects exactly `http://192.168.1.100:11434/v1/embeddings`, Vitest will throw an error because the strings don't match.

sunba91-su (changes requested):

tests/unit/embeddingsCore.test.js
HaoNgo232: I've updated this test to use expect(fetch).toHaveBeenCalledWith(...). Thanks for the tip!

tests/unit/embeddingsCore.test.js
HaoNgo232: I've updated the assertion here as well to cleanly match the expected URL using expect(fetch).toHaveBeenCalledWith(...). Thanks again for the review!

PR #2838, opened 2026-07-26
fix(kiro): canonicalize tool history and route API keys correctly

nguyenha935:
## Summary

- Route Kiro API-key inference through Amazon Q before the legacy CodeWhisperer and Kiro runtime surfaces.
- Canonicalize Claude/OpenAI tool history after session replay and immediately before building the Kiro wire payload.
- Enforce adjacent one-to-one `toolUses`/`toolResults`, including partial parallel calls, duplicate IDs/results, orphan results, trailing unanswered calls, and compacted history.
- Preserve valid structured pairs while flattening only interactions that can no longer be represented safely.
- Normalize Kiro tool names, descriptions, and JSON schemas without mutating the client request; preserve Claude `tool_result.is_error` as Kiro `status: "error"`.
- Build endpoint-specific headers: `X-Amz-Target` is sent only to `codewhisperer.*`; API keys use `TokenType: API_KEY`.
- Treat payload-invalid HTTP 400 as terminal instead of resending the same malformed body to every Kiro surface.
- Validate API keys against the Amazon Q model catalog, reject empty catalogs, and avoid inventing or persisting a `profileArn`.

## Why

The endpoint fix alone did not cover long agent sessions. Kiro rejects a payload when an assistant turn contains multiple parallel tool uses but the immediately following user turn does not contain exactly one result for every tool-use ID. The previous shared helper accepted a partial match as complete and, on the direct Claude→Kiro route, inserted OpenAI `role: "tool"` messages that the translator ignored. Reconciliation also ran before session replay, allowing replay to reintroduce an invalid history.

The new shared Kiro canonicalizer runs after replay, checks each adjacent assistant/user pair, remaps IDs together, keeps complete pairs structured, and converts only missing/orphaned pieces to readable text. It never fabricates a successful tool result.

## Validation

- 81/81 direct Kiro translator, canonicalization, replay, and endpoint tests passed.
- 200 related translator/Kiro tests passed; 2 existing `it.fails` cases remained expected.
- Regression coverage includes a 30-parallel-tool turn with one missing result.
- ESLint: 0 errors (one pre-existing anonymous default-export warning in the Kiro registry).
- `git diff --check` passed.
- Production build passed.
- Full five-patch production stack applied cleanly and `auto-update.sh --check` passed.
- Live API-key smoke after deployment: simple request, complete parallel tools, and partial 30-tool history all returned HTTP 200.

The unrelated golden snapshot/environment failures reproduce on the clean base commit and are not updated by this PR.

### DIFF
--- a/open-sse/config/kiroConstants.js
+++ b/open-sse/config/kiroConstants.js
@@ -20,6 +20,12 @@ import { effortToBudget } from "../translator/concerns/thinking.js";
 
 export const KIRO_AGENTIC_SUFFIX = "-agentic";
 export const KIRO_THINKING_SUFFIX = "-thinking";
+export const KIRO_TOOL_NAME_MAX_LENGTH = 64;
+export const KIRO_TOOL_DESCRIPTION_MAX_LENGTH = 10237;
+export const KIRO_TOOL_ID_MAX_LENGTH = 64;
+export const KIRO_CODEWHISPERER_TARGET =
+  "AmazonCodeWhispererStreamingService.GenerateAssistantResponse";
+export const KIRO_ENDPOINT_FALLBACK_STATUSES = new Set([401, 403, 404]);
 
 // Public default CodeWhisperer profile ARNs (us-east-1), keyed by auth method.
 // Used when an account cannot resolve its own profileArn. Builder ID and social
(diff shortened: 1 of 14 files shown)


decolua:

PR #2848, opened 2026-07-26
fix(claude): harden OAuth usage quota polling

ryanngit:
## Summary

- normalizes current and legacy Claude OAuth five-hour, weekly, and model-scoped quota windows
- hashes credential cache keys and bounds cache/in-flight state to 128 entries
- coalesces concurrent polls and caches successful results for 65 seconds
- honors bounded `Retry-After` and Anthropic reset cooldowns
- serves stale known quota on 429, reports 401 expiry, and uses legacy endpoints only for 404/405

This normalized quota shape is the data dependency for #2850 inactive-window auto-ping.

## Verification

- `tests/unit/claude-usage.test.js` and `tests/unit/usage-dispatch.test.js`: 22/22 passed
- changed-file ESLint: passed
- `git diff --check`: passed
- exact patch replay verified with `git range-diff`
- Gitleaks: no leaks

Base: `v0.5.45` (`6fcd2733`).


ryanngit:
## v0.5.45 refresh

- Old head: `e3308cfa`
- New head: `36f7a367`
- Dropped old-base history; behavior patch is range-diff equivalent
- Focused quota/dispatch suite: 22/22 passed
- ESLint, diff check, and Gitleaks passed

Issue #2592, opened 2026-07-14
Auto-Ping doesn't work


MFA-G:
I've used the auto-ping feature for both Codex and Claude, and both didn't work (the countdown timer wouldn't start), but since Codex has now removed the session limit, I don't think the Codex auto-ping feature will be as important as Claude's auto-ping feature.

I think it's because claude's auto-ping uses the Haiku model, and I just tested that model and it's giving me errors.

PR #2850, opened 2026-07-26
fix(claude): guard inactive quota auto-ping

ryanngit:
## Summary

- sends one tiny request only for explicitly enrolled Claude OAuth connections whose five-hour session row exists without `resetAt`
- requires known non-exhausted weekly capacity and no exhausted blocking quota
- fails closed when selected proxy resolution is unavailable
- drains accepted responses, cancels rejected bodies, and persists success only after completion
- prevents duplicate accepted pings for five hours and applies a 15-minute failure cooldown

## Dependencies and ceiling

- #2848 supplies normalized Claude OAuth quota windows
- #2764 or equivalent runtime bootstrap is required for API-only process startup
- this PR does not select a replacement ping model for open issue #2592

## Verification

- quota/usage/provider-display matrix: 35/35 passed
- changed-file ESLint: passed
- `git diff --check`: passed
- exact patch replay verified with `git range-diff`
- Gitleaks: no leaks

Base: `v0.5.45` (`6fcd2733`).


ryanngit:
## v0.5.45 refresh

- Old head: `e1de3bac`
- New head: `37c73c73`
- Dropped old-base history; behavior patch is range-diff equivalent
- Focused matrix: 35/35 passed
- ESLint, diff check, and Gitleaks passed
- Dependency notes for #2848/#2764 and #2592 are now explicit

PR #2491, opened 2026-07-09
feat(featherless): add OpenAI-compatible provider presets

newnol:
Adds a Featherless provider using the OpenAI-compatible API with a small curated preset list for DeepSeek, GLM, and Kimi models.\n\nIncluded presets:\n- deepseek-ai/DeepSeek-V4-Pro\n- deepseek-ai/DeepSeek-V4-Flash\n- zai-org/GLM-5.2\n- zai-org/GLM-5.1\n- moonshotai/Kimi-K2.7-Code\n- moonshotai/Kimi-K2.6\n- moonshotai/Kimi-K2.5\n\nNotes:\n- base URL: https://api.featherless.ai/v1/chat/completions\n- validate URL: https://api.featherless.ai/v1/models\n- kept intentionally small instead of importing the full Featherless catalog

PR #2854, opened 2026-07-27
feat(providers): add Poolside provider (OpenAI-compatible)

whale9820:
## Summary
Adds [Poolside](https://poolside.ai) as an API-key provider. Poolside exposes an OpenAI-compatible Responses/Chat API at `https://inference.poolside.ai/v1`, so this uses the standard registry pattern with the default OpenAI transport — no custom executor or translator required.

## Changes
- `open-sse/providers/registry/poolside.js` — new registry entry (`category: "apikey"`, `format: "openai"`, alias `poolside` / `ps`).
- `open-sse/providers/registry/index.js` — register the entry (`p100`), appended the same way as prior additions.

## Models
All three are currently free, 262K context, 32K max completion, with `tools` + `reasoning`:

| id | note |
|----|------|
| `poolside/laguna-s-2.1` | 2nd-gen S class — set as default |
| `poolside/laguna-xs-2.1` | 2nd-gen XS class |
| `poolside/laguna-m.1` | most-capable flagship, but carries `deprecation_date: 2026-07-28`, so it is listed last rather than defaulted |

## Verification
- `GET /v1/models` and `POST /v1/chat/completions` confirmed working against the live API with a key (OpenAI-shaped responses, including `reasoning_content`).
- Loaded the registry locally: `PROVIDERS.poolside` builds as `{ baseUrl, validateUrl, format: "openai" }` and `PROVIDER_MODELS.poolside` resolves all three models; `ps` alias wires through the standard alias mechanism.
- Follows the same two-file footprint as the previously merged Featherless provider PR (#2491).

### DIFF
--- a/open-sse/providers/registry/index.js
+++ b/open-sse/providers/registry/index.js
@@ -99,6 +99,7 @@ import p96 from "./xiaomi-mimo.js";
 import p97 from "./xiaomi-tokenplan.js";
 import p98 from "./youcom.js";
 import p99 from "./alims-intl.js";
+import p100 from "./poolside.js";
 
 export default [
   p0,
@@ -201,4 +202,5 @@ export default [
   p97,
   p98,
   p99,
+  p100,
 ];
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @whale9820 for the contribution! Reviewed and merged into master. 🙏

PR #2858, opened 2026-07-27
feat(kiro): add Claude Opus 5 models

Jordannst:
## Description
Add Claude Opus 5 to the Kiro provider model registry.

## Changes
- Add normal, thinking, agentic, and thinking + agentic variants.
- Declare 1M context, 128K output, and adaptive-thinking capabilities.
- Cover all four variants with focused capability tests.

## Verification
- `git diff --check`
- Ran an assert-based registry/capability check for all four model IDs.
- The checkout does not include a Vitest runner, so the full test suite was not run.

## Reference
- https://kiro.dev/blog/opus-5/


decolua:

PR #2864, opened 2026-07-28
i18n(pt-BR): traducao parcial — 986 termos (+793 novos)

FuryusTK:


decolua:
Thanks @FuryusTK for the contribution! Reviewed and merged into master. 🙏

Issue #2595, opened 2026-07-14
❌ antigravity [400]: [400]: { "error": { "code": 400, "message": "Invalid value at 'request.tools[0].function_d


Puh1001:
[11:01:16] ⚠️ [AUTH] xxx@gmail.com locked modelLock_gemini-3.5-flash-extra-low for 30s [400]
❌ antigravity [400]: [400]: { "error": { "code": 400, "message": "Invalid value at 'request.tools[0].function_d
[11:01:16] ⚠️ [FALLBACK] ⇄ ACC:xxx@gmail.com UNAVAILABLE (400) → NEXT ACCOUNT
[11:01:16] ⚠️ [AUTH] antigravity | all 1 accounts locked for gemini-3.5-flash-extra-low (reset after 3s) | lastError=[400]: { "error": { "code": 400, "messag
[11:01:16] ⚠️ [CHAT] [antigravity/gemini-3.5-flash-extra-low] [400]: { "error": { "code": 400, "message": "Invalid value at 'request.tools[0].function_declarations[39].parameters.properties[3].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[40].parameters.properties[3].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"", "status": "INVALID_ARGUMENT", "details": [ { "@type": "type.googleapis.com/google.rpc.BadRequest", "fieldViolations": [ { "field": "request.tools[0].function_declarations[39].parameters.properties[3].value", "description": "Invalid value at 'request.tools[0].function_declarations[39].parameters.properties[3].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"" }, { "field": "request.tools[0].function_declarations[40].parameters.properties[3].value", "description": "Invalid value at 'request.tools[0].function_declarations[40].parameters.properties[3].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"" } ] } ] } } (reset after 3s)
[11:01:16] ⚠️ [COMBO] Model ag/gemini-3.5-flash-extra-low failed, trying next {"status":400}
[11:01:16] ℹ️ [COMBO] Trying model 8/38: ag/gemini-3.5-flash-low
[...]

Issue #2489, opened 2026-07-09
Antigravity fails with 400 INVALID_ARGUMENT due to invalid function declaration schema


tuananhehou:
I’m getting a 400 INVALID_ARGUMENT error when using antigravity/gemini-pro-agent. The request appears to fail because several tool/function declaration parameter schemas contain invalid value fields with "object".

[antigravity/gemini-pro-agent] [400]: {
  "error": {
    "code": 400,
    "message": "Invalid value at 'request.tools[0].function_declarations[53].parameters.properties[6].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[54].parameters.properties[11].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[56].parameters.properties[19].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[57].parameters.properties[8].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[60].parameters.properties[42].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[61].parameters.properties[11].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[63].parameters.properties[53].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[65].parameters.properties[4].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"\nInvalid value at 'request.tools[0].function_declarations[75].parameters.properties[5].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\"",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.BadRequest",
        "fieldViolations": [
          {
            "field": "request.tools[0].function_declarations[53].parameters.properties[6].value",
            "description": "Invalid value at 'request.tools[0].function_declarations[53].parameters.properties[6].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), \"object\""
          },
          {
[...]

PR #2870, opened 2026-07-28
fix(gemini): resolve schema translation error for Vertex AI function calls

cokkyturnip:
### Vertex AI Gemini Function Declaration Schema Error (HTTP 400)

**Fixes/Related to Issues:** #2595, #2489

#### Overview
This PR resolves a critical issue where calls to Gemini/Vertex AI models (specifically via the `gemini-cli` provider and `Antigravity` models) fail with an `HTTP 400 INVALID_ARGUMENT` error.

#### Symptom
The upstream API (e.g., `cloudcode-pa.googleapis.com/v1internal:streamGenerateContent`) returns:
```
Invalid value at 'request.tools[0].function_declarations[XX].parameters.properties[X].value' (type.googleapis.com/google.cloud.aiplatform.master.Schema), "object"
```

#### Root Cause Analysis
1. **Orphaned Schemas after Keyword Removal:** 9router's schema translation pipeline (`cleanJSONSchemaForAntigravity`) removes unsupported JSON Schema keywords like `$defs` and `$ref`. If a property only contains a `$ref`, it becomes an empty object `{}`.
2. **Missing Type Inference for Empty Objects:** The internal `ensureObjectType` and `addPlaceholders` functions in 9router's Gemini translator only process schemas that already have a `type: "object"` field or have a `properties` field.
3. **The Resulting "Orphan" `{}`:** Empty objects `{}` resulting from `$ref` removal are left as-is. Vertex AI's protobuf JSON parser seems to interpret these empty or un-typed objects incorrectly in certain nested contexts (like Map entries), leading to the validation error.

#### Fix & Patch Implementation
We ensure that **any** empty object or un-typed object in the schema tree is explicitly promoted to an `object` type with a valid placeholder before being sent to the Gemini API.

Modified the `ensureObjectType` logic (Phase 4 of the cleaner) to catch empty objects:
- **Old Logic:** `obj.properties && !obj.type && (obj.type = "object")`
- **New Logic:** `!obj.type && (obj.properties || Object.keys(obj).length === 0) && (obj.type = "object")`

This ensures that:
1. Empty `{}` schemas are converted to `{ type: "object" }`.
2. `addPlaceholders` (Phase 5) then sees the `object` type and populates it with a mandatory `reason` property.
3. The final output is a fully qualified Schema object instead of an ambiguous empty one.

#### Verification
Verified using a test suite simulating various `$ref` removal scenarios (including deep nesting in arrays and complex objects). All test cases that previously produced empty `{}` schemas now produce valid, typed objects with placeholders, which are accepted by the Vertex AI Schema validation.

### DIFF
--- a/open-sse/translator/formats/gemini.js
+++ b/open-sse/translator/formats/gemini.js
@@ -353,6 +353,19 @@ export function cleanJSONSchemaForAntigravity(schema) {
   function addPlaceholders(obj) {
     if (!obj || typeof obj !== "object") return;
 
+    // Empty schema {} (no type, no properties) after $ref removal — treat as object with placeholder
+    if (Object.keys(obj).length === 0) {
+      obj.type = "object";
+      obj.properties = {
+        reason: {
+          type: "string",
+          description: "Brief explanation of why you are calling this tool"
+        }
+      };
+      obj.required = ["reason"];
+      return;
+    }
+
     if (obj.type === "object") {
       if (!obj.properties || Object.keys(obj.properties).length === 0) {
         obj.properties = {


decolua:

Issue #2767, opened 2026-07-22
Caveman/Ponytail breaks Cursor provider (empty content / OUT 0); need per-provider token-saver scope


Qinruiy:
## Summary

Enabling **Caveman** (and/or **Ponytail**) globally causes **Cursor OAuth provider** (`cu/*`) requests to return **HTTP 200 with empty assistant content** (`content: null`, `OUT 0`). The same savers work fine for Grok (`gcli/*` / `xai/*`).

There is currently **no per-provider switch** for Caveman/Ponytail — only global settings — so users who want Caveman for Grok cannot safely leave it on while using Cursor as an upstream.

## Environment

- Deploy: Docker (`decolua/9router:latest`)
- App version observed: `0.5.40` (from local `data.sqlite` meta)
- Cursor provider: OAuth token import (`authMethod: imported`, Pro subscription)
- Headroom optional sidecar present (not required to reproduce)

## Reproduction

1. Dashboard → enable **Caveman** (level `full`) and/or **Ponytail** (`full`).
2. Connect Cursor as provider (OAuth import or dashboard Connect Cursor).
3. Call OpenAI-compatible chat:

```bash
curl -sS http://localhost:20128/v1/chat/completions \
  -H "Authorization: Bearer $API_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "cu/claude-4.5-sonnet",
    "messages": [{"role": "user", "content": "Reply with exactly: pong"}],
    "max_tokens": 64,
    "stream": false
  }'
```

### Actual

- HTTP **200**
- Body like:
  ```json
  {
    "choices": [{
      "message": { "role": "assistant", "content": null },
      "finish_reason": "stop"
    }],
    "usage": { "completion_tokens": 0 }
  }
  ```
- Logs:
  ```text
  POST cu/claude-4.5-sonnet → cursor/... · JSON · 1 MSG · ACC:Cursor ...
  ⚙ CAVEMAN:full · PONYTAIL:full
  DONE ~300–700ms · OUT 0
  ```
- With tools/agents (many tools), same pattern: `STREAM · N MSG · many TOOL · CAVEMAN:full · OUT 0`.

### Expected

- Non-empty assistant content for Cursor models when Caveman/Ponytail are on, **or**
- Caveman/Ponytail should **skip** providers whose transport breaks with system-prompt injection (Cursor protobuf / `openai→cursor`), and/or allow **per-provider** enablement.

## Control tests

| Config | Model | Result |
[...]

PR #2878, opened 2026-07-28
fix(cursor): stop leaking agent tool errors as text

kylewelsby:
Follow-up to the secondary observation on #2767.

`cursor.js` acknowledges `ExecServerMessage.request_context_args` (field 10) with an empty context. Every other variant — the editor-backed tools 9router structurally cannot service — took an error path that emitted the message as an assistant **content delta**, then closed with `finish_reason: "stop"`.

Two consequences:

1. The user reads `[Cursor AgentService requested an unsupported IDE tool]` as the model's reply.
2. The turn is recorded as a success with a normal token count, so combo and account fallback never see a failure to route around.

Observed on 0.5.40, 4 of 22 cursor requests over a day. Intermittent — it depends on Cursor deciding the request warrants an IDE tool, so it only surfaces in agent mode (`tools.length > 0`), which Claude Code always triggers.

The error now emits an SSE error frame, matching the precedent in `kiro.js`.

**Second, latent bug fixed in the same path.** `decodeAgentFrames` processes every frame in a read batch. When an exec frame and a text frame arrive in one TCP chunk — likely, since `request_context_args` and the first text delta are adjacent — the trailing frame reached the already-closed controller and threw `TypeError: Invalid state: Controller is already closed`, replacing the response with a stream error. Guarded on `finished`. There is a regression test for this; it fails without the guard.

To be clear about scope: this does not make the turn recoverable. Failover in `base.js` happens at the HTTP response level, before the body streams, so a post-headers protocol failure cannot be routed around. What changes is that the failure is now visible to the client and to usage tracking rather than silently recorded as a success.

Tests: `tests/unit/cursor-agent-exec-request.test.js`, 4 cases, pure codec plus a stubbed h2 session, no network. Verified non-vacuous — reverting the source reproduces the exact string seen in production.

### DIFF
--- a/open-sse/executors/cursor.js
+++ b/open-sse/executors/cursor.js
@@ -12,7 +12,7 @@ import {
 import { buildCursorHeaders } from "../utils/cursorChecksum.js";
 import { estimateUsage } from "../utils/usageTracking.js";
 import { SSE_DONE, SSE_HEADERS } from "../utils/sseConstants.js";
-import { chatChunkSse } from "../utils/sse.js";
+import { chatChunkSse, sseChunk } from "../utils/sse.js";
 import { FORMATS } from "../translator/formats.js";
 import { proxyAwareFetch } from "../utils/proxyFetch.js";
 import zlib from "zlib";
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @kylewelsby for the contribution! Reviewed and merged into master. 🙏

PR #2923, opened 2026-07-30
fix(codex): normalize replayed call item IDs

Jaaay50:
## Problem

Codex Responses rejects replayed history when function or custom tool call items include item IDs.

## Fix

When forwarding stateless requests, remove the optional `id` from `function_call`, `function_call_output`, `custom_tool_call`, and `custom_tool_call_output`. Preserve `call_id` and all payload fields.

## Validation

Covers function and custom tool call/output pairs, long replay histories, and non-call item ID invariance.

### DIFF
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -55,13 +55,15 @@ function convertSystemToDeveloperRole(body) {
   }
 }
 
-// Strip server-generated item IDs (rs_/fc_/resp_/msg_) from input — avoids 404 with store=false
+// Strip invalid or stored item IDs before sending a store=false request.
 function stripStoredItemReferences(body) {
   if (!Array.isArray(body.input)) return;
   body.input = body.input.filter((item) => {
     if (typeof item === "string" && SERVER_ID_PATTERN.test(item)) return false;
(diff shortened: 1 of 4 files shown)


Jaaay50:
Rebased onto latest master (v0.5.50) and resolved the conflict in `open-sse/translator/schema/blocks.js` — kept both `ADDITIONAL_TOOLS` (from master) and `ITEM_REFERENCE` (from this PR).

Note on the overlap with master: the `stripStoredItemReferences` that landed upstream only strips IDs matching server-generated prefixes (`rs_/fc_/resp_/msg_`). This PR is still needed because it removes the optional `id` from `function_call` / `function_call_output` / `custom_tool_call` / `custom_tool_call_output` items regardless of prefix — Codex Responses rejects replayed stateless histories when those call items carry any item ID, and it also covers a wider set of stored-reference prefixes (`at_/amsg_/lsh_/tsc_/fco_/ctc_/ctco_/tso_/ws_/ig_/cmp_`).

Verified: PR's unit tests pass (9/9), and the full tests/unit suite shows zero new failures versus master (the 59 pre-existing failing files are identical on both branches).

Ready for review 🙏

PR #2946, opened 2026-07-31
fix(chat): replay Envoy buffer overflow on same account

ryanngit:
## Summary

- detect Envoy HTTP 507 request replay-buffer overflow precisely
- replay once on same provider connection before normal fallback handling
- avoid account lock/fallback when the same request-level error repeats

Envoy emits this response when an upstream retry is waiting while the incoming request exceeds its replay buffer. The selected account is not unhealthy, and cross-account fallback can break account-bound context.

## Verification

- `tests/unit/account-fallback-request-replay.test.js`
- `tests/unit/chat-request-replay.test.js`
- focused adjacent suite: 30/30 passed


ryanngit:
## v0.5.45 verification

- Base: `6fcd2733` (`v0.5.45`)
- Head: `872b06bd`
- Focused tests: request replay/account fallback - 3/3 passed
- Changed-file ESLint: passed
- `git diff --check`: passed
- Merge state before verification: CLEAN

PR #2966, opened 2026-07-31
fix(oauth): resolve ReferenceError searchParams is not defined in POST handler

alfep:
### Summary
This PR fixes a `ReferenceError: searchParams is not defined` in `POST /api/oauth/[provider]/[action]`.

### Details
In `src/app/api/oauth/[provider]/[action]/route.js`, under the `action === "register-session"` block in the `POST` handler, `searchParams.get("state")` was called without declaring `const searchParams = new URL(request.url).searchParams;`.

This caused a runtime `500 Internal Server Error` returning an HTML error response (`Internal Server Error...`), leading to `SyntaxError: Unexpected token 'I', "Internal S"... is not valid JSON` in the frontend UI modal when attempting to parse the response.

### Solution
Initialized `const searchParams = new URL(request.url).searchParams;` inside the `register-session` block.


decolua:
Thanks @alfep for the contribution! Reviewed and merged into master. 🙏

PR #2970, opened 2026-08-01
fix(translator): stop prompt cache being invalidated every turn

beprintvietnam:
## Problem

Anthropic matches a cached prefix in the order `tools` → `system` → `messages`, and only up to a breakpoint. Two things broke that invariant, so on most turns the **entire context was re-billed as `cache_creation`** instead of being read back.

Measured on production traffic before the fix — ~688k tokens re-written 29 seconds after the previous request, while the context had grown only 985 tokens:

```
cacheW=688,619  gap=29s  contextGrowth=985     totalIn=770,831
cacheW=687,090  gap=49s  contextGrowth=-1,529  totalIn=769,302
cacheW=681,795  gap=28s  contextGrowth=318     totalIn=764,007
```

Ruled out first: retry amplification (0 duplicate requests), TTL expiry (99% of requests were <5 min apart), and the `extended-cache-ttl` beta header (`ttl:"1h"` is GA).

## Cause 1 — the single breakpoint moved every turn

`prepareClaudeRequest` deleted all client `cache_control`, then re-added exactly **one** breakpoint on the last assistant message. That position advances each turn, so nothing was left at the previous position to read from:

```
turn N:   [system][tools][msg 1..40][BP @ 40]  → writes cache up to 40
turn N+1: [system][tools][msg 1..42][BP @ 42]  → no BP at 40 to read from
                                               → rewrites the whole prefix
```

Anthropic allows 4 breakpoints precisely so a **fixed** one can be kept at the old position to read from while a new one at the tail extends the cache.

**Fix:** cascade two breakpoints within the budget (`system` and `tools` take one each) — an anchor on the second-to-last assistant message, which is exactly where the previous turn put its tail, plus the tail itself. The anchor absorbs everything up to last turn, so only this turn's delta is written. Messages that cannot hold a breakpoint (empty, or thinking-only) are skipped, since `thinking` blocks reject `cache_control` and a wasted slot costs a full rewrite.

## Cause 2 — mid-conversation system messages were hoisted into `system`

The Messages API rejects `role:"system"` inside `messages`, so `normalizeClaudePassthrough` moved them into the top-level `system`. But `system` sits in front of every message in the cached prefix, and clients inject reminders mid-conversation — so each one grew `system` by a block and invalidated every cached message behind it.

Over 62 consecutive production request pairs, `system` grew 7 times. All 7 re-wrote ~133k tokens with cache reads collapsing to the system+tools floor (exactly 39,218), while the 53 turns with an unchanged `system` wrote 400–1,400:

| `system` | requests | cache write | cache read |
|---|---|---|---|
| grew | 7 | ~133,000 | 39,218 (floor) |
| unchanged | 53 | 400–1,400 | full prefix |

**Fix:** convert them to `role:"user"` in place, so growth only ever appends at the tail — which is what prompt caching is designed for. Two consecutive `user` messages are accepted by the API (verified against the live endpoint), and `fixToolUseOrdering` merges same-role neighbours on the translator path anyway.

Verified on live traffic after deploy: upstream `system` went from `array[21]` to a fixed `array[3]`.

## Result

Production traffic, same API key, priced with one fixed rate card so the model switch we happened to make cannot distort the comparison:

| | requests | wasteful rewrites | cache write/req | cache hit | $/1M ctx |
|---|---|---|---|---|---|
| before | 957 | **17.3%** | 44,736 | 81.6% | 1.558 |
| after | 68 | **0%** | 13,089 | 88.5% | **1.158 (−25%)** |

"Wasteful rewrite" = re-writing >20k tokens when the context grew only a few hundred and the 5-minute TTL was still alive. The remaining large writes after the fix are all new sessions or genuine TTL expiry.

⚠️ The "after" sample is 68 requests against 957 — the direction is consistent and the mechanism is confirmed on captured payloads, but the exact −25% figure deserves more data.

## Tests

Adds `tests/translator/cache-breakpoints.test.js` (15 tests): breakpoint cascade positions, the 4-breakpoint budget, thinking-block skipping, and prefix stability as reminders accumulate.

Full `translator/` + `unit/` suite: **1582 → 1597 passing, failure list unchanged** (86 pre-existing failures on a clean v0.5.45 checkout — verified the same set fails before and after with `comm -13` on sorted test names, so nothing new broke and nothing suspiciously "fixed itself").

The one existing test asserting the hoist behaviour is updated to assert in-place conversion. Its actual invariant — no `role:"system"` reaches the API inside `messages` — is unchanged.

## Not included

This branch is scoped to the cache fix. Unrelated local changes (cost guards, a `updateApiKey` column bug, a missing PXPIPE token-saver guard) are held back for separate PRs if you want them.

### DIFF
--- a/open-sse/translator/formats/claude.js
+++ b/open-sse/translator/formats/claude.js
@@ -22,6 +22,69 @@ export function hasValidContent(msg) {
   return false;
 }
 
+// Anthropic allows at most 4 cache breakpoints per request. system + tools take
+// one each, leaving two for messages.
+const MAX_MESSAGE_CACHE_BREAKPOINTS = 2;
+
+// thinking/redacted_thinking blocks cannot carry cache_control.
+function canHoldCacheControl(block) {
+  return block?.type !== CLAUDE_BLOCK.THINKING && block?.type !== CLAUDE_BLOCK.REDACTED_THINKING;
+}
+
+// True when a breakpoint can actually be attached to this message. Used to skip
+// messages that would silently waste one of the scarce breakpoint slots.
+function acceptsCacheBreakpoint(msg) {
+  return Array.isArray(msg?.content) && msg.content.some(canHoldCacheControl);
+}
+
+// Attach a cache breakpoint to the last cacheable block of a message.
+function setCacheBreakpoint(msg) {
+  if (!Array.isArray(msg?.content)) return false;
+  for (let j = msg.content.length - 1; j >= 0; j--) {
+    if (canHoldCacheControl(msg.content[j])) {
+      msg.content[j].cache_control = { type: "ephemeral" };
+      return true;
+    }
+  }
+  return false;
+}
+
+// Pick which assistant messages get a cache breakpoint.
+//
+// A single breakpoint on the last assistant message is worst-case for prompt
+// caching: Anthropic only reads cache up to a breakpoint, so when that
+// breakpoint advances each turn there is nothing left at the previous position
+// to read from — the whole prefix is re-written every turn. Observed in
+// production: ~688k cache-creation tokens re-written 29s after the prior
+// request, with the context only growing ~985 tokens.
+//
+// Fix: cascade two breakpoints, positioned relative to the END of the
+// conversation:
+//   anchor = second-to-last assistant → this is exactly where the PREVIOUS turn
+//            put its tail breakpoint, so the cache written last turn is read
+//            back in full.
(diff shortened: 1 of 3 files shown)


beprintvietnam:
Closing to re-open from my own account (@dzungprintway) — the fork this was pushed from belongs to a different account of mine. Same commit, identical tree hash, no content change. Superseded by the PR linked below.

PR #2974, opened 2026-08-01
fix: include complete API artifacts in CLI package

### LINKED ISSUE #2945: bug(api): 500 Internal Server Error across API endpoints due to missing Webpack chunks in v0.5.45
### Description
In `9router@0.5.45`, requests to API endpoints such as `/v1/chat/completions` for any AI provider return `HTTP 500 Internal Server Error`.

### Root Cause Analysis
1. **Missing Webpack Chunk Dependencies**: In `.next-cli-build`, route entry points like `server/app/api/v1/chat/completions/route.js` call `b.X(0, [<chunk_ids>])` to load dependent modules. The array omits required provider chunks (`8499` translator core, `3535` provider configs, `6832` OAuth registry, `1882` Grok/Kiro resolvers, `3610` provider definitions).
2. **Route-Isolated `machine-id` Module**: Module `19713` (`node-machine-id`) was compiled only into specific route files (such as `api/keys/route.js`) instead of a shared server chunk. When required by provider handlers in other routes, Webpack fails to resolve module `19713`, throwing:
   `TypeError: Cannot read properties of undefined (reading 'call')`

### Proposed Fix
1. Extract module `19713` into `server/chunks/19713.js`.
2. In `server/webpack-runtime.js`, dynamically register all chunk files in `server/chunks/` during initialization to guarantee all provider modules are available across all routes:
   ```javascript
   try {
     var fs = require('fs'), path = require('path');
     var cDir = path.join(__dirname, 'chunks');
     if (fs.existsSync(cDir)) {
       fs.readdirSync(cDir).forEach(function(f) {
         if (f.endsWith('.js')) try { h(require('./chunks/' + f)); } catch(e) {}
       });
     }
   } catch(e) {}
   ```
3. Update chunk dependency arrays `b.X(0, [...])` in API route entrypoints to include missing chunk IDs.

mvanhorn:
The published `9router@0.5.45` CLI contains an incomplete `.next-cli-build/server` tree: provider chunks needed by API routes are absent, and the compiled `/api/v1/messages` route is missing entirely. As a result, OpenAI-compatible and Anthropic-compatible requests can reach the packaged server but fail with `500` errors when Webpack cannot resolve a chunk or route module. The source routes already exist, so the defect is in `cli/scripts/build-cli.js`, where the CLI package is assembled from the pruned standalone output while only static and vendor assets are supplemented from the complete Next build. There are no prior or competing PRs in the supplied issue evidence.

## Summary

Extend the existing CLI build packaging stage to merge the complete generated server artifact tree from `.next-cli-build/server` into the package's matching `.next-cli-build/server` destination after the standalone copy, preserving Next's generated route layout and chunk relationships. Keep this logic in the production build path and make the filesystem operation testable with temporary fixture directories; do not patch generated route files, enumerate transient numeric chunk IDs, or eagerly load every chunk from `webpack-runtime.js`. Add a post-copy integrity check for the API route artifacts that the shipped CLI promises, including both chat completions and messages, so a future incomplete package fails during `pack:cli` rather than at runtime.

## Tests

- Given a standalone fixture missing the messages route and provider chunks while the complete build fixture contains them, packaging copies the full server tree so both `/api/v1/chat/completions` and `/api/v1/messages` compiled routes and their referenced chunks are present at the destination.
- Given complete standalone output, merging the server tree is idempotent and preserves existing generated files instead of deleting or relocating them.
- Given build output with a required API route artifact still absent, the integrity check fails with a path-specific error rather than producing a publishable but broken CLI package.
- Given the legacy and Next 16 nested standalone layouts already handled by the script, server artifacts land under the same `.next-cli-build/server` path used by the packaged server in either layout.

Fixes #2945

### DIFF
--- a/cli/scripts/build-cli.js
+++ b/cli/scripts/build-cli.js
@@ -81,201 +81,268 @@ function copyRecursive(src, dest) {
   }
 }
 
-console.log("📦 Building 9Router CLI package with Next.js...\n");
+function resolveStandaloneBuild(appDir, buildDistDir) {
+  const legacyStandaloneRoot = path.join(appDir, ".next", "standalone");
+  const resolvedStandaloneRoot = path.join(buildDistDir, "standalone");
+  let standaloneRoot = fs.existsSync(resolvedStandaloneRoot)
+    ? resolvedStandaloneRoot
+    : legacyStandaloneRoot;
 
-fs.mkdirSync(buildHomeDir, { recursive: true });
-fs.mkdirSync(path.join(buildHomeDir, "AppData", "Roaming"), { recursive: true });
-fs.mkdirSync(path.join(buildHomeDir, "AppData", "Local"), { recursive: true });
-
-// Step 0: Sync version from app/cli/package.json to app/package.json
-console.log("0️⃣  Syncing version to app/package.json...");
-const cliPkg = JSON.parse(fs.readFileSync(path.join(cliDir, "package.json"), "utf8"));
-const appPkgPath = path.join(appDir, "package.json");
-const appPkg = JSON.parse(fs.readFileSync(appPkgPath, "utf8"));
-if (appPkg.version !== cliPkg.version) {
-  appPkg.version = cliPkg.version;
-  fs.writeFileSync(appPkgPath, JSON.stringify(appPkg, null, 2) + "\n");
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @mvanhorn for the contribution! Reviewed and merged into master. 🙏

PR #2975, opened 2026-08-01
feat: Ollama Cloud quota tracker + proactive background OAuth refresh

b1nhm1nh:
## Summary

Two independent improvements, one branch.

### 1. Real Ollama Cloud quota tracker
The Ollama (`ollama`) usage handler was an informational stub claiming "no public quota API". That's incorrect — Ollama Cloud exposes:
- `GET https://ollama.com/api/usage` (Bearer apiKey) → `limits.session.usage` (5h) and `limits.weekly.usage` (7d) as a **0..1 ratio** (1.0 = limit reached), plus per-model request counts.
- `POST https://ollama.com/api/me` (Bearer apiKey) → account `Plan` (free / pro / max).

The handler now renders two quota bars — **Session (5h)** and **Weekly (7d)** — and the plan label.

Also fixes two latent gaps that kept apikey connections from working at all:
- The handler was bound to `connection.accessToken` (always empty for an apikey provider) → now reads `apiKey`.
- The `ollama` registry gained `features.usageApikey` so apikey connections pass the usage-route eligibility gate.

Verified live against a maxed account (weekly `usage=1` → **100% used / 0% remaining**) and a fresh one (`usage=0` → 0% used). The `usage` field is a 0..1 ratio, not a 0..100 percentage (caught and fixed in a follow-up commit).

Note: Ollama's API does **not** expose a reset timestamp (only a rolling 4-week `activity.period`), so the reset column shows `N/A` — this is accurate, not a bug. grok-cli's billing endpoint, by contrast, returns a real `currentPeriod.end` so its rows show a date.

### 2. Proactive background OAuth token refresh
OAuth access tokens (notably **grok-cli**'s ~6 h TTL) previously only refreshed in the last ~5 min of their TTL, and only when a request or the 60 s quota-poll landed in that window. Idle or closed-dashboard connections silently lapsed and surfaced as "authentication expired".

New `src/sse/services/backgroundTokenRefresh.js`: a server-side scheduler that refreshes OAuth connections within `max(provider refresh lead, 30 min)` of expiry, independently of inbound traffic.
- Ticks 10 s after boot, then every 5 min. `unref()`'d timers; `DISABLE_BACKGROUND_TOKEN_REFRESH` kill-switch.
- Idempotent start guard; fail-open per tick and per connection (one failure never aborts the batch).
- Runtime-guarded so it never starts during `next build` / static export / edge.
- Registered from `custom-server.js` (on `listening`) **and** `src/shared/services/initializeApp.js` (covers `next start` / standalone).
- `checkAndRefreshToken` gains an opt-in `{ force }` used only by the scheduler; the on-request path is unchanged.

The refresh method itself was already correct (public PKCE client, `grant_type=refresh_token`); this change is purely **when** refresh fires.

Verified live: ran the scheduler's real tick against the live DB — it selected grok-cli, force-refreshed via xAI, and persisted a new 6 h token (`expiresAt` rolled forward, `lastRefreshAt` stamped).

## Test coverage
- `tests/unit/ollama-usage.test.js` — 5 tests (ratio conversion, apiKey binding, 401, missing-key, parseQuotaData).
- `tests/unit/background-token-refresh.test.js` — 10 tests (selection logic, fail-open, mocked tick).

Full token-refresh + quota suite: 51/51 passing.

## Commits
- `fa7a57a` feat(providers): real Ollama Cloud quota tracker
- `ceb8fea` feat(token-refresh): proactive background OAuth refresh
- `7a550ba` fix(providers): ollama usage is a 0..1 ratio, not 0..100 percent


decolua:
Thanks @b1nhm1nh for the contribution! Reviewed and merged into master. 🙏

Issue #2969, opened 2026-08-01
Cloudflare (cloudflare-ai) provider card shows 'No connections' because registry entry is missing authModes


techysy:
## Bug

On the **Providers** grid page, the **Cloudflare** card always shows **"No connections" / 无连接**, even when a Cloudflare API-key connection exists and is active. Other free-tier providers (e.g. NVIDIA, api-airforce, bazaarlink, kilo-gateway) correctly show their connection count.

### Root cause

The provider catalog entry for `cloudflare-ai` in `open-sse/providers/registry/cloudflare-ai.js` is **missing `authModes`** (and `authType`). Working free-tier providers all declare `authType: "apikey", authModes: ["apikey"]` (e.g. `open-sse/providers/registry/nvidia.js`).

The grid page computes each card's connection count via `dualAuthTypes(info, key)` in `src/app/(dashboard)/dashboard/providers/page.js`:

```js
const dualAuthTypes = (info, key) => {
  if (key === "kiro") return ["oauth", "apikey", "api_key"];
  const modes = info?.authModes;
  if (!Array.isArray(modes) || !modes.includes("apikey")) return "oauth";  // <-- Cloudflare lands here
  return ["oauth", "apikey", "api_key"];
};
```

Because `cloudflare-ai` has no `authModes`, this returns `"oauth"`, so the `apikey`-typed connection is filtered out of the card's connected count → the card shows "No connections" even though the detail page correctly lists the active connection.

### Suggested fix

Add the missing fields to `open-sse/providers/registry/cloudflare-ai.js`:

```js
category: "freeTier",
authType: "apikey",
authModes: ["apikey"],
hasProviderSpecificData: true,
```

### Environment
- 9Router v0.5.45
- Verified the connection is active in `providerConnections` (isActive=1, testStatus="active") and successfully routes `cf/@cf/...` requests; only the grid card display is wrong.

techysy:
**对比图 / Before & After**
修复前 (Before) - 卡片显示"无连接":

<img width="2430" height="1437" alt="Image" src="https://github.com/user-attachments/assets/984474cd-4824-4fd4-bf7c-68f2d98a91a9" />

<img width="2229" height="1341" alt="Image" src="https://github.com/user-attachments/assets/46fbff65-81c7-4a8e-bbcd-b659462df51e" />

修复后 (After) - 卡片显示"1已连接":

[...]

PR #2993, opened 2026-08-02
fix: freeTier/apikey providers without authModes default to apikey in dualAuthTypes

techysy:
## Summary

Fix provider cards showing "No connections" on the **Providers** grid for free-tier providers whose registry entries omit `authModes`.

## Problem

`dualAuthTypes()` in `src/app/(dashboard)/dashboard/providers/page.js` returns `"oauth"` for any provider without `authModes`:

```js
if (!Array.isArray(modes) || !modes.includes("apikey")) return "oauth";
```

Several `category: "freeTier"` providers omit `authType`/`authModes` in the registry (e.g. `cloudflare-ai`, `byteplus`, `ollama`, `vertex`). Their `apikey` connections are then filtered out of the card's connected count → the card shows **"No connections"** even though the provider detail page correctly lists the active connection, and the connection routes fine.

Affected: `cloudflare-ai`, `byteplus`, `ollama`, `vertex` (all `category: "freeTier"`). `category: "apikey"` entries were unaffected because they're counted via the dedicated apikey section.

## Fix

When `authModes` is absent, default to supporting `apikey` for providers that belong to `FREE_TIER_PROVIDERS` or `APIKEY_PROVIDERS`:

```js
if (!Array.isArray(modes)) {
  return key in FREE_TIER_PROVIDERS || key in APIKEY_PROVIDERS
    ? ["oauth", "apikey", "api_key"]
    : "oauth";
}
```

This fixes all affected free-tier cards at once, keeps oauth-only providers (e.g. `kimchi`, which declares `authModes:["oauth"]`) unchanged, and preserves existing behavior for providers that do declare `authModes`.

## Related

- Issue #2969 (original Cloudflare report)

### DIFF
--- a/src/app/(dashboard)/dashboard/providers/page.js
+++ b/src/app/(dashboard)/dashboard/providers/page.js
@@ -283,7 +283,15 @@ export default function ProvidersPage() {
   const dualAuthTypes = (info, key) => {
     if (key === "kiro") return ["oauth", "apikey", "api_key"];
     const modes = info?.authModes;
-    if (!Array.isArray(modes) || !modes.includes("apikey")) return "oauth";
+    // Free-tier and API-key providers default to supporting apikey even when the
+    // registry entry omits authModes (e.g. cloudflare-ai, byteplus, ollama,
+    // vertex) — otherwise their apikey connections are invisible on the grid card.
+    if (!Array.isArray(modes)) {
+      return key in FREE_TIER_PROVIDERS || key in APIKEY_PROVIDERS
+        ? ["oauth", "apikey", "api_key"]
+        : "oauth";
+    }
+    if (!modes.includes("apikey")) return "oauth";
     return ["oauth", "apikey", "api_key"];
   };
 


techysy:
**环境信息 / Environment (for review)**

This was reproduced and verified in a **fnOS (飞牛 NAS) packaged 9Router app**:

- **fnOS**: Debian 12 (bookworm) based, kernel `6.18.18.c938-trim` (x86_64)
- **9Router**: `9router-app` **v0.5.45** (standalone Next.js server)
- **Deployment**: installed via fnOS App Center as a third-party app, running under the app's own system user on port 20128
- **Access**: Providers grid at `/dashboard/providers`, connections stored in SQLite (`providerConnections`)

**Verification**:
- The affected connections are present and active (`providerConnections.isActive=1`, `testStatus="active"`) and route requests successfully (e.g. `cf/@cf/...`).
- After applying the fix, the affected cards (`cloudflare-ai`, `byteplus`, `ollama`, `vertex`) correctly show their apikey connection count instead of "No connections".

Feel free to ping me if you need the same fix applied to the built bundle or any additional repro details.

techysy:
**补充 / Related upstream fix**

Heads-up: upstream commit `1319dea` ("fix(providers): count apikey connections for freeTier providers", Jul 29) already addresses this class of bug — but with the **per-provider** approach, and it only covers **openrouter, nvidia, gemini**. It does **not** cover the other affected `freeTier` providers that still omit `authModes`:

- `cloudflare-ai`
- `byteplus`
- `ollama`
- `vertex`

Those still hit the exact symptom (active apikey connection, but the grid card shows "No connections") until their registry entries gain `authType:"apikey", authModes:["apikey"]` — which is what my runtime patch did to verify the fix.

This PR takes the **systemic** route instead: default `dualAuthTypes()` to apikey for any provider in `FREE_TIER_PROVIDERS`/`APIKEY_PROVIDERS` that lacks `authModes`. That fixes all present **and future** free-tier providers in one place, without needing to remember to add `authModes` to every new registry entry. It's complementary to (and strictly more robust than) 1319dea, and touches a different file, so there's no conflict.

Either approach is fine — if you prefer to keep the per-provider style, I can add the missing `authModes` to `cloudflare-ai`, `byteplus`, `ollama`, `vertex` instead. Let me know which you'd like.

decolua:
Thanks @techysy for the contribution! Reviewed and merged into master. 🙏

Issue #2098, opened 2026-06-25
[Bug] Compatible provider (openai/anthropic) custom models not visible in combo model selector


fogetpassw-art:
## Description

Custom models added via the "Add Model" button on an OpenAI-compatible or Anthropic-compatible provider detail page do **not** appear in the combo creation/edit model selector, even though they are correctly saved and work in chat.

## Root Cause

In `ModelSelectModal.js`, the `groupedModels` logic has three branches for different provider types:

| Branch | Condition | Model Sources |
|--------|-----------|---------------|
| 1 | `passthroughModels` | modelAliases ✅ + customModels ✅ |
| 2 | **openai/anthropic-compatible** (UUID starts with `openai-compatible-` / `anthropic-compatible-`) | **modelAliases only** ❌ |
| 3 | Other (standard built-in providers) | built-in models ✅ + modelAliases ✅ + customModels ✅ |

Commit `707a915` ("fix(models): store provider custom models by provider scope", Jun 20) changed the "Add Model" button storage mechanism from `modelAliases` to `customModels` in the provider detail page (`CompatibleModelsSection.js`, `PassthroughModelsSection.js`, `provider/[id]/page.js`). However, **Branch 2 of `ModelSelectModal.js` was never updated to read `customModels`** — it only reads `modelAliases`.

This means:
- Models added **before** the change (stored in `modelAliases`) still appear ✅
- Models added **after** the change (stored in `customModels`) do not appear ❌

## Fix

Add `customModels` reading logic to Branch 2 in `ModelSelectModal.js`, mirroring the pattern used in Branch 3:

```javascript
// Read customModels for this provider (filtered by providerId/UUID)
const customModelsForProvider = customModels
  .filter((m) => m.providerAlias === providerId)
  .map((m) => ({ 
    id: m.id, 
    name: m.name || m.id, 
    value: `${nodePrefix}/${m.id}`, 
    isCustom: true 
  }));

// Merge: custom models first, then aliases (deduped by value)
const merged = customModelsForProvider.length > 0 || nodeModels.length > 0
  ? [...customModelsForProvider, ...nodeModels.filter(m => !customModelsForProvider.some(cm => cm.value === m.value))]
  : [];

// Use merged list for display and hasModels check
```

## Affected Versions

Since v0.5.8 (commit `707a915` was included in this release).

## Related Issues

- #2028 Custom provider models cannot be selected directly from the dropdown
- #1993 model-id not detected when using custom provider (anthropic and openai compatible)
[...]

Issue #2028, opened 2026-06-23
[Bug] Custom provider models cannot be selected directly from the dropdown; require manual input of the model ID


2doright:
**Description:**  
Custom models from providers (e.g., `wf/gpt-5.5`) are displayed only as raw ID strings (e.g., `wf/model-id`) in the model selection list. Unlike built-in models, they cannot be selected with a single click. Users are forced to manually click and type the full model ID to use them in features like model combination or terminal tool configuration.  
As shown in the UI screenshots：

<img width="1867" height="595" alt="Image" src="https://github.com/user-attachments/assets/4a50959c-dc6f-4877-af7f-1976b0f5adaf" />

<img width="828" height="755" alt="Image" src="https://github.com/user-attachments/assets/7180d032-f02c-4bd5-8712-968ff08f4701" />

**Expected behavior:**  
Custom models should be directly selectable as clickable options from the dropdown menu, eliminating the need for manual input.  

**Environment:**  
9router v0.5.8

Skuwwy:
+1 this is frustrating when the custom models list is quite lengthy.

Also sometimes, I am unable to click on the button to enter manually, it just doesn't work.

iklilzaki177:
yes it seems bug on v0.5.8

Issue #2013, opened 2026-06-23
New Custom model can't add to combo


edhyrg2:
Custom providers created via the UI are successfully saved to the database but do not appear in the provider dropdown/combo when creating or editing connections.

Issue #1993, opened 2026-06-22
[BUG] model-id not detected when using custom provider (anthropic and openai compatible)


jmahiswara1:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ee0781a4-2574-4c5d-b596-95f2ea12cf1b" />

<img width="507" height="212" alt="Image" src="https://github.com/user-attachments/assets/748f5446-155b-454f-9103-f676607eb445" />

4dityo:
same with me

OctaLectzz:
same broo, last update 9router yesterday

onizukashonan14-png:
also only one api key we can add in these custom endpoints. do you guys face same problem?

PR #2995, opened 2026-08-02
fix(cli-tools): enable Apply button for dynamic OpenAI/Anthropic-compatible providers

### LINKED ISSUE #2994: [Bug] CLI Tools Apply button permanently disabled for OpenAI/Anthropic-compatible providers
## Description

On the CLI Tools detail page (`/dashboard/cli-tools/claude`, `/codex`, etc.) the **Apply** button is permanently grayed out as soon as the only active provider connection is a **dynamically-registered OpenAI-compatible or Anthropic-compatible provider** (id looks like `openai-compatible-chat-<uuid>` / `anthropic-compatible-chat-<uuid>`).

Routing through 9router keeps working fine — the connection's `testStatus` is `active` and requests flow through. Only the dashboard blocks the user from clicking **Apply**, which is the button that actually writes `~/.claude/settings.json` (`ANTHROPIC_BASE_URL`, `ANTHROPIC_AUTH_TOKEN`, default models, Exa MCP).

## Steps to reproduce

1. Add a new provider connection of type "OpenAI Compatible" in the dashboard. The connection gets an auto-generated provider id like `openai-compatible-chat-d0cd250f-228d-437e-a813-6cd8391f3f0e`. Configure `baseUrl`, `apiKey`, `defaultModel`, verify `Test Connection` shows **active**.
2. Remove / deactivate every other provider connection.
3. Go to **Dashboard → CLI Tools → Claude Code**.
4. Observe that **Apply** is grayed out and **Select Model** returns no entries.

Expected: Apply is enabled, since the connection is healthy and routing works.
Actual: Apply stays disabled (`disabled={!hasActiveProviders}`).

## Root cause

`src/app/(dashboard)/dashboard/cli-tools/[toolId]/ToolDetailClient.js`:

```js
const getAllAvailableModels = () => {
  const activeProviders = getActiveProviders();
  const models = [];
  activeProviders.forEach(conn => {
    const alias = PROVIDER_ID_TO_ALIAS[conn.provider] || conn.provider;
    const providerModels = getModelsByProviderId(conn.provider);   // <-- static catalog lookup
    providerModels.forEach(m => models.push({ value: `${alias}/${m.id}`, ... }));
  });
  return models;
};
const availableModels = getAllAvailableModels();
const hasActiveProviders = availableModels.length > 0;             // <-- gates the Apply button
```

`getModelsByProviderId` (open-sse/config/providerModels.js) only looks in the static `PROVIDER_MODELS` registry, which has no key for the random UUID-style ids used by compatible providers, so it returns `[]`.

Consequence: `availableModels.length === 0` ⇒ `hasActiveProviders === false` ⇒ `<Apply disabled>` in `ClaudeToolCard.js` (and all sibling tool cards that take `hasActiveProviders`).

The actual routing layer never consults this catalog for compatible providers — it uses the connection's own `defaultModel`, `providerSpecificData.prefix`, and the request body's model name — which is why 9router keeps forwarding successfully even while the button is unusable.

## Environment

- 9router v0.5.45 (macOS, npm install)
- Single provider connection: `openai-compatible-chat-<uuid>`, `testStatus: active`, `defaultModel: Kimi-K3`
- Claude Code `~/.claude/settings.json` already contains `ANTHROPIC_BASE_URL=http://127.0.0.1:20128/v1` and works — I just can't re-Apply because the button is disabled.

## Related issues

Same class of bug (dynamic provider ids aren't honored by the model catalog) in other UI surfaces:

- #2028 Custom provider models cannot be selected directly from the dropdown
- #2098 Compatible provider custom models not visible in combo model selector
- #1990 / #1993 / #2013 — model-id not detected on OpenAI/Anthropic compatible nodes

#2098 was fixed on master via `src/shared/components/ModelSelectModal.js` (combo selector merges `modelAliases` + `customModels`). The CLI-tools Apply button hasn't received the same treatment, so this is a fresh report, not a duplicate.

## Proposed fix

Fall back to the connection's own models when the static catalog returns nothing. I'm preparing a PR that mirrors the ModelSelectModal pattern:

- if `providerModels.length === 0`, use `conn.defaultModel`
- then `conn.providerSpecificData?.customModels`
- then, when `conn.testStatus === "active"`, a `${prefix}/model-id` placeholder so Apply isn't permanently disabled

This restores the same UX combo creation already has for compatible providers, without changing behavior for built-in providers (Claude, Groq, OpenRouter, etc. — they continue to use the static catalog).

lazysaltyfish:
## Description

On the CLI Tools detail page (`/dashboard/cli-tools/claude`, `/codex`, …) the **Apply** button is permanently grayed out whenever the only active provider connection is a dynamically-registered OpenAI-compatible / Anthropic-compatible provider (id like `openai-compatible-chat-<uuid>`).

Routing keeps working fine — the connection's `testStatus` is `active` and traffic flows through 9router. Only the dashboard blocks the user from writing `~/.claude/settings.json`, `~/.codex/auth.json`, etc. via the **Apply** button.

## Root Cause

`src/app/(dashboard)/dashboard/cli-tools/[toolId]/ToolDetailClient.js` gates Apply on `hasActiveProviders`, which is derived from `getAllAvailableModels()`. That helper only consults the static `PROVIDER_MODELS` registry via `getModelsByProviderId(conn.provider)`:

| Provider id | In `PROVIDER_MODELS`? | Catalog result | `hasActiveProviders` |
|---|:-:|:-:|:-:|
| `claude`, `groq`, `openrouter`, … | ✅ | model list | `true` (correct) |
| `openai-compatible-chat-<uuid>` | ❌ (random UUID) | `[]` | `false` (wrong) |
| `anthropic-compatible-chat-<uuid>` | ❌ | `[]` | `false` (wrong) |

The actual routing path never reads this catalog for compatible providers — it forwards using the connection's own `defaultModel`, `providerSpecificData.prefix`, and the request body's model name. So 9router happily serves traffic while the UI keeps Apply disabled.

This is the same catalog-miss pattern as #2098 (combo page), which was fixed in `ModelSelectModal.js` by falling back to `modelAliases` + `customModels`. The CLI-tools pages never received the equivalent treatment.

## Changes

In `getAllAvailableModels()` (ToolDetailClient.js), when the static catalog returns no models for a connection, fall back in this order:

1. `conn.defaultModel` — set when the user clicked "Test Connection" on the compatible provider, always present in practice.
2. `conn.providerSpecificData?.customModels` — models registered via the provider detail page's **Add Model** flow (same source `ModelSelectModal` reads).
3. `${prefix}/model-id` placeholder when `conn.testStatus === "active"` — last-resort escape hatch so the button isn't permanently bricked even if the user hasn't registered any model yet. Mirrors the placeholder already used in `ModelSelectModal.js`.

The model value is built with `providerSpecificData.prefix` (e.g. `ixgbe/Kimi-K3`) instead of the raw provider UUID, matching how the router actually routes for these connections — same convention as `ModelSelectModal`.

Built-in providers are untouched: the fallback branch only executes when `getModelsByProviderId()` returned `[]`, and for built-ins that branch never fires.

## Related Issues

- Closes #2994
- Related: #2028, #2098, #1990, #1993, #2013

## Testing

- [x] Reproduced the bug locally on 9router v0.5.45 (macOS): single active connection of type `openai-compatible-chat-d0cd250f-…` with `defaultModel: Kimi-K3`, Apply button disabled.
- [x] Simulated the patched `getAllAvailableModels()` against the real connection record from `~/.9router/db/data.sqlite` (`defaultModel: Kimi-K3`, `providerSpecificData.prefix: ixgbe`): fallback produces `["ixgbe/Kimi-K3"]`, so `hasActiveProviders` flips to `true`.
- [x] Built-in providers unaffected: for `providerModels.length > 0` the new branch is skipped entirely, so Claude / Groq / OpenRouter behavior is unchanged.

### DIFF
--- a/src/app/(dashboard)/dashboard/cli-tools/[toolId]/ToolDetailClient.js
+++ b/src/app/(dashboard)/dashboard/cli-tools/[toolId]/ToolDetailClient.js
@@ -81,6 +81,33 @@ export default function ToolDetailClient({ toolId, machineId }) {
           models.push({ value: modelValue, label: `${alias}/${m.id}`, provider: conn.provider, alias, connectionName: conn.name, modelId: m.id });
         }
       });
+
+      // openai/anthropic-compatible providers are registered with a random UUID (e.g.
+      // "openai-compatible-chat-<uuid>") that has no entry in the static PROVIDER_MODELS
+      // catalog, so `getModelsByProviderId` returns []. Routing still works because the
+      // request path uses the connection's own model config, but `hasActiveProviders`
+      // below would flip to false and disable the Apply button. Fall back to the
+      // connection's own models so these providers are usable from CLI tool pages.
+      if (providerModels.length === 0) {
+        const prefix = conn.providerSpecificData?.prefix || alias;
+        const fallbackModels = [];
+        if (conn.defaultModel) fallbackModels.push({ id: conn.defaultModel, name: conn.defaultModel });
+        (conn.providerSpecificData?.customModels || []).forEach(m => {
+          if (m?.id && !fallbackModels.some(f => f.id === m.id)) fallbackModels.push({ id: m.id, name: m.name || m.id });
+        });
+        if (fallbackModels.length === 0 && conn.testStatus === "active") {
+          // Provider is confirmed reachable but exposes no model info anywhere;
+          // still let the user apply so they aren't stuck on a permanently disabled button.
+          fallbackModels.push({ id: "model-id", name: `${prefix}/model-id` });
+        }
+        fallbackModels.forEach(m => {
+          const modelValue = `${prefix}/${m.id}`;
+          if (!seenModels.has(modelValue)) {
+            seenModels.add(modelValue);
+            models.push({ value: modelValue, label: `${prefix}/${m.id}`, provider: conn.provider, alias: prefix, connectionName: conn.name, modelId: m.id });
+          }
+        });
+      }
     });
     return models;
   };


decolua:
Thanks @lazysaltyfish for the contribution! Reviewed and merged into master. 🙏

PR #3002, opened 2026-08-03
fix(server): support IntelliJ IDEA OpenAI clients over HTTP

dajinglingpake:
## Summary

- handle JBR 25 `Upgrade: h2c` requests before the Next.js upgrade handler closes the socket
- replay IntelliJ IDEA OpenAI-compatible GET and content-length POST requests through the existing HTTP/1.1 handler
- preserve streaming chat request bodies and add a regression test for `/v1/chat/completions` SSE

## Problem

IntelliJ IDEA 2026.2 third-party OpenAI integrations use Koog/Ktor on JBR 25. When the configured 9Router base URL uses plain HTTP, the JDK HTTP client sends an h2c upgrade request. The existing upgrade path closes that connection without an HTTP response, so model discovery and chat fail with:

```text
HTTP/1.1 header parser received no bytes
EOF reached while reading
```

This change handles only h2c upgrades, strips the upgrade headers, and replays the request through the existing HTTP/1.1 handler. Other upgrade protocols keep their current behavior.

## Test

- `node --check custom-server.js`
- `node --test tests/unit/custom-server-h2c.test.cjs`

### DIFF
--- a/custom-server.js
+++ b/custom-server.js
@@ -26,7 +26,54 @@ http.createServer = (...args) => {
     if (viaProxy) req.headers["x-9r-via-proxy"] = "1";
     return handler(req, res);
   };
-  return origCreate(...rest, wrapped);
+  const server = origCreate(...rest, wrapped);
+  const origEmit = server.emit;
+  // JBR 25 sends h2c upgrades that the HTTP/1.1 server would otherwise close.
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @dajinglingpake for the contribution! Reviewed and merged into master. 🙏

PR #3005, opened 2026-08-03
fix(auth): redirect active sessions from /login

### LINKED ISSUE #3004: [Bug] /login does not redirect users with an active dashboard session
## Problem

Opening `/login` while a valid dashboard session is still present shows the password form instead of redirecting to `/dashboard`.

The login page calls `/api/auth/status`, but currently only redirects when `requireLogin === false`. The status endpoint validates the `auth_token` cookie, but the login page does not use that result.

## Expected behavior

If the `auth_token` cookie is valid, visiting `/login` should navigate to `/dashboard` without showing the login form. An invalid or expired token should continue to show the login page.

## Reproduction

1. Sign in to the dashboard.
2. Navigate directly to `/login` in the same browser session.
3. Observe that the password form is displayed even though the dashboard session is still active.

## Relevant code

- `src/app/login/page.js` only checks `data.requireLogin === false`.
- `src/app/api/auth/status/route.js` already validates `auth_token` through `getDashboardAuthSession`, but does not expose whether the session is authenticated.

## Proposed fix

- Return an explicit `authenticated` boolean from `/api/auth/status`.
- Redirect from `/login` when `authenticated === true` or when login is disabled.
- Add a regression test covering valid and invalid sessions.

DaDecky:
## Summary

- expose whether `/api/auth/status` has a valid dashboard session
- redirect authenticated users from `/login` to `/dashboard`
- keep invalid or expired sessions on the login form
- add regression coverage for valid sessions, invalid sessions, and status failures

Fixes #3004

## Validation

- `npx vitest run unit/auth-status.test.js --config vitest.config.js`
- `npx vitest run unit/auth-status.test.js unit/dashboard-guard.test.js --config vitest.config.js`
- `npx eslint src/app/login/page.js src/app/api/auth/status/route.js tests/unit/auth-status.test.js`
- `npm run build`


decolua:
Thanks @DaDecky for the contribution! Reviewed and merged into master. 🙏

Issue #2942, opened 2026-07-31
bug(ui): CSS bundle 404 error causes unstyled dashboard in v0.5.45 (missing cc29e6f232c039ff.css)


alfep:
### Description
In v0.5.45, accessing the dashboard UI (http://localhost:20128/dashboard or /login) results in a broken, unstyled HTML page across all browsers.

### Root Cause Analysis
When the server renders the Next.js HTML page, the <head> section requests:
\/_next/static/css/cc29e6f232c039ff.css\`n
However, in the published package bundle (\.next-cli-build/static/css\), the main stylesheet asset is generated with a different hash filename (\17f873d769dc530.css\). Because \cc29e6f232c039ff.css\ does not exist in \.next-cli-build/static/css/\, the server returns a \404 Not Found\ for the CSS asset, causing the browser to render unstyled raw HTML.

### Steps to Reproduce
1. Install \9router@0.5.45\ (\
pm i -g 9router@latest\ or \un add -g 9router@latest\)
2. Run \9router\\
3. Open \http://localhost:20128/dashboard\ or \http://localhost:20128/login\\
4. Inspect Network tab in DevTools: \cc29e6f232c039ff.css\ returns 404 Not Found.

### Workaround / Temporary Fix
Copy/duplicate \.next-cli-build/static/css/a17f873d769dc530.css\ to \cc29e6f232c039ff.css\ inside the global \
ode_modules/9router/app/.next-cli-build/static/css/\ directory.

### Environment
- Version: \9router@0.5.45\\
- OS: Windows 11 / Node 20 / Bun 1.3

PR #3007, opened 2026-08-03
fix(build): include assets in standalone output

### LINKED ISSUE #3006: bug(build): standalone source build omits static and public assets
## Description

A source checkout built with Next.js `output: "standalone"` can serve the rendered HTML and API routes, but the standalone server does not serve the browser assets required by `/login` and `/dashboard`.

When PM2 runs `.next/standalone/server.js`, the generated HTML references `/_next/static/...` files, but those files are only present under the root `.next/static` directory. The standalone directory does not contain `.next/standalone/.next/static` or `.next/standalone/public`, so the browser receives 404 responses for JavaScript, CSS, fonts, and favicon files. React then remains on the login page's `Loading...` state because the client bundle cannot hydrate.

## Steps to reproduce

1. Clone the repository and install dependencies.
2. Run `npm run build`.
3. Start `.next/standalone/server.js` through PM2 with the production port.
4. Open `/login` and inspect the browser Network tab.
5. Observe 404 responses for `/_next/static/...` and `/favicon.svg`.

## Expected behavior

A source standalone build should be directly runnable from `.next/standalone/server.js`, or the build command should clearly perform the required asset assembly.

## Proposed direction

Copy the generated static and public assets into the standalone runtime directory after a normal source build:

```text
.next/static -> .next/standalone/.next/static
public       -> .next/standalone/public
```

The CLI packaging build uses a separate `NEXT_DIST_DIR` and already has its own asset-copying stage, so the source-build fix should not duplicate or interfere with `cli:pack`.

Related to #2942, which reports a different but related static-asset artifact problem in the published CLI package.

## Environment

- 9Router v0.5.45
- Next.js standalone output
- PM2
- Linux / Node.js 22

DaDecky:
## Summary

Fix source deployments that run the Next.js standalone server directly.

With `output: "standalone"`, `next build` creates `.next/standalone/server.js`, but the generated browser assets remain in the root `.next/static` and `public` directories. Starting the standalone server from PM2 therefore returns 404 for JavaScript, CSS, fonts, and favicon files. The HTML renders, but the client cannot hydrate and `/login` remains stuck on `Loading...`.

This adds a post-build asset assembly step that copies:

```text
.next/static -> .next/standalone/.next/static
public       -> .next/standalone/public
```

The script supports `NEXT_DIST_DIR` and explicitly skips workspace-traced CLI builds because `cli/scripts/build-cli.js` already owns the CLI packaging flow and copies its assets separately.

## Changes

- Add `postbuild` and `postbuild:bun` asset-copy lifecycle scripts.
- Add an idempotent standalone asset copy helper.
- Add tests for the default dist directory, custom dist directory, and CLI-build guard.

## Verification

- `npm run build`
- `npx vitest run unit/standalone-assets.test.js`
- `npx eslint scripts/copy-standalone-assets.mjs tests/unit/standalone-assets.test.js`
- `git diff --check`

Fixes #3006
Related to #2942

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,8 @@
     "dev": "next dev --port 20127",
     "dev:webpack": "next dev --webpack --port 20127",
     "build": "next build --webpack",
+    "postbuild": "node scripts/copy-standalone-assets.mjs",
+    "postbuild:bun": "node scripts/copy-standalone-assets.mjs",
     "start": "next start --port 20127",
     "dev:bun": "bun --bun next dev --webpack --port 20127",
     "build:bun": "bun --bun next build --webpack",
--- a/scripts/copy-standalone-assets.mjs
+++ b/scripts/copy-standalone-assets.mjs
@@ -1,0 +1,36 @@
+import { cpSync, existsSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { dirname, resolve } from "node:path";
+
+export function copyStandaloneAssets({ projectRoot = process.cwd(), distDir = process.env.NEXT_DIST_DIR || ".next" } = {}) {
+  if (process.env.NEXT_TRACING_ROOT_MODE === "workspace") {
+    console.log("[standalone-assets] Skipping workspace-traced CLI build; CLI packaging handles assets");
+    return;
+  }
+
+  const buildDir = resolve(projectRoot, distDir);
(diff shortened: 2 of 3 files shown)


decolua:
Thanks @DaDecky for the contribution! Reviewed and merged into master. 🙏

PR #3009, opened 2026-08-03
fix(providers): count apikey connections for ollama freeTier provider

cokkyturnip:
## Problem

On the `/dashboard/providers` page, the **Ollama Cloud** card always shows **"No Connections"** even when API-key connections exist and are active. The provider detail page (`/dashboard/providers/ollama`) correctly lists them, so the card stats were wrong.

## Root Cause

The dashboard card stats are computed via `dualAuthTypes()` in `src/app/(dashboard)/dashboard/providers/page.js`:

```js
const dualAuthTypes = (info, key) => {
  if (key === "kiro") return ["oauth", "apikey", "api_key"];
  const modes = info?.authModes;
  if (!Array.isArray(modes) || !modes.includes("apikey")) return "oauth";
  return ["oauth", "apikey", "api_key"];
};
```

Since the ollama registry entry (`open-sse/providers/registry/ollama.js`) lacked `authModes`, `dualAuthTypes` fell back to `"oauth"`, so only `authType: "oauth"` connections were counted — the ollama API-key connections were invisible on the card.

## Fix

Add `authType: "apikey"` and `authModes: ["apikey"]` to the ollama freeTier registry entry — the same pattern already applied upstream to `gemini.js`, `nvidia.js`, and `openrouter.js` (commit 1319dea6).

## Verification

- Rebuilt the app from source; the served catalog chunk now includes `authModes:["apikey"]` for ollama.
- Toggling the card off sets both ollama connections `isActive: false`; toggling on restores `isActive: true` (verified via `/api/providers`).


decolua:

PR #3011, opened 2026-08-03
fix(oauth): keep `open` external so xAI/Grok token refresh works on Windows

Tomauskasz:
## Summary

On Windows, **xAI/Grok token refresh has never worked**. Not "sometimes fails" — the code
throws before it opens a socket, on every install, every time. The refresh token is never
sent to `auth.x.ai`, so a 6h access token simply expires and the user is told to
re-authorize. Daily.

One line fixes it: `open` must not be bundled.

## Root cause

`open` derives its own directory from `import.meta.url`:

```js
const __dirname = path.dirname(fileURLToPath(import.meta.url));
const xdgOpenPath = path.join(__dirname, 'xdg-open');
```

That is **module scope**, and webpack replaces `import.meta.url` with the absolute path of
the *build* machine as a string literal. I built current `master` and confirmed it:

```
.next/server/chunks/5852.js
  file:///C:/…/9router-upstream/node_modules/open/index.js
```

A Windows build bakes a drive letter, so it works. **A release built on macOS bakes
`file:///Users/…`, and `fileURLToPath` rejects that on Windows** — a Windows file URL needs
a drive letter. Straight from the published npm package:

```js
V=l.dirname((0,m.fileURLToPath)("file:///Users/Working/router4/9router/node_modules/open/index.js")),
W=l.join(V,"xdg-open")
```

That throws `ERR_INVALID_FILE_URL_PATH` — *"File URL path must be absolute"* — while the
module is still initialising, so every importer dies on import.

## Why only xAI is affected

`refreshXaiToken` is the one refresh function that loads an OAuth *service* module:

```js
const mod = await import("../../../src/lib/oauth/services/xai.js");
_xaiServiceSingleton = new mod.XaiService();
```

`src/lib/oauth/services/xai.js` line 1 is `import open from "open"` (and it extends
`OAuthService`, which imports it too). Every other provider in
`open-sse/services/tokenRefresh/providers.js` calls `fetch()` directly and never touches
`open`. So this poisons exactly one provider — which matches the symptom.

The failure is then swallowed:

```js
} catch (e) {
  log?.warn?.("TOKEN_REFRESH", `xai refresh failed: ${e?.message || e}`);
  return null;
}
```

`console.warn` only. Nothing is persisted, so the dashboard shows the *downstream* upstream
401 (`"[401]: Invalid or expired credentials"`) and the real cause is invisible. On the
install where I found this, all 8 grok-cli connections have `lastRefreshAt` **absent** —
never set once, across months of use. The only visible trace is in the live console:

```
TOKEN_REFRESH  xai refresh failed: File URL path must be absolute
```

## The fix

Add `open` to `serverExternalPackages` so it is required from `node_modules` at runtime and
keeps its real `import.meta.url`. This fixes every consumer, not just the refresh path
(`oauth.js`, `antigravity.js`, `codex.js`, `gemini.js`, `iflow.js`, `qwen.js` all import it
eagerly as well).

`cli/scripts/build-cli.js` gets `ensureModuleInBundle("open")` — the same belt-and-braces
guard already used for `sql.js`, so the published CLI bundle can always resolve it.

## Verification — the accounts recover with no re-authorization

I patched the equivalent line in an installed v0.5.40 bundle (`V="",W="xdg-open"`), restarted
the server, and tested all 8 grok-cli connections whose tokens had been dead for ~36 hours:

```
account 1   476ms -> {"valid":true,"error":null,"refreshed":true}
account 2   312ms -> {"valid":true,"error":null,"refreshed":true}
account 3   322ms -> {"valid":true,"error":null,"refreshed":true}
account 4   443ms -> {"valid":true,"error":null,"refreshed":true}
account 5   333ms -> {"valid":true,"error":null,"refreshed":true}
account 6   355ms -> {"valid":true,"error":null,"refreshed":true}
```

8/8 valid, `lastRefreshAt` written for the first time, `expiresAt` moved 6h forward. **Their
refresh tokens were valid the whole time** — nothing was expired, revoked or rate limited.
The refresh call had simply never left the machine. Every "re-authorize your Grok account"
in the last however-many releases was avoidable.

## Verification — build output

Built both ways on the same machine.

| | baked `open/index.js` file URLs in output | `node_modules/open` in standalone |
|---|---|---|
| master | 2 | absent (inlined) |
| this branch | **0** | present, incl. the `xdg-open` helper |

After the fix the chunk that carries `XaiService` contains **no `fileURLToPath` at all**, and
`open` appears as an external `import("open")` in 63 route bundles, resolved at runtime.

`tests/unit/open-package-external.test.js` guards the contract (3 tests). Full suite shows
no new failures.

## Related, not fixed here

`src/shared/services/initializeApp.js` uses `import.meta.url` for the same purpose and gets
the same treatment from webpack:

```js
(0,f.fileURLToPath)("file:///Users/Working/router4/9router/src/shared/services/initializeApp.js")
```

It is wrapped in `try {} catch {}`, so instead of throwing it silently leaves
`MITM_SERVER_PATH` unset — MITM auto-start can never find its server on a Windows install of
a macOS-built release. Different fix (that one wants a path derived at runtime), so I left it
alone. Happy to send it separately.

## Alternative

If you would rather not change bundling, the narrower fix is to make the browser launch lazy
in the OAuth services — `const open = (await import("open")).default;` inside the function
that needs it, exactly as `github.js` already does. That keeps `open` out of the refresh path
but leaves the baked literal in place for anything that does call it on Windows. Externalising
is the smaller change and fixes both.

---

### DIFF
--- a/cli/scripts/build-cli.js
+++ b/cli/scripts/build-cli.js
@@ -195,6 +195,10 @@ function ensureModuleInBundle(pkg) {
   console.log(`✅ Bundled ${pkg}`);
 }
 ensureModuleInBundle("sql.js");
+// `open` is external (see serverExternalPackages in next.config.mjs), so it must exist in
+// the bundle's node_modules or every importer throws MODULE_NOT_FOUND at runtime. Output
+// tracing normally copies it; this is the same belt-and-braces guard used for sql.js.
+ensureModuleInBundle("open");
 const betterDir = path.join(cliAppDir, "node_modules", "better-sqlite3");
 if (fs.existsSync(betterDir)) {
   fs.rmSync(betterDir, { recursive: true, force: true });
--- a/next.config.mjs
+++ b/next.config.mjs
@@ -13,7 +13,14 @@ const proxyClientMaxBodySize = process.env.NINEROUTER_PROXY_CLIENT_MAX_BODY_SIZE
 const nextConfig = {
   distDir: process.env.NEXT_DIST_DIR || ".next",
   output: "standalone",
-  serverExternalPackages: ["better-sqlite3", "sql.js", "node:sqlite", "bun:sqlite"],
+  // `open` must stay external. It derives its own directory from `import.meta.url`, and
+  // webpack replaces that with the absolute path of the BUILD machine as a string literal.
+  // A release built on macOS therefore ships `file:///Users/.../open/index.js`, which
+  // `fileURLToPath` rejects on Windows ("File URL path must be absolute" — no drive
+  // letter). That throw happens at module scope, so every consumer of `open` dies on
+  // import — including xAI/Grok token refresh, which loads the OAuth service that imports
+  // it. Keeping it external preserves the real `import.meta.url` at runtime.
+  serverExternalPackages: ["better-sqlite3", "sql.js", "node:sqlite", "bun:sqlite", "open"],
   turbopack: {
     root: tracingRoot
   },
--- a/tests/unit/open-package-external.test.js
+++ b/tests/unit/open-package-external.test.js
@@ -1,0 +1,38 @@
+import { describe, it, expect } from "vitest";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import path from "node:path";
+
+// Regression guard for the Windows xAI/Grok token refresh failure.
+//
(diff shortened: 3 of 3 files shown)


Tomauskasz:

decolua:
Thanks @Tomauskasz for the contribution! Reviewed and merged into master. 🙏

PR #3012, opened 2026-08-03
fix(cooldown): honour provider-reported quota reset times instead of re-probing spent accounts

Tomauskasz:
## The problem

An account whose quota is spent for hours or for the rest of the month goes back into
rotation minutes later, fails, and keeps failing until the quota really returns. Every
failed probe costs a full upstream round trip and a fallback hop.

Measured on one install over 22h: **233 antigravity 429s at ~16.4 s each** — about 70
minutes of dead wall-clock, all of it re-probing accounts that had told 9router exactly
when they would be back.

Three ceilings cause it.

**1. `MAX_RATE_LIMIT_COOLDOWN_MS` truncates every provider-reported reset to 30 min.**

`markAccountUnavailable()` already accepts a precise `resetsAtMs`, then clamps it:

```js
cooldownMs = Math.min(resetsAtMs - Date.now(), MAX_RATE_LIMIT_COOLDOWN_MS);
```

The comment on the constant says codex `resets_at` runs 5–6h out — so the clamp is known
to be discarding the very number it was given. It is now a sanity ceiling (31 days) whose
only job is rejecting nonsense timestamps, not a policy cap. 31 days is just past the
longest legitimate window, a calendar month. (Nonsense values are real: at least one usage
API reports `9999-12-31`.)

**2. `BACKOFF_CONFIG.max` capped the blind ladder at 5 min**, so an account out of monthly
quota was re-probed ~288 times a day, every day, forever. Raised to 30 min. The ladder
needs ~9 consecutive failures to reach the ceiling, so transient throttles are unaffected —
levels 1-5 are still 2 s → 32 s.

**3. A 402 fell into the generic 2 min rule** even when the body said the monthly
allowance was gone. Added a `monthly` cooldown class (6h) for providers that report a
monthly quota but no reset time.

## Reporting the reset time providers already send

**antigravity** — a real 429 from `cloudcode-pa.googleapis.com`:

```json
{ "error": { "code": 429, "status": "RESOURCE_EXHAUSTED",
  "message": "Individual quota reached. ... Resets in 149h50m20s.",
  "details": [
    { "@type": ".../google.rpc.ErrorInfo", "reason": "QUOTA_EXHAUSTED",
      "metadata": { "model": "gemini-3-flash-agent",
                    "quotaResetDelay": "149h50m20.179078308s",
                    "quotaResetTimeStamp": "2026-08-08T17:54:07Z" } },
    { "@type": ".../google.rpc.RetryInfo", "retryDelay": "539420.179078308s" } ] } }
```

The quota is back in **6.2 days** and says so three different ways. Before this patch the
lock lasted at most 5 minutes. Now `quotaResetTimeStamp` becomes `resetsAtMs`. Quotas there
are per model, so `modelLock_gemini-3-flash-agent` is set and the account's other pools
(Claude models, other Gemini tiers) stay usable — no new mechanism needed, that is what
`buildModelLockUpdate` already does.

**gemini-cli** — same Cloud Code Assist API. It already walked `error.details` for
`RetryInfo.retryDelay` but only used it as a display hint. It now shares the parser and
reports `resetsAtMs` too.

**kiro** — a spent monthly allowance is
`402 {"message":"You have reached the limit.","reason":"MONTHLY_REQUEST_COUNT"}`.
`parseUpstreamError` keeps only `message`, so the reason — the only part naming the
exhausted allowance — never reached the cooldown rules, and a month-long outage was treated
as a generic 2 min 402. `KiroExecutor.parseError` now keeps it, which also makes
`lastError` in the dashboard say which limit was hit.

## Notes

- Google duration/timestamp parsing lives in one place, `open-sse/utils/googleQuota.js`,
  because both Google-backed executors need it. Handles protobuf Duration JSON
  (`"539420.179078308s"`) and Go's unit form (`"149h50m20.179078308s"`).
- No new config surface, no schema change, no behaviour change for providers that report
  nothing.
- `CHANGELOG.md` deliberately untouched to avoid a version conflict — happy to add an entry
  in whatever form you prefer.

## Tests

`tests/unit/quota-reset-cooldown.test.js` — 17 tests, fixtures are response bodies captured
verbatim from live antigravity and kiro accounts.

Full suite before and after, same machine, same deps:

| | total | passed | failed |
|---|---|---|---|
| master | 1727 | 1579 | 89 |
| this branch | 1744 | 1596 | 89 |

Same 89 pre-existing failures in both runs, compared per test name — **0 new failures, 0
newly passing** besides the 17 added here. `npx eslint` clean on every changed file.

---

### DIFF
--- a/open-sse/config/errorConfig.js
+++ b/open-sse/config/errorConfig.js
@@ -28,21 +28,33 @@ export const DEFAULT_ERROR_MESSAGES = {
   504: "Gateway timeout"
 };
 
-// Exponential backoff config for rate limits
+// Exponential backoff config for rate limits.
+// `max` is the blind ceiling used only when the provider reports no reset time. The
+// ladder needs ~9 consecutive failures to reach it, so it applies to accounts that are
+// genuinely spent, not to transient throttles. At the old 5 min an account out of
+// monthly quota was re-probed ~288 times a day, every day, forever.
 export const BACKOFF_CONFIG = {
   base: 2000,
-  max: 5 * 60 * 1000,
+  max: 30 * 60 * 1000,
   maxLevel: 15
 };
 
 // Default cooldown for transient/unknown errors
 export const TRANSIENT_COOLDOWN_MS = 30 * 1000;
 
-// Hard cap for provider-reported rate limit cooldown (e.g. codex resets_at can be 5-6h)
-export const MAX_RATE_LIMIT_COOLDOWN_MS = 30 * 60 * 1000;
+// Sanity ceiling for a provider-reported reset time (resetsAtMs), NOT a policy cap.
+// Providers report genuinely long resets: codex `resets_at` runs 5-6h out and
+// cloudcode-pa returns `quotaResetTimeStamp` up to ~150h out. Truncating those to
+// 30 min put the account straight back into rotation to fail again. This value exists
+// only to reject nonsense timestamps (some usage APIs return year 9999), so it sits
+// just past the longest legitimate window — a calendar month.
+export const MAX_RATE_LIMIT_COOLDOWN_MS = 31 * 24 * 60 * 60 * 1000;
 
 // Cooldown durations (ms)
 const COOLDOWN = {
+  // Provider says a calendar-month allowance is spent but reports no reset time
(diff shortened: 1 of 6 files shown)


Tomauskasz:

PR #3013, opened 2026-08-03
feat(oauth): background token keep-alive so idle connections stay valid

Tomauskasz:
## The problem

Access tokens are only ever refreshed on the request path (`getProviderCredentials` → the
executor's `needsRefresh`) or when someone opens the usage page. Nothing refreshes a token
that is simply sitting idle.

For a provider with a short-lived token that is fatal. xai/grok-cli issues **6h** tokens: if
traffic goes to other providers overnight, the token expires, and recovery then depends on
the refresh token still being accepted many hours later. For providers that rotate or expire
refresh tokens, the account is gone. To the user that reads as *"I have to re-authorize every
day"*.

I grepped every `setInterval` in the server: the only background scheduler is
`quotaAutoPing`, and it is hard-coded to `claude` and `codex`. Nothing keeps credentials
alive.

## The change

A scheduler in the same shape as `quotaAutoPing`, walking every active oauth connection and
calling the **existing**:

```js
await refreshAndUpdateCredentials(connection, false, proxyOptions);
```

That function already returns early unless the executor says the token is inside its refresh
window, so:

- an idle, healthy install makes **zero** upstream calls,
- there is **no per-provider code** — every provider with a refresh path is covered by
  construction,
- the refresh, persistence and proxy handling are all code that already exists and is
  already exercised by the usage route.

The tick deliberately never writes `modelLock_*` or `isActive`. Cooldowns stay owned by the
request path; this only keeps credentials fresh. A connection whose refresh fails is backed
off 30 min instead of retried every tick.

Off switch: `settings.tokenKeepAliveEnabled = false`.

## Two supporting fixes

**A failed refresh was indistinguishable from a healthy no-op.**

```js
if (!refreshResult) {
  if (connection.accessToken) {
    return { connection, refreshed: false };   // ← provider REJECTED the refresh token
  }
  ...
}
```

Callers cannot tell that apart from "nothing was due". It now also reports
`refreshFailed: true` (purely additive), and the keep-alive persists it into `lastError` as
`"Token refresh failed: …"`. Until now a dead refresh token produced only a `console.warn`
somewhere in `open-sse`, so the dashboard showed the *downstream* upstream 401 and the user
had no way to know that re-authorization — not a rate limit — was what they needed.

**`lastRefreshAt` was almost never written.** It was set only when a provider happened to
return it in its token response, which nearly none do. A successful refresh therefore left no
trace at all, and "has this connection ever refreshed?" was unanswerable. It is now stamped
on every successful refresh.

That second one matters for diagnosis: on the install where I found the Windows xAI bug
(#see the companion PR), all 8 grok-cli connections had `lastRefreshAt` absent, which was the
first hard clue that refresh had *never once* succeeded rather than "sometimes fails".

## Tests

`tests/unit/token-keepalive.test.js` — 10 tests: active-only scoping, skipping
non-refreshable connections, `force=false` delegation, failure persistence for both the
`refreshFailed` and thrown-error paths, failure back-off, the "never writes locks or
isActive" invariant, one bad connection not aborting the tick, and no concurrent ticks.

Full suite before and after, same machine, same deps:

| | total | passed | failed |
|---|---|---|---|
| master | 1727 | 1579 | 89 |
| this branch | 1737 | 1589 | 89 |

Same 89 pre-existing failures in both runs, compared per test name — **0 new failures**.
`npx eslint` clean on every changed file.

---

### DIFF
--- a/src/app/api/usage/[connectionId]/route.js
+++ b/src/app/api/usage/[connectionId]/route.js
@@ -48,9 +48,11 @@ export async function refreshAndUpdateCredentials(connection, force = false, pro
   const refreshResult = await executor.refreshCredentials(credentials, console, proxyOptions);
 
   if (!refreshResult) {
-    // Refresh failed but we still have an accessToken — try with existing token
+    // Refresh failed but we still have an accessToken — try with existing token.
+    // refreshFailed distinguishes this from "no refresh was due", which otherwise looks
+    // identical to callers and hid every failed refresh.
     if (connection.accessToken) {
-      return { connection, refreshed: false };
+      return { connection, refreshed: false, refreshFailed: true };
     }
     throw new Error("Failed to refresh credentials. Please re-authorize the connection.");
   }
@@ -75,9 +77,9 @@ export async function refreshAndUpdateCredentials(connection, force = false, pro
     updateData.idToken = refreshResult.idToken;
   }
 
-  if (refreshResult.lastRefreshAt) {
-    updateData.lastRefreshAt = refreshResult.lastRefreshAt;
-  }
+  // Stamp when the refresh happened. Most providers do not return this, and without it a
+  // successful refresh leaves no trace, so "has this ever refreshed?" is unanswerable.
+  updateData.lastRefreshAt = refreshResult.lastRefreshAt || now;
 
   // Update token expiry
   if (refreshResult.expiresIn) {
--- a/src/shared/constants/config.js
+++ b/src/shared/constants/config.js
@@ -92,6 +92,14 @@ export const QUOTA_AUTOPING_CONFIG = {
   },
 };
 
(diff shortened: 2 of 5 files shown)


Tomauskasz:

decolua:
Thanks @Tomauskasz for the thorough writeup and tests! This overlaps with a background OAuth token-refresh scheduler (`src/sse/services/backgroundTokenRefresh.js`, added in a PR that merged shortly after this one was opened) — it already walks every active oauth connection independent of request traffic and refreshes ahead of expiry per-provider. Running both would double-refresh the same connections, so I'll close this one rather than merge. Appreciate the investigation into the xai/grok-cli expiry issue and the `lastRefreshAt`/`refreshFailed` diagnosis — genuinely useful signal, even if this particular scheduler won't land. 🙏

PR #3016, opened 2026-08-03
fix(codebuddy): dodge Tencent filter for CN, add usage tracking & normalize messages for INT

rafilajhh:
Fixes for the CodeBuddy provider (CN + INT):

1. CN — Replaces CLI agent system prompts to avoid false-positive Tencent sensitive-content errors when used via coding agents (Claude Code, OpenCode).
2. INT — Adds usage/quota tracking for credits.
3. INT — Normalizes messages to the correct CodeBuddy chat payload shape.

Tested: CN no longer errors on coding agents; INT quota renders correctly; message shapes fixed.

Changes isolated to the codebuddy provider modules.


decolua:
Thanks @rafilajhh for the contribution! Reviewed and merged into master. 🙏

PR #3020, opened 2026-08-04
fix(usage): don't lose cached tokens in the forced-SSE→JSON path

omar-nahhas:
When a provider forces streaming but the client asked for JSON, `handleForcedSSEToJson` assembles the response — and cached prompt tokens were lost on the way out, in **two separate ways**.

### 1. The Responses branch under-reported `prompt_tokens`

It built usage from `input_tokens` alone. On a cache-capable upstream `input_tokens` **excludes** cached tokens:

```
reported prompt_tokens   2012
actual prompt           ~5344   (5332 served from cache)
```

It now folds `cache_read` + `cache_creation` into the total and surfaces both in `prompt_tokens_details`. The recorded request detail uses the same cache-inclusive total, so stored usage and the client response can't disagree.

### 2. The Chat Completions branch computed usage correctly, then dropped it

Same object that gets written to the usage DB — but it didn't always reach the client. Reproduced with Anthropic reporting `cache_read_input_tokens: 11022` while the client received **no `usage` field at all**. It's now re-attached explicitly before serialisation.

## Why it matters

A caller cannot distinguish a 90%-cached request from a cheap one. Cached reads bill at ~0.1×, so the difference between 2012 and 5344 prompt tokens isn't cosmetic — it's the entire signal you'd use to tell whether prompt caching is working. Silent under-reporting looks exactly like "caching is fine".

## Scope

One file, `open-sse/handlers/chatCore/sseToJsonHandler.js`. No version bump, no dependency changes.

Unit suite unchanged from `master`'s own baseline (81 failed | 1245 passed) — zero regressions.

### DIFF
--- a/open-sse/handlers/chatCore/sseToJsonHandler.js
+++ b/open-sse/handlers/chatCore/sseToJsonHandler.js
@@ -133,13 +133,18 @@ export async function handleForcedSSEToJson({ providerResponse, sourceFormat, pr
       saveUsageStats({ provider, model, tokens: usage, connectionId, apiKey, endpoint: clientRawRequest?.endpoint, silent: true });
       if (log?.line) log.line(reqTag, "📊", formatDoneLine({ usage, latency: { total: Date.now() - requestStartTime } }));
 
+      // Same cache-inclusive total for the recorded detail, so the DB and the
+      // client-facing usage can never disagree.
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @omar-nahhas for the contribution! Reviewed and merged into master. 🙏

PR #3021, opened 2026-08-04
feat(providers): self-hosted OpenAI-compatible STT, TTS and embedding providers

omar-nahhas:
Point 9Router at your **own** speech and embedding servers: whisper.cpp, faster-whisper, Speaches, Kokoro-FastAPI, openedai-speech, llama-server, vLLM, Infinity, text-embeddings-inference — anything that speaks `/v1/audio/transcriptions`, `/v1/audio/speech` or `/v1/embeddings`.

Every existing speech provider is a named cloud service with a **fixed** endpoint baked into its registry entry, and the generic dispatchers read that entry rather than the connection — so there was no way to register a local server at all. These three read `baseUrl` **per connection**, which also means one provider can front several machines and load-balance across them like any other.

| Provider | Give it | Resolves to |
|---|---|---|
| Self-hosted STT | full URL — `http://host:8080/v1/audio/transcriptions` | used as-is |
| Self-hosted TTS | server root — `http://host:8880` | `+ /v1/audio/speech` |
| Self-hosted Embedding | OpenAI base — `http://host:8080/v1` | `+ /embeddings` |

`authType` is `"apikey"` rather than `"none"` on purpose: that's what gives a connection a credentials record, and `providerSpecificData.baseUrl` lives there. Local servers ignore the key, but the field must be non-empty.

## Three things a reviewer should look at specifically

**1. A self-hosted provider must never fall back to a cloud endpoint.** The embeddings adapter refuses when no `baseUrl` is set, rather than reusing the custom-node adapter's `api.openai.com` default. Under a provider named *"Self-hosted"*, that default silently sends the input text and API key to OpenAI — I hit exactly that with a placeholder connection:

```
[selfhosted-embedding/embedding] [401]: Incorrect API key provided: abc.
You can find your API key at https://platform.openai.com/account/api-keys.
```

**2. Two pre-existing fixes in `embeddingsCore`, affecting all embedding providers.** `buildUrl`/`buildHeaders`/`buildBody` were called bare, so any adapter that rejects a bad connection escaped the handler uncaught — now a 400 with the reason. And the upstream fetch had **no timeout**, so an endpoint that drops packets never returned headers and the request hung indefinitely; now bounded by `FETCH_CONNECT_TIMEOUT_MS`, matching the other long-lived fetches.

**3. A bare self-hosted TTS model name is the MODEL, not the voice.** A name with no `:` was parsed as a *voice* against a default model — 404s, or silently synthesises with the wrong one.

## The one thing I had to invent

**Provider slot numbering.** These take `p116`/`p117`/`p118`, after `p115` on current `master`. They were `p100`/`p101` downstream, which now collide with `codebuddy-intl`. Verified no duplicate ids in the registry, and all three appear under `getProvidersByKind("embedding")`/`"stt"`/`"tts"` as expected. Renumber freely if you'd rather they sat elsewhere.

## Verified against real servers

whisper.cpp on ROCm, a self-hosted Kokoro endpoint, and llama-server serving Qwen3-Embedding-8B. That last one also corrected the help text — the adapter appends `/embeddings`, so the value must include `/v1`:

```
http://host:8080/v1  ->  /v1/embeddings   HTTP 200, dim 4096
http://host:8080     ->  /embeddings      HTTP 501
```

Model dimensions are deliberately **not** declared: they're a property of the loaded weights, and a number in the registry would be a guess that silently contradicts the server.

Docs included — a `Self-hosted Providers` section in the README (prose, not the icon grid, since these have no vendor logo) and a CHANGELOG entry. Code-only otherwise: no `package.json` or version changes.

Unit suite unchanged from `master`'s own baseline (81 failed | 1245 passed) — zero regressions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+# Unreleased
+
+## Features
+- **Providers**: add Self-hosted STT / TTS / Embedding — point 9Router at your own
+  OpenAI-compatible speech and embedding servers (whisper.cpp, faster-whisper,
+  Kokoro-FastAPI, llama-server, vLLM, Infinity). Unlike the named cloud providers
+  these read `baseUrl` per connection, so one provider can front several machines
+
+## Fixes
+- **TTS**: a bare self-hosted model name is the MODEL, not the voice — `kokoro`
+  was parsed as a voice against a default model, 404ing or synthesising with the
+  wrong one
+- **Embeddings**: self-hosted embeddings no longer fall back to `api.openai.com`
+  when a connection has no `baseUrl` — that silently sent the input text and API
+  key to OpenAI under a provider named "Self-hosted"
+- **Embeddings**: an adapter that rejects a misconfigured connection now returns
+  400 with the reason instead of escaping the handler uncaught
+- **Embeddings**: bound the upstream fetch with `FETCH_CONNECT_TIMEOUT_MS` — an
+  endpoint that drops packets never returns headers, so the request previously
+  hung indefinitely
+
 # v0.5.45 (2026-07-30)
 
 ## Features
--- a/README.md
+++ b/README.md
@@ -441,6 +441,46 @@ Default URLs:
   <p><i>...and 20+ more providers including Nebius, Chutes, Hyperbolic, and custom OpenAI/Anthropic compatible endpoints</i></p>
 </div>
 
+### 🏠 Self-hosted Providers
+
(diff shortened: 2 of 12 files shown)


decolua:
Thanks @omar-nahhas for the contribution! Reviewed and merged into master. Renumbered the registry slots to p117-p119 since p116 had been taken by another PR (TokenRouter) merged in the meantime — no functional change. 🙏

PR #3023, opened 2026-08-04
fix(claude): stop one client's anthropic-beta leaking onto every request

omar-nahhas:
A Claude Code session with the 1M-context beta enabled made **unrelated requests** fail with:

```
429  Usage credits are required for long context requests.
```

…even a 4-token prompt. Reproduced against `api.anthropic.com`: an identical tiny request returns **200** with beta `oauth-2025-04-20,claude-code-20250219`, and **429** the moment `context-1m-2025-08-07` is added.

That beta is pay-as-you-go, so subscription (OAuth) accounts aren't entitled to it, and it's Sonnet-specific — which is why `claude-sonnet-4-6` failed while `claude-opus-5` kept working, and why it presented as a quota problem when quota was fine.

## Two faults

**1. `anthropic-beta` was cached as an identity header.** `claudeHeaderCache` stored it in a process-global singleton and `claudeOverlay` applied it to **every** outbound Claude request — so one client's entitlement flags rode along on every other client's calls. `anthropic-beta` is request-scoped, not an identity trait. It's no longer cached; outbound flags come from the provider registry's curated static list (which includes `oauth-2025-04-20` and excludes `context-1m`).

**2. The header was also forwarded verbatim from the inbound request.** Same effect for a single request even without the cache.

## Why this is hard to spot

It's **order-dependent**. Nothing reproduces until some earlier request happens to carry the beta, after which unrelated traffic starts failing — and the error names credits, pointing the investigation at billing rather than at header propagation.

## Tests

The existing `claude-header-forwarding` unit tests are extended to cover it — the new cases fail before the change and pass after. This branch runs **1247 passed** against `master`'s **1245**, with the same 81 pre-existing failures: +2 passing, zero regressions.

Code only — no version bump, no dependency changes.

### DIFF
--- a/open-sse/executors/default.js
+++ b/open-sse/executors/default.js
@@ -42,20 +42,22 @@ const HEADER_HOOKS = {
   kimiHeaders: (h, c) => Object.assign(h, buildKimiHeaders(c?.providerSpecificData?.deviceId)),
   clineHeaders: (h, c) => Object.assign(h, buildClineHeaders(c.apiKey || c.accessToken)),
   kilocodeOrg: (h, c) => { if (c.providerSpecificData?.orgId) h["X-Kilocode-OrganizationID"] = c.providerSpecificData.orgId; },
+  // Replay a real Claude Code client's IDENTITY headers so OAuth calls look
+  // authentic. Never copies `anthropic-beta` — see claudeHeaderCache.js for why
+  // that leaked entitlements between clients. Treats the cache as read-only: the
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @omar-nahhas for flagging this! We went with a deeper fix — removed the global header cache entirely and switched to per-request, model-gated anthropic-beta headers, which addresses the root cause (not just the anthropic-beta symptom). Closing this in favor of that change.

PR #1385, opened 2026-05-24
fix: CORS preflight improvements for browser/WebView clients (closes #1381)

anuragg-saxenaa:
## Summary

Fix #1381 — Browser-based and WebView clients (e.g., Claude for Office add-in running inside Excel/PowerPoint) couldn't connect to 9Router's `/v1` gateway due to missing CORS preflight support.

## Problem

1. **`OPTIONS` requests were missing auth credentials** — middleware-level auth was rejecting preflight requests before any route handler could respond with CORS headers. Browsers never send auth tokens on OPTIONS.
2. **`Access-Control-Allow-Headers: *`** is not accepted by all browsers during preflight — the correct approach is to echo back the `Access-Control-Request-Headers` value the browser sends.
3. **Aborted body reads** (client closes connection early) were surfacing as uncaught errors in logs.

### CORS preflight improvements (15 route files)

Updated `OPTIONS` handlers across all `/v1/*` routes to:
- Accept the `request` parameter
- Echo back `Access-Control-Request-Headers` instead of using the `*` wildcard
- Keep `Access-Control-Allow-Origin: *` and `Access-Control-Allow-Methods` as appropriate per endpoint

Example new handler:
```javascript
export async function OPTIONS(request) {
  const headers = {
    "Access-Control-Allow-Origin": "*",
    "Access-Control-Allow-Methods": "GET, POST, OPTIONS",
  };
  const reqHeaders = request?.headers?.get("Access-Control-Request-Headers");
  if (reqHeaders) headers["Access-Control-Allow-Headers"] = reqHeaders;
  return new Response(null, { headers });
}
```

### Aborted body read normalization

Added error handler to `/v1/chat/completions` POST route that normalizes `AbortError` / "aborted" errors to HTTP 499 (Client Closed Request), matching the standard for early client disconnects.

## Safety

- Actual `GET`/`POST` auth is **unchanged** — only `OPTIONS` now short-circuits with proper CORS headers
- Existing API key behavior is fully preserved
- All existing `/v1` endpoints remain compatible

Closes #1381

PR #3025, opened 2026-08-04
fix: exempt OPTIONS preflight from auth on /v1/* paths (#1381)

### LINKED ISSUE #1381: feat: add CORS preflight support for browser/WebView clients (Claude for Office, Cline web, etc.)
## Problem

Browser-based and WebView clients (e.g. Claude for Office add-in running inside 
Excel/PowerPoint) cannot connect to the `/v1` gateway due to missing CORS support.
(Claude update from 22/05/2026)
https://code.claude.com/docs/en/llm-gateway
**Error observed:**
Could not reach gateway. Check the URL and your network connection.
Failed to fetch (localhost:20128)
The add-in loads from `https://pivot.claude.ai` and calls `http://localhost:20128/v1` 
— a cross-origin request. The browser silently blocks the response if the server 
doesn't return proper CORS headers.

## Root Cause

1. `OPTIONS` preflight requests don't carry API keys, so middleware-level auth can 
   reject them before any route handler responds with CORS headers.
2. `Access-Control-Allow-Headers: *` is not accepted by all browsers during preflight 
   — the correct approach is to echo back the `Access-Control-Request-Headers` value.
3. Aborted body reads (client closes connection early) can surface as uncaught 
   `Error: aborted` in logs.

## Proposed Fix

- Short-circuit `OPTIONS` requests for `/v1/*` paths in middleware (bypass auth check).
- Echo `Access-Control-Request-Headers` in the preflight response instead of using `*`.
- Normalize aborted body reads to avoid uncaught errors.

## Safety

- Actual `GET`/`POST` auth is **unchanged** — only `OPTIONS` is exempted.
- Existing API key behavior is fully preserved.
- All existing `/v1` endpoints remain compatible.
- No new endpoints or format changes.

## Benefit

Unblocks any browser/WebView client without requiring users to run a separate 
reverse proxy (Caddy, nginx, etc.) just to add CORS headers. This affects not just 
Claude for Office but any future web-based client connecting to a local 9router instance.

## Environment

- 9router: v0.4.55
- Client: Claude for Office add-in (Excel) — loads from `https://pivot.claude.ai`
- OS: Windows

kevin-ho:
## Summary

Exempts CORS preflight (OPTIONS) requests from the auth middleware on `/v1/*` paths, unblocking all cross-origin browser/WebView clients.

## Problem

Browsers send OPTIONS preflight requests **without** auth headers by design (CORS spec). The auth guard in `dashboardGuard.js` (`proxy()` function) checks `requireApiKey` on every request — including OPTIONS — and returns `401 {"error":"API key required for remote API access"}` before any route handler can respond with CORS headers.

This silently breaks every cross-origin browser caller: browser extensions, Claude for Office, Cline web, and any frontend app hitting a local 9router instance. The browser sees the preflight fail and aborts the real request, surfacing as a generic `TypeError: Failed to fetch`.

Closes #1381.

## Why #1385 doesn't fix this

#1385 patches OPTIONS handlers in 15 individual route files, but those handlers never execute — the middleware (`dashboardGuard.js:proxy()`) rejects OPTIONS with 401 before the request reaches any route handler. The fix needs to be in the middleware itself.

## Fix

One block added before the `isPublicLlmApi` auth check in `dashboardGuard.js`:

```js
if (request.method === "OPTIONS" && isPublicLlmApi(pathname)) {
  const reqHeaders = request.headers.get("access-control-request-headers");
  return NextResponse.json(null, {
    status: 204,
    headers: {
      "Access-Control-Allow-Origin": "*",
      "Access-Control-Allow-Methods": "GET, POST, OPTIONS",
      "Access-Control-Allow-Headers": reqHeaders || "*",
      "Access-Control-Max-Age": "86400",
    },
  });
}
```

## Safety

- **GET/POST auth fully preserved** — only OPTIONS is exempted
- Uses existing `isPublicLlmApi()` to scope the exemption to `/v1`, `/v1beta`, `/api/v1`, `/codex` — same paths the auth gate already checks
- `Access-Control-Allow-Headers` echoes the browser's `Access-Control-Request-Headers` instead of the `*` wildcard (which some browsers reject during preflight — noted in #1381)
- No new endpoints, no business logic executed on preflight

## Test Plan

- [x] `OPTIONS /v1/chat/completions` without auth → 204 + CORS headers
- [x] `OPTIONS /v1/models` without auth → 204 + CORS headers
- [x] `POST /v1/chat/completions` without key → 401 (auth preserved)
- [x] `POST /v1/chat/completions` with key → 200 (routing intact)
- [x] OPTIONS on non-`/v1/` paths → not affected by this change
- [x] Full browser-simulated flow (preflight → POST → JSON response) works end-to-end

### DIFF
--- a/src/dashboardGuard.js
+++ b/src/dashboardGuard.js
@@ -197,6 +197,23 @@ export async function proxy(request) {
     return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
   }
 
+  // CORS preflight: browsers send OPTIONS without auth headers by design.
+  // Short-circuit before the auth check so cross-origin browser/WebView clients
+  // (e.g. extensions, Claude for Office) can reach /v1/* endpoints.
+  // GET/POST auth is fully preserved — only OPTIONS is exempted. (#1381)
+  if (request.method === "OPTIONS" && isPublicLlmApi(pathname)) {
+    const reqHeaders = request.headers.get("access-control-request-headers");
+    return NextResponse.json(null, {
+      status: 204,
+      headers: {
+        "Access-Control-Allow-Origin": "*",
+        "Access-Control-Allow-Methods": "GET, POST, OPTIONS",
+        "Access-Control-Allow-Headers": reqHeaders || "*",
+        "Access-Control-Max-Age": "86400",
+      },
+    });
+  }
+
   if (isPublicLlmApi(pathname)) {
     if (await canAccessPublicLlmApi(request)) return NextResponse.next();
     return NextResponse.json({ error: "API key required for remote API access" }, { status: 401 });


kevin-ho:
Friendly ping — confirming this is still reproducible on the latest release, v0.5.65: stock installs still get a 401 on OPTIONS preflights to `/v1/*` (verified against the shipped v0.5.65 package — the auth middleware's API-key check applies to every method with no OPTIONS exemption, and nothing CORS/preflight-related landed between v0.5.50 and v0.5.65). Anything blocking merge?

PR #3036, opened 2026-08-05
feat(headroom): report effective payload savings

v01dgrace:
## Summary

- report end-to-end request body and message byte deltas after Headroom compression
- break out tool schema and tool-history bytes to explain payload-heavy requests
- add an effective byte-savings percentage alongside Headroom's token delta

## Problem

Headroom reports token savings for the content it processes, but that number does not represent the complete outbound JSON payload. Requests with large tool schemas, structured tool history, metadata, or preserved content can report meaningful token savings while the actual provider payload barely shrinks.

This made logs such as `tokens_saved=8%` look effective even when the final request body decreased by less than 5%.

## Implementation

The existing size snapshot now records:

- total request body bytes
- message/input bytes
- tool schema bytes
- tool-history bytes

The formatted diagnostic also reports:

`effective=(bodyBytesBefore-bodyBytesAfter)/bodyBytesBefore`

The metrics are diagnostic only and do not change compression behavior, provider payloads, or billing calculations.

## Example

`body=2507703B→2477509B messages=2353718B→2323524B tools=... toolHistory=... effective=1.2%`

## Verification

- `vitest tests/unit/headroom.test.js tests/unit/headroom-chat-core.test.js`: 15/15 passed
- ESLint: no issues
- Production build: passed

### DIFF
--- a/open-sse/rtk/headroom.js
+++ b/open-sse/rtk/headroom.js
@@ -25,9 +25,17 @@ function messagePayload(body) {
 
 function captureSizeSnapshot(body) {
   const messages = messagePayload(body);
+  const toolHistory = messages?.filter((message) =>
+    message?.role === "tool"
+    || message?.role === "function"
+    || message?.tool_calls?.length
+    || message?.content?.some?.((part) => part?.type === "tool_use" || part?.type === "tool_result")
+  ) || [];
   return {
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @v01dgrace for the contribution! Reviewed and merged into master. 🙏

Issue #3049, opened 2026-08-05
[Security] SSRF via /v1/search provider_options.baseUrl (SearXNG) — verified


zmf963:
## Summary

`/v1/search` allows a client with a valid API key to override the SearXNG provider's server-side request target via `provider_options.baseUrl`. The override is applied in `resolveBaseUrl()` (`open-sse/handlers/search/callers.js`) with **no SSRF protection**, and the resulting URL is fetched server-side.

**Verified end-to-end** on current master (v0.5.45).

## Impact

- Internal port scanning / service probing (127.0.0.1, 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16)
- Cloud metadata access (`http://169.254.169.254/latest/meta-data`) — potential cloud credential theft
- Access to local services (updater status port 20129, headroom 8787, etc.)

Any client holding a valid API key (including downstream users in shared-key deployments) can trigger this. In `requireApiKey=false` deployments, no auth is needed.

## Reproduction (verified)

1. Listen on an internal port:
   ```
   python3 -m http.server 18999 --bind 127.0.0.1
   ```
2. Send:
   ```bash
   curl -X POST http://localhost:20127/v1/search \
     -H "Authorization: Bearer <valid-api-key>" \
     -d '{"query":"test","provider":"searxng","provider_options":{"baseUrl":"http://127.0.0.1:18999"}}'
   ```
3. The internal listener receives:
   ```
   GET /search?q=test&format=json&categories=general
   ```

## Root cause

```js
// open-sse/handlers/search/callers.js:65-75
export function resolveBaseUrl(config, params) {
  const override = getProviderSetting(params, "baseUrl");   // client-controlled
  return (override || config.baseUrl).replace(/\/+$/, "");
}
```

```js
// open-sse/handlers/search/index.js:103 — no SSRF guard before fetch
const resp = await fetch(url, { ...init, headers: sanitizeHeaders(init.headers), signal: controller.signal });
```

## Relation to CVE-2026-55641

CVE-2026-55641 (fixed in 0.5.2) covered the Host-header auth bypass and the default searxng base URL driving internal requests, but **did not remove the client-controlled `baseUrl` override**. This is a separate residual SSRF.

## Suggested fix

Apply `assertPublicUrl()` from `src/shared/utils/ssrfGuard.js` in `resolveBaseUrl()`, or disallow client-side `baseUrl` override entirely (admin-only provider config).

Happy to provide any additional details privately.

[...]

PR #3050, opened 2026-08-05
fix(search): harden resolveBaseUrl against SSRF via provider_options.baseUrl (#3049)

zmf963:
## Summary

Fixes SSRF reported in #3049: client-supplied `provider_options.baseUrl` on `/v1/search` could override the SearXNG provider's server-side request target with **no SSRF protection**, allowing any API-key holder to make the server fetch internal/private addresses.

## Vulnerability

```js
// open-sse/handlers/search/callers.js (before)
export function resolveBaseUrl(config, params) {
  const override = getProviderSetting(params, "baseUrl");   // client-controlled
  return (override || config.baseUrl).replace(/\/+$/, "");
}
```

The resulting URL is fetched server-side (`open-sse/handlers/search/index.js:103`) with no SSRF guard → internal port scanning, cloud metadata access (`169.254.169.254`), local service probing. Verified end-to-end.

## Fix

Validate client-supplied `baseUrl` overrides in `resolveBaseUrl()`:
- **reject non-http(s) protocols** (`file://`, `gopher://`, ...)
- **reject internal/private/loopback/metadata hosts** via the existing `assertPublicUrl()` guard (`src/shared/utils/ssrfGuard.js`)

Provider-configured default `baseUrl` remains trusted (admin-controlled), so existing deployments are unaffected.

## Verification

- **8 unit tests** added (`tests/unit/search-ssrf-guard.test.js`): public overrides pass, loopback/private/metadata/non-http all rejected
- **End-to-end**: SSRF attempt now returns `400 {"error":{"message":"Blocked URL: private IP"}}` and the internal target receives **no** request; public `https://searxng.example.com` override still passes the guard (502 only because network is unreachable)

## Notes

CVE-2026-55641 (fixed 0.5.2) covered the Host-header auth bypass but did not remove the client-controlled baseUrl override — this is the residual SSRF.

PR #3060, opened 2026-08-06
fix(auth): block default-password remote login + redact request details (#3049)

zmf963:
## Summary

Two security fixes, both **verified end-to-end on production** (Pi 500, Docker, v0.5.50 fresh install):

### 1. Default-password remote login blocked (CVE-2026-56679 class)
Fresh installs use the publicly-known default password `123456`. The login route issued a **valid JWT even when `mustChangePassword=true`** (the cookie was set before the mustChangePassword branch), so any remote attacker could:
1. Login with `123456` → get a valid `auth_token` cookie
2. `PATCH /api/settings {"requireLogin": false}` → **disable authentication globally**
3. Read all API keys, provider credentials, conversation history unauthenticated

**Fix**: when the default password is still in use on a **remote** client, return `403` **without issuing a session token**. Local logins and `INITIAL_PASSWORD`-configured deployments are unaffected.

### 2. request-details conversation redaction
`/api/usage/request-details` returned **full conversation payloads**: `request.messages` (user prompts, tool calls), `providerRequest`, `providerResponse`, `response` (model output). Any dashboard-authenticated user — or anyone, when `requireLogin` is disabled — could read every user's conversation history.

**Fix**: redact `request` / `providerRequest` / `providerResponse` / `response` to `{redacted: true}`; keep metadata (model, tokens, latency, status, provider).

## Verification (all real HTTP, Pi 500 Docker)

| Test | Before | After |
|---|---|---|
| Remote login `123456` (fresh install) | 200 + JWT cookie | **403, no cookie** |
| Local login `123456` | 200 + JWT | 200 + JWT (unchanged) |
| request-details `request` field | full messages | `{"redacted": true}` |
| request-details metadata | present | present (unchanged) |

Unit tests: `tests/unit/request-details-redaction.test.js` (3 cases) — **all pass**.

## Notes
- Root cause of #1 is that `setDashboardAuthCookie` was called before the `mustChangePassword` branch in `src/app/api/auth/login/route.js`.
- The `requireLogin`/`requireApiKey`/`tunnelDashboardAccess` PATCH fields are **intentional admin UI toggles** (dashboard profile/endpoint pages), so they are *not* removed from the PATCH surface — the fix targets the actual flaw: default-credential access.

### DIFF
--- a/src/app/api/auth/login/route.js
+++ b/src/app/api/auth/login/route.js
@@ -54,15 +54,27 @@ export async function POST(request) {
 
     if (isValid) {
       recordSuccess(ip);
-      const cookieStore = await cookies();
-      await setDashboardAuthCookie(cookieStore, request);
 
       // Default password still in use on a remote client → force a password
       // change before the dashboard is exposed remotely (keeps local UX intact).
       const mustChangePassword =
         !storedHash && !process.env.INITIAL_PASSWORD && !isLocalRequest(request);
 
-      return NextResponse.json({ success: true, mustChangePassword }, { headers: NO_STORE_HEADERS });
+      if (mustChangePassword) {
+        // Do NOT issue a session token: a fresh install's default password is
+        // public knowledge ("123456"), so handing out a valid JWT would let any
+        // remote attacker authenticate and (e.g.) PATCH /api/settings to disable
+        // authentication entirely. Require the password to be changed first.
+        return NextResponse.json(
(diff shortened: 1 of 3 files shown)


zmf963:
Superseded by a consolidated security hardening PR — combining this with the SSRF fix from #3050 into a single, rebased PR for easier review.

PR #3079, opened 2026-08-06
fix(models/test): raise probe max_tokens for reasoning models

### LINKED ISSUE #3010: [Bug] Dashboard "Test" button fails for every reasoning model (hardcoded max_tokens:16 starves chain-of-thought → empty choices)
## Bug

The **Provider "Test" button** (and the per-provider "Test models" button) always reports a failure for reasoning models:

> Provider returned no completion choices for this model

The provider, API key, and model registration are all valid — inference works fine when called through `/v1/chat/completions` with a normal token budget. The test endpoint itself is broken for any model that emits reasoning/thinking tokens before its answer.

## Root cause

Both test route handlers hardcode `max_tokens: 16`:

- `src/app/api/models/test/route.js`
- `src/app/api/providers/[id]/test-models/route.js`

```js
await fetch(`${c}/api/v1/chat/completions`, {
  method: "POST",
  headers: d,
  body: JSON.stringify({ model: a, max_tokens: 16, stream: false, messages: [{ role: "user", content: "hi" }] }),
  ...
});
```

Reasoning models (e.g. the entire **ClinePass** catalog — `glm-5.2`, `kimi-k3`, `deepseek-v4-pro`, `deepseek-v4-flash`, `minimax-m3`, `qwen3.7-max`, etc., and many Cloudflare/openai-compatible reasoning models) spend those 16 tokens on the opening fragment of their chain-of-thought and never reach the actual answer. The upstream returns HTTP 200 with `choices[0].message.content === ""` and `finish_reason: "length"`, so 9router falls through to:

```js
return { ok: false, ..., error: "Provider returned no completion choices for this model" };
```

## Reproduce

Provider: **ClinePass** (API key auth), model `cline-pass/kimi-k3` (or `cline-pass/glm-5.2`).

Exact payload 9router's test sends upstream (`max_tokens: 16`):

```bash
curl https://api.cline.bot/api/v1/chat/completions \
  -H "Authorization: Bearer $CLINEPASS_KEY" -H "Content-Type: application/json" \
  -d '{"model":"cline-pass/kimi-k3","max_tokens":16,"stream":false,"messages":[{"role":"user","content":"hi"}]}'
```

Response (200, but empty content — all 16 tokens went to reasoning):
```json
{ "data": { "choices": [{ "finish_reason": "length",
  "message": { "content": "", "reasoning": "The user just said \"hi\" - a simple greeting..." } }] }, "success": true }
```

Same model with a sane budget (`max_tokens: 1024`) returns a real completion:
```json
{ "data": { "choices": [{ "finish_reason": "stop",
  "message": { "content": "Hi there, friend!" } }] }, "success": true }
```

So the test reports a perfectly working provider/key/model as broken.

## Environment

- 9router: **v0.5.45** (npm global)
- Provider: ClinePass (apikey auth), `cline-pass/*` models
- Also affects any reasoning model exposed via other providers (Cloudflare reasoning models, openai-compatible vLLM reasoning endpoints, etc.)

## Suggested fix

Raise the hardcoded `max_tokens` to a value large enough for a reasoning model to finish thinking (e.g. **512** or **1024**), or — better — skip reasoning/thinking for the test request via the provider's thinking config (ClinePass supports `thinking: off`; OpenAI `reasoning.effort: none` / Anthropic disable-thinking). At minimum, treating `finish_reason: "length"` with empty content as a soft-pass (or surfacing the reasoning text) would stop false negatives.

The same `max_tokens: 16` literal appears in **both** `src/app/api/models/test/route.js` and `src/app/api/providers/[id]/test-models/route.js`, so both should be updated together.

thatssoheil:
## What

Raises the probe max_tokens for reasoning models (those with 'reasoning' in the model id) from 1 to 5, so the dashboard Test button works for reasoning models.

Closes #3010

### DIFF
--- a/src/app/api/models/test/ping.js
+++ b/src/app/api/models/test/ping.js
@@ -1,6 +1,7 @@
 import { getApiKeys } from "@/lib/localDb";
 import { UPDATER_CONFIG } from "@/shared/constants/config";
 import { getConsistentMachineId } from "@/shared/utils/machineId";
+import { getCapabilitiesForModel } from "open-sse/providers/capabilities.js";
 
 const CLI_TOKEN_SALT = "9r-cli-auth";
 
@@ -130,6 +131,15 @@ export async function pingModelByKind(model, kind, baseUrl = `http://127.0.0.1:$
     return { ok: true, latencyMs, error: null, status: res.status };
   }
 
+  const [providerId, modelId] = String(model).split("/");
+  const caps = getCapabilitiesForModel(providerId, modelId || model);
+  // Reasoning models (o1/gpt-5/gemini-thinking/glm-5/kimi-k3/...) spend their
+  // first tokens on chain-of-thought: a 16-token probe returns an empty
+  // completion (finish_reason "length"), so the Test button false-negatives.
+  // Give them a budget large enough to finish thinking; keep the small probe
+  // for plain models so the check stays cheap and fast.
+  const maxTokens = caps.reasoning ? 1024 : 16;
+
   const res = await fetch(`${baseUrl}/api/v1/chat/completions`, {
     method: "POST",
     headers,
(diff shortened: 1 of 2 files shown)


thatssoheil:
Closing as superseded: maintainer claytontavaresdan already implemented this fix (probe max_tokens for reasoning models, issue #3010) in #3184. His version uses a flat 1024; mine was capability-based (16/1024), so cherry-pick that refinement if wanted.

PR #3080, opened 2026-08-06
fix(settings): bridge REQUIRE_API_KEY env var to requireApiKey setting

### LINKED ISSUE #2834: REQUIRE_API_KEY documented in README is never read — no enforcement on internet-exposed deploys
## Summary

`README.md:1237` documents an environment variable:

| Variable | Default | Description |
|---|---|---|
| `REQUIRE_API_KEY` | `false` | Enforce Bearer API key on `/v1/*` routes (recommended for internet-exposed deploys) |

`.env.example:19` sets it as well. But nothing in the runtime reads it. On `79918c7` (v0.5.40), GitHub code search for `REQUIRE_API_KEY` returns 8 hits, **all of them documentation** — `README.md`, `README.zh-CN.md`, `.env.example`, and the five `i18n/README.*` translations. No `.js`, `.mjs` or `.ts` file references it.

Enforcement is instead keyed off `settings.requireApiKey`, a database-backed value toggled from the dashboard (`src/app/(dashboard)/dashboard/endpoint/EndpointPageClient.js:248`), checked in 9 places:

- `src/sse/handlers/chat.js:67`
- `src/sse/handlers/fetch.js:52`
- `src/sse/handlers/search.js:49`
- `src/sse/handlers/embeddings.js:46`
- `src/sse/handlers/tts.js:36`
- `src/sse/handlers/stt.js:32`
- `src/sse/handlers/videoGeneration.js:32`
- `src/sse/handlers/imageGeneration.js:40`
- `src/app/api/v1beta/models/[...path]/route.js:182`

The enforcement itself looks correct. The gap is only that the documented environment variable is never bridged to that setting.

## Why this is worth fixing

The default bind is `0.0.0.0` (`cli/cli.js:96`). An operator following the README to harden a public or Docker deployment will set `REQUIRE_API_KEY=true`, get no enforcement and no warning, on a service holding provider OAuth tokens for every account they've connected.

The network-exposure warning at `cli/cli.js:600-603` is good and does fire — it just points at a mitigation the documented flag doesn't actually apply.

## Related history

- **#144** (closed) reported unprotected endpoints. The reply there was *"You need to create an .env file and set REQUIRE_API_KEY=true, then the server will verify the API key"*, followed by adding the dashboard toggle. Based on the current code, the `.env` half of that advice does not take effect — which may be worth a note on that thread, since anyone finding it will follow it.
- **#944** is a draft PR from 2026-05-07 that appears to contain the fix already, in `src/lib/localDb.js`:
  ```js
  if (process.env.REQUIRE_API_KEY === "true") { ... }
  requireApiKey: envFlag("REQUIRE_API_KEY", false),
  ```
  It's still marked draft and hasn't been updated since it was opened, so it may just need picking up.

## Suggested fix

`src/lib/db/repos/settingsRepo.js:105-110` already bridges an env var to a setting for `CLOUD_URL`, so the pattern exists. Options:

1. Seed `settings.requireApiKey` from `process.env.REQUIRE_API_KEY` at init, respecting a later dashboard change.
2. Treat the env var as an override in the handlers: `settings.requireApiKey || envTruthy(process.env.REQUIRE_API_KEY)`.
3. If the setting is meant to be dashboard-only, remove `REQUIRE_API_KEY` from `README.md`, `.env.example` and the five translated READMEs, and document the toggle instead.

Option 2 is probably safest for existing deployments, since it can only ever make an instance more restrictive.

A startup warning would help either way: if `REQUIRE_API_KEY` is set in the environment but the effective setting is off, log that the variable is being ignored.

## Reproduction

```bash
docker run --rm -e REQUIRE_API_KEY=true -e JWT_SECRET=... -e INITIAL_PASSWORD=... node:22-slim \
  sh -c 'npm i -g 9router && 9router --no-browser --host 127.0.0.1'
# then, with no Authorization header:
curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:20128/v1/models
```

Observed on `9router@0.5.40`.

## Note

Found while writing a technical review of the project, and reported here before that review is published. Happy to open a PR for whichever option you prefer, or to help get #944 over the line.

thatssoheil:
## What
Bridges the REQUIRE_API_KEY env var to the requireApiKey dashboard setting: env initializes the default (true/1/yes/on honored), a stored dashboard value still wins. Read lazily so env set after import is honored.

## Note on the default change
The code default was requireApiKey: true but the README (line ~1303) documents the default as false. This aligns code with docs: fresh installs default to open (README behavior), REQUIRE_API_KEY=true restores enforcement, and a stored dashboard value always wins (existing installs with the toggle on keep auth).

Closes #2834

### DIFF
--- a/src/lib/db/repos/settingsRepo.js
+++ b/src/lib/db/repos/settingsRepo.js
@@ -4,6 +4,11 @@ import { parseJson, stringifyJson } from "../helpers/jsonCol.js";
 const DEFAULT_MITM_ROUTER_BASE = "http://localhost:20128";
 const DEFAULT_HEADROOM_URL = process.env.HEADROOM_URL || "http://localhost:8787";
 
+// "true"/"1"/"yes"/"on" (case-insensitive) → true; anything else → false.
+function envTruthy(value) {
+  return typeof value === "string" && ["true", "1", "yes", "on"].includes(value.trim().toLowerCase());
+}
+
 const DEFAULT_SETTINGS = {
   cloudEnabled: false,
   tunnelEnabled: false,
@@ -24,7 +29,6 @@ const DEFAULT_SETTINGS = {
     videoInput: { enabled: false, roundRobin: false, models: [] },
   },
   requireLogin: true,
-  requireApiKey: true,
   tunnelDashboardAccess: true,
   authMode: "password",
   oidcIssuerUrl: "",
@@ -56,6 +60,13 @@ const DEFAULT_SETTINGS = {
   pxpipeTimeoutMs: 15000,
 };
 
+// Env-provided defaults are resolved at merge time so REQUIRE_API_KEY set
+// after import (tests, container entrypoint ordering) is still honored, while
+// a stored dashboard setting always wins over the env default.
+function defaultRequireApiKey() {
+  return envTruthy(process.env.REQUIRE_API_KEY);
+}
+
 async function readRaw() {
   const db = await getAdapter();
   const row = db.get(`SELECT data FROM settings WHERE id = 1`);
@@ -77,6 +88,9 @@ function mergeWithDefaults(raw) {
         merged[key] = defVal;
       }
     }
+  }
+  if (merged.requireApiKey === undefined) {
+    merged.requireApiKey = defaultRequireApiKey();
   }
   return merged;
 }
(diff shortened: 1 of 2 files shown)


thatssoheil:
Closing as superseded: maintainer claytontavaresdan already bridged REQUIRE_API_KEY to requireApiKey in #3120 (settingsRepo defaults). His reads the env at module load and only accepts exactly "true"; mine accepted true/1/yes/on at merge time, so cherry-pick envTruthy() if wanted.

PR #3081, opened 2026-08-06
fix(stream): inject stream_options.include_usage for OpenAI-compatible streaming

### LINKED ISSUE #3017: Usage/token tracking shows 0 for streamed responses (STREAM mode), works fine for JSON/non-stream
## Title
Usage/token tracking shows 0 for streamed responses (STREAM mode), works fine for JSON/non-stream

## Description

Requests going through a combo (tested with `opencode` combo → `deepseek-v4-flash-free`) that get relayed in `STREAM` mode consistently log `IN 0 · OUT 0` in Console Log and don't increment `usageHistory`, even though the response itself completes successfully with correct content.

The same API key/combo, when hit with a plain non-streaming JSON request (1 message, no `stream` param), correctly reports token counts (e.g. `IN 5 · OUT 16`) and shows up in the Usage dashboard.

### Repro
1. Add a working combo (e.g. free-tier `opencode`).
2. Send a request with 2+ messages (system + user) via `/v1/chat/completions`, no explicit `stream` param.
3. Console Log shows `STREAM · 2 MSG` → `DONE ... IN 0 · OUT 0`.
4. Send a single-message request instead → Console Log shows `JSON · 1 MSG` → `DONE ... IN 5 · OUT 16` (correct).

### Observation from source
`open-sse/translator/concerns/usage.js` has per-provider extractors (`claude`, `gemini`, `kiro`, `ollama`, `commandcode`) but no explicit passthrough case for OpenAI-shaped upstreams — suspect the trailing SSE usage chunk isn't being captured/parsed for this code path when the response is streamed.

### Impact
- `usageHistory` table only ever recorded the very first successful request for a given key; never incremented again despite 15 subsequent successful requests (confirmed via sqlite `requestDetails` table — all logged as `success`, but `usageHistory` stuck at 1 row).
- Cost estimation and dashboard "Total Requests" become unreliable for API-key-based usage (as opposed to the in-app "Test Connection" button, which uses non-stream JSON and reports correctly).

### Environment
- 9Router v0.5.45, self-hosted via Docker
- Combo: `opencode` (OpenCode Free provider)

thatssoheil:
## What

Injects `stream_options: { include_usage: true }` into upstream OpenAI-compatible requests when streaming, so usage chunks are actually returned by upstreams that require the opt-in (fixes streaming usage showing 0).

Closes #3017

### DIFF
--- a/open-sse/executors/default.js
+++ b/open-sse/executors/default.js
@@ -67,7 +67,7 @@ export class DefaultExecutor extends BaseExecutor {
     super(provider, PROVIDERS[provider] || PROVIDERS.openai);
   }
 
-  transformRequest(model, body) {
+  transformRequest(model, body, stream) {
     const transformed = this.applyJsonSchemaFallback(body);
 
     if (transformed && typeof transformed === "object") {
@@ -76,6 +76,12 @@ export class DefaultExecutor extends BaseExecutor {
         delete transformed.client_metadata;
       }
       stripUnsupportedParams(this.provider, model, transformed);
+      // Ask OpenAI-compatible upstreams to include usage in the final stream
+      // chunk so /v1 streaming requests record real token counts instead of
+      // IN 0 · OUT 0 (issue #3017). Same approach as the iflow executor.
+      if (stream && transformed.messages && !transformed.stream_options) {
+        transformed.stream_options = { include_usage: true };
+      }
     }
 
     return injectReasoningContent({ provider: this.provider, model, body: transformed });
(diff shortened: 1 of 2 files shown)


chisewaguri:
heads up, on the responses to chat path the translated body has no stream field, so just injecting stream_options makes openai compatible upstreams (deepseek) 400 with "stream_options should be set along with stream = true". we set transformed.stream = true alongside the injection, probably worth doing here too so responses clients dont break.

thatssoheil:
Good catch, thanks. The executor-level `stream` flag is derived from `body.stream !== false` (chatCore.js), so streaming-capable clients that omit `stream` (Responses->chat path, `Accept: text/event-stream`) can reach the executor with `stream=true` while the translated body has no `stream` field.

Fixed in becd573f: injection is now gated on `transformed.stream === true`, so `stream_options` only goes out with bodies that actually declare `stream: true`. Added a test covering the Responses->chat case (body without `stream`, executor flag true) - 4/4 pass.

PR #3083, opened 2026-08-06
fix(usage): read cached_tokens from nested prompt_tokens_details

### LINKED ISSUE #2873: [Bug] Streaming path drops cache-read tokens: usage records cached_tokens: 0 and bills cache at full input rate
## Environment

- 9Router version: **v0.5.40** (Docker `decolua/9router:latest`, image revision `79918c7`)
- Bug also present on current `master`
- Affected provider in my case: `codex` / `gpt-5.6-sol` (but the root cause is provider-agnostic — see below)

## Describe the bug

On the **streaming** path, cache-read tokens are silently dropped when usage is persisted. `usageHistory` and `usageDaily` always record `cached_tokens: 0` / `cachedTokens: 0`, even though the upstream response reported a large cache hit and `requestDetails` stored it correctly.

Because cost is computed from the canonicalized usage, **all cache-read input is billed at the full uncached input rate**. For `gpt-5.6-sol` (`input: 5.00` vs `cached: 0.50`) that is a 10× overcharge on the cached portion.

Concretely, one of my rows — `requestDetails.data.tokens` (correct):

```json
{"prompt_tokens":27985,"completion_tokens":60,"total_tokens":28045,
 "prompt_tokens_details":{"cached_tokens":27136}}
```

…and the matching `usageHistory` row (cache lost):

```
promptTokens=27985  completionTokens=60  cost=0.141725
tokens={"prompt_tokens":27985,"completion_tokens":60,"total_tokens":28045,
        "cached_tokens":0,"cache_creation_input_tokens":0}
```

`0.141725 = 27985 × 5.00/1e6 + 60 × 30.00/1e6` — i.e. the 27,136 cached tokens were charged at `input` instead of `cached`. Correct cost is `0.019613`, a **7.2× overcharge on this request**.

Aggregated over one day on my instance, `gpt-5.6-sol` reported **$332.03** where ~84% of prompt tokens were cache hits; the true cost is roughly **$87**.

## Root cause

`buildUsage()` writes cache-read **only** into the nested `prompt_tokens_details.cached_tokens`, never as a top-level `cached_tokens` — [`open-sse/translator/concerns/usage.js#L3-L14`](https://github.com/decolua/9router/blob/master/open-sse/translator/concerns/usage.js#L3-L14):

```js
if (cachedTokens > 0 || cacheCreationTokens > 0) {
  usage.prompt_tokens_details = {};
  if (cachedTokens > 0) usage.prompt_tokens_details.cached_tokens = cachedTokens;
  if (cacheCreationTokens > 0) usage.prompt_tokens_details.cache_creation_tokens = cacheCreationTokens;
}
```

`canonicalizeUsage()` reads **only** the top-level field — [`open-sse/utils/usageTracking.js#L174-L193`](https://github.com/decolua/9router/blob/master/open-sse/utils/usageTracking.js#L174-L193):

```js
// cache_creation HAS the nested fallback:
const cacheCreation = num(usage.cache_creation_input_tokens ?? usage.prompt_tokens_details?.cache_creation_tokens);
...
} else {
  // OpenAI/Gemini path (or already-canonical input): prompt already includes cached_tokens.
  cached = num(usage.cached_tokens);   // ← no nested fallback → always 0 for buildUsage() output
}
```

The nested fallback that was deliberately added for `cache_creation_tokens` is missing for `cached_tokens`.

Full chain for a Codex/Responses stream:

1. [`translator/response/openai-responses.js#L474`](https://github.com/decolua/9router/blob/master/open-sse/translator/response/openai-responses.js#L474) — correctly extracts `input_tokens_details.cached_tokens` and calls `buildUsage({ ..., cachedTokens: cacheReadTokens })`
2. `state.usage` → `onStreamComplete(contentObj, usage, ttftAt)`
3. [`handlers/chatCore/streamingHandler.js#L139`](https://github.com/decolua/9router/blob/master/open-sse/handlers/chatCore/streamingHandler.js#L139) — `saveUsageStats({ ..., tokens: usage })`
4. `handlers/chatCore/requestDetail.js` — `canonicalizeUsage(tokens)` → `cached_tokens: 0`
5. cost computed from that → cache-read billed at full `input` rate

**This is not Codex-specific.** Every provider whose streaming usage flows through `buildUsage()` loses cache-read the same way, including the `claude` and `gemini` extractors in `concerns/usage.js` (their `cachedTokens` also lands nested-only, and `canonicalizeUsage`'s Claude branch keys off `usage.cache_read_input_tokens !== undefined`, which `buildUsage()` output never has). On my instance the `antigravity` provider also shows `cachedTokens: 0` across the board.

The **non-streaming** path is unaffected, because `extractUsageFromResponse()` in `requestDetail.js` explicitly pulls `responseBody.usage.prompt_tokens_details?.cached_tokens` into the top-level field.

## To Reproduce

Against unmodified `master` sources, no server needed:

```js
import { canonicalizeUsage } from "./open-sse/utils/usageTracking.js";
import { buildUsage } from "./open-sse/translator/concerns/usage.js";

const u = buildUsage({ promptTokens: 27985, completionTokens: 60, totalTokens: 28045, cachedTokens: 27136 });
console.log("buildUsage       :", JSON.stringify(u));
console.log("canonicalizeUsage:", JSON.stringify(canonicalizeUsage(u)));
```

Output:

```
buildUsage       : {"prompt_tokens":27985,"completion_tokens":60,"total_tokens":28045,"prompt_tokens_details":{"cached_tokens":27136}}
canonicalizeUsage: {"prompt_tokens":27985,"completion_tokens":60,"total_tokens":28045,"cached_tokens":0,"cache_creation_input_tokens":0}
```

End-to-end: send a repeated large-prefix streaming request through a Codex connection, then compare

```sql
SELECT json_extract(data,'$.tokens.prompt_tokens_details.cached_tokens') FROM requestDetails ORDER BY timestamp DESC LIMIT 1;
SELECT tokens FROM usageHistory ORDER BY id DESC LIMIT 1;
```

The first is non-zero, the second reports `"cached_tokens":0`.

## Expected behavior

`cached_tokens` should survive canonicalization, so `usageHistory` / `usageDaily` record the real cache-read count and cost applies the `cached` rate to it.

## Suggested fix

One line in `canonicalizeUsage()`, mirroring the existing `cacheCreation` fallback:

```diff
-    cached = num(usage.cached_tokens);
+    cached = num(usage.cached_tokens ?? usage.prompt_tokens_details?.cached_tokens);
```

Note the `??` matters — `buildUsage()` omits the key entirely when cache is 0, so `??` keeps the already-canonical passthrough branch idempotent.

Alternatively (or additionally) have `buildUsage()` emit the top-level `cached_tokens` / `cache_creation_input_tokens` alongside the nested details, which would also make the `canonicalizeUsage` Claude-vs-OpenAI discriminator behave predictably for streamed Claude usage.

Happy to send a PR if you'd like.

thatssoheil:
## What

`canonicalizeUsage` now reads cached_tokens from the nested `prompt_tokens_details.cached_tokens` shape that buildUsage writes for OpenAI Responses streaming, so cache-read tokens are not dropped.

Closes #2873

### DIFF
--- a/open-sse/utils/usageTracking.js
+++ b/open-sse/utils/usageTracking.js
@@ -176,6 +176,12 @@ export function canonicalizeUsage(usage) {
   let prompt = num(usage.prompt_tokens ?? usage.input_tokens);
   let cached;
 
+  // buildUsage() (openai-responses streaming) writes cache-read only into the
+  // nested prompt_tokens_details.cached_tokens — treat that as top-level
+  // cached_tokens so streaming usage records cache hits instead of billing
+  // them at full input rate (issue #2873).
+  const nestedCached = num(usage.prompt_tokens_details?.cached_tokens);
+
   // Claude path: prompt excludes cache; cache_read_input_tokens and/or
   // cache_creation_input_tokens are separate. A cache-miss "first write" only
   // carries cache_creation_input_tokens (no cache_read_input_tokens yet), so
@@ -190,7 +196,7 @@ export function canonicalizeUsage(usage) {
     prompt = prompt + cached + cacheCreation;
   } else {
     // OpenAI/Gemini path (or already-canonical input): prompt already includes cached_tokens.
-    cached = num(usage.cached_tokens);
+    cached = num(usage.cached_tokens) || nestedCached;
   }
 
   const result = {
--- a/tests/unit/canonicalize-nested-cache.test.js
+++ b/tests/unit/canonicalize-nested-cache.test.js
@@ -1,0 +1,39 @@
+import { describe, expect, it } from "vitest";
+import { canonicalizeUsage } from "../../open-sse/utils/usageTracking.js";
+
+describe("canonicalizeUsage nested prompt_tokens_details (#2873)", () => {
+  it("reads cached_tokens from nested prompt_tokens_details (buildUsage shape)", () => {
+    const out = canonicalizeUsage({
+      prompt_tokens: 27985,
+      completion_tokens: 60,
+      total_tokens: 28045,
+      prompt_tokens_details: { cached_tokens: 27136 },
+    });
+    expect(out.cached_tokens).toBe(27136);
+    // prompt already includes cache in the OpenAI shape; keep as-is
+    expect(out.prompt_tokens).toBe(27985);
+  });
+
+  it("prefers top-level cached_tokens when both are present", () => {
+    const out = canonicalizeUsage({
+      prompt_tokens: 100,
+      completion_tokens: 10,
+      cached_tokens: 90,
+      prompt_tokens_details: { cached_tokens: 50 },
+    });
+    expect(out.cached_tokens).toBe(90);
+  });
+
+  it("does not break the Claude fold path", () => {
+    const out = canonicalizeUsage({
+      input_tokens: 100,
+      output_tokens: 20,
+      cache_read_input_tokens: 30,
+      cache_creation_input_tokens: 5,
+    });
+    // prompt = input + cacheRead + cacheCreation
+    expect(out.prompt_tokens).toBe(135);
+    expect(out.cached_tokens).toBe(30);
+    expect(out.cache_creation_input_tokens).toBe(5);
+  });
+});


kushaldotdev:
+1 — this is the missing piece for cache stats end-to-end.

Confirming the two-part problem from a real client (pi agent → 9Router → commandcode provider endpoint):

1. `buildUsage()` writes cache-read only into nested `prompt_tokens_details.cached_tokens` (per #2873's root cause), and the client-side parser reads that same nested shape — so **the bug is observable by clients**: the sidebar cache row stays blank and cache-read gets billed at full input rate.
2. My probe of the commandcode provider endpoint confirms the upstream already returns `prompt_tokens_details.cached_tokens` (0 in my tests, but present) — so fixing the canonicalization makes the whole chain work end-to-end.

The added test case (`prompt_tokens_details: { cached_tokens: 27136 }` → `cached_tokens: 27136`) matches the exact shape I see. LGTM — hoping this lands soon, since it also affects billing accuracy (10× overcharge case in #2873).

PR #3086, opened 2026-08-06
fix(auth): allow local JWT-less browser on local-only routes when login disabled

thatssoheil:
## What

Allows JWT-less browser access to local-only routes (e.g. headroom start/stop) when the request is loopback Host+Origin and requireLogin is disabled, while still blocking tunnel/CSRF.

Fixes local dashboard access when login is disabled.

### DIFF
--- a/src/dashboardGuard.js
+++ b/src/dashboardGuard.js
@@ -141,8 +141,16 @@ async function canAccessPublicLlmApi(request) {
 
 async function canAccessLocalOnlyRoute(request) {
   if (await hasValidCliToken(request)) return true;
-  // Browser on host: loopback Host + Origin (blocks tunnel/CSRF) + auth (JWT or requireLogin=false)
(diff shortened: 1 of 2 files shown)


claytontavaresdan:
Verified: `canAccessLocalOnlyRoute()` in `dashboardGuard.js` already allows local JWT-less browser access to local-only routes (headroom, etc.) when `requireLogin` is disabled. The check at line 149 handles this case correctly — local request + requireLogin=false → allow without JWT.

thatssoheil:
Closing as redundant: on master, canAccessLocalOnlyRoute() already allows the JWT-less local browser when requireLogin is disabled - isLocalRequest(request) && await isAuthenticated(request), and isAuthenticated() returns true when settings.requireLogin === false (dashboardGuard.js:166). The behavior this PR adds is already reachable through the existing path, so the change is a no-op. Thanks for the verification, @claytontavaresdan.

PR #3112, opened 2026-08-07
fix(kiro): report real output tokens, stop discarding usable turns, and keep a malformed body from cooling accounts down

nguyenha935:
## Summary
- Count tool-call bytes into `state.totalContentLength`, so a turn whose entire answer is a tool call no longer reports `completion_tokens: 0` through the `/4` estimator in `finish()`.
- Add `KIRO_TRUNCATION_STOP_REASONS` and close a turn that was truncated *after* producing output as `finish_reason: "length"`, keeping the bytes instead of failing the turn.
- Validate tool calls per tool rather than per turn, keep the buffered map on a rejected fragment, and drop the latched early return — one unusable fragment no longer discards the complete calls (and the streamed text) alongside it.
- Forward `cache_read_input_tokens` / `cache_creation_input_tokens` through `kiro-to-claude`, in both the streaming and the non-streaming exit.
- Classify `improperly formed request` / `request_body_invalid` / `invalid_request_error` as `fallback: false`, so a malformed body stops cooling down every account that rejects it.
- Return `null` from the two Kiro request translators when `canonicalizeKiroConversation()` reports an invalid conversation, matching the translator local-fail contract in `chatCore.js`.

## Notes
- Kiro's executor is two layers: `transformEventStreamToSSE()` parses EventStream frames and enqueues OpenAI SSE deltas as they arrive, and `attachIntegrityGate()` wraps it, buffers the whole body, and re-derives a verdict from the reported `stop_disposition` before releasing anything. A transform-level exit that reports `terminal_incomplete` therefore causes the gate to discard bytes the transform had already produced — which is why block 2 has to change what the exit *reports*, not only what it emits.
- Tool arguments are billed output like any other completion bytes, but were never added to `totalContentLength`. With no text in the turn the estimator saw `0`, so `Math.max(1, …)` was not even reached and usage went out as `OUT 0`.
- Of the reasons `stopDisposition()` folds into `terminal_incomplete`, only `model_context_window_exceeded` and `max_tokens` mean "usable as far as it got, then the budget ran out". `cancelled` and `pause_turn` are abandoned turns whose partial content must stay private, so they are deliberately excluded from the new set; the gate is additionally conditioned on `state.chunkIndex > 0`.
- Three separate sites compounded into the tool-loss symptom: a per-turn `try` around `emitTools()`, a `state.tools.clear()` in the frame-loop catch, and a latched early return in the `toolUseEvent` branch. The "nothing usable was produced" check has to run *after* `emitTools()` — before it, the rejected tool was still buffered and `tools.size` was never `0`.
- Both cache spellings are accepted on the `kiro-to-claude` path because the Kiro executor emits the Chat shape (`prompt_tokens_details.cached_tokens`) while passthrough responses use the nested details form.
- A malformed body is a property of the request, not of the account: every account rejects the identical payload the same way. Observed on this instance, 39 of 44 error rows carried `errorCode 400` and 12 Kiro connections were pulled out of rotation within 18 seconds, each costing one upstream call for a guaranteed failure. Same reasoning as the existing `context_length_exceeded` rule.
- The two translator guards are defence-in-depth and are **unreachable by construction** on the current code path: `normalizeTurns()` guarantees strict alternation, coerces every user `content` to a non-empty string, and `flattenAllStructuredTools()` strips structured tools on retry, so `canonicalizeKiroConversation()` cannot return `valid: false`. They are covered in the tests through the exported, non-normalizing `validateKiroConversation()`. Included so the contract is enforced at the boundary rather than assumed.
- One further issue was investigated and deliberately **not** fixed here: `KiroExecutor.transformRequest()` is a passthrough with no request-size guard, so an over-large conversation is only rejected by the upstream. That is upstream-owned behaviour and a local cap would need a limit this repo cannot know; no code change.

## Test Plan
- `CI=true npx vitest run -c tests/vitest.config.js tests/unit/kiro-usage-and-tool-integrity.test.js` -> 16 passed (16), 1 file passed
- `CI=true npx vitest run -c tests/vitest.config.js tests/unit/kiro-terminal-integrity.test.js` -> 2 failed | 68 passed (70); both failures are pre-existing on the unpatched tree (verified by stashing the change set and re-running) and are the two retry-HTTP tests that time out at the 5000 ms default
- Differential full suite against the unpatched tree: 118 failing before, 118 failing after, zero regressions and zero newly-passing tests outside the new file
- `npx eslint` on the 6 changed files -> no findings
- `git diff --check` -> clean
- Deployed to a production instance and verified `service healthy`; the reporting user has since confirmed correct behaviour on the traffic that originally reproduced blocks 1-3

### DIFF
--- a/open-sse/config/codexConstants.js
+++ b/open-sse/config/codexConstants.js
@@ -1,0 +1,8 @@
+export const CODEX_RESPONSES_LITE_HEADER = "x-openai-internal-codex-responses-lite";
+export const CODEX_RESPONSES_LITE_HEADER_VALUE = "true";
+export const CODEX_MODELS_BASE_INSTRUCTIONS = "You are Codex, a coding agent. Follow the user's instructions and use the provided tools to complete the task.";
+export const CODEX_FORWARD_HEADER_ALLOWLIST = Object.freeze([CODEX_RESPONSES_LITE_HEADER]);
+export const CODEX_RESPONSES_LITE_INPUT_TYPES = Object.freeze([
+  "additional_tools", "function_call", "function_call_output", "custom_tool_call",
+  "custom_tool_call_output", "tool_search_call", "tool_search_output", "reasoning", "message",
+]);
--- a/open-sse/config/errorConfig.js
+++ b/open-sse/config/errorConfig.js
@@ -28,6 +28,14 @@ export const DEFAULT_ERROR_MESSAGES = {
   504: "Gateway timeout"
 };
 
+export const CODEX_REQUEST_SCHEMA_ERROR_CODES = new Set([
+  "invalid_request_error",
+  "unknown_parameter",
+  "unsupported_value",
+]);
+
+export const CODEX_REQUEST_SCHEMA_ERROR_PATTERN =
+  /unknown[_ ]parameter|unsupported[_ ]value|(?:invalid|unsupported)[_ ]?(?:parameter|schema)|(?:parameter|schema).*(?:invalid|unsupported)/i;
 // Exponential backoff config for rate limits
 export const BACKOFF_CONFIG = {
   base: 2000,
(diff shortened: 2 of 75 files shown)


nguyenha935:
Closing: this branch was cut on top of three other open patch branches, so the diff showed 75 files instead of the 6 Kiro files it is about. Reopening from a branch based directly on master.

PR #3114, opened 2026-08-07
fix(antigravity): preserve schema-keyword property names

### LINKED ISSUE #2884: [Bug] Schema cleaner treats the properties name-map as a schema node: MCP tools with a parameter named "properties" 400 every Gemini request; parameters named "title" are silently deleted
## Environment

- 9Router version: **v0.5.40** (Docker `decolua/9router:latest`)
- Bug also present on current `master` (`open-sse/translator/formats/gemini.js:299-303` and `:132-153`)
- Affected provider: `antigravity` (and any provider that runs `cleanJSONSchemaForAntigravity`)
- Trigger: OpenAI-format requests whose `tools` come from **MCP servers** (Notion / Linear official MCP), routed to a Gemini model

## Describe the bug

`cleanJSONSchemaForAntigravity()` walks the schema with naive recursive walkers that treat **every** object node as a schema node. But a JSON Schema tree alternates:

```
schema node → (key "properties") → property-name map → (any key) → schema node
```

The property-name map is **not** a schema — its keys are user-defined parameter names. When a tool parameter name collides with a schema keyword, two walkers corrupt the request:

### Bug 1 — parameter named `properties` → whole request 400s

`ensureObjectType()`:

```js
if (obj.properties && !obj.type) obj.type = "object";
```

When a tool has a parameter literally named `properties`, the *property-name map* satisfies this condition (it has a `properties` key and no `type` key), so the walker injects a bogus `"type": "object"` **into the map**. Vertex then rejects the entire request:

```
400 INVALID_ARGUMENT
Invalid value at 'request.tools[0].function_declarations[62].parameters.properties[2].value'
(type.googleapis.com/google.cloud.aiplatform.master.Schema), "object"
```

Notion's official MCP server has 4 such tools (`API-patch-page`, `API-post-page`, `API-create-a-data-source`, `API-update-a-data-source` — Notion pages literally have a `properties` field), so **every Gemini request with Notion MCP attached fails**, no matter which tool the model would have used. The 4 `fieldViolations` in the 400 correspond 1:1 to those 4 tools.

### Bug 2 — parameter named `title` (etc.) → silently deleted

`removeUnsupportedKeywords()` deletes keyword-named keys at *all* levels, including inside the property-name map. A parameter named `title`, `format`, `default`, `const`, … is treated as a schema keyword and dropped. With Linear + Notion MCP, 8 real tools lose their `title` parameter (e.g. `linear/save_issue` can no longer set an issue title). No error — just a silently degraded tool.

## Minimal repro

```js
const schema = {
  type: "object",
  properties: {
    page_id: { type: "string" },
    properties: {              // parameter named "properties" (Notion MCP)
      type: "object",
      description: "Page property values",
      additionalProperties: true
    },
    title: { type: "string" }  // parameter named "title" (Linear/Notion MCP)
  }
};
cleanJSONSchemaForAntigravity(schema);
// schema.properties is now:
//   { page_id: {...}, properties: {...}, type: "object" }  ← bogus entry, Vertex 400s
//   and "title" has been deleted entirely
```

## Suggested fix

Thread an `isSchema` flag through the walkers: children under a schema node's `properties` key are a name map (skip keyword deletion / type injection at that level), the map's children are schema nodes again, arrays pass the flag through.

```js
function ensureObjectType(obj, isSchema = true) {
  if (!obj || typeof obj !== "object") return;
  if (isSchema && !Array.isArray(obj) && obj.properties &&
      typeof obj.properties === "object" && !obj.type) obj.type = "object";
  for (const [k, v] of Object.entries(obj))
    if (v && typeof v === "object")
      ensureObjectType(v, Array.isArray(obj) ? isSchema : isSchema ? k !== "properties" : true);
}
```

(same pattern for `removeUnsupportedKeywords`; the other walkers — `convertConstToEnum`, `flattenAnyOfOneOf`, `mergeAllOf`, `flattenTypeArrays` — are guarded by value-shape checks and happen not to fire on name maps in practice, but would benefit from the same flag.)

## Verification

Ran the full tool lists of Linear's official MCP (57 tools) and Notion's official MCP (24 tools) through the cleaner:

- current code: exactly **4 invalid outputs** (matching the production 400's 4 `fieldViolations`) + **8 tools** with a deleted `title` parameter
- with the fix: **0 invalid outputs**, all parameters preserved; replaying the previously-failing request (all 81 tools attached) against Vertex via antigravity returns 200, and a `save_issue` tool call correctly carries `title`

CBOSSX:
## Summary

- Fixes: Gemini requests fail with HTTP 400 when an MCP tool has a parameter named properties, while parameters such as title are silently removed from the submitted tool schema.
- Root cause: The recursive cleaner treats the JSON Schema properties name map as a schema node, so ensureObjectType injects a bogus type entry into the map and removeUnsupportedKeywords deletes user-defined names that collide with schema keywords.

Closes #2884.

## Regression evidence

- Before: `cd tests && npx vitest run unit/antigravity-schema-cleaner.test.js --reporter=verbose` exited `1`

- After: `cd tests && npx vitest run unit/antigravity-schema-cleaner.test.js --reporter=verbose` exited `0`

## Verification

- `cd tests && npx vitest run unit/antigravity-schema-cleaner.test.js translator/bugs-antigravity.test.js unit/gemini-36-integration.test.js unit/antigravity-stream-options.test.js --reporter=verbose`
- `npx eslint open-sse/translator/formats/gemini.js tests/unit/antigravity-schema-cleaner.test.js`
- `npm run build`
- CI-mode full Vitest comparison: the frozen base and patched tree have identical 129-test failure sets; the patched tree adds one passing regression test.

## Scope

- 2 files changed, +54 / -8 lines

### DIFF
--- a/open-sse/translator/formats/gemini.js
+++ b/open-sse/translator/formats/gemini.js
@@ -135,25 +135,27 @@ export function generateProjectId() {
 
 // Helper: Remove unsupported keywords recursively from object/array
 // Also strips all vendor extension fields (x- prefixed) not supported by Gemini
-function removeUnsupportedKeywords(obj, keywords) {
+function removeUnsupportedKeywords(obj, keywords, isSchema = true) {
   if (!obj || typeof obj !== "object") return;
 
   if (Array.isArray(obj)) {
     for (const item of obj) {
-      removeUnsupportedKeywords(item, keywords);
+      removeUnsupportedKeywords(item, keywords, isSchema);
     }
     return;
   }
 
   for (const key of Object.keys(obj)) {
-    if (keywords.includes(key) || key.startsWith("x-")) {
+    if (isSchema && (keywords.includes(key) || key.startsWith("x-"))) {
       delete obj[key];
       continue;
     }
 
     const value = obj[key];
     if (value && typeof value === "object") {
-      removeUnsupportedKeywords(value, keywords);
+      // `properties` contains user-defined names, not schema keywords. Its
+      // values resume the schema tree one level below the name map.
+      removeUnsupportedKeywords(value, keywords, isSchema ? key !== "properties" : true);
     }
   }
 }
@@ -302,10 +304,18 @@ function flattenTypeArrays(obj) {
 }
 
 // Infer missing type=object when properties exist (Gemini requires explicit type)
-function ensureObjectType(obj) {
+function ensureObjectType(obj, isSchema = true) {
   if (!obj || typeof obj !== "object") return;
-  if (obj.properties && !obj.type) obj.type = "object";
-  for (const v of Object.values(obj)) if (v && typeof v === "object") ensureObjectType(v);
+  if (isSchema && obj.properties && !obj.type) obj.type = "object";
+  if (Array.isArray(obj)) {
+    for (const item of obj) ensureObjectType(item, isSchema);
+    return;
+  }
+  for (const [key, value] of Object.entries(obj)) {
+    if (value && typeof value === "object") {
+      ensureObjectType(value, isSchema ? key !== "properties" : true);
+    }
+  }
 }
 
 // Clean JSON Schema for Antigravity API compatibility - removes unsupported keywords recursively
@@ -396,4 +406,3 @@ export function cleanJSONSchemaForAntigravity(schema) {
 
   return cleaned;
 }
-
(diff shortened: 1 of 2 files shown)


CBOSSX:
Merged current upstream master (17c4cc76877bd1755030a8414f8d0083f48dcccf) into the existing branch in 92c693e4, resolving the Gemini translator conflict while preserving the upstream additions. The original schema-keyword regression still fails on that upstream base and passes on the merged branch. Validation: seven affected Vitest files passed (42 passed, five expected failures), ESLint passed, and the Next.js production build passed.

Issue #2797, opened 2026-07-23
Antigravity image models drop input images (`inlineData`) — image editing silently falls back to text-to-image


nriefff:
The antigravity **image adapter** (`open-sse/handlers/imageProviders/antigravity.js`) supports an input image for editing: it reads `body.image` / `body.images[0]`, converts it via `resolveImageInput()` to a Gemini `inlineData` part, and unshifts it into `parts`.

However, the antigravity **executor** (`open-sse/executors/antigravity.js`) rebuilds the request for image models and keeps **text parts only**, so the input image never reaches the model:

```js
// executors/antigravity.js — image generation branch
if (isImageModel(model)) {
  ...
  for (const c of srcContents) {
    const textParts = (c.parts || []).filter(p => p.text !== undefined).map(p => ({ text: p.text }));
    if (textParts.length > 0) {
      contents.push({ role: c.role || "user", parts: textParts });
    }
  }
```

As a result, an **edit request silently degrades to text-to-image**: the model generates a brand-new image from the prompt alone. There is no error, which makes this hard to notice — e.g. sending a photo of a card with prompt *"change the background to green"* returns a random green-background image unrelated to the input.

The same happens on the chat path (`/v1/chat/completions` with an `image_url` part to an image model), since it goes through the same executor branch.

## Steps to reproduce

```bash
curl -s -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{
    "model": "ag/gemini-3.1-flash-image",
    "prompt": "change the background to bright green",
    "image": "data:image/png;base64,<any image containing distinctive content>"
  }' http://localhost:20128/v1/images/generations
```

**Expected:** the returned image is the input image with a green background (Gemini image models support editing via `inlineData`).
**Actual:** an unrelated, newly generated image.

## Suggested fix

Keep `inlineData` parts alongside text parts in the image branch:

```js
const textParts = (c.parts || [])
  .filter(p => p.text !== undefined || p.inlineData)
  .map(p => p.inlineData ? { inlineData: p.inlineData } : ({ text: p.text }));
```

I verified this one-line change against a live deployment (docker `decolua/9router:latest`, patched the corresponding minified chunk in `.next/server/chunks/`): with it, `ag/gemini-3.1-flash-image` performs a **true edit** — the input image is preserved and only the requested change is applied (text in the input image stays intact). Aspect-ratio suffixes (e.g. `-16x9`) keep working.

[...]

Issue #2014, opened 2026-06-23
Antigravity + 9Router works, but image/vision detection fails


ryankosegeran-byte:
## Summary
Connecting Antigravity to 9Router works for text — requests succeed and responses are returned normally. However, image (vision) detection does not work: the model cannot read or analyze attached images.

## Steps to Reproduce
1. Configure Antigravity to use 9Router as the provider.
2. Send a normal text prompt → works correctly.
3. Send a prompt with an image attachment (vision request).

## Expected
The model reads the image and returns an analysis/description.

## Actual
Text requests succeed, but image detection fails — the image content is not processed.

## Environment
- Client: Antigravity
- Provider: 9Router
- Request type that fails: vision / image input

## Notes
Text-only flow is fully functional, so the issue appears specific to image/multimodal handling through 9Router.

<img width="1919" height="1079" alt="Image" src="https://github.com/user-attachments/assets/1c26070a-758e-4240-9df4-e06ddcf2b194" />

<img width="496" height="702" alt="Image" src="https://github.com/user-attachments/assets/623a73d6-6dea-4520-98cc-55238e112f07" />

PR #3117, opened 2026-08-07
fix(translator): preserve image blocks when routing Claude -> Antigravity

FadhliRajwaaRahmana:
## Summary

Fixes image/vision detection for the **Antigravity** provider when the client is **Claude Code** (or any Anthropic-format client). Today, pasted/attached images are silently dropped — text works, but the model never sees the image.

## Root cause

The generic pipeline pivots through OpenAI (`claude → openai → antigravity`), which is lossy for images in **two** places:

1. **`openai-to-gemini.js` → `wrapInCloudCodeEnvelopeForClaude()`** had **no `CLAUDE_BLOCK.IMAGE` branch** (0 occurrences). Image blocks in a user message were silently discarded for Claude-backed models (`claude-opus-4-6`, `claude-sonnet-4-6`, …).
2. **`claude-to-openai.js`** stringified image blocks *inside* `tool_result` into JSON text (`[{"type":"image",...}]`), so the model received a base64 blob as plain text — not a readable image.

Additionally, `convertOpenAIContentToParts` (`formats/gemini.js`) emitted `inlineData.mime_type` (snake_case) while the rest of the codebase — and the Antigravity API path that works (`imageProviders/antigravity.js`) — reads `mimeType` (camelCase).

## Approach

Add a **direct `claude:antigravity` translator** (`open-sse/translator/request/claude-to-antigravity.js`) that feeds the Anthropic Messages body straight into the Cloud Code envelope when images are present, mirroring how [CLIProxyAPI](https://github.com/router-for-me/CLIProxyAPI)'s `antigravity/claude` translator preserves images:

- **User-message image** → `inlineData` part (`mimeType` camelCase)
- **`tool_result` image** → `functionResponse.parts[].inlineData`; text stays in `response.result`

Requests **without** images fall back to the exact existing double-hop, so text-only traffic is untouched (no regression).

Also normalizes `convertOpenAIContentToParts` to emit `mimeType`, matching the rest of the codebase.

## Verification

- New regression tests in `tests/translator/bugs-antigravity.test.js`:
  - Claude-backed model keeps image in user message as `inlineData`
  - Gemini-backed model uses camelCase `mimeType`
  - image inside `tool_result` lands in `functionResponse.parts`
  - non-image request still goes through the fallback (text preserved)
- `npx vitest run translator/bugs-antigravity.test.js unit/multimodal-drop-lock.test.js unit/file-block-routing.test.js` → all pass
- Updated `golden-request` snapshot (`mime_type` → `mimeType`)
- `npx eslint` on changed files → clean
- Empirically verified: `claude-opus-4-6` + image now produces `inlineData` (was: only text); image in `tool_result` now lands in `functionResponse.parts`

## Related issues

Closes image/vision failures reported in #2014 and #2797 (inlineData drop for image models).


FadhliRajwaaRahmana:
Hi maintainers 👋

Just a friendly bump on this PR. It's been open for over a week now with no review or feedback.

This fix is important because **Claude vision/image support is completely broken when routing through Antigravity** — all image blocks are silently stripped, causing Claude to respond as if no image was provided. Issue #3148 tracks this bug.

The fix is minimal and non-breaking: it adds a direct `claude:antigravity` translator that preserves image content blocks instead of going through the generic OpenAI conversion path that drops them.

I noticed the project has shipped several npm updates (up to v0.5.55) since this PR was opened, so I'd really appreciate it if someone could take a look. Happy to address any feedback or make changes if needed.

Thanks!

PR #3132, opened 2026-08-08
feat(elevenlabs): full TTS panel — v3 audio tags, AI enhance, stability, language override, usage credits

ntdatt812:
Brings the ElevenLabs TTS panel up to what the v3 API actually offers: v3 audio tags, an "Enhance" pass that rewrites a line into directed v3 script, stability/speed/output_format controls, a language override, and the credits balance on the connection card.

Rebased on `v0.5.55`, and **verified end to end against a live ElevenLabs key** — details below.

## The line count is 69% locale files

`+4777` reads alarming, so here is what it actually is:

| | files | lines |
|---|---|---|
| **Routing engine** (`open-sse/`) | 8 | **+271** |
| **Dashboard UI** (`src/`) | 7 | +968 |
| Tests | 2 | +238 |
| Locale literals (32 languages) | 32 | +3302 |

The engine side is the part that can break routing, and it is 271 lines across 8 files. Nothing outside `elevenlabs` changes behaviour; `ttsProviders/index.js` (+4) and `services/usage.js` (+2) are registration lines.

`c570fe33 feat(tts): add Xiaomi MiMo text-to-speech support` is already on master and touches the same set — `TtsExampleCard.js`, `ttsProviders.js`, `sse/handlers/tts.js`, the same 32 locale files — at 44 files, `+396/-20`. This is the same shape, bigger because the panel is bigger. If the 32-locale spread is the objection rather than the code, say so and I will drop the locales.

## Live verification

Driving the PR's own engine path with a real key (short input, so the numbers are byte sizes of returned audio):

```
voices: 21 (fetched, not the fallback roster)
PASS  v3 + stability                                    12582B  format=mp3
PASS  v3 + language_code=vi                             17598B  format=mp3
PASS  multilingual_v2 + language_code=vi (dropped)      15090B  format=mp3
PASS  flash_v2_5 + output_format=mp3_22050_32            4198B  format=mp3
PASS  flash_v2_5 + speed=1.1                            15926B  format=mp3
PASS  bare voice id (default model)                     15926B  format=mp3
```

Two of those are worth calling out.

**The `language_code` guard is load-bearing.** Sending the field to a model that does not declare `langCode` is not ignored upstream, it is a hard failure:

```
POST /v1/text-to-speech/{voice}  {"model_id":"eleven_multilingual_v2", …,"language_code":"vi"}
→ HTTP 400
  {"detail":{"code":"invalid_parameters",
             "message":"Model 'eleven_multilingual_v2' does not support language_code 'vi'.",
             "status":"unsupported_language"}}
```

With the guard the same request returns 200 and 15090B of audio. That is why the field is dropped per-model instead of failing the whole synthesis over an optional hint.

**`output_format` really is applied** — `mp3_22050_32` came back at 4198B against ~15–17KB for the default 128 kbps, i.e. the requested bitrate, which is also why the container is labelled from the requested format rather than sniffed from `Content-Type`.

Then through the dashboard itself, on a throwaway `DATA_DIR`:

- the connection card shows `7,818 / 10,000 credits left · resets 8/26/2026` — the `services/usage/elevenlabs.js` reader against the real account
- clicking the `excited` chip inserts `[excited]` at the caret and the request preview updates to match
- **Run** → `RESPONSE ⚡ 1555ms`, a 2.16 s audio element, and a `HISTORY (1)` entry reading `pNInz6obpgDQGcFmaJgB · eleven_v3 · just now`
- a starter script fills the editor and moves stability to Creative, and the preview follows to `"stability":0`

What is still not demonstrable here: "Enhance with AI" needs an LLM connection configured in 9Router, which the throwaway instance had none of. Its prompt construction is covered by unit tests, not by a live call.

## TtsExampleCard is no longer where the panel lives

`TtsExampleCard.js` is shared by every TTS provider and this work had grown it from 612 to 1313 lines. The v3 tag palette, Enhance vibes and director prompt, language roster, starter scripts and the settings row are pure data plus one presentational component, so the last commit moves them to `elevenlabsPanel.js`:

```
TtsExampleCard.js   | 204 +-------------------
elevenlabsPanel.js  | 213 +++++++++++++++++++++
```

The card keeps all state and handlers — only module-level constants moved, which is why this is safe to do late. I re-ran the whole browser flow above after the move: same panel, same generated audio, same history entry, no console errors.

## Suggested reading order

1. `open-sse/handlers/ttsProviders/elevenlabs.js` (+100) — the request builder, including the `language_code` and `speed` per-model guards
2. `open-sse/config/ttsModels.js` (+48), `config/elevenlabsVoices.js` (+41) — pure data
3. `open-sse/services/usage/elevenlabs.js` (+54) — credits, isolated behind the usage registry
4. `src/app/.../elevenlabsPanel.js` (new, 213) — the panel's presets, skimmable
5. `src/app/.../TtsExampleCard.js` — what is left is the wiring
6. the rest is registration, history persistence, and the connection card

## Suite

`npx vitest run unit translator` on this base: **93 failed before and after, failing sets byte-identical** (`comm` empty both directions). `npx eslint` clean on the changed files — `TtsExampleCard.js` carries 5 pre-existing `react-hooks` warnings, identical in count and rule before and after the extraction.

`unit/elevenlabs-tts.test.js` (+234) covers model resolution, the `language_code` guard, and the usage handler.


ntdatt812:
Hi **@decolua**,

Just wanted to provide a bit more context on this PR.

I've actually been building and using this ElevenLabs integration in my own setup for quite a while, and it's been working well for my use cases.

I decided to contribute it back to 9Router because I thought these improvements could be useful to the wider community as well. I've tested the features extensively in my own environment, including the TTS panel, audio tags, AI enhancement, stability options, language override, credit tracking, and multiple API key support.

Of course, I'm happy to adjust anything that doesn't fit the project's architecture, coding style, or direction. This is my first contribution to this part of the project, so any feedback is very welcome.

Whenever you have a chance, I'd really appreciate a review of the PR.

Thanks for maintaining 9Router and for giving the community a great project to build on! ❤️

photo_2026-08-08_14-16-52
photo_2026-08-08_14-16-56
photo_2026-08-08_14-17-01
photo_2026-08-08_14-17-04

ntdatt812:
Closing this myself. Since it was opened, master gained ElevenLabs support of its own (`open-sse/providers/registry/elevenlabs.js`, the TTS handler, the voices route and the brand icon), so the overlap here is no longer worth reviewing as one change. What is left un-covered upstream is the dashboard side — the panel, TTS history, per-voice usage tracking — but that is a ~1.2k-line feature, which is a lot to ask of a review queue this deep. If a TTS panel is something you want, say so and I will send it as a small, self-contained change instead.

PR #3149, opened 2026-08-08
fix(kiro): resolve broken MITM interception and stream decoding in Kiro IDE v1.0.228+

rm1dev:
Recent updates to Kiro IDE (v1.0.228+) introduced breaking changes that completely broke MITM interception, caused client-side stream decoding crashes, and bypassed model routing. This PR fixes these critical issues:

1. **Fixed Broken Request Interception (Endpoint Migration):**
   Kiro Runtime moved chat requests away from `/generateAssistantResponse` to the root path (`POST /`), identifying operations via the `x-amz-target: KiroRuntimeService.GenerateAssistantResponse` header. Previously, all chat turns were treated as non-chat and passed through unhandled. Added `isChatRequest()` to properly detect and intercept these requests.

2. **Fixed Client-Side Stream Decoding Crashes (Smithy Decoder):**
   The client-side `SmithyMessageDecoderStream` strictly requires the event stream to begin with an `initial-response` frame (`application/x-amz-json-1.0`). Without this frame, Kiro failed to parse the stream and crashed the response. Added `withInitialFrame()` to inject this mandatory frame at stream start.

3. **Fixed Agent/Vibe Mode Bypassing Router:**
   In Agent and Vibe modes, Kiro sends requests with `modelId: "auto"` (and `"simple-task"` for sub-tasks). Because this slot was unmapped, `getMappedModel` returned `null` and leaked requests directly to AWS. Added the `auto` model slot in `cliTools.js` to ensure agent turns are routed as configured.

4. **Target Domain Update:**
   Updated Kiro's MITM domain to `runtime.us-east-1.kiro.dev` matching the new runtime infrastructure.

### DIFF
--- a/src/mitm/config.js
+++ b/src/mitm/config.js
@@ -26,9 +26,27 @@ const TARGET_HOSTS = [
 const URL_PATTERNS = {
   antigravity: [":generateContent", ":streamGenerateContent"],
   copilot: ["/chat/completions", "/v1/messages", "/responses"],
+  // Legacy path form. Kiro IDE 1.0.228+ posts to `/` with x-amz-target instead —
+  // see isChatRequest() for the header-based match.
   kiro: ["/generateAssistantResponse"],
   cursor: ["/BidiAppend", "/RunSSE", "/RunPoll", "/Run"],
 };
+
+/**
+ * Whether this request is a chat turn we should intercept (vs passthrough).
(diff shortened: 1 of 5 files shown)


decolua:
Thanks @rm1dev for the contribution! Reviewed and merged into master. 🙏

Issue #3090, opened 2026-08-06
[Bug] Streaming request details stuck at "[Streaming in progress...]" / 0 tokens when client disconnects before upstream EOF


ZeroOttantotto:
## Summary

Streaming requests (STREAM mode) that the **client disconnects before upstream EOF** leave request details stuck at `[Streaming in progress...]` with `IN 0 · OUT 0` forever. The first save happens immediately (`handleStreamingResponse`), but the completion save (`onStreamComplete`) is never called, so the Usage dashboard / request details show no content and zero tokens even hours later.

## Evidence

Self-hosted v0.5.45, provider `opencode-go` (`ocg/deepseek-v4-flash`). SQLite `requestDetails`:

- **803 total opencode-go rows: 253 permanently `[Streaming in progress...]` with tokens 0/0** (87% of rows from the affected session)
- Pattern by model: `deepseek-v4-pro` rows all completed (442), `deepseek-v4-flash` rows 253/355 stuck — consistent with a client that closes the connection as soon as it has the content it needs (tool call / first useful chunk)
- Non-streaming (`stream:false`) requests complete correctly — different save path, works
- Console never logs the `📊 DONE` line for these requests

## Root cause

In `open-sse/handlers/chatCore/streamingHandler.js`, `handleStreamingResponse` saves a placeholder row immediately, then `onStreamComplete` (from `buildOnStreamComplete`) updates it with content+tokens. `onStreamComplete` is invoked **only inside the `flush()` of the SSE TransformStream** (`open-sse/utils/stream.js`).

`createDisconnectAwareStream` (`open-sse/utils/streamHandler.js`) has:

```js
cancel(reason) {
  streamController.handleDisconnect(reason || "cancelled");
  reader.cancel();   // ← kills the transform stream
  writer.abort();    // ← abort() never runs flush()
}
```

When the client disconnects, `reader.cancel()` + `writer.abort()` destroy the transform stream **before its `flush()` can run** → `onStreamComplete` is never called → the placeholder row is never updated. Also, `handleDisconnect` schedules `abortController.abort()` after 500ms, which tears down the upstream fetch the same way.

## Proposed fix

In `createDisconnectAwareStream.cancel()` (and the pull-catch branch that handles `"already closed"` / `"Invalid state"`), instead of cancelling the reader / aborting the writer, **drain the transform stream to EOF** so `flush()` fires and `onStreamComplete` persists the real content+tokens:

```js
cancel(reason) {
[...]

PR #3175, opened 2026-08-09
fix(stream): finalize interrupted streaming request details

brimob-sowax:
Part of #3090. Implements the semantic gap fix discussed there (marking the placeholder row from the disconnect/error handler).

## Problem

Streaming requests save a placeholder `requestDetails` row at stream start (`[Streaming in progress...]`, tokens 0/0, status `success`). That row is only updated from the SSE transform `flush()` via `onStreamComplete`. `flush()` never runs when:

* the client disconnects (`cancel()` on the response body),
* the upstream stall watchdog aborts (`stream stall timeout`),
* the upstream connection resets midstream (`reader.cancel()` + `writer.abort()` in the error path).

Those rows stay as fake `success` placeholders forever and never reach `usageHistory`. On a busy gateway this was the majority of streaming rows (measured: ~80% of placeholders never finalized over an 8h window).

## Change

`buildOnStreamComplete` (open-sse/handlers/chatCore/streamingHandler.js) now also returns `onStreamAbandoned(reason)`:

* Reuses the same `streamDetailId`, so the existing `INSERT ... ON CONFLICT(id) DO UPDATE` overwrites the placeholder in place (no extra row).
* Sets `status: "cancelled"` and `response.content` / `providerResponse` to `[Streaming interrupted: <reason>]` (reasons: client provided disconnect reason, `stall_timeout`, `stream_error`).
* A `completed` flag guards against overwriting a row that already finalized normally (flush ran first).

chatCore.js wires it into the stream controller's `onDisconnect`/`onError` callbacks; it is only set once the response turns out to be streaming, so nonstreaming paths are untouched.

## Correction to my earlier comment in this issue

My "duplicate row / id mismatch" datapoint (comment on 2026-08-09) was wrong: rechecking timestamps, the placeholder and the "real" row belonged to two different requests (one dead stream + one completed stream). The id propagation already works on v0.5.50. The remaining bug is purely the abandoned placeholder semantics this PR addresses.

## Tests

New `tests/unit/streaming-interrupted-detail.test.js` (4 tests): abandoned finalizes with same id + `cancelled`; no overwrite after normal completion; normal completion unchanged; abandon fires once. All pass. Nearby streaming/request details suites checked: no new failures vs master (the 2 preexisting reds in request-details-tab and 2 in force-stream-config fail identically on a clean checkout).

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -282,12 +282,21 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   const msgCount = translatedBody.messages?.length || translatedBody.input?.length || translatedBody.contents?.length || translatedBody.request?.contents?.length || 0;
   log?.debug?.("REQUEST", `${provider.toUpperCase()} | ${model} | ${msgCount} msgs`);
 
+  // Set when the response turns out to be streaming; called by the stream
+  // controller on disconnect/error to finalize the placeholder requestDetail
+  // row (flush() never runs on those paths).
+  let abandonStreamingDetail = null;
+
   const streamController = createStreamController({
     onDisconnect: (reason) => {
       trackPendingRequest(model, provider, connectionId, false);
+      abandonStreamingDetail?.(typeof reason?.reason === "string" ? reason.reason : "client_disconnected");
       if (onDisconnect) onDisconnect(reason);
     },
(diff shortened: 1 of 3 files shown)


brimob-sowax:
Heads up: this change is fully included in #3348 (rebased on current master, with usageHistory recovery added on top). Applying #3348 alone is sufficient; keeping this one open for reference.

Issue #3024, opened 2026-08-04
Fusion combo strategy triggers "stream_options should be set along with stream = true" error with DeepSeek models


JosuaMnl:
When using the **Fusion** combo strategy, requests to DeepSeek models fail with a 400 error about `stream_options`. The error does not appear when using **Fallback** or **Round Robin** strategies with the exact same combo and models.

## Environment
- 9router version: v0.5.45
- Client: Hermes CLI
- Endpoint: `http://localhost:20128/v1`
- Provider type: custom (direct API)
- OS: Windows 11

## Combo Configuration
- Combo name: GemSeek
- Models in combo:
  - `ds/deepseek-v4-flash`
  - `ds/deepseek-v4-pro`
  - `gemini/gemma-4-31b-it`
  - `gemini/gemini-3.5-flash-lite`
- Strategy tested: Fusion

## Steps to Reproduce
1. Create a combo with the models listed above.
2. Set the combo strategy to **Fusion**.
3. Set this combo as the active model in Hermes CLI.
4. Send any prompt through Hermes.

## Expected Behavior
The Fusion strategy should query all panel models and synthesize one answer, without triggering a provider-level validation error.

## Actual Behavior
The request fails with:
HTTP 400: `[deepseek/deepseek-v4-pro] [400]: {"error":{"message":"stream_options should be set along with stream = true","type":"invalid_request_error"}}`
The same error also occurs with `deepseek/deepseek-v4-flash` when it is queried as part of the same Fusion panel.

## Workaround Found
Switching the combo strategy from **Fusion** to **Fallback** or **Round Robin** resolves the issue immediately, using the same combo and same models.

## Suspected Root Cause
Fusion mode queries all panel models in parallel and likely disables `stream` internally to collect full responses before synthesis. It appears the `stream_options` field (sent by the client) is not stripped from the request before it is forwarded to DeepSeek's API, causing DeepSeek's own validation to reject the request since `stream` is not `true`.

## Additional Context
A similar unanswered report of the same error message exists in Discussion #169, but without any connection to Fusion mode specifically:
https://github.com/decolua/9router/discussions/169#discussioncomment-17784019

claytontavaresdan:
## Correção aplicada — Issue #3024

### ❌ Problema
Fusion combo triggers 'stream_options should be set along with stream = true' with DeepSeek

### ✅ O que foi feito
Injecao automatica de stream_options para o judge model quando body.stream===true

### 📍 Onde no código
[...]

PR #3184, opened 2026-08-09
fix: batch of community bug fixes (SSRF guard, Fusion stream_options, reasoning test probe)

claytontavaresdan:
## Summary

Three concrete bug fixes pulled from the open issue tracker, each with a regression test.

### 1. Security: SSRF guard on `/v1/search` baseUrl override (#3049)
`resolveBaseUrl()` in `open-sse/handlers/search/callers.js` accepted a client-controlled `provider_options.baseUrl` with no validation, allowing internal/metadata SSRF. Now applies the existing `assertPublicUrl()` guard. Blocks `127.0.0.1`, `169.254.169.254`, RFC1918 ranges, and `localhost`.

### 2. Fusion combo strips `stream_options` (#3024)
Fusion runs panel models with `stream: false` but left the client's `stream_options` in the body, so DeepSeek rejected with *"stream_options should be set along with stream = true"*. Now dropped alongside `tools`/`tool_choice`.

### 3. Dashboard "Test" button works for reasoning models (#3010)
The model-test probe hardcoded `max_tokens: 16`, which starves reasoning models (ClinePass/kimi-k3, deepseek-v4-pro, etc.) — their whole budget goes to chain-of-thought and the answer comes back empty, producing a false "no completion choices" failure. Raised to `1024` and added a soft-pass: a `finish_reason: "length"` response with non-empty `reasoning`/`thinking` is treated as a successful connection.

## Test plan
- `tests/unit/ssrf-guard-search-baseurl-3049.test.js` — 8 cases (loopback, metadata, RFC1918, localhost, trailing-slash strip)
- `tests/unit/fusion-strip-stream-options-3024.test.js` — 2 cases (strip verified via mocked panel call)
- `tests/unit/ping-reasoning-models-3010.test.js` — 4 cases (1024 budget, reasoning-only soft-pass, empty failure, normal pass)

## Acceptance criteria
- [ ] SSRF override rejected with 400 (not fetched server-side)
- [ ] Fusion + DeepSeek no longer 400s on `stream_options`
- [ ] Reasoning models report "Test" success

Happy to split into separate PRs if you prefer.

### DIFF
--- a/open-sse/handlers/search/callers.js
+++ b/open-sse/handlers/search/callers.js
@@ -31,6 +31,8 @@
 
 // ── Helpers ─────────────────────────────────────────────────────────────
 
+import { assertPublicUrl } from "../../../src/shared/utils/ssrfGuard.js";
+
 /**
  * Split domain filter into includes / excludes (excludes prefixed with "-").
  * @param {string[]} [domainFilter]
@@ -69,7 +71,15 @@ export function getProviderSetting(params, key) {
  */
 export function resolveBaseUrl(config, params) {
   const override = getProviderSetting(params, "baseUrl");
-  return (override || config.baseUrl).replace(/\/+$/, "");
+  const resolved = (override || config.baseUrl).replace(/\/+$/, "");
+  // SSRF guard: reject client-controlled baseUrl targeting internal/private hosts.
+  // See issue #3049 — residual SSRF via provider_options.baseUrl.
+  try {
+    assertPublicUrl(resolved);
+  } catch (err) {
+    throw new Error(`Invalid baseUrl (SSRF guard): ${err.message}`);
+  }
+  return resolved;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


claytontavaresdan:
Hi @decolua 👋

Wanted to share a batch of small, self-contained bug fixes I put together while dogfooding 9Router. Each one comes with regression tests and a conventional commit — all targeting `origin/master` cleanly (CLEAN mergeable).

What's in this PR:
- **#3049** — SSRF guard on `/v1/search` `baseUrl` override (blocks `127.0.0.1`, `169.254.x`, RFC1918, `localhost`)
- **#3024** — Fusion now strips `stream_options` from panel fan-out (was 400ing on DeepSeek)
- **#3010** — Dashboard "Test" button works for reasoning models (raised probe budget + soft-pass on reasoning-only responses)

All new tests pass; the pre-existing `golden-url-header` translator failures are unrelated (present on `master`).

Completely understand if you'd prefer a different split, scope, or style — happy to adjust, squash, or split further. If anything needs rework before merge, just say the word. 🙏

Thanks for maintaining 9Router — it's been a joy to build on top of.

decolua:
Thanks @claytontavaresdan! Merged the Fusion `stream_options` strip (#3024) and the reasoning-model test probe (#3010) into master, including both regression tests. The SSRF guard for `/v1/search` baseUrl (#3049) was already landed separately via another PR with a protocol check on top, so that part was skipped here to avoid a duplicate. 🙏

Issue #2752, opened 2026-07-21
Qwen models via OpenAI-compatible passthrough (LimitRouter) fail with 400: enable_thinking / thinking_budget not stripped


prajadiputra:
## Problem

When routing Qwen models (e.g. `qwen3.7-plus`) through an OpenAI-compatible passthrough provider like LimitRouter, 9router v0.5.40 injects `enable_thinking: true` and `thinking_budget: 8192` into the request body. LimitRouter (and other strict OpenAI-compatible providers) reject these fields with HTTP 400:

```
{"error":{"message":"2 request validation errors: Extra inputs are not permitted, field: 'enable_thinking', value: True; Extra inputs are not permitted, field: 'thinking_budget', value: 8192","type":"invalid_request_error","code":"invalid_request"}}
```

## Root Cause

In `chunks/52136.js`, the Qwen thinking format adapter unconditionally injects `enable_thinking` and `thinking_budget` for any model whose name contains "qwen":

```javascript
case "qwen": {
    if (g && h) { b.enable_thinking = false; break; }
    b.enable_thinking = true;
    let a = k(i, d.thinkingRange);
    Number.isFinite(a) && a > 0 && (b.thinking_budget = a);
    break;
}
```

This is correct when the downstream provider is Qwen-native (DashScope/Qwen API). But when the downstream is an OpenAI-compatible passthrough like LimitRouter, these fields are not in the OpenAI Chat Completions spec and get rejected.

## Same bug class as #1468 and #1442

- #1468: `context_management` injected → Anthropic-compatible provider rejected → PR #1500 fix strips it
- #1442: `client_metadata` injected → provider rejected
- This issue: `enable_thinking` / `thinking_budget` injected → OpenAI-compatible passthrough rejected

## Expected behavior

9router should strip `enable_thinking` and `thinking_budget` when the downstream provider is `openai-compatible-chat` (passthrough), similar to how PR #1500 strips `context_management` for Anthropic-compatible providers.

Alternatively, the thinking format adapter should only apply when the provider is Qwen-native (`qwen` provider type), not when routing through `openai-compatible-chat`.

## Reproduction

```bash
# Via 9router with limit/qwen3.7-plus combo:
# 9router injects enable_thinking → LimitRouter returns 400

# Direct to LimitRouter (works without enable_thinking):
curl https://limitrouter.com/v1/chat/completions \
  -H "Authorization: Bearer $KEY" \
[...]

PR #1500, opened 2026-05-27
fix: strip Claude context management for compatible providers

Delcado19:
## Summary
- strip Claude Code context_management before forwarding requests to anthropic-compatible providers
- keep context_management intact for the first-party Claude provider
- add focused regression coverage for the compatibility behavior

Fixes #1468.

## Validation
- node --check open-sse/translator/helpers/claudeHelper.js
- npx vitest run --config ./tests/vitest.config.js tests/unit/claude-context-management.test.js
- git diff --check
- npm run build

Issue #1468, opened 2026-05-26
Claude Code fails with "context_management: Extra inputs are not permitted" on anthropic-compatible endpoint


ohahe52-dot:
## Problem
When using Claude Code with the 9router anthropic-compatible endpoint, requests fail with error 400:

context_management: Extra inputs are not permitted

## Details
- Claude Code automatically sends a `context_management` field in API requests
- The gateway returns 400 error instead of ignoring the unknown field
- Same model works fine with Codex CLI because it sends simpler requests

## Expected behavior
Gateway should forward or ignore unknown fields instead of rejecting the entire request

## Workaround
None available on client side — Claude Code cannot disable `context_management` field

## Environment
- Tool: Claude Code (latest)
- Endpoint: anthropic-compatible gateway
- Error: `context_management: Extra inputs are not permitted`

LorhanSohaky:
Up!

Issue #1442, opened 2026-05-26
Opencode Go kimi-2.6 Model Error


yanpaing007:
[13:57:13] [PENDING] END (ERROR) | provider=opencode-go | model=kimi-k2.6
[ERROR] [400]: {"error":{"object":"error","type":"invalid_request_error","code":"invalid_request_error","message":"Error from provider: Extra inputs are not permitted, field: 'client_metadata'"}}

PR #3188, opened 2026-08-09
fix: strip Qwen thinking params for OpenAI-compatible passthrough (#2752)

claytontavaresdan:
## Summary

Qwen models routed through an OpenAI-compatible passthrough provider (LimitRouter, etc.) fail with HTTP 400 because 9Router's Qwen thinking adapter unconditionally injects `enable_thinking` / `thinking_budget` into the request body. These fields are not in the OpenAI Chat Completions spec and strict passthrough providers reject them.

### Fix
Extended `stripUnsupportedParams` in `open-sse/translator/concerns/paramSupport.js` to drop `enable_thinking` and `thinking_budget` (alongside `reasoning_effort` / `reasoning`) for `openai-compatible-*` / `custom` providers. Qwen-native providers keep these fields untouched.

Also fixed the provider-match logic to accept regex matchers (the existing `/openai-compatible|custom/i` rule was silently skipped because the check only compared strict equality).

### Test plan
- `tests/unit/param-support.test.js` — 2 new cases: drop for `openai-compatible-chat-*` passthrough, preserve for `qwen` native

## Acceptance criteria
- [ ] Qwen models via LimitRouter / OpenAI-compatible passthrough no longer 400 on `enable_thinking`
- [ ] Qwen-native provider still receives thinking params

Same bug class as #1468 / #1442 (PR #1500).

### DIFF
--- a/open-sse/translator/concerns/paramSupport.js
+++ b/open-sse/translator/concerns/paramSupport.js
@@ -21,6 +21,11 @@ const STRIP_RULES = [
   // "integer above maximum value, expected <= 32768". Pin an explicit endpoint cap;
   // min() with the model ceiling still applies if a variant's own limit is lower.
   { provider: "volcengine-ark", match: /kimi/i, maxOutputCap: 32768, clampToModelMaxOutput: true },
+  // Custom OpenAI-compatible providers (Codex App, LimitRouter, etc.) reject
+  // thinking params that aren't in the OpenAI Chat Completions spec. Drop them to
+  // avoid HTTP 400 "Extra inputs are not permitted: enable_thinking / thinking_budget"
+  // and "Unsupported parameter: reasoning_effort". #2752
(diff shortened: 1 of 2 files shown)


claytontavaresdan:
Hi @decolua 👋

Small fix for Qwen models routed through OpenAI-compatible passthrough providers (LimitRouter, etc.).

What's in this PR:
- **#2752** — Qwen's thinking adapter unconditionally injects `enable_thinking`/`thinking_budget`, which strict passthrough providers reject with HTTP 400. Now stripped for `openai-compatible-*` / `custom` providers (Qwen-native keeps them). Also fixed the provider-match logic to accept regex rules (the existing `/openai-compatible|custom/i` rule was silently skipped on strict-equality check).

Ships with regression tests; pre-existing translator failures are unrelated (present on `master`).

Same bug class as #1468 / #1442 (PR #1500). Happy to adjust or split if you prefer — and if anything needs rework before merge, just let me know. 🙏

Thanks for maintaining 9Router!

Issue #2876, opened 2026-07-28
[BUG] Fusion doesn't really work


AlexDochioiu:
```console
[16:07:42] ⚠️ [AUTH] Claude locked modelLock_claude-fable-5 for 30s [400]
❌ claude [400]: [400]: {"type":"error","error":{"type":"invalid_request_error","message":"This model does not suppor
[16:07:42] ⚠️ [FALLBACK] ⇄ ACC:Claude UNAVAILABLE (400) → NEXT ACCOUNT
[16:07:42] ⚠️ [AUTH] claude | all 1 accounts locked for claude-fable-5 (reset after 30s) | lastError=[400]: {"type":"error","error":{"type":"invalid_re
[16:07:42] ⚠️ [CHAT] [claude/claude-fable-5] [400]: {"type":"error","error":{"type":"invalid_request_error","message":"This model does not support assistant message prefill. The conversation must end with a user message."},"request_id":"req_011CdUWNSoH4rQcmfUVKfdr3"} (reset after 30s)
[16:07:58] 🟣 📊 DONE 17972ms · IN 108959 · OUT 513
[16:07:58] ℹ️ [FUSION] fan-out collected in 17980ms
[16:07:58] ⚠️ [FUSION] Panel cc/claude-fable-5 failed {"status":400}
[16:07:58] ℹ️ [FUSION] Panel cx/gpt-5.6-sol(max) ok (386 chars)
[16:07:58] ⚠️ [FUSION] Panel combo_grok-4.5 returned empty content
[16:07:58] ℹ️ [FUSION] Only cx/gpt-5.6-sol(max) succeeded — answering directly (no fusion)
```

I love the idea of fusion as part of 9 router. However, most models constantly just fail.

P.s. All of them work fine on occasions. None of the used models are just constantly failing. But seems that at no turn they all really work together.

Paololalap:
Same on my end... Fusion is not working properly

PR #3189, opened 2026-08-09
fix(fusion): trim trailing assistant messages so panel ends on user turn (#2876)

claytontavaresdan:
## Summary

Fusion combos fail on Claude/Anthropic panels with `400 This model does not support assistant message prefill. The conversation must end with a user message.` (#2876). The Fusion fan-out forwards the full inbound `messages` array to each panel model, and when the client's request ends on an assistant message (e.g. it echoed a partial assistant reply), Claude rejects it.

### Fix
Added `trimTrailingAssistant()` in `open-sse/services/combo.js`, called on the panel body's `messages` / `input` after `flattenToolHistory`. It removes trailing `assistant` roles so the panel request terminates on a user turn. If stripping would leave an empty array (all-assistant input), it falls back to the original so we never send an empty `messages` array.

This complements the earlier Fusion fix (#3024, strip `stream_options`).

### Test plan
- `tests/unit/fusion-trim-trailing-assistant-2876.test.js` — 5 cases (user-ending untouched, single/multi trailing assistant trimmed, all-assistant fallback, empty/null input)

## Acceptance criteria
- [ ] Fusion panel requests no longer 400 on Claude with "must end with a user message"
- [ ] Normal (user-ending) conversations unaffected

Happy to split if you prefer separate PRs.

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -58,6 +58,21 @@ function flattenToolHistory(messages) {
     });
 }
 
+// Remove trailing assistant messages so the conversation ends on a user turn.
+// Anthropic/Claude reject requests whose last message is from the assistant
+// ("conversation must end with a user message"). When a client echoes a partial
+// assistant reply into the next request, trim those trailing assistant roles. #2876
+export function trimTrailingAssistant(messages) {
+  if (!Array.isArray(messages) || messages.length === 0) return messages;
+  const out = messages.slice();
+  while (out.length > 0 && out[out.length - 1]?.role === "assistant") {
+    out.pop();
+  }
+  // If we stripped everything (all-assistant input), keep at least the last user-ish
+  // msg by falling back to the original so we never send an empty array.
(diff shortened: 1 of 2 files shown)


claytontavaresdan:
Hi @decolua 👋

Fix for a Fusion reliability issue I hit while testing combo routing.

What's in this PR:
- **#2876** — Fusion panel requests failed on Claude/Anthropic with `400 This model does not support assistant message prefill. The conversation must end with a user message.` because the fan-out forwarded the full inbound `messages` array (which can end on an assistant turn). Added `trimTrailingAssistant()` so the panel body terminates on a user turn. (Complements the earlier Fusion fix in #3184 that strips `stream_options`.)

Ships with regression tests; pre-existing translator failures are unrelated (present on `master`).

Happy to adjust or split if you prefer — and if anything needs rework before merge, just say the word. 🙏

Thanks for maintaining 9Router!

PR #3201, opened 2026-08-10
fix(providers): add llm7 to provider test support

cokkyturnip:
Add `llm7` case to `testApiKeyConnection()` in `testUtils.js` so the provider test no longer returns "Provider test not supported".

The test hits `https://api.llm7.io/v1/models` (from the llm7 provider registry) with Bearer auth and reports validity based on the response status.

Tested locally: provider test now passes after rebuilding and restarting the service.


decolua:

PR #3213, opened 2026-08-10
fix(auto-ping): select Codex model from live catalog

### LINKED ISSUE #3212: fix(auto-ping): select Codex model from live account catalog
## Problem

Codex auto-ping used a fixed `gpt-5.5` request to begin the next 5-hour quota window. Codex model access is account-specific and changes over time, so a fixed model can be unavailable for an otherwise valid account.

## Fix

- Read the authenticated account's live Codex model catalog from the official `GET /backend-api/codex/models` endpoint.
- Select the catalog's preferred model from API-supported entries.
- Send the tiny auto-ping request only when the catalog returns a usable model.
- Do not infer model access from Free, Plus, or Pro plan labels.
- Update Codex auto-ping help text so it no longer claims a fixed model is used.

## Verification

- Catalog parser filters unsupported models and respects catalog preference order.
- Auto-ping uses the selected live model.
- Empty catalog skips the upstream ping.

emi-ran:
## Summary

Makes Codex auto-ping choose a model dynamically from each account's live Codex catalog instead of sending a fixed `gpt-5.5` request.

Closes #3212.

## Behavior

- Calls the authenticated Codex `models` catalog with the connection's account binding.
- Filters models not supported in the API.
- Uses the catalog's preferred model for the tiny request that starts the next 5-hour window.
- Skips the ping when no supported model is returned.
- Removes fixed-model wording from Codex auto-ping tooltips.

## Verification

- `npx vitest run unit/quota-auto-ping.test.js unit/codex-models.test.js`
- `git diff --check`

Lint has no errors. Existing React hook dependency warnings remain in the two touched UI files.

### DIFF
--- a/open-sse/providers/registry/codex.js
+++ b/open-sse/providers/registry/codex.js
@@ -40,6 +40,8 @@ export default {
     },
     usage: {
       url: "https://chatgpt.com/backend-api/wham/usage",
+      modelsUrl: "https://chatgpt.com/backend-api/codex/models",
+      clientVersion: "0.136.0",
       resetCreditsUrl: "https://chatgpt.com/backend-api/wham/rate-limit-reset-credits",
       resetCreditsConsumeUrl: "https://chatgpt.com/backend-api/wham/rate-limit-reset-credits/consume",
     },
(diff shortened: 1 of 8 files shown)


emi-ran:
@decolua

PR #3221, opened 2026-08-10
fix(chat): key error state to its model, and preserve the upstream status class

haumanto:
## Problem

Two related defects in how a failed request is classified and reported. They are in one PR because the second is only safe to fix once the first is.

### 1. An error from one request is served to a different, unrelated request

Model locks are per-model (`modelLock_${model}`), but `lastError` and `errorCode` are flat **per-connection** fields, overwritten by whichever model failed on that connection most recently. `getProviderCredentials` then hands that stored error to any request whose model happens to be locked on the same connection.

Observed in production. A throwaway probe for a nonexistent model produced a 401 ModelError. Three minutes later, a live conversation on a **different** model was served that probe's error for the whole backoff window:

```
06:56:09  503 - {'error': {'message':
  '[opencode-go/kimi-k2.7-code] [401]: {"type":"error","error":{"type":"ModelError",
  "message":"Model does-not-exist-xyz is not supported"}} (reset after 26s)'}}
06:56:12  ... same body ... (reset after 22s)
06:56:18  ... same body ... (reset after 16s)
```

Three things to note:

- The envelope is correctly tagged `[opencode-go/kimi-k2.7-code]`; the body inside describes a model that appears nowhere in the caller's config.
- `reset after` counts **down** across the retries — one shared timer replaying a stored body, not three fresh upstream attempts.
- The stored error crossed request boundaries: created by a different process, for a different model.

On a shared router this is a cross-tenant fault — anyone who fat-fingers a model name briefly degrades everyone else on the same key. It also misdirects diagnosis: the 401 made the caller report "authentication failed" for credentials that were verified fine against `/v1/chat/completions` and `/v1/models`.

### 2. Upstream 4xx reported as 503

Permanent failures were surfacing as transient. Both of these returned 503:

```
temperature the model rejects   ->  upstream said 400
model that does not exist       ->  upstream said 401 (ModelError)
```

503 means "retry later", so clients spend their whole retry budget on hopeless requests. There is a subtler cost: clients keep reactive repair paths keyed to error class — on a 4xx naming a rejected parameter, drop it and retry once — and a 5xx bypasses that logic entirely, suppressing the automatic recovery that would have fixed the request unnoticed.

## Fix

**Key the error state to its model.** Record `lastErrorModel` alongside the error, and surface the stored error only when it provably belongs to the model being asked about; otherwise return a generic unavailable message. Also select the connection that actually owns the earliest expiry rather than the first locked one, so the countdown and the error can no longer come from different accounts.

**Preserve the upstream class.** `clientStatusForUpstream` keeps 4xx as 4xx and 5xx as 5xx, because the contract clients depend on is simply *4xx means stop, 5xx means retry*.

Unknown-model errors are the one deliberate re-mapping, normalised to **404**: providers report them as 400, as 404, and as 401-with-a-ModelError-body, and passing that 401 through is what makes good credentials look broken. `createErrorResult` gained a separate client status so internal classification (fallback rules, cooldowns) keeps seeing the true upstream code.

**A rejected parameter no longer cools down the account.** Same harm as defect 1 within a single model: a bad `temperature` locked the model, so the next caller's good request failed for the cooldown window. Matched on the *inner* error class, because this provider wraps transient socket errors and permanent parameter errors in the same `Upstream request failed:` prefix — the bracketed class after it is the only thing separating them:

```
permanent : Upstream request failed: [invalid_request_error] invalid temperature: ...
transient : Upstream request failed: connection reset by peer
```

## Verification

The leak is timing-dependent and did not reproduce on demand (ten attempts), so it was verified by **forcing the exact reported state** — a stale error for model A, a lock on model B:

| check | result |
|---|---|
| model A's error served for model B | withheld, generic message |
| model A's error served for model A | still surfaced, status preserved |

Live matrix against a real provider:

| case | expected | got |
|---|---|---|
| temperature rejected by model | 400 | 400 |
| model does not exist | 404 | 404 |
| `max_tokens: 900000` | 400 | 400 |
| malformed JSON body | 400 | 400 |
| good request straight after two bad ones | 200 | 200 |
| bogus model name leaked anywhere | no | no |

`npx eslint` clean. Full vitest suite identical to a clean-HEAD baseline (128 failed / 1616 passed both before and after — the suite is not green on a plain checkout, so this is a baseline comparison, not zero).

## Note

This replaces the previous contents of this PR, which mapped every non-429 4xx to 503. That was the wrong lever: the mislabelled statuses it was chasing were defect 1 replaying stale state, and flattening the class made permanent failures look retryable. Fixing the root cause let the blanket mapping be removed.

The 404 normalisation for unknown models is the main judgement call and is easy to change if you would rather pass the provider's status through.

### DIFF
--- a/open-sse/config/errorConfig.js
+++ b/open-sse/config/errorConfig.js
@@ -56,7 +56,77 @@ const COOLDOWN = {
  *   - cooldownMs: fixed cooldown duration
  *   - backoff: true = use exponential backoff (rate limit)
  */
+/**
+ * Phrases that identify a permanently wrong model name, whatever status the
+ * provider chose to attach to it. Providers are wildly inconsistent here: the
+ * same class of failure arrives as 400 ("model is not supported"), as 401 with
+ * a ModelError body, or as 404. Clients key their retry behaviour off the
+ * status, so these are normalised to one permanent status rather than passed
+ * through as an auth failure the caller cannot act on.
+ */
+export const PERMANENT_MODEL_ERROR_PATTERNS = [
+  "model is not supported",
+  "model not supported",
+  "model not found",
+  "model does not exist",
+  "unknown model",
+  "invalid model",
+  // Matches a lowercased `"type":"ModelError"` body, which is how at least one
+  // provider reports an unknown model — on a 401, of all statuses.
+  "modelerror",
+];
+
+/**
+ * Regex matchers for the same class, needed where the model NAME sits in the
+ * middle of the phrase: "Model does-not-exist-xyz is not supported". A plain
+ * substring cannot span that.
+ */
+export const PERMANENT_MODEL_ERROR_REGEXES = [
+  /\bmodel\s+\S+\s+is\s+not\s+supported\b/,
+  /\bmodel\s+\S+\s+(?:does\s+not\s+exist|not\s+found)\b/,
+];
+
+/**
+ * Request-scoped failures that are permanent but are NOT about the model — a
+ * parameter the model rejects, for instance. These must not cool down the
+ * account: the request is the caller's fault, so retrying on another account
+ * repeats the same failure, and locking the model makes one client's bad
+ * parameter break every other client's good request for the cooldown window.
+ *
+ * Matched on the INNER error class, because this provider wraps everything in
+ * "Upstream request failed:" — including genuinely transient socket errors. The
+ * bracketed class after that prefix is what separates them:
+ *   permanent : "Upstream request failed: [invalid_request_error] invalid temperature: ..."
+ *   transient : "Upstream request failed: connection reset by peer"
+ * Matching "invalid_request_error" alone would wrongly catch the transient case,
+ * whose envelope also carries that type.
+ */
+export const PERMANENT_REQUEST_ERROR_REGEXES = [
+  /upstream request failed:\s*\[invalid_request_error\]/,
+];
+
+/**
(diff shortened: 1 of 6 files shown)


haumanto:
Force-pushed a rewrite of this branch.

Further production evidence showed the original approach here was treating a symptom. The mislabelled statuses this PR was chasing turned out to be stale per-connection error state being replayed across unrelated requests — one client's probe for a bogus model being served to another client's live conversation. Fixing that root cause meant the blanket 4xx→503 mapping could be removed entirely, which also restores the 4xx/5xx contract clients rely on.

The PR title and description have been updated to match. Same base, still independent of my other open PRs.

haumanto:
Superseded by #3771, which carries this same change plus two things it needs.

Running this in production surfaced a failure mode in the approach here. Preserving the upstream class is right, but a **bodyless 404** from a gateway hop is not a model error, so it cools the account down — and the preserved 404 then tells the client "this will never work" about a state that clears itself:

```
404  [codex/gpt-5.6-luna] [404]: HTTP 404 (reset after 1m 58s)
```

The model was in `/v1/models` throughout and served the request two minutes later. A client classifying on status burned three retries in six seconds against a two-minute breaker.

#3771 adds `clientStatusForBreakerOpen` for that one case, extends the same correction to the combo path, and adds 19 tests that are shown non-vacuous in both directions by mutation. Closing this in favour of that one so there is only one branch to review.

PR #3222, opened 2026-08-10
fix(stream): synthesize a terminal when upstream drops mid-response

haumanto:
## Problem

An upstream dying mid-stream reaches the same EOF path as a completed one, so the client stream simply closes:

```js
if (done) {
  streamController.handleComplete();
  controller.close();
  return;
}
```

The caller receives a truncated body with no `finish_reason` and no error, and cannot tell "the model stopped" from "the connection died". A client waiting for a terminal event blocks until its own timeout. Observed: one such stall wedged a turn for 11 minutes and needed a gateway restart, with sockets left in `CLOSE-WAIT`. Downstream logs it as:

```
WARNING agent.chat_completion_helpers: Stream ended with no finish_reason after
delivering text with no tool calls; treating as a mid-stream drop.
```

The network-reset path had the same hole: `emitTerminal` no-ops unless `onAbortTerminal` is configured, which only the Responses passthrough sets. So an OpenAI or Claude stream cut by `ECONNRESET` also closed silently.

## Fix

Both paths now append an explicit error frame plus a terminator, keeping whatever partial content already arrived.

A false positive here would be worse than the bug, since decorating a **healthy** stream breaks working traffic. So detection is deliberately narrow:

- Keyed on the real terminal marker per format (`finish_reason`, `message_stop`, `response.completed`/`failed`), **not** on `[DONE]`. A live provider here emits `finish_reason` and never sends `[DONE]`, so keying on `[DONE]` would have flagged every good stream as dropped.
- Markers split across chunk boundaries are joined via a bounded tail.
- Formats whose terminal is ambiguous (gemini, kiro, cursor, ollama, ...) get **no tracker at all** and keep their existing behaviour byte for byte.
- Suppressed once a real terminal has gone out, so a completed stream is never decorated and never double-terminated.

The Responses passthrough reuses the existing `buildAbortedResponsesTerminalBytes`, so both paths emit an identical terminal.

## Verification

28 targeted tests driving real chunks through the real close path:

- healthy: `finish_reason` alone, `finish_reason` + `[DONE]`, `[DONE]` alone, `tool_calls`, Claude `message_stop`, Responses `response.completed` — none decorated
- markers split mid-token across two chunks — still detected, no false positive
- truncated: error frame appended, partial content preserved, reported as error rather than complete
- `ECONNRESET` mid-stream: error frame appended; after a real terminal: nothing appended, exactly one `[DONE]`
- gemini/kiro/cursor/ollama: no tracker, output byte-identical

Live after deploy: a normal stream returns 200 with its `finish_reason` and no added error, a Claude-format stream returns 200 with its `message_stop` and no added error, exactly one `[DONE]`, non-streaming unaffected.

`npx eslint` clean. Full vitest suite identical to a clean-HEAD baseline (128 failed / 1616 passed both before and after).

## Note

The format list is the reviewable part. I only added trackers for formats whose terminal marker I could confirm; extending it to gemini and the rest is straightforward if you know their terminals well enough to be certain.

### DIFF
--- a/open-sse/handlers/chatCore/streamingHandler.js
+++ b/open-sse/handlers/chatCore/streamingHandler.js
@@ -5,6 +5,7 @@ import { pipeWithDisconnect } from "../../utils/streamHandler.js";
 import { PROVIDERS } from "../../config/providers.js";
 import { STREAM_STALL_TIMEOUT_MS } from "../../config/runtimeConfig.js";
 import { buildAbortedResponsesTerminalBytes } from "../../utils/responsesStreamHelpers.js";
+import { createTerminalTracker } from "../../utils/streamTerminal.js";
 import { buildRequestDetail, extractRequestConfig, saveUsageStats, formatDoneLine } from "./requestDetail.js";
 import { saveRequestDetail } from "@/lib/usageDb.js";
 import { SSE_HEADERS_CORS as SSE_HEADERS } from "../../utils/sseConstants.js";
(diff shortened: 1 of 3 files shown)


chisewaguri:
the terminal tracker gets created with targetFormat but it watches the client facing stream, which is sourceFormat after translation. so for a /v1/responses client it never sees response.completed and fires "upstream stream ended without a terminal event" on every request. content is fine, but the fake error also stops request detail saving. switching to createTerminalTracker(sourceFormat) fixed it for us.

haumanto:
@chisewaguri you are right, and thank you — this was a real bug, not a nitpick.

Confirmed the mechanism: `buildTransformStream` translates `targetFormat → sourceFormat` (`translateResponse(targetFormat, sourceFormat, …)`, `initState(sourceFormat)`), so the frames the tracker inspects are already in the client's format while I was keying it on the provider's. A `/v1/responses` client never shows `response.completed` after translation, the tracker saw no terminal, and appended a spurious "ended without a terminal event" frame to healthy streams. Precisely the false positive this change was supposed to avoid.

Pushed a fix. `buildTransformStream` now returns the format it emits alongside the stream, because its three branches emit three different ones and a caller cannot pick correctly without re-deriving the branch condition:

| branch | emits |
|---|---|
| codex translation | `codexTarget` |
| translation | `sourceFormat` |
| passthrough | `targetFormat` |

Worth naming why my tests missed it: they passed a format straight into `createTerminalTracker` with matching frames. That proves the tracker works and says nothing about whether the caller hands it the right format — I tested the unit and not the wiring. The new check drives the real path and asserts each branch reports its own format.

Verified live across all four client shapes (`/chat/completions`, `/messages`, `/responses`, and an opencode-go provider): none carry an injected frame now.

PR #3227, opened 2026-08-11
fix(qoder): detect billing blocks at stream start, return 403 for failover

brimob-sowax:
**Issue**: Qoder billing errors (quota exhausted, queue throttle) return HTTP 200 + error text wrapped in SSE body → piped to chat as model response → no failover. Pool stuck on dead keys requiring manual disable.

**Root cause**: `wrapQoderSSE` treats non-200 `statusCodeValue` as chat content instead of error. chatCore never sees failure, connection stays active.

**Fix**: Peek first SSE frame; if `statusCodeValue != 200` + billing signature (code 112/10605 or `pricingUrl`), return synthetic 403 response so chatCore marks connection unavailable and triggers combo fallback.

**Test coverage**: `unit/qoder-billing.test.js` (10 new tests), existing qoder suite updated for async `wrapQoderSSE` (54 tests pass, zero regression).

**Billing signatures**:
- code 112: quota exhausted
- code 10605: queue throttle
- pricingUrl field

Production verified: pool with 27/30 unavailable keys, error leaked to chat.

**Files changed**: 3 (+245/-12)
- `open-sse/executors/qoder.js`: peek logic + async wrapQoderSSE
- `tests/unit/qoder-billing.test.js`: new suite
- `tests/unit/qoder.test.js`: async call updates

### DIFF
--- a/open-sse/executors/qoder.js
+++ b/open-sse/executors/qoder.js
@@ -216,6 +216,52 @@ async function buildQoderRequestBody({ model, body, credentials, log, proxyOptio
 }
 
 /**
+ * Check if a qoder error message indicates a billing/quota block.
+ * Signatures: code 112 (quota exhausted), code 10605 (queue throttle), pricingUrl field.
+ */
+function isBillingBlock(inner) {
+  if (!inner || typeof inner !== "string") return false;
+  const lowerMsg = inner.toLowerCase();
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @brimob-sowax for the contribution! Reviewed and merged into master. 🙏

PR #3236, opened 2026-08-11
fix(responses): don't close message on empty tool_calls array

### LINKED ISSUE #3234: /v1/responses sends output_text.done after the first delta on cbcn reasoning models
When streaming a reasoning model through `/v1/responses`, 9router sends `response.output_text.done` after the first content delta, then keeps sending the rest of the `output_text.delta` events. The `done` event's `text` only has the first token, so a spec-compliant client truncates the answer to one token.

`/v1/messages` handles the same model fine, so this is specific to the Responses serializer.

**Environment**
- 9router local instance (`http://localhost:20128`), also seen on the hosted gateway
- Client: Codex CLI v0.147.0 (`wire_api = "responses"`)
- Affected: `cbcn/kimi-k3`, `cbcn/glm-5.2`
- Not affected: `ds/deepseek-v4-flash` (reasoning, works), `ds/deepseek-chat` (non-reasoning)
- So this looks specific to the `cbcn/` reasoning models, not reasoning models in general.

**Reproduce (raw curl)**
```bash
curl -N -X POST http://localhost:20128/v1/responses \
  -H "Authorization: Bearer $KEY" \
  -d '{"model":"cbcn/kimi-k3","input":"Reply with exactly: codex-ok","stream":true}'
```

Content events, `cbcn/kimi-k3` (broken):
```
seq 61  output_text.delta  'cod'
seq 62  output_text.done   text='cod'
seq 65  output_text.delta  'ex'
seq 66  output_text.delta  '-'
seq 67  output_text.delta  'ok'
```

`ds/deepseek-chat` (correct):
```
seq 5  output_text.delta  'code'
seq 6  output_text.delta  'x'
seq 7  output_text.delta  '-'
seq 8  output_text.delta  'ok'
seq 9  output_text.done   text='codex-ok'
```

Same model through `/v1/messages` (correct):
```bash
  -H "Authorization: Bearer $KEY" \
  -d '{"model":"cbcn/kimi-k3","max_tokens":64,"stream":true,"messages":[{"role":"user","content":"Reply with exactly: codex-ok"}]}'
```
```
[0] content_block_start  thinking
[0] (thinking deltas)
[0] content_block_stop
[1] content_block_start  text
[1] text_delta 'cod'
[1] text_delta 'ex'
[1] text_delta '-'
[1] text_delta 'ok'
[1] content_block_stop
message_stop
```

**Likely cause**

A reasoning model's stream has a reasoning phase then an answer phase. The Responses translator seems to send `output_text.done` when the reasoning phase ends (or on the first answer delta) instead of after the answer finishes. Non-reasoning models have no reasoning phase, so the order stays correct.

**Client symptom (Codex CLI)**
```
codex exec -m cbcn/kimi-k3 "Reply with exactly: codex-ok"

codex
cod
ERROR codex_core::util: OutputTextDelta without active item  (x3)
```
Codex treats `done` as message complete and drops the three deltas after it (`ex`, `-`, `ok`), so only `cod` survives.

**Related but different**
- #2713 is Responses stream reconstruction for tool calls, not this ordering.
- #976 is reasoning deltas in the Responses to Chat direction. This is upstream to Responses.

I can provide full raw SSE captures if useful.

chisewaguri:
## problem

streaming a reasoning model through `/v1/responses` truncates the answer to its first token.

some providers (e.g. codebuddy / cbcn) attach `tool_calls: []` to every streaming chunk. an empty array is truthy in js, so `if (delta.tool_calls)` fired on every content chunk and called `closeMessage()`, emitting `response.output_text.done` after the first content token. the remaining deltas then arrive after the message is closed and get dropped. codex cli for example answered only "cod" instead of "codex-ok" with `cbcn/kimi-k3`.

`/v1/messages` handled the same model fine, so the bug was isolated to the responses serializer.

## fix

guard on a non-empty array. empty `tool_calls: []` no longer closes the message; `finish_reason` still closes it, and real tool calls still close it before emitting `function_call` items.

```diff
-  if (delta.tool_calls) {
+  if (delta.tool_calls && delta.tool_calls.length) {
```

## why it's safe

real openai and other spec-compliant providers only send `tool_calls` with actual elements when a call is happening, so `.length` is always > 0 for them. the check only excludes the empty-array case. existing responses-translator tests (custom tools, non-stream, terminal events, multi-turn, headroom) all pass.

## test

added a regression test mirroring the real repro: codebuddy-style chunks with `tool_calls: []` on every chunk must not emit `output_text.done` early, and real tool calls still work. it fails without the fix, passes with it.

fixes #3234

### DIFF
--- a/open-sse/translator/response/openai-responses.js
+++ b/open-sse/translator/response/openai-responses.js
@@ -99,8 +99,8 @@ export function openaiToOpenAIResponsesResponse(chunk, state) {
     }
   }
 
-  // Handle tool_calls
-  if (delta.tool_calls) {
+  // Handle tool_calls (empty array is truthy; require a real call)
+  if (delta.tool_calls && delta.tool_calls.length) {
     closeMessage(state, emit, idx);
     for (const tc of delta.tool_calls) {
       emitToolCall(state, emit, tc);
--- a/tests/unit/openai-responses-empty-toolcalls.test.js
+++ b/tests/unit/openai-responses-empty-toolcalls.test.js
@@ -1,0 +1,52 @@
+/**
+ * Some providers (e.g. codebuddy / cbcn) attach `tool_calls: []` to every
+ * streaming chunk. An empty array is truthy in JS, so the guard
+ * `if (delta.tool_calls)` closed the message on the first content token,
+ * emitting `output_text.done` early and truncating the answer. This mirrors
+ * the real repro: `codex exec -m cbcn/kimi-k3` answered only "cod" instead
+ * of "codex-ok".
+ */
+import { describe, it, expect } from "vitest";
+import { openaiToOpenAIResponsesResponse } from "../../open-sse/translator/response/openai-responses.js";
+import { initState } from "../../open-sse/translator/index.js";
+import { FORMATS } from "../../open-sse/translator/formats.js";
+
+describe("OpenAI Chat stream → Responses: empty tool_calls arrays", () => {
+  it("does not emit output_text.done early when every chunk carries tool_calls: []", () => {
+    const state = initState(FORMATS.OPENAI_RESPONSES);
+    const chunks = [
+      { id: "cmb-test", choices: [{ index: 0, delta: { role: "assistant", content: "", reasoning_content: "", tool_calls: [] }, finish_reason: null }] },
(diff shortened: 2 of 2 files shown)


decolua:
Thanks @chisewaguri for the contribution! Reviewed and merged into master. 🙏

PR #3243, opened 2026-08-12
fix(translator): nest reasoning effort for openai-responses targets

### LINKED ISSUE #3154: Responses API: `reasoning_effort` sent flat instead of nested `reasoning.effort` (400 on gpt-5.6 official OpenAI)
## Summary

When using an OpenAI-compatible **Responses API** provider (`https://api.openai.com/v1/responses`) with a reasoning model (`gpt-5.6-luna`), 9Router sends the reasoning parameter in the flat Chat Completions shape (`reasoning_effort`) instead of the nested Responses shape (`reasoning.effort`). OpenAI rejects the request with a 400.

## Version

- 9Router **v0.5.50**
- Provider: OpenAI (official), OpenAI-compatible **Responses** endpoint `https://api.openai.com/v1/responses`
- Model: `openai-api/gpt-5.6-luna`

## Steps to reproduce

1. Add an OpenAI-compatible provider, type **Responses API**, base URL `https://api.openai.com/v1/responses`.
2. Add a valid official OpenAI API key.
3. Add model `gpt-5.6-luna` with any Thinking level (also reproduces with **Thinking: Auto**).
4. Run **Test Connection** on the key / model.

## Actual result

```
[openai-compatible-responses/gpt-5.6-luna] [400]: {
  "error": {
    "message": "Unsupported parameter: 'reasoning_effort'. In the Responses API, this parameter has moved to 'reasoning.effort'. Try again with the new parameter. See the API documentation for more information: https://platform.openai.com/docs/api-reference/responses/create.",
    "type": "invalid_request_error",
    "param": null,
    "code": "unsupported_parameter"
  }
}
```

## Expected result

For the Responses endpoint, the reasoning effort should be serialized as:

```json
{ "reasoning": { "effort": "high" } }
```

instead of the flat `reasoning_effort: "high"` (which is Chat Completions only).

## Notes

- Setting **Thinking: Auto** does not help — the reasoning field is still emitted (looks related to the `thinking:{type:adaptive}` / `output_config.effort` mapping from v0.5.35).
- Chat Completions endpoint (`/v1/chat/completions`) with the same key works, since flat `reasoning_effort` is still valid there.

## Suggested fix

In the translator path that builds the request body for the **openai-responses** target (the same path touched in v0.5.20 "preserve developer instructions in openai-responses conversion" and v0.5.35 "strip client_metadata when converting openai-responses to openai"), the reasoning field is not being remapped for the Responses schema.

When the target endpoint is `/v1/responses`, the serializer should:

1. Remove the flat `reasoning_effort` key from the outgoing body.
2. Emit it nested instead:

```js
// Chat Completions shape (do NOT send to /v1/responses):
body.reasoning_effort = effort;

// Responses shape (correct for /v1/responses):
delete body.reasoning_effort;
body.reasoning = { ...(body.reasoning ?? {}), effort };
```

3. For gpt-5.6 models, `reasoning.mode` (`standard` / `pro`) also belongs under the same `reasoning` object, so keep both `effort` and `mode` nested there rather than as top-level keys.

Ref: OpenAI Responses API docs — `reasoning.effort` / `reasoning.mode` (https://platform.openai.com/docs/api-reference/responses/create).

ntdatt812:
Closes #3154.

A Responses target gets the Chat-Completions-only top-level `reasoning_effort` instead of the nested `reasoning: { effort }` the endpoint expects, so thinking is silently dropped.

**Credit where it is due: #3183 by @Jordannst found this three days before I did.** I said there that I had no claim, and that stands. What follows is not a claim of priority — it is the two concrete cases #3183 does not cover, both of which I verified by applying its patch locally rather than by reading it.

## 1. #3183 turns three existing tests on master red

`tests/unit/codex-native-passthrough-thinking.test.js` is already in the repo. Applying #3183 to a clean `origin/master`:

```
× native Codex passthrough thinking suffixes > forwards Ultra for Sol
  AssertionError: expected undefined to deeply equal { effort: 'ultra' }
× ... > converts unsupported Luna Ultra to Max without dropping reasoning metadata
  AssertionError: expected { effort: 'low', summary: 'detailed' }
                to deeply equal { effort: 'max', summary: 'detailed' }
× ... > forwards Ultra through a Terra review alias
  Tests  3 failed (3)
```

The cause is in `chatCore.js`, which #3183 does not touch. The codex native-passthrough branch calls `applyThinking` on a scratch object and reads the **flat** key back off it:

```js
const suffixThinking = {};
applyThinking(sourceFormat, upstreamModel, suffixThinking, provider);
if (suffixThinking.reasoning_effort) { … }   // never true once the key is nested
```

Once `resolveFormat` maps a Responses `sourceFormat` to the nested shape, that `if` stops firing. Note the second failure: it is not an error, the request still succeeds — the effort just reverts from `max` to the caller's `low`. A user's `(ultra)` suffix stops working with nothing in the logs.

This PR pins that call to `FORMATS.OPENAI` with a comment saying why, so the branch keeps getting the flat key it nests itself.

## 2. #3183 misses the singular `openai-response` target

`FORMAT_TO_NATIVE` carries both spellings, and the singular one is dispatched on in `utils/usageTracking.js:105` and `translator/concerns/modality.js:152`:

```js
"openai-responses": "openai",
"openai-response": "openai",
```

#3183 keys its mapping on `targetFormat === "openai-responses"`, so `openai-response` still ships the flat key. Here it is a set:

```js
const RESPONSES_TARGETS = new Set([FORMATS.OPENAI_RESPONSES, FORMATS.OPENAI_RESPONSE]);
```

I checked this is a real difference and not a reading error: copying this PR's test file onto #3183's implementation, seven of the eight pass and exactly one fails —

```
× nests for the singular openai-response target too
  AssertionError: expected 'high' to be undefined
```

## Everything #3183 covers is covered here

- Responses-only siblings survive: `preserves sibling reasoning keys set by the caller` starts from `reasoning: {effort:"low", mode:"pro", summary:"detailed"}`, applies a `(high)` suffix and asserts exact equality with `{effort:"high", mode:"pro", summary:"detailed"}`.
- A native Responses request is not flattened: `leaves a native Responses request nested rather than flattening it`.
- End to end through `translateRequest`, the outgoing body has no flat key and keeps `summary: "auto"`.

## Approach

#3183 resolves the format up front — `resolveFormat()` returns `"openai-responses"` and `applyFormat` grows a matching case. This PR leaves format resolution alone and rewrites the emitted key afterwards:

```js
if (RESPONSES_TARGETS.has(targetFormat)) nestReasoningEffort(body, priorReasoning);
```

That is one line at the end of `applyThinking`, no new `applyFormat` case, and it catches anything that produces a flat `reasoning_effort` on a Responses target regardless of which branch produced it. It also uses `FORMATS.*` constants rather than the literal `"openai-responses"`, per the "never hardcode format strings" rule in `open-sse/AGENTS.md`.

If you prefer #3183's shape, the two fixes above still need to ride along with it — that is the part worth carrying over whichever you take.

## Verification

8 tests in `unit/thinking-responses-nested-3154.test.js`, plus `translator/thinking-unified.test.js` updated for the nested shape.

Full suite (`npx vitest run unit translator`) rebased on current master: failing set byte-identical to the base, `codex-native-passthrough-thinking.test.js` green. `npx eslint` clean.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -169,7 +169,11 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     translatedBody = { ...body, model: stripThinkingSuffix(upstreamModel) };
     if (provider === "codex") {
       const suffixThinking = {};
-      applyThinking(sourceFormat, upstreamModel, suffixThinking, provider);
+      // Pinned to OPENAI on purpose: this branch reads the flat reasoning_effort key
+      // off the scratch object and nests it itself, so applyThinking must not nest.
+      // Passing sourceFormat here would hand back {reasoning:{effort}} for a Responses
+      // client and the suffix would silently stop applying.
+      applyThinking(FORMATS.OPENAI, upstreamModel, suffixThinking, provider);
       if (suffixThinking.reasoning_effort) {
         const reasoning = translatedBody.reasoning;
         translatedBody.reasoning = {
--- a/open-sse/translator/concerns/thinkingUnified.js
+++ b/open-sse/translator/concerns/thinkingUnified.js
@@ -5,6 +5,7 @@
 import { getCapabilitiesForModel } from "../../providers/capabilities.js";
 import { getThinkingLevels } from "../../providers/thinkingLevels.js";
 import { PROVIDERS } from "../../providers/index.js";
+import { FORMATS } from "../formats.js";
 import { LEVEL_TO_BUDGET, budgetToLevel, effortToBudget, effortToThinkingLevel } from "./thinking.js";
 
 // Map a target wire-format to its native thinking format (when capability has none).
@@ -20,6 +21,8 @@ const FORMAT_TO_NATIVE = {
   antigravity: "gemini-budget",
   kiro: "kiro",
 };
+
+const RESPONSES_TARGETS = new Set([FORMATS.OPENAI_RESPONSES, FORMATS.OPENAI_RESPONSE]);
 
 // Strip a trailing thinking suffix "model(value)" → "model" (no-op when absent).
 export function stripThinkingSuffix(model) {
@@ -347,7 +350,17 @@ export function applyThinking(targetFormat, model, body, provider = null, intent
 
   const fmt = resolveFormat(targetFormat, cleanModel, provider);
   const supportedLevels = getThinkingLevels(provider, cleanModel);
+  const prior = body.reasoning;
+  const priorReasoning = prior && typeof prior === "object" ? prior : null;
   stripAll(body);
   applyFormat(fmt, body, cfg, caps, supportedLevels);
+  if (RESPONSES_TARGETS.has(targetFormat)) nestReasoningEffort(body, priorReasoning);
   return body;
+}
+
+function nestReasoningEffort(body, priorReasoning) {
+  if (typeof body.reasoning_effort !== "string") return;
+  const effort = body.reasoning_effort;
+  delete body.reasoning_effort;
+  body.reasoning = { ...priorReasoning, effort };
 }
--- a/tests/translator/thinking-unified.test.js
+++ b/tests/translator/thinking-unified.test.js
@@ -172,11 +172,13 @@ describe("applyThinking per provider format", () => {
     ["gpt-5.6-luna", "ultra", "max"],
   ])("normalizes Codex %s effort %s to %s", (model, effort, expected) => {
     const out = apply("openai-responses", model, { reasoning: { effort } }, "codex");
-    expect(out.reasoning_effort).toBe(expected);
+    expect(out.reasoning_effort).toBeUndefined();
+    expect(out.reasoning.effort).toBe(expected);
   });
   it("applies a supported Codex Ultra suffix", () => {
     const out = apply("openai-responses", "gpt-5.6-sol(ultra)", {}, "codex");
-    expect(out.reasoning_effort).toBe("ultra");
+    expect(out.reasoning_effort).toBeUndefined();
+    expect(out.reasoning.effort).toBe("ultra");
   });
   it("keeps Codex-only GPT-5.6 levels out of Kiro translation", () => {
     const out = apply("openai", "gpt-5.6-sol", { reasoning_effort: "max" }, "kiro");
--- a/tests/unit/thinking-responses-nested-3154.test.js
+++ b/tests/unit/thinking-responses-nested-3154.test.js
@@ -1,0 +1,87 @@
+import { describe, expect, it } from "vitest";
+
+const { applyThinking } = await import("../../open-sse/translator/concerns/thinkingUnified.js");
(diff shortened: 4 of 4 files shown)


ntdatt812:
@decolua heads-up: #3183 by @Jordannst predates this by three days and fixes the same issue. I missed it before opening this — please take theirs if you prefer, I have no claim here.

The two differ in approach, so recording it in case it's useful:

- **#3183** resolves the format: `resolveFormat()` returns `"openai-responses"` when `targetFormat` is a Responses endpoint and the resolved format would otherwise be `"openai"`, then adds a matching `applyFormat` case. The capability override (`*gpt-5*` → `thinkingFormat: "openai"`) is handled because the mapping happens *after* caps are consulted.
- **This PR** leaves format resolution alone and rewrites the emitted key afterwards (`nestReasoningEffort`), so anything that produces a flat `reasoning_effort` on a Responses target gets nested regardless of where the format came from.

Both are correct for the reported case. #3183 is more in keeping with how the file already models formats; this one covers a provider that declares a non-`openai` `thinkingFormat` while still transporting over `/v1/responses`, which is hypothetical today.

One thing worth carrying over whichever you take: the codex native-passthrough branch in `chatCore.js` calls `applyThinking` on a scratch object and reads `reasoning_effort` off it, so it needs updating too or Ultra/Max suffixes silently stop applying. #3183 does not touch it; here it is pinned to `FORMATS.OPENAI`. `tests/unit/codex-native-passthrough-thinking.test.js` catches this.

ntdatt812:
Update — my comment above is now out of date, please read this one instead.

I said I had no claim here and that #3183 was the natural one to take. #3183 still found this first and the PR body says so. But I applied it to a clean `origin/master` to check my own carry-over note, and it is worse than I described:

```
× native Codex passthrough thinking suffixes > forwards Ultra for Sol
× ... > converts unsupported Luna Ultra to Max without dropping reasoning metadata
    expected { effort: 'low', summary: 'detailed' } to deeply equal { effort: 'max', summary: 'detailed' }
× ... > forwards Ultra through a Terra review alias
  Tests  3 failed (3)
```

Those are **existing tests already on master** (`tests/unit/codex-native-passthrough-thinking.test.js`), not tests I wrote. The codex native-passthrough branch in `chatCore.js` reads the flat `reasoning_effort` off a scratch object and nests it itself, so once `resolveFormat` maps a Responses source to the nested shape that `if` never fires. Note the second one: no error, the request succeeds, the effort just falls back from `max` to the caller's `low`. A `(ultra)` suffix stops working with nothing in the logs.

Second gap: `FORMAT_TO_NATIVE` carries both `"openai-responses"` and `"openai-response"`, and the singular one is dispatched on in `usageTracking.js:105` and `modality.js:152`. #3183 keys on the plural only. Copying this PR's tests onto its implementation, exactly one of the eight fails — `nests for the singular openai-response target too`.

Both are fixed here. If you prefer #3183's shape, these two need to ride along with it either way.

ntdatt812:

PR #3245, opened 2026-08-12
fix(rtk): inject the system prompt with the target API's content-part type

### LINKED ISSUE #3202: CAVEMAN injects `input_text` into array-content system message → StepFun 400 "Unrecognized chat message"
## Summary

When **CAVEMAN** is enabled and the incoming request has a **system message with array content** (`content: [{type:"text",...}]`), 9router injects a Responses-API-style part `{type:"input_text"}` into the chat-format `messages[]`. Strict chat/completions providers (e.g. StepFun) reject this type and return:

```json
400 {"error":{"message":"Unrecognized chat message.","type":"request_params_invalid"}}
```

## Environment

- 9router v0.5.50 (Docker, ghcr.io/decolua/9router:latest)
- CAVEMAN enabled
- Client sends `system` message with **array** content (e.g. QwenPaw serializes system content as `[{"type":"text",...}]`)
- Upstream: StepFun (`/chat/completions`, strict validation)

## Root cause

`appendToOpenAIMessage()` in `open-sse/rtk/systemInject.js`:

```js
    // Responses-style array of parts {type:"input_text"|"text", text}
    msg.content.push({ type: "input_text", text: prompt });
}
```

It always pushes `{type:"input_text"}` (OpenAI Responses API type). Chat Completions format expects `{type:"text"}`. StepFun strictly validates and rejects `input_text`.

The fixes in #2503 / #2508 covered the `instructions` string path and the `input[]` path, but **this array-content branch is still present on master** (verified 2026-08-10).

## Reproduction

```bash
curl -s -w "\n---HTTP %{http_code}---\n" -X POST "http://127.0.0.1:20128/v1/chat/completions" \
  -H "Authorization: Bearer <your-9router-key>" \
  -d '{"model":"stf/step-3.5-flash","messages":[{"role":"system","content":[{"type":"text","text":"你是助手"}]},{"role":"user","content":[{"type":"text","text":"你好"}]}],"stream":false}'
```

Expected: 200. Actual with CAVEMAN on: 400 `Unrecognized chat message`.

Note: a plain-string system content (`{"role":"system","content":"hi"}`) takes a different branch and does NOT trigger this bug — the array content is required.

## Suggested fix

One-line change in `appendToOpenAIMessage()`:

```js
// before
msg.content.push({ type: "input_text", text: prompt });
// after
msg.content.push({ type: "text", text: prompt });
```

## Verified locally

Patched the built chunk (`input_text` → `text` in this single injection site) on v0.5.50; the repro above returns 200 after the patch. The other two `input_text` occurrences in the codebase (vision/pdf filtering, Responses array conversion) are legitimate and must stay untouched.

## Related

- #2497 — same family of "injection produces wrong schema" 400s, but on the Responses `input[]` path
- #2503, #2508 — fixed `instructions`/`input[]` paths, not this one

ntdatt812:
Closes #3202.

`injectSystemPrompt` (caveman / ponytail) pushed `{type:"input_text"}` into every array-shaped system message, including Chat Completions bodies where the valid part type is `text`. The provider rejects the request or drops the part.

**Credit where it is due: #3204 by @2Bdou found this two days before I did.** I said there that theirs was the more complete fix and offered to close this one. That was true at the time. It is no longer — this now covers everything #3204 does plus two cases it misses, and I verified both by applying its patch to a clean `origin/master` and running this PR's tests against it:

```
× creates instructions when the Responses body carries a bare string input
  → expected undefined to be 'speak like a caveman'
× does not leave blank lines in front of an empty system message
  → expected '\n\nspeak like a caveman' to be 'speak like a caveman'
  the other 9 pass
```

## 1. A Responses body with a string input drops the prompt silently

`/v1/responses` accepts `input` as a bare string, not only as an array — `{model, input: "hello", stream: false}` is the exact shape `tests/unit/codex-native-passthrough-thinking.test.js` already sends. On master, on #3204, and in this PR's first revision, that body walks straight past both branches:

```js
const arr = Array.isArray(body.messages) ? body.messages
  : Array.isArray(body.input) ? body.input
  : null;
if (!arr) return;          // ← string input lands here
```

`instructions` is absent too, so there is nothing to append to and the caveman/ponytail prompt is dropped with no error and no log. Toggling the feature on simply does nothing for native Codex traffic. Fixed by creating the Responses system channel in that case:

```js
if (typeof body.input === "string") body.instructions = prompt;
```

## 2. An empty system message leaves the prompt behind two blank lines

The other two string branches in this same file already guard it — `injectMessagesSystem` does `body.instructions ? … : prompt`, `injectClaudeSystem` checks `body.system.length > 0`. `appendToOpenAIMessage` was the only one that did not, so `content: ""` produced `"\n\nspeak like a caveman"`. Now guarded the same way.

## 3. The untyped unshift (this is #3204's find)

With no system item in a Responses `input[]`, the old code unshifted a bare `{role, content}`, which is invalid there. #3204 fixed it; this PR now does too, keyed off which array was selected rather than a separate shape probe, and built from `RESPONSES_ITEM.MESSAGE` / `ROLE.SYSTEM` instead of literals.

## Shape, not format label

Both fixes derive the part type from the body, and that is deliberate rather than incidental — `detectFormatByEndpoint` labels `/v1/chat/completions` + `input[]` as `FORMATS.OPENAI` on purpose, because Cursor CLI posts a Responses body to the chat endpoint:

```js
// /v1/chat/completions + input[] → treat as openai (Cursor CLI sends Responses body via chat endpoint)
if (pathname.includes("/v1/chat/completions") && Array.isArray(body?.input)) return FORMATS.OPENAI;
```

So keying on `format` would put `text` parts into a Responses body for every Cursor CLI request. #3204 gets this right too, but nothing pins it in either PR's tests, and a later "simplify to use the format we were handed" refactor would break Cursor silently. `uses Responses part types for a Cursor-CLI body labelled FORMATS.OPENAI` pins it.

Per `open-sse/AGENTS.md` — "Never hardcode role/block/model strings — use `open-sse/translator/schema/`" — the part types and roles come from `OPENAI_BLOCK.TEXT`, `RESPONSES_ITEM.INPUT_TEXT`, `RESPONSES_ITEM.MESSAGE`, `ROLE.SYSTEM` / `ROLE.DEVELOPER` rather than literals.

## Not changed on purpose

`appendToOpenAIMessage` still overwrites when `content` is some other object, which loses the original. That is the whole file's convention for unexpected shapes (`injectClaudeSystem` and `injectGeminiSystem` both end the same way), so changing it in one branch would make the file less predictable, not more. Worth doing as its own change if you want it.

## Verification

11 tests in `unit/system-inject-part-type-3202.test.js`: chat vs Responses part types, `developer` role, string content, plain unshift, typed unshift, the Cursor CLI body, string `input`, a chat body with no `messages[]` left untouched, the empty-content guard, and `instructions` appending.

Full suite (`npx vitest run unit translator`) rebased on current master: failing set byte-identical to the base. `npx eslint` clean. RTK stays fail-open — no new throw paths.

### DIFF
--- a/open-sse/rtk/systemInject.js
+++ b/open-sse/rtk/systemInject.js
@@ -3,6 +3,7 @@
 // native-passthrough flows. Used by caveman.js and ponytail.js.
 
 import { FORMATS } from "../translator/formats.js";
+import { OPENAI_BLOCK, RESPONSES_ITEM, ROLE } from "../translator/schema/index.js";
 
 const SEP = "\n\n";
 
@@ -36,25 +37,38 @@ function injectMessagesSystem(body, prompt) {
     return;
   }
 
-  const arr = Array.isArray(body.messages) ? body.messages
+  const isChatMessages = Array.isArray(body.messages);
+  const arr = isChatMessages ? body.messages
     : Array.isArray(body.input) ? body.input
     : null;
-  if (!arr) return;
+  if (!arr) {
+    // Responses also accepts a bare string input with no instructions field. There is
+    // no array to append a part to, so the system channel has to be created instead —
+    // otherwise the prompt is dropped without a trace.
+    if (typeof body.input === "string") body.instructions = prompt;
+    return;
+  }
 
-  const idx = arr.findIndex(m => m && (m.role === "system" || m.role === "developer"));
+  const partType = isChatMessages ? OPENAI_BLOCK.TEXT : RESPONSES_ITEM.INPUT_TEXT;
+
+  const idx = arr.findIndex(m => m && (m.role === ROLE.SYSTEM || m.role === ROLE.DEVELOPER));
   if (idx >= 0) {
-    appendToOpenAIMessage(arr[idx], prompt);
+    appendToOpenAIMessage(arr[idx], prompt, partType);
+  } else if (isChatMessages) {
+    arr.unshift({ role: ROLE.SYSTEM, content: prompt });
   } else {
-    arr.unshift({ role: "system", content: prompt });
+    // A Responses input[] item carries its own type; upstream rejects a bare {role,content}.
+    arr.unshift({ type: RESPONSES_ITEM.MESSAGE, role: ROLE.SYSTEM, content: [{ type: partType, text: prompt }] });
   }
 }
 
-function appendToOpenAIMessage(msg, prompt) {
+function appendToOpenAIMessage(msg, prompt, partType) {
   if (typeof msg.content === "string") {
-    msg.content = `${msg.content}${SEP}${prompt}`;
+    // Guarded like the instructions and Claude branches: an empty string would
+    // otherwise leave the prompt behind two blank lines.
+    msg.content = msg.content ? `${msg.content}${SEP}${prompt}` : prompt;
   } else if (Array.isArray(msg.content)) {
-    // Responses-style array of parts {type:"input_text"|"text", text}
-    msg.content.push({ type: "input_text", text: prompt });
+    msg.content.push({ type: partType, text: prompt });
   } else {
     msg.content = prompt;
   }
--- a/tests/unit/system-inject-part-type-3202.test.js
+++ b/tests/unit/system-inject-part-type-3202.test.js
@@ -1,0 +1,118 @@
+import { describe, expect, it } from "vitest";
+
+const { injectSystemPrompt } = await import("../../open-sse/rtk/systemInject.js");
+const { FORMATS } = await import("../../open-sse/translator/formats.js");
+
+const PROMPT = "speak like a caveman";
+
+describe("#3202 system injection uses the content-part type of the target API", () => {
+  it("appends {type:'text'} to an array-content system message in messages[]", () => {
+    const body = {
+      messages: [
+        { role: "system", content: [{ type: "text", text: "you are a helper" }] },
+        { role: "user", content: [{ type: "text", text: "hi" }] },
+      ],
+    };
+    injectSystemPrompt(body, FORMATS.OPENAI, PROMPT);
+
+    expect(body.messages[0].content).toEqual([
+      { type: "text", text: "you are a helper" },
+      { type: "text", text: PROMPT },
+    ]);
+    expect(JSON.stringify(body)).not.toContain("input_text");
+  });
+
+  it("appends {type:'input_text'} to an array-content system item in input[]", () => {
+    const body = {
+      input: [
+        { role: "system", content: [{ type: "input_text", text: "you are a helper" }] },
+        { role: "user", content: [{ type: "input_text", text: "hi" }] },
+      ],
+    };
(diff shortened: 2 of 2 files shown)


ntdatt812:
@decolua heads-up: #3204 by @2Bdou predates this by two days and fixes the same issue. I missed it before opening this — please take theirs, it is the more complete fix.

#3204 covers everything here **plus** the branch this PR leaves untouched: when no system message exists at all, `injectMessagesSystem` unshifts a bare `{role, content}` item, which is invalid for a Responses `input[]`. #3204 unshifts a properly typed `{type:"message", role, content:[{type:"input_text"}]}` in that case. That is a real second bug on the same code path and this PR does not address it.

Happy to close this one — say the word, or just close it yourself.

ntdatt812:
Update — my comment above is now out of date, please read this one instead.

I said #3204 was the more complete fix and offered to close this. That was true then. I have since rebased this on current master and covered two cases #3204 misses, both verified by applying its patch to a clean `origin/master` and running this PR's tests against it (9 of 11 pass there, these two fail):

1. `/v1/responses` accepts `input` as a bare string — `{model, input: "hello", stream: false}`, the exact shape `codex-native-passthrough-thinking.test.js` already sends. That body matches neither array branch and has no `instructions`, so the caveman/ponytail prompt is **dropped silently**. Turning the feature on does nothing for native Codex traffic.
2. `content: ""` produced `"\n\n<prompt>"`. The `instructions` and Claude branches in this same file already guard that; `appendToOpenAIMessage` was the only one that did not.

#3204's own find — the untyped unshift into a Responses `input[]` — is covered here too, built from `RESPONSES_ITEM.MESSAGE` / `ROLE.SYSTEM` rather than literals per the schema rule in `open-sse/AGENTS.md`.

@2Bdou found this first and that is stated in the PR body. If you would rather take theirs, the two cases above still need to ride along.

ntdatt812:
Closing this — `master` has since implemented it, and better.

`injectMessagesSystem` was split into `injectInstructionsSystem` / `injectChatSystem` / `injectResponsesInputSystem`, dispatching on the wire shape. That covers three of the four things this PR was for:

- the part type now follows the array that was selected (`OPENAI_BLOCK.TEXT` vs `RESPONSES_ITEM.INPUT_TEXT`), via the schema constants;
- an empty system message no longer yields `"\n\n<prompt>"` — `dedupStringAppend` returns the prompt alone when the current content is falsy;
- an unshifted Responses item carries `type: "message"` and a typed content part.

On the fourth, a bare string `input`, `master` deliberately goes the other way: `// string input must stay untouched`. That is your call to make and I am not going to argue it from outside.

Running this PR's tests against `master` left two failures. One is that string-input case, which is now intended behaviour. The other turned out to be a real gap in the new code — a message item that omits `type` is skipped, so a second system message gets prepended beside the caller's own. I have sent that separately as #3652 with the evidence, rather than reviving this branch around it.

Thanks for taking the problem on directly.

PR #3247, opened 2026-08-12
fix(handlers): honor the client format for non-streaming binary-transport replies

### LINKED ISSUE #3199: Bug: Non-streaming /v1/messages returns OpenAI format instead of Anthropic format (Kiro provider)
## Description

When sending a non-streaming request (`"stream": false`) to `/v1/messages` with a Kiro provider model, the response is returned in OpenAI `chat.completion` format instead of the expected Anthropic message format.

Streaming requests to the same endpoint work correctly — they return proper Anthropic SSE format.

## Version

9Router v0.5.50

## Steps to Reproduce

1. Send a non-streaming request to `/v1/messages`:

```bash
curl -s http://localhost:20128/v1/messages \
  -H "x-api-key: <your-key>" \
  -d '{"model":"kr/claude-haiku-4.5","max_tokens":20,"stream":false,"messages":[{"role":"user","content":"Say hi"}]}'

## Expected Response (Anthropic format)
{
  "id": "msg_...",
  "type": "message",
  "role": "assistant",
  "content": [{"type": "text", "text": "Hey there."}],
  "model": "claude-haiku-4.5",
  "stop_reason": "end_turn",
  "usage": {"input_tokens": 10, "output_tokens": 8}
}


## Actual Response (OpenAI format — incorrect)
{
  "id": "chatcmpl-1786352438097",
  "object": "chat.completion",
  "created": 1786352438,
  "model": "claude-haiku-4.5",
  "choices": [{"index": 0, "message": {"role": "assistant", "content": "Hey there."}, "finish_reason": "stop"}],
  "usage": {}
}

## Additional Context

- **Streaming works correctly:** Without `"stream": false` (or with `"stream": true`), the response is proper Anthropic SSE format with `message_start`, `content_block_delta`, etc.
- **Impact:** LiteLLM's `anthropic/` provider prefix expects Anthropic response format from `/v1/messages`. The non-streaming path fails with `KeyError: 'content'` because it receives OpenAI format instead.
- The `translateNonStreamingResponse` function with `openAICompletionToClaudeMessage` should handle this case (OpenAI response → Claude format when sourceFormat is CLAUDE), but it appears to not be triggered for non-streaming requests with explicit `"stream": false`.

### PR DESCRIPTION
**Pull request by ntdatt812**
Fixes #3199

## Problem

```bash
curl -s http://localhost:20128/v1/messages -H "anthropic-version: 2023-06-01" \
  -d '{"model":"kr/claude-haiku-4.5","max_tokens":20,"stream":false,"messages":[{"role":"user","content":"Say hi"}]}'
```

returns `{"id":"chatcmpl-…","object":"chat.completion","choices":[…]}` instead of an Anthropic message. Streaming against the same route is correct.

Providers on a binary or proprietary transport decode their upstream reply to OpenAI Chat Completions inside their own executor — the kiro executor emits `chat.completion.chunk` SSE, which `handleNonStreamingResponse` collapses via `parseSSEToOpenAIResponse()`. By the time `translateNonStreamingResponse()` runs, `targetFormat` is `kiro` and the body is already OpenAI-shaped.

`translateNonStreamingResponse()` branches on `openai`, `gemini`/`vertex`/`antigravity`, `claude` and `ollama`. `kiro` (and `cursor`, `commandcode`) match none of them, so control reached the trailing `return responseBody` and the OpenAI shape went straight to a Claude client. The existing `targetFormat === OPENAI && sourceFormat === CLAUDE` branch was the right conversion — it just was not reachable for these targets.

Downstream impact: LiteLLM's `anthropic/` prefix expects an Anthropic body from `/v1/messages` and fails with `KeyError: 'content'`.

## Fix

The "body is OpenAI Chat Completions — what does the client speak?" dispatch existed once already, keyed on `targetFormat === OPENAI`. Rather than write it a second time, it is extracted into `fromOpenAICompletion()` and called from both places: the existing `targetFormat === OPENAI` branch, and a new trailing guard for bodies that carry a `choices` array under a target format with no branch of its own.

Net result is one list of client formats instead of two. The Responses arm covers the identical gap for a Responses client on the same targets. An OpenAI client is unaffected, and a body without `choices` is returned untouched.

## Scope note — the deeper seam, deliberately not taken

The codebase already has the right long-term mechanism for this: `result.responseFormat`, honored in `chatCore.js` ("Most executors return their registry format. Cursor AgentService is an exception: it is decoded by the executor into OpenAI-compatible output."). Today only `executors/cursor.js` sets it.

`executors/kiro.js` and `executors/commandcode.js` both qualify — kiro's `execute()` always runs `transformEventStreamToSSE`, and commandcode's always runs `wrapNdjsonAsOpenAISse()`, so both unconditionally return OpenAI Chat Completions. Declaring `responseFormat: FORMATS.OPENAI` on those two executors would fix this issue with no new code at all, since the existing `targetFormat === OPENAI` branch would then handle them.

I did not do that here because `providerResponseFormat` also feeds `handleStreamingResponse`, so it would reroute kiro's streaming path off the direct `kiro:claude` route onto `openai:claude` and make `translator/response/kiro-to-claude.js` dead. That should be behavior-neutral, but it is a much larger change on a heavily used provider and belongs in its own PR. Flagging it so the direction is on record.

## Tests

New: `tests/unit/nonstream-binary-transport-source-format-3199.test.js` — a kiro-decoded body converts to an Anthropic message (text, `stop_reason`, `usage`), tool calls become `tool_use` blocks, a Responses client gets `object:"response"`, an OpenAI client's body is returned by identity, and a body with no `choices` is untouched.

Full suite (`npx vitest run unit translator`) against `master` as the baseline: **1670 → 1675 passed, 91 failed on both sides** — no pass→fail, the 5 new passes are this PR's tests.

### DIFF
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -141,17 +141,17 @@ function openAICompletionToResponses(responseBody, customToolNames = null) {
 /**
  * Translate non-streaming response body from provider format → OpenAI format.
  */
+// Provider responded in OpenAI Chat Completions shape — hand the client whatever
+// dialect it speaks, or the body unchanged when that dialect is OpenAI itself.
+function fromOpenAICompletion(responseBody, sourceFormat, customToolNames) {
+  if (sourceFormat === FORMATS.OPENAI_RESPONSES) return openAICompletionToResponses(responseBody, customToolNames);
+  if (sourceFormat === FORMATS.CLAUDE) return openAICompletionToClaudeMessage(responseBody);
+  return responseBody;
+}
+
 export function translateNonStreamingResponse(responseBody, targetFormat, sourceFormat, customToolNames = null) {
   if (targetFormat === sourceFormat) return responseBody;
-  // Provider responded in OpenAI Chat Completions shape but the client speaks
-  // Responses API — convert so tool_calls/text surface as Responses `output`.
-  if (targetFormat === FORMATS.OPENAI && sourceFormat === FORMATS.OPENAI_RESPONSES) {
-    return openAICompletionToResponses(responseBody, customToolNames);
-  }
-  if (targetFormat === FORMATS.OPENAI && sourceFormat === FORMATS.CLAUDE) {
-    return openAICompletionToClaudeMessage(responseBody);
-  }
-  if (targetFormat === FORMATS.OPENAI) return responseBody;
+  if (targetFormat === FORMATS.OPENAI) return fromOpenAICompletion(responseBody, sourceFormat, customToolNames);
 
   // Gemini / Antigravity
   if (targetFormat === FORMATS.GEMINI || targetFormat === FORMATS.ANTIGRAVITY || targetFormat === FORMATS.GEMINI_CLI || targetFormat === FORMATS.VERTEX) {
@@ -273,6 +273,10 @@ export function translateNonStreamingResponse(responseBody, targetFormat, source
   // Ollama
   if (targetFormat === FORMATS.OLLAMA) {
     return ollamaBodyToOpenAI(responseBody);
+  }
+
+  if (Array.isArray(responseBody?.choices)) {
+    return fromOpenAICompletion(responseBody, sourceFormat, customToolNames);
   }
 
   return responseBody;
--- a/tests/unit/nonstream-binary-transport-source-format-3199.test.js
+++ b/tests/unit/nonstream-binary-transport-source-format-3199.test.js
@@ -1,0 +1,88 @@
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("@/lib/usageDb.js", () => ({
+  appendRequestLog: vi.fn(async () => {}),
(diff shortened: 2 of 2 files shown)


momotseuk:
will this fix be incorporated into next version?

ntdatt812:
I wrote the patch, so I cannot answer for the release schedule — that is @decolua's call. What I can give you is the current state and a workaround.

**State.** Open since 12 Aug, `MERGEABLE` / `CLEAN` against `master` (`699edac3`), waiting on review. Nothing is blocking it from my side.

I re-checked it just now rather than assuming it had kept: merged the current `master` into the branch — no conflicts — and ran its tests on the result.

```
tests/unit/nonstream-binary-transport-source-format-3199.test.js
Test Files  1 passed (1)
     Tests  5 passed (5)
```

**Workaround until it lands.** Only the non-streaming path is affected; streaming through the same route already returns the right shape. So on a binary-transport provider (`kr/*`, `cursor/*`) sending `"stream": true` and collapsing the SSE client-side gives you a correct Anthropic message today.

ntdatt812:

Issue #3135, opened 2026-08-08
enabledModels is read in models-listing but never persisted — custom provider model selection is a no-op


Crpedia:
## Title
`enabledModels` field is read in models-listing logic but never written anywhere server-side — custom provider model restriction is a no-op

## Environment
- 9Router version: 0.5.50
- Install method: `npm install -g 9router`
- Runtime: Node 20+, systemd-managed (`ExecStart=/usr/bin/9router -p 20128 -H 0.0.0.0 -l --skip-update`)
- DB driver: better-sqlite3 (`~/.9router/db/data.sqlite`)
- Affected provider type: custom OpenAI-compatible (`authType: "apikey"`, `provider: "openai-compatible-chat-*"`)

## Summary
`providerSpecificData.enabledModels` is consumed by the model-listing route but there is no code path in the server bundle that persists it. The per-model enable/disable control in the provider dashboard page therefore has no effect for custom OpenAI-compatible providers — it neither survives a page reload, nor changes `/v1/models` output, nor changes which models are actually routable.

## Evidence

**1. DB record for an affected connection** (`providerConnections` table, `data` column, JSON-decoded):
```json
{
  "defaultModel": "DeepSeek-V4-Pro",
  "apiKey": "sk-...",
  "testStatus": "active",
  "providerSpecificData": {
    "prefix": "hn",
    "apiType": "chat",
    "baseUrl": "https://api.hcnsec.cn/v1",
    "nodeName": "OpenAI Compatible",
    "connectionProxyEnabled": false,
    "connectionProxyUrl": "",
    "connectionNoProxy": ""
  }
}
```
No `enabledModels` key present, despite toggling individual models on/off in the dashboard UI multiple times across page reloads and full app reinstalls (`npm uninstall -g 9router && rm -rf ~/.9router && npm install -g 9router@latest`).

**2. Read-side logic exists and is correct** — found in `server/chunks/5612.js` (minified), corresponds to this readable equivalent from the `v0.4.59` tag, `src/app/api/v1/models/route.js`:
```js
const enabledModels = conn?.providerSpecificData?.enabledModels;
const hasExplicitEnabledModels =
  Array.isArray(enabledModels) && enabledModels.length > 0;
...
let rawModelIds = hasExplicitEnabledModels
  ? Array.from(new Set(enabledModels.filter(m => typeof m === "string" && m.trim() !== "")))
  : providerModels.map(model => model.id);

[...]

PR #3252, opened 2026-08-12
fix(models): let a curated custom-provider model list suppress the live catalog

### LINKED ISSUE #3115: Custom provider /v1/models returns all models instead of only enabled ones
## Describe the bug
Custom OpenAI-compatible provider exposes ALL models from its `/models` endpoint via 9Router's `/v1/models`, even when only 1 model is manually kept/enabled in the dashboard. Manually deleting the other models from the provider's model list in the dashboard does not filter them out of `/v1/models` — they reappear.

## Steps to reproduce
1. Add a custom OpenAI-compatible provider (in my case: provider `hnsec`) via bulk-import from its `/models` endpoint
2. In the dashboard, delete/disable all models except 1
3. Run `curl http://localhost:20128/v1/models`
4. Observe that all models from that provider are still returned, not just the 1 kept

## Expected behavior
`/v1/models` should only return the model(s) actually configured/enabled for that provider in the dashboard.

## Actual behavior
`/v1/models` returns the full raw model list from the provider's endpoint regardless of dashboard state.

## Troubleshooting already done
- Fully uninstalled (`npm uninstall -g 9router`), deleted `~/.9router` data directory, reinstalled fresh (`npm install -g 9router@latest`) — bug persisted with a completely clean install and fresh database
- Manually deleted unwanted models from the provider's model list in the dashboard (not just toggling disable) — bug persisted

## Environment
- **9Router version:** 0.5.50
- **Installation method:** npm (global)
- **Operating system:** Linux (Ubuntu VPS)
- **Node.js version:** (jalanin `node -v` buat isi ini)
- **Provider involved:** Custom OpenAI-compatible provider (`hnsec`)
- **Deployment:** systemd service (`9router.service`), exposed via Cloudflare quick tunnel

## Additional context
Possibly related to recent release note "Models: store provider custom models by provider scope" — this may have changed how custom provider models are cached/filtered.

ntdatt812:
Fixes #3115

## Problem

Add a custom OpenAI-compatible provider, bulk-import its models, delete all but one in the dashboard, then `curl /v1/models` — every model from the upstream is still listed, and the deleted ones reappear after every reload. A clean reinstall with a fresh database does not help.

## Root cause

In `buildModelsList()`, a custom node has no entry in the static `PROVIDER_MODELS` registry, so:

```js
let rawModelIds = hasExplicitEnabledModels ? … : providerModels.map(m => m.id);   // []

if (isCompatibleProvider && rawModelIds.length === 0 && !skipDynamicFetch) {
  rawModelIds = await fetchCompatibleModelIds(conn);   // the upstream's ENTIRE catalogue
}
```

The user's own list lives in `customModels` (written by `/api/models/custom`, which is what the dashboard's model management calls). But `customModelIds` is only computed ~45 lines *below* this fallback and merged in at the end:

```js
const mergedModelIds = Array.from(new Set([...modelIds, ...customModelIds, ...aliasModelIds]));
```

So the fallback fires even when the user has curated a list, and the merge then unions the curated models with the full upstream catalogue. Deleting a model in the dashboard removes it from `customModels` — and the live fetch adds it right back. That is the "they reappear" symptom.

## Fix

Move the `customModelIds` computation above the fallback and include it in the guard. The block was already independent of everything between the two points (it only reads `customModels`, `kindFilter`, and the alias variables), so this is a pure reorder plus one added condition.

```js
if (isCompatibleProvider && rawModelIds.length === 0 && customModelIds.length === 0 && !skipDynamicFetch) {
```

A provider with nothing configured still falls back to the live catalogue, so first-time discovery and bulk-import are unchanged. `skipDynamicFetch` (the cross-instance recursion guard) is unaffected.

Note this also covers the `enabledModels` half of #3135: that field is read here but never written anywhere server-side, so it can never be the mechanism that filters a custom provider's list — `customModels` is.

## Tests

New: `tests/unit/custom-provider-model-list-3115.test.js` — the first test coverage for `buildModelsList()`, with the DB layer and `fetch` mocked. Curated list wins and the upstream endpoint is never called; an empty curated list still falls back and does call it; curated models belonging to another provider do not suppress the fallback; `skipDynamicFetch` still short-circuits.

The first test fails on `master` with `expected [ 'keep-me', 'drop-me-1', 'drop-me-2' ] to deeply equal [ 'keep-me' ]` — the exact reported symptom — and passes with this patch.

Full suite (`npx vitest run unit translator`) against `master` as the baseline: **1670 → 1674 passed, 91 failed on both sides** — no pass→fail, the 4 new passes are this PR's tests.

### DIFF
--- a/src/app/api/v1/models/route.js
+++ b/src/app/api/v1/models/route.js
@@ -375,7 +375,25 @@ export async function buildModelsList(kindFilter, options = {}) {
           )
         : providerModels.map((model) => model.id);
 
-      if (isCompatibleProvider && rawModelIds.length === 0 && !skipDynamicFetch) {
+      const customModelKindById = new Map();
+      const customModelIds = customModels
+        .filter((m) => {
+          if (!m?.id) return false;
+          const kind = getModelKind(m) || LLM_KIND;
+          // imageToText custom models are vision-capable chat models: expose them
+          // both in the default LLM list and in /v1/models/image-to-text.
+          if (!kindFilter.includes(kind) && !(kind === "imageToText" && kindFilter.includes(LLM_KIND))) return false;
+          const alias = m.providerAlias;
+          return alias === staticAlias || alias === outputAlias || alias === providerId;
+        })
+        .map((m) => {
+          const modelId = String(m.id).trim();
+          if (modelId) customModelKindById.set(modelId, getModelKind(m) || LLM_KIND);
+          return modelId;
+        })
+        .filter((modelId) => modelId !== "");
+
+      if (isCompatibleProvider && rawModelIds.length === 0 && customModelIds.length === 0 && !skipDynamicFetch) {
         rawModelIds = await fetchCompatibleModelIds(conn);
       }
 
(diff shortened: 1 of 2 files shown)


ntdatt812:

PR #3263, opened 2026-08-12
fix(cli): attach to existing server instead of killing it; enable macOS auto-restart

kundeng:
## Summary

Two related CLI robustness fixes that make 9router behave correctly when run alongside a supervisor-managed (launchd/systemd) always-on instance.

### 1. Avoid clash with an already-running instance
When 9router is managed by launchd/systemd with `KeepAlive`, running the CLI used to `killAllAppProcesses(port)` → `killProcessOnPort(port)` → `startServer()`. Killing the supervised server makes the supervisor instantly restart it, which then races the CLI for the port and crashes the CLI.

Now the CLI reuses the existing `waitServerReady` (with a short 500ms timeout) to detect whether a server is already listening. If one is, it attaches to it (`attachOnly` mode) and just shows the menu — it never kills the supervised server. It only kills/restarts when no server is present.

`attachOnly` mode:
- does not spawn a server process
- `cleanup()` does not kill the shared server
- crash/restart handlers are not attached (the supervisor owns restarts)

### 2. Enable auto-restart on macOS
The autostart launchd plist now writes `KeepAlive=true` so the tray/server process is restarted by launchd if it crashes, matching the robust always-on behavior users expect.

## Test plan
- With a launchd-managed server running, `9router` now prints `ℹ 9Router already running on port 20128. Attaching to existing instance.` and shows the menu without killing the server.
- With no server running, `9router` starts fresh as before.

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -540,9 +540,20 @@ if (!fs.existsSync(serverPath)) {
 
 // Start server immediately; run update check in parallel (not on the critical path).
 const updatePromise = checkForUpdate();
-killAllAppProcesses(port)
-  .then(() => killProcessOnPort(port))
-  .then(() => startServer(updatePromise));
+// If a server is already listening (e.g. a launchd/systemd-managed always-on
+// instance), attach to it instead of killing/restarting — killing races with the
+// supervisor's KeepAlive and crashes the CLI. Short timeout: a live server
+// resolves immediately; a free port resolves false after ~500ms.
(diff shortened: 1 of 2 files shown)


kundeng:
Superseded: both attach + watchdog changes are now merged into the fork's master (kundeng/9router) and published as @kundeng/9router. Closing this upstream PR since the fork is the source of truth.

PR #3264, opened 2026-08-12
feat(cli): add built-in watchdog to restart a frozen server

kundeng:
## Summary

The CLI's crash/restart handlers only fire when the server **process exits**. A server that stays alive but freezes (App Nap, event-loop stall, hung upstream) never triggers them — which is why users end up adding a separate external launchd watchdog daemon that polls the port and kickstarts the server.

This PR adds a **built-in watchdog** to the CLI that:
- Polls the HTTP health endpoint (`/api/settings`) every 30s with a 5s timeout
- Restarts the server after 3 consecutive failures (alive-but-frozen detection)
- Skips checks while a restart is already in flight, so it never stacks restarts while the fresh server boots
- Resets its failure counter as soon as the server responds again

## Why

With this, a **single** launchd agent (the existing autostart plist with `KeepAlive=true`) fully covers:
1. **Autostart** on boot (existing)
2. **Crash restart** on process exit (existing `KeepAlive` + `tryRestart`)
3. **Frozen restart** on a hung-but-alive process (new watchdog)

The separate external watchdog daemon becomes unnecessary and can be removed.

## Notes
- The watchdog only runs when the CLI owns the server process (it is wired into `startServer` alongside `attachServerEvents`).
- Health check treats any HTTP response (including 401) as healthy — only connection failure counts as frozen, matching the old external watchdog's behavior.

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -3,6 +3,7 @@
 const { spawn, exec, execSync } = require("child_process");
 const path = require("path");
 const fs = require("fs");
+const http = require("http");
 const https = require("https");
 const net = require("net");
 const os = require("os");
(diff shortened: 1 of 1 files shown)


kundeng:
Superseded: both attach + watchdog changes are now merged into the fork's master (kundeng/9router) and published as @kundeng/9router. Closing this upstream PR since the fork is the source of truth.

PR #3268, opened 2026-08-13
feat(codex): support fast service tier for image generation

trananhtung:
## Problem

The Codex chat path already honours `service_tier: "fast"` — `CodexExecutor.transformRequest()` maps it to the backend's `"priority"` tier ([`open-sse/executors/codex.js`](https://github.com/decolua/9router/blob/master/open-sse/executors/codex.js)).

Image generation does not. `imageProviders/codex.js` builds its own Responses payload in `buildBody()` and never sets `service_tier`, so a fast-tier text-to-image request against `gpt-5.5-image` (or `gpt-5.4-image` / `gpt-5.3-image`) is silently downgraded to the standard tier. `service_tier` is already in `RESPONSES_API_ALLOWLIST` and both paths POST to the same `chatgpt.com/backend-api/codex/responses` endpoint, so the field was simply being dropped on the image side.

## Change

- Add `open-sse/shared/codexServiceTier.js` with `normalizeCodexServiceTier()`, and use it from **both** the executor and the image adapter so the two paths can't drift apart.
- Unknown tiers are dropped rather than forwarded — an unrecognized `service_tier` makes the backend reject the whole request with `routing_unsupported`. This preserves the executor's previous behaviour exactly.
- Expose `service_tier` in the `params` list of the three Codex image models.
- Add a "Speed" selector to the image example card in the dashboard.

Behaviour is opt-in: with no `service_tier` the payload is byte-identical to before.

## Testing

New `tests/unit/codex-image-service-tier.test.js` (17 cases) covers the normalizer (case/whitespace handling, rejection of `flex`/`default`/non-strings) and the adapter payload (fast → priority, omission when absent, unknown tier dropped, rest of the payload untouched).

```
npx vitest run unit/codex-image-service-tier.test.js unit/codex-fast-capacity.test.js unit/capabilities.test.js
Test Files  3 passed (3)
     Tests  32 passed (32)
```

The pre-existing `unit/codex-fast-capacity.test.js` still passes, confirming the executor refactor is behaviour-preserving.

Full-suite comparison before vs. after this branch, on the same checkout:

| | before | after |
|---|---|---|
| failed | 88 | 88 |
| passed | 1656 | 1673 |

Same 24 failing files in both runs — all pre-existing per `CLAUDE.md`. The +17 are this PR's new tests. `npx eslint` on the touched files reports 0 errors. Provider/alias baselines are unaffected (`params` is not snapshotted; `verify-alias.mjs` is byte-for-byte equal, and the one `verify-providers.mjs` diff — `tokenrouter.thinkingFormat` — reproduces on master without this branch).

## Note

I could not exercise this against a live ChatGPT Plus account, so the wire-level assertion that the image endpoint honours `service_tier` the same way the chat endpoint does is unverified. The field is already allowlisted for the same endpoint on the chat path, and the change is inert unless a caller opts in.

Codex is flagged `deprecated: true` in the registry — happy to close this if the provider isn't taking new features.

### DIFF
--- a/open-sse/executors/codex.js
+++ b/open-sse/executors/codex.js
@@ -12,6 +12,7 @@ import { getThinkingLevels } from "../providers/thinkingLevels.js";
 import { DEFAULT_RETRY_CONFIG, HTTP_STATUS, resolveRetryEntry } from "../config/runtimeConfig.js";
 import { dbg } from "../utils/debugLog.js";
 import { resolveSessionId } from "../utils/sessionManager.js";
+import { normalizeCodexServiceTier } from "../shared/codexServiceTier.js";
 
 // SSE error patterns inside 200-OK bodies. Some retry same account first; capacity rotates accounts.
 const CODEX_SSE_RETRY_PATTERNS = ["server_is_overloaded", "service_unavailable_error"];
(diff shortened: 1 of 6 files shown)


wecatclaude-bit:
@decolua giúp tí, mình muốn tăng tốc độ tạo ảnh lên bằng /fast

vutranHS:
> @decolua giúp tí, mình muốn tăng tốc độ tạo ảnh lên bằng /fast

Thực tế tạo ảnh là chậm như nhau cho dù có để service_tier cũng thế thôi, đóng PR đi

trananhtung:
OK. close

PR #3272, opened 2026-08-13
feat: add Oh My Pi (omp) CLI tool integration

krzemienski:
## What

Add **Oh My Pi (omp)** as a first-class 9Router **CLI tool** (consumer), not an upstream provider.

Dashboard → CLI Tools → Oh My Pi lets users select 9Router models and assign every official OMP role:

- `default`
- `smol`
- `slow`
- `vision`
- `plan`
- `designer`
- `commit`
- `tiny`
- `task`
- `advisor`

Apply writes:

- `~/.omp/agent/models.yml` — `providers.9router` with `api: openai-completions`, `baseUrl: http://127.0.0.1:20128/v1`, and `discovery.type: openai-models-list`
- `~/.omp/agent/config.yml` — selected `modelRoles.<role>: 9router/<model-id>` assignments

Unset and Reset remove only 9Router-owned official-role assignments. Foreign-provider assignments and custom role keys remain unchanged.

## How it mirrors OpenCode

| OpenCode | Oh My Pi |
|---|---|
| `~/.config/opencode/opencode.json` | `~/.omp/agent/models.yml` + optional role assignments in `config.yml` |
| `provider["9router"].options.baseURL` | `providers.9router.baseUrl` |
| JSON/JSONC writer | YAML document merge (`yaml` ^2.9.0) |
| Apply / Reset / status batch | Same dashboard card + `/api/cli-tools/omp-settings` + `all-statuses` |

Writer preserves YAML comments and unrelated keys, creates timestamped backups, rejects corrupt YAML without overwriting it, and validates role/model selections before either config file is written.

## Config sample (no secrets)

```yaml
# ~/.omp/agent/models.yml
providers:
  9router:
    baseUrl: http://127.0.0.1:20128/v1
    api: openai-completions
    apiKey: NINEROUTER_KEY
    authHeader: true
    discovery:
      type: openai-models-list
```

```yaml
# ~/.omp/agent/config.yml
modelRoles:
  default: 9router/kimi/kimi-k3
  vision: 9router/gcli/grok-4.6
```

## Live validation

- Initial branch base: **v0.5.35**, the latest GitHub Release when work began
- Conflict-resolution merge: current `upstream/master` at `15223724` (**v0.5.50**)
- `omp` **17.2.15**
- Isolated `HOME`; live dashboard and live 9Router catalog; user `~/.omp` hashes unchanged
- Dashboard assigned all 10 official roles, hard-reloaded, switched assignments, unset assignments, and applied changes
- `env HOME=<tmp> omp config get modelRoles` from `/tmp` matched the generated role map
- `env HOME=<tmp> omp models 9router` returned **46** live 9Router model IDs
- `env HOME=<tmp> omp -p --no-session --no-tools --model 9router/kimi/k3 "Reply with exactly: OK"` exited 0 and returned `OK`

## Verification

```text
Focused OMP tests: 35/35 passed
Changed OMP files: 0 ESLint findings
Production build: passed
Independent post-merge review: PASS
```

The repository-wide suite and lint remain red on clean `upstream/master`. Exact baseline attribution against the same toolchain found:

- Tests: merged tree and clean upstream each have the same **118** failing assertions; **0 introduced**
- ESLint: merged tree and clean upstream each have the same **331** findings; **0 introduced**

No GitHub Actions workflow was registered for the fork branch commit.

## Docs

- README + locale CLI matrices (vi, zh-CN, ja, ru, th, fa)
- CHANGELOG Unreleased
- `docs/ARCHITECTURE.md` client list

## Links

- https://github.com/can1357/oh-my-pi/blob/main/docs/models.md
- 9Router OpenAI `/v1` contract: `http://127.0.0.1:20128/v1`


src/app/(dashboard)/dashboard/cli-tools/components/OmpToolCard.js
```diff
@@ -0,0 +1,482 @@
...
+        ? effectiveSelectedApiKey
+        : (!cloudEnabled ? "sk_9router" : effectiveSelectedApiKey);
+      const capabilities = {};
+      for (const id of models || []) {
+        const m = catalogModels.find((x) => x.fullModel === id);
+        if (!m?.caps) continue;
+        capabilities[id] = { vision: !!m.caps.vision, search: !!m.caps.search, reasoning: !!m.caps.reasoning };
+      }
+      const roles = pruneRoleAssignments(roleAssignments, models);
+      await fetch("/api/cli-tools/omp-settings", {
```
krzemienski: `saveModels` ignores the POST response. If the user deselects the last model and closes the modal, this endpoint returns 400 because `models` is empty, but local state still shows no models while the previous `models.yml` contents remain persisted. Check `res.ok`; on failure, surface the error and rehydrate via `checkStatus()` (or make deselect-all a supported persistence path).

krzemienski (commented):
Independent review found one correctness issue at 100/100 confidence.

krzemienski:
## Update: model catalog + scoped role management

Pushed `6bb7e7fd`. The card now covers the full omp surface — add/remove models with real capabilities, and assign roles at the scope omp actually reads from.

### API split by scope

The two config files have different lifetimes, so they no longer share a request:

| | target | scope |
|---|---|---|
| `POST` | `~/.omp/agent/models.yml` | always global |
| `PATCH` | `config.yml` | `{scope, projectIndex}` |
| `DELETE` | catalog + roles | cleans every scope, transactionally |
| `GET` | both layers | reports omp's own `modelRoleStorage` |

### Bugs fixed (each found by driving the dashboard, not by reading code)

- **Capabilities dropped end to end.** The UI narrowed caps to `{vision, search, reasoning}` before POSTing, and `/api/models` hand-picked 5 fields and discarded `thinkingFormat`. Context window, max output and thinking mode never reached `models.yml`; every adaptive/budget model silently degraded to a generic `effort` ladder.
- **Roles written to the wrong file.** The writer hardcoded the global config, so with `modelRoleStorage: project` omp read `<project>/.omp/config.yml` and ignored everything the dashboard wrote.
- **Apply cleared untouched roles.** The card sent its whole pruned role map, wiping any assignment whose model was not in the picker. It now sends only the role ids the user edited.
- **Custom roles invisible.** omp treats the ten built-ins as a floor, not a ceiling. Roles beyond them are now read, rendered and assignable, and `DELETE` cleans them too.
- Removed a duplicate `omp` entry in `cliTools.js` (pointing at a non-existent `.png`) and a duplicate `ompGet` import that broke the build.

### Project scope is opt-in

Project roles write outside the app's data dir, so the feature is disabled unless the operator sets `OMP_PROJECT_ROOTS`. There is no project registry in the app and `process.cwd()` is the dashboard's own install dir, so neither can be trusted as a write target. When unset, the selector shows Project disabled with inline guidance rather than silently writing somewhere plausible.

Hardening: no client-supplied paths (index selection only), `O_NOFOLLOW` reads, symlink refusal in both directions, atomic same-directory renames, prototype-pollution guards, and a preflight that aborts before any write if a layer is corrupt or unsafe.

### Verification

Driven against a real dashboard under an isolated `HOME` with a fixture project root:

- add-model autosave writes `contextWindow: 1000000`, `maxTokens: 128000`, `mode: anthropic-adaptive`
- delete cleans role refs in **both** scopes, leaving custom and foreign-provider roles intact
- per-scope role drafts do not leak across layers (set in project → Global unaffected → back to project retained)
- Apply writes only the selected file; the other layer is byte-identical

88 tests, lint clean, build green. Live configs unchanged (checksums identical pre/post).

krzemienski:
### Code review

Found 1 issue:

1. Clicking a model chip to set or clear the Default role never marks `default` as touched, so Apply silently drops the change (bug due to `handleApply` building the PATCH body solely from `touchedRoleIds`: `for (const id of touched) roles[id] = pruned[id] ?? "";`)

The chip's `onClick` calls `setRoleAssignments` only, while the official Default `<select>` at [L644-L648](https://github.com/decolua/9router/blob/6bb7e7fdf613b0fab2ea18f86ad3f3e2049887cd/src/app/(dashboard)/dashboard/cli-tools/components/OmpToolCard.js#L644-L648) additionally calls `markRoleTouched`. A chip-only edit therefore produces `roles = {}`, which [omp-settings/route.js#L731-L734](https://github.com/decolua/9router/blob/6bb7e7fdf613b0fab2ea18f86ad3f3e2049887cd/src/app/api/cli-tools/omp-settings/route.js#L731-L734) treats as a 200 no-op. Apply then clears the dirty keys and `checkStatus()` rehydrates from disk, so the star reverts after a success toast.

https://github.com/decolua/9router/blob/6bb7e7fdf613b0fab2ea18f86ad3f3e2049887cd/src/app/(dashboard)/dashboard/cli-tools/components/OmpToolCard.js#L499-L507

Repro: select a model, click its chip to star it, click Apply, reload. Default is unchanged on disk.

Fix: call `markRoleTouched(currentScopeKey, "default")` in the chip handler on both set and clear.

Two lower-confidence findings scored below the reporting threshold and are noted only in passing: DELETE preflight skips allowlisted project roots that fail `resolveProjectTarget` and still writes the catalog ([route.js#L806-L809](https://github.com/decolua/9router/blob/6bb7e7fdf613b0fab2ea18f86ad3f3e2049887cd/src/app/api/cli-tools/omp-settings/route.js#L806-L809), scored 50 — opt-in and recoverable), and `OMP_PROJECT_ROOTS` is absent from `.env.example` and the ARCHITECTURE.md env matrix (scored 25 — neither file's env section is modified by this PR).

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

krzemienski:
## Hardening pass — `9e549e4f`

Follow-up to the initial omp integration, fixing two defects found while driving the real dashboard.

### 1. DELETE could orphan roles in an unreachable layer
When a configured `OMP_PROJECT_ROOTS` entry could not be resolved, the preflight skipped only that layer and deleted the model from the catalog anyway — leaving roles in the unreachable layer pointing at a model that no longer exists.

The preflight now distinguishes an *unreadable* root from an *absent* one (errno carried on a `code` field rather than matched by message text, which was one reworded string away from silently reverting) and aborts the whole operation with **409** before any write.

### 2. Model chip set `default` but it never reached disk
The chip set the `default` role without marking it touched, so Apply's delta PATCH omitted it while still clearing the scope's dirty flag — the assignment was silently dropped. The chip now marks on both set and clear.

The ✕ button deliberately does **not** mark: DELETE already blanks those roles server-side, and marking would make a later Apply PATCH `""` over any external reassignment that `checkStatus()` cannot re-read while the scope is dirty.

### Verification

| Gate | Result |
|---|---|
| Production build | `BUILD_EXIT=0` |
| omp suite | 92/92 pass |
| Full suite | 1869 passed (baseline 1865, **+4**) |
| Pre-existing failures | 92 at `6bb7e7fd`, 92 at `9e549e4f` — **Δ 0** |
| Lint | clean on both changed source files |
| `git diff-tree --check` | clean |

Baseline was measured in an isolated worktree at the parent commit, not recalled. The 92 failures (cursor DB, windsurf, protobuf, mitm) are unrelated to this change.

**Mutation-tested:** the 4 new regression cases were run against the pre-fix route — 2 fail without the fix, all pass with it, and the route was restored byte-exact (SHA-256 verified) afterward.

### End-to-end validation
Driven in a real browser against an isolated `HOME`, so no user config was touched:

- Add models → chips render → click chip sets `default` → Apply → `config.yml` contains `default: 9router/m1` → survives full page reload
- `<select>` role assignment → Apply → `smol: 9router/m2` written
- Apply enriches the catalog with capabilities: `input`, `supportsTools`, `compat.maxTokensField`, `compat.thinkingFormat`
- ✕ delete → 200, `cleanedScopes: ["global"]`, `smol` cleared, `default` preserved
- **The 409 fix fired on a real user path**: with an unresolvable project root, DELETE refused with "nothing was modified" and left the catalog intact

All 10 omp roles (`default, smol, slow, vision, plan, designer, commit, tiny, task, advisor`) render across both global and project scope, matching `model-roles.d.ts`.

krzemienski:
## Rebased on `master` (v0.5.59) + endpoint-preservation fix

Two commits since the last update:

| Commit | What |
|---|---|
| `833f116a` | Merge `upstream/master` (`90b52e06`, v0.5.59) — resolves the conflict that made this PR unmergeable |
| `58eba2ae` | `fix(cli-tools): keep a configured custom endpoint on re-init` |

PR state went `CONFLICTING`/`DIRTY` → `MERGEABLE`/`CLEAN`.

### The merge

`master` had moved 41 commits ahead. The only **conflict** was `CHANGELOG.md` — purely additive (this PR's `# Unreleased` section vs the new `# v0.5.59` heading), resolved by keeping both. Three of this PR's files changed as a result of the merge: `CHANGELOG.md` (conflict resolution) plus `package.json` and `src/shared/constants/cliTools.js`, which auto-merged to take upstream's edits. The remaining 26 are byte-identical to their pre-merge blobs.

### The fix (found by driving the dashboard, not by reading code)

While validating the Oh My Pi card end-to-end I hit a real bug, and it is **not** specific to this PR — it lives in the shared `BaseUrlSelect`:

1. Select **Custom URL** and enter an endpoint.
2. Add a model — the model-modal autosave fires and calls `checkStatus()`.
3. Click **Apply**.

The autosave `POST` carried the custom endpoint, but the Apply `POST` carried the **default** one, and that is what landed in `~/.omp/agent/models.yml`. The user's chosen endpoint was silently overwritten.

Cause: on a fresh mount / re-initialization after a status refresh, `BaseUrlSelect`'s initialization effect had no way to represent a configured URL that matches none of its options — i.e. a custom endpoint. It fell back to the first option and pushed that to the parent through `onChange()`, resetting `customBaseUrl`.

Fix: initialization now honours an unmatched `currentUrl` as the custom value instead of falling back to the first option, and `OmpToolCard` passes the endpoint omp is actually configured with. The selection logic moved into a pure exported `resolveInitialSelection()`. Behaviour is unchanged for the 12 other cards that already pass `currentUrl`, and for the unconfigured case (still lands on the first option).

### Validation — real system, no mocks

Ran against this branch on an isolated port (`:20129`) with an isolated `HOME` and `DATA_DIR`, driving a real headless Chrome over CDP.

| Check | Result |
|---|---|
| Dashboard → CLI Tools → **Oh My Pi** card renders | PASS |
| Add Model correctly disabled with 0 providers, enabled with a real one | PASS |
| Apply writes `~/.omp/agent/models.yml` | PASS |
| Endpoint written = the one selected in the UI | **PASS (was the bug)** |
| Pre-existing unrelated provider preserved | PASS |
| Foreign role + unrelated `config.yml` key preserved | PASS |
| `omp models 9router` lists the models | PASS — real omp **v18.1.2** |
| Real completion through the gateway | PASS — below |

The completion was driven through the real `omp` binary against this branch:

```
$ omp -p --model "9router/prod9r/cf/@cf/meta/llama-3.1-8b-instruct-fp8-fast" "Reply with exactly: OK"
Working...
OK
$ echo $?
0
```

`prod9r/*` is a second 9Router attached to the test instance as an `openai-compatible` upstream node, so the path is `omp → this branch (:20129) → upstream → model`.

### Regression posture

Three arms, same machine, `CI` unset (with `CI=true`, vitest refuses to write the missing `golden-url-header` snapshot and fails 124 tests on *any* checkout, including clean `master`):

| Arm | Failed | Passed |
|---|---|---|
| this PR before the merge (`9e549e4f`) | 90 | 1871 |
| **pure `upstream/master` (`90b52e06`)** | **85** | 1923 |
| **this branch merged (`58eba2ae`)** | **85** | **2015** |

Merged failures equal pure upstream exactly. A normalized failure-set diff of merged vs upstream gives **0 regressions** and **0 failing tests unique to this branch**; the PR contributes **+92 passing tests**.

- `npm run build` — passes.
- `npx eslint` on both changed files — **2** `react-hooks/set-state-in-effect` errors, both **pre-existing**: clean `upstream/master` reports the same 2 in this file. Not introduced by this change, and not reduced by it.
- `tests/__baseline__/verify-no-regression.mjs` is not usable outside Docker — line 17 does `f.name.split("/app/")[1]`, so every test name resolves to `undefined` and it exits 1 on pure `upstream/master` too. The 3-arm diff above was used instead.

### Suggested follow-up (not included)

A regression test for `resolveInitialSelection()` covering the exact sequence — set custom URL → presets re-sync → selection survives — would be worth adding; my tooling blocked me from creating the test file. Happy to add it if you'd like it in this PR.

Two upstream issues surfaced during validation, both outside this PR's scope and untouched by it:

- Streaming to `ollama-local` fails: the executor rejects Ollama's `application/x-ndjson` as non-SSE (`BLOCKED 200 · non-SSE`). Non-streaming works (HTTP 200). Since `omp` always streams, an Ollama-backed model can't serve it today.
- `ollama-local` has no static model list and no discovery, so it contributes nothing to `/v1/models` and can only be reached by explicit id.

krzemienski:
## Verification pass — real `omp` 18.1.4 against a live 9Router

Re-verified this branch's **integration route and CLI path** in a clean worktree at the PR head, with the dashboard fully isolated (`HOME` + `DATA_DIR` in temp dirs) so nothing touched real user config. Scope and gaps are stated precisely below.

### What was proven

| Check | Result |
|---|---|
| `POST /api/cli-tools/omp-settings` (the call the card's Apply button makes) | 200, `has9Router` false → true |
| Generated `~/.omp/agent/models.yml` | matches the documented contract exactly |
| `omp models 9router` | **exit 0**, 63 models discovered via `openai-models-list` |
| `omp -p --model 9router/cc/claude-fable-5 "Reply with exactly: OK"` | **exit 0**, output `OK` |

Generated YAML (verbatim, key is an env-var name — no secret written):

```yaml
providers:
  9router:
    baseUrl: http://127.0.0.1:20128/v1
    api: openai-completions
    apiKey: NINEROUTER_KEY
    authHeader: true
    discovery:
      type: openai-models-list
```

Capabilities are translated from the gateway, not invented: `glm/glm-5.2` (vision:false) → `input: [text]` with `thinking.mode: effort`; `cc/claude-fable-5` (vision:true) → `input: [text, image]` with `thinking.mode: anthropic-adaptive` and a legal `compat.thinkingFormat: openai`.

### Topology of the run (stated explicitly)

- The **dashboard and the `omp-settings` route under test ran from this branch** on port **20139** (isolated `HOME`/`DATA_DIR`).
- Port 20128 was held by a separately installed 9Router instance, so the **model discovery and the completion were served by that gateway**, not by this branch's process. This branch produced the config; the installed gateway answered the traffic.

### Quality gates

- OMP unit tests: **88/88 passed** (4 files)
- Regression delta vs `upstream/master` (90b52e06), same environment: baseline **209** failing assertions / this branch **209** → **0 new, 0 fixed**. Note `tests/__baseline__/known-fails.txt` catalogues only 24 of them; the remaining failures are pre-existing in this environment but not individually catalogued. The zero delta is the load-bearing claim, not the absolute count.
- `npm run build`: exit 0, `/api/cli-tools/omp-settings` present in the route manifest
- ESLint on the 5 OMP files: exit 0, zero problems
- Full-repo `npx eslint .`: **fails** — `✖ 333 problems (135 errors, 198 warnings)` across 213 files, **0 of them in OMP paths**. Pre-existing, unchanged by this PR.
- `open-sse/` untouched — no routing-engine or upstream-provider risk

### Change in this push (`7d3f443a`)

`docs(omp)`: the `models.yml` example writes the env-var *name* as `apiKey`, but the verify block never exported it, so a copy-paste run hit an unresolved key. Added `export NINEROUTER_KEY=…` to the English, zh-CN and vi sections, then re-ran the documented sequence (exit 0, 63 models).

### Base provenance — FAIL against the stated requirement

This branch is **not** based on the latest *published GitHub Release*. `gh release list` shows the newest published Release is **v0.5.35**. This PR head is **180 commits past v0.5.35**, based on `master`, which carries the **v0.5.59 git tag** (a tag, not a published Release).

Recorded as a **deviation**, not a pass. Rationale: rebasing onto v0.5.35 would base the work on a ~4-month-older tree and would very likely break the currently CLEAN mergeability against `master`. Happy to rebase onto v0.5.35 if maintainers prefer that base.

### Not verified

- **Dashboard Add Model → Apply click-path: UNVERIFIED.** The modal is gated on `activeProviders?.length`, which requires an OAuth-connected provider. I had no credentials for one and would not use the user's. The route it calls is proven above, but the button-level journey was not driven.
- **This branch did not serve a completion.** See topology above.

krzemienski:
### Verification refresh — final HEAD `20e48f35`

Superseding the figures in my earlier comment, which were captured at `7d3f443a`. All values below were measured at the current head.

| Gate | Result at `20e48f35` |
|---|---|
| OMP unit tests | `4 files / 88 tests passed`, exit 0 (`npx vitest run unit/omp`) |
| `npm run build` | exit 0; `/api/cli-tools/omp-settings` present in the route manifest |
| `npx eslint .` (whole repo) | `✖ 333 problems (135 errors, 198 warnings)` |
| Same command on `upstream/master` (clean worktree) | `✖ 333 problems (135 errors, 198 warnings)` — identical, so this branch introduces none |
| `npx eslint` scoped to the 5 OMP files | exit 0, empty output — zero findings |
| Test regression delta | 209 failing before / 209 after — 0 new, 0 fixed |
| GitHub checks | none configured on this PR (`Checks: 0`), so the above is the only verification signal |

The two ESLint rows are separate measurements: a full-repo run on each of the two trees, plus a scoped run naming `omp-settings/route.js`, `OmpToolCard.js`, `ompModelSelection.js`, `ompRoles.js`, `ompModelSchema.js` explicitly.

### End-user validation (dashboard click-path)

Previously reported as UNVERIFIED. Now exercised in an **isolated headless Chrome instance** (separate profile, no access to any live browser session) driving this branch on an isolated port (`20140`) with a sandboxed `HOME` and a transactionally consistent (`sqlite3 .backup`) copy of a populated provider database:

Dashboard → CLI Tools → Oh My Pi → **Add Model** → select model → **Apply** → assign the `default` role → **Apply** → `Applied — roles written to global config`.

Artifacts written by that click-path:
- `config.yml` → `modelRoles: { default: 9router/minimax/MiniMax-M2.1 }`
- `models.yml` → `9router` provider with `baseUrl`, `api: openai-completions`, `authHeader: true`, `discovery.type: openai-models-list`, plus per-model capabilities (context window, max tokens, thinking efforts, compat flags)

Consumed by the real CLI: `omp models find 9router` → **exit 0, 68 lines**.

### Docs fix in `20e48f35`

- `CHANGELOG.md` — removed an Oh My Pi bullet that had been inserted into the already-shipped `v0.5.50` section. Verified against primary sources: `git show v0.5.50:CHANGELOG.md` contains zero OMP mentions, and after removal the section is byte-identical to the tagged release. The correct entries remain under `# Unreleased`.
- `README.zh-CN.md` / `i18n/README.vi.md` — added the two blocks that existed only in the English README (the optional default-role pin, and the `127.0.0.1`-over-`localhost` note with the upstream docs link). Three-way parity is now verified.

### Known, unchanged

- **Base provenance** — as disclosed above, this branch is 180 commits past published Release `v0.5.35` and sits on `master`. Left unchanged by preference; a rebase is available on request.
- **Behavior note for maintainers:** selecting a stored dashboard key writes that key literally into `models.yml`, while the README documents the env-var form (`apiKey: NINEROUTER_KEY`). Both work. This matches the existing OpenCode card's behavior, so it is left as-is — flagged as a design question rather than changed unilaterally.
- **Pre-existing, outside this PR's scope:** `i18n/README.vi.md` references `cc/claude-opus-4-6` where the English and Chinese READMEs use `cc/claude-opus-4-7`. That drift predates this branch and was left untouched.



---

*Evidence provenance: the dashboard click-path above was driven in an isolated headless Chrome instance against a sandboxed server on port 20140. Verification that this comment and the PR state render correctly on github.com was performed separately through a live browser session. No production 9Router instance, user configuration, or credential store was modified at any point.*

PR #3275, opened 2026-08-13
fix(combo): detect images from Hermes and attachment payloads for Vision Adapter

berthojoris:
## Summary
Fixes image capability detection in `detectRequiredCapabilities` so that Vision Adapter auto-switch is correctly triggered when images are sent via Hermes (e.g. Telegram bot interface) or other agents using non-standard attachment shapes.

## Root Cause
Hermes and related agent tools transmit images using Ollama `images: [...]` arrays, Vercel AI SDK `experimental_attachments: [...]` / `attachments: [...]`, or inline `data:image/` content URIs rather than standard OpenAI `content` block arrays (`[{ type: "image_url", ... }]`).

`detectRequiredCapabilities` was previously skipping `m.images` and attachment fields on user messages. As a result:
1. Requests with images from Hermes were classified as text-only (`requiredCapabilities = []`).
2. 9Router did not trigger auto-switching to the configured Vision Adapter pool (e.g. `Kimi-K3`).
3. The request proceeded to text-only models (like `deepseek-v4-pro`), which stripped the image before forwarding.

## Changes Made
- **`open-sse/services/combo.js`**:
  - Updated `detectRequiredCapabilities` to inspect `m.images`, `m.experimental_attachments`, `m.attachments`, direct `m.image`/`m.image_url` properties, and inline `data:image/` URIs on trailing user turns.
- **`open-sse/translator/concerns/modality.js`**:
  - Updated `stripOpenAI` to clean up `msg.images` and attachment fields when forwarding to models without vision capability.
- **`tests/unit/hermes-vision-detection.test.js`**:
  - Added unit test suite covering Hermes/Ollama/attachment image detection and Vision Adapter auto-switch behavior.

## Test Plan
- [x] Ran unit tests: `npx vitest run unit/hermes-vision-detection.test.js` (6/6 tests passed).
- [x] Verified existing combo auto-switch logic is preserved.

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -138,8 +138,42 @@ export function detectRequiredCapabilities(body) {
     if (Array.isArray(content)) for (const b of content) scanBlock(b);
   };
 
+  const scanMessage = (m) => {
+    if (!m || typeof m !== "object") return;
+
+    // Ollama / Hermes images array (strings or objects)
+    if (Array.isArray(m.images) && m.images.length > 0) {
+      required.add("vision");
+    }
+
+    // Vercel AI SDK / Hermes attachments / experimental_attachments
+    const attachments = m.experimental_attachments || m.attachments;
+    if (Array.isArray(attachments)) {
+      for (const att of attachments) {
+        if (!att) continue;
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @berthojoris for the contribution! Reviewed and merged into master. 🙏

PR #3276, opened 2026-08-13
fix(cursor): tunnel AgentService through HTTP/2 proxy so vendor models stream

craboy4:
## Summary

Cursor's AgentService is HTTP/2-only and **gates both the model catalog and chat by client IP**. `http2.connect()` ignores proxy pools, `HTTP_PROXY`, and the system proxy. This PR tunnels **GetUsableModels and AgentService/Run** through the same HTTP CONNECT / SOCKS path so a proxied Cursor connection can list *and* use international vendor models.

### 1. Catalog was proxied; chat was not

`GetUsableModels` used a direct `http2.connect()` to `agent.api5.cursor.sh`, so the picker stayed on the regional catalog (Grok / Kimi / GLM) even after a proxy pool was bound.

This PR (first commit) tunnels that unary call: CONNECT/SOCKS → TLS ALPN h2 → HTTP/2, using the same proxy order as other providers (connection pool first, then outbound / env). The catalog cache key includes the proxy URL so direct and proxied lists do not overwrite each other.

### 2. Agent Run still went out on the machine IP

ChatService (`api2.cursor.sh`) is retired (HTTP 464). Pure-text turns already go to `AgentService/Run` on `agent.api5.cursor.sh`, but Run still used a raw `http2.connect()`. Cursor then accepted Claude / GPT / Fable with **HTTP 200 and an empty stream** (no error frame). Grok/Composer on the same token worked because they are in the regional catalog.

Run now uses `connectHttp2(url, { proxyUrl })`. A bound proxy that cannot be resolved **fails closed** instead of silently going direct. The tunnel waits for the h2 `connect` event, sets `TCP_NODELAY`, and destroys the TLS socket when the session closes.

### 3. Claude / GPT need a KV blob handshake

`AgentServerMessage` field 4 is `kv_server_message` (`get_blob_args` / `set_blob_args`). Ignoring it leaves heartbeats only, then an empty assistant. Grok often does not wait on this; Anthropic / OpenAI models do.

The executor now keeps an in-memory blob store and acks get/set via `AgentClientMessage.kv_client_message`. Proto3: omit KV `id` when the server did not send one (do not write `id=0`).

### 4. `RequestedModel` field 7 was not a flag

`agent.v1.RequestedModel` is `model_id = 1`, `max_mode = 2`, `parameters = 3`. Field 7 on nearby messages is Bedrock credentials. Writing `agentBool(7, true)` was incorrect. The run frame now sends `model_id` only.

### 5. Fable has no `*-fast` catalog slugs

Live `GetUsableModels` includes `claude-opus-5-thinking-max-fast` and `gpt-5.6-sol-max-fast`, but **no** `claude-fable-*-fast`. Clients that copy the Opus name (`claude-fable-5-thinking-max-fast`) get HTTP 200 + empty stream. Real ids are `claude-fable-5-thinking-max` and `claude-fable-5-max`.

`resolveCursorAgentModel()` strips a trailing `-fast` **only** for Fable before `RequestedModel.model_id`. Opus / GPT / Grok are unchanged.

### 6. Other stream fixes

- Abort no longer emits `done` with empty content (was recorded as a successful blank turn).
- `interaction_update` thinking deltas (field 4) are forwarded as `reasoning_content`.
- Stats log: frames, text chars, server/update field numbers, to debug empty streams.

## 中文说明

Cursor 按 **客户端 IP** 同时裁剪「可用模型」和「对话」。`http2.connect()` 不走 9Router 代理池，也不读 `HTTP_PROXY`。

先前只有 **GetUsableModels** 做了 HTTP/2 隧道，所以面板里能刷出国际目录（Opus / GPT / Fable），但 **AgentService/Run 仍直连本机 IP**。表现是：Grok / Composer 正常出字，Claude / GPT / Fable 返回 HTTP 200、没有任何错误帧、内容为空。旧 ChatService（`api2.cursor.sh`）已经 464，纯文本都走 `agent.api5.cursor.sh` 的 Run。

这次把 Run 接到同一条 CONNECT/SOCKS 隧道。绑定了代理却解析失败会直接报错，不再静默直连。另外补了 Claude/GPT 必需的 KV blob 握手（忽略的话只剩心跳然后空回复）、去掉误写进 `RequestedModel` field 7 的 bool（那不是 max-mode 开关）、以及 Fable 没有 `*-fast` 目录 ID 的映射（`claude-fable-5-thinking-max-fast` → `claude-fable-5-thinking-max`）。Opus/GPT/Grok 的 `-fast` 不动。

## Test plan

- [x] `cd tests && npx vitest run unit/cursor-models.test.js unit/cursor-agent-exec-request.test.js`
- [x] Bind an HTTP proxy pool to a Cursor connection; `/v1/models` lists international ids (`claude-opus-5-*`, `gpt-5.6-sol-*`, `claude-fable-5-*`)
- [x] Same connection without proxy: catalog stays regional (expected Cursor IP gate)
- [ ] Live chat (`/v1/chat/completions` or `/v1/responses`):
  - `cu/cursor-grok-4.6-xhigh-fast` still streams
  - `cu/claude-opus-5-thinking-max-fast` streams (not empty 200)
  - `cu/gpt-5.6-sol-max-fast` streams
  - `cu/claude-fable-5-thinking-max` streams
  - `cu/claude-fable-5-thinking-max-fast` also streams (mapped to `thinking-max`)
- [ ] Logs show `GetUsableModels via HTTP/2 proxy` and `CURSOR_AGENT Run via HTTP/2 proxy`
- [ ] Fable-fast logs `resolved …-fast → …`
- [ ] Bound proxy that cannot be resolved fails closed
- [ ] Aborting a stream is not recorded as an empty successful completion

### DIFF
--- a/open-sse/executors/cursor.js
+++ b/open-sse/executors/cursor.js
@@ -14,7 +14,8 @@ import { estimateUsage } from "../utils/usageTracking.js";
 import { SSE_DONE, SSE_HEADERS } from "../utils/sseConstants.js";
 import { chatChunkSse, sseChunk } from "../utils/sse.js";
 import { FORMATS } from "../translator/formats.js";
-import { proxyAwareFetch } from "../utils/proxyFetch.js";
+import { proxyAwareFetch, resolveOutboundProxyUrl } from "../utils/proxyFetch.js";
+import { connectHttp2 } from "../utils/http2Connect.js";
 import zlib from "zlib";
 import crypto from "crypto";
 
@@ -46,6 +47,15 @@ const AGENT_RUN_PATH = "/agent.v1.AgentService/Run";
 const PROTOBUF_LEN = 2;
 const PROTOBUF_VARINT = 0;
 
+function maskProxyUrl(proxyUrl) {
+  try {
+    const parsed = new URL(proxyUrl);
+    return `${parsed.protocol}//${parsed.hostname}${parsed.port ? `:${parsed.port}` : ""}`;
+  } catch {
+    return "proxy";
+  }
+}
+
 function concatBuffers(...parts) {
   const length = parts.reduce((total, part) => total + part.length, 0);
   const result = new Uint8Array(length);
@@ -59,7 +69,6 @@ function concatBuffers(...parts) {
 
 const agentString = (field, value) => encodeField(field, PROTOBUF_LEN, value);
 const agentMessage = (field, value) => encodeField(field, PROTOBUF_LEN, value);
-const agentBool = (field, value) => encodeField(field, PROTOBUF_VARINT, value ? 1 : 0);
 
 function textFromContent(content) {
   if (typeof content === "string") return content;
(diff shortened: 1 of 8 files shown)


AMSeify:
This looks like the same Cursor AgentService empty-reply / hang cluster. Tracking the framing fix in https://github.com/decolua/9router/issues/4016 and PR https://github.com/decolua/9router/pull/4017. cc @decolua @anuragg-saxenaa

PR #3278, opened 2026-08-13
feat(opencode-go): route by request format via transports + per-model guard

### LINKED ISSUE #3279: feat: route opencode-go by client request format (Claude Code / Codex / OpenAI)
## Problem

opencode-go exposes multiple upstream endpoints (`/chat/completions`, `/messages`, `/responses`), but 9router currently hard-codes `targetFormat: "claude"` for a fixed model set (MiniMax/Qwen). As a result:

- Every client calling MiniMax/Qwen is forced through `/messages` — even Codex/OpenAI clients, which then get translated Responses→OpenAI→Claude (two lossy hops: thinking blocks, tool ids).
- DeepSeek, which supports `/responses`, can't be reached natively by Codex.
- The bespoke `OpenCodeGoExecutor` + `MESSAGES_FORMAT_MODELS` set duplicates routing logic that the `transports` mechanism already handles generically.

## Proposal

Pick the opencode-go upstream endpoint by the **client's request format**, using the existing `transports` mechanism (the same one other multi-endpoint providers already use):

- Claude Code (`/v1/messages`) → `/messages` (native Anthropic, thinking block preserved)
- Codex (`/v1/responses`) → `/responses` (native Responses)
- OpenAI (`/v1/chat/completions`) → `/chat/completions` (native OpenAI chat)

Because opencode-go models differ in endpoint support (kimi/glm/mimo only do `/chat/completions`; minimax/qwen also do `/messages`; deepseek also does `/responses`), add a per-model `supportedFormats` guard so a model is only routed to a transport it actually supports — undeclared models keep current behavior.

## Benefit

- Lossless routing for the common cases (Claude Code↔Anthropic, Codex↔Responses).
- Removes a bespoke executor + hard-coded model set → less special-casing.
- opencode-go becomes consistent with every other multi-endpoint provider.

## Implementation

PR: #3278

KiMelody:
## What

opencode-go currently hard-codes `targetFormat: "claude"` per-model (MiniMax/Qwen always hit `/messages` regardless of caller). This wires it into the existing upstream multi-endpoint `transports` mechanism, so the endpoint is picked by the **client's request format** — skipping the lossy OpenAI-pivot double translation:

| Client | Hits | Upstream |
|---|---|---|
| Claude Code (`/v1/messages`) | `/messages` | native Anthropic |
| Codex (`/v1/responses`) | `/responses` | native Responses |
| OpenAI (`/v1/chat/completions`) | `/chat/completions` | native OpenAI chat |

## Why

- **Lossless**: a Codex/OpenAI request to MiniMax was force-routed to `/messages`, translating Responses→OpenAI→Claude (two hops, with thinking/tool_id loss). Now it uses whichever endpoint matches the caller.
- **Consistency**: opencode-go was the only multi-endpoint provider not yet using the `transports` mechanism. This removes its bespoke `OpenCodeGoExecutor` and the hard-coded `MESSAGES_FORMAT_MODELS` set.

## Changes

- **`registry/opencode-go.js`**: declare `transports: [openai, claude, openai-responses]`; per-model `supportedFormats`; drop hard-coded `targetFormat: "claude"`.
- **`chatCore.js`**: per-model guard on `resolveTransport` — only use the sourceFormat-matched transport when the model declares support (opencode-go models differ: kimi/glm/mimo only do `/chat/completions`). Undeclared models keep the upstream default.
- **`schema.js` + `providerModels.js`**: `modelSupportedFormats` / `getModelSupportedFormats`.
- **`executors`**: drop `OpenCodeGoExecutor` — opencode-go now uses `DefaultExecutor`, which already reads `runtimeTransport` and calls `injectReasoningContent`.

## Behavior change (intended)

- **MiniMax/Qwen + OpenAI/Codex client**: was `/messages` (force-translated), now `/chat/completions` (native). Codex→MiniMax drops a 2-hop translation to 1-hop.
- **MiniMax/Qwen + Claude Code**: still `/messages` (unchanged), thinking block preserved natively.
- **DeepSeek + Codex**: `/responses` (new — DeepSeek declares all three formats).

## Per-model endpoint matrix

| model | supportedFormats |
|---|---|
| glm / kimi / mimo | `["openai"]` |
| minimax / qwen | `["openai", "claude"]` |
| deepseek-v4-pro / flash | `["openai", "claude", "openai-responses"]` |

## Tests

- Rewrote `tests/unit/opencode-go-models.test.js`: catalog, `supportedFormats`, `transports`, per-model guard matrix. **11/11 pass**.
- **Zero regression**: full unit suite run head-to-head against `upstream/master` in the same environment — **93 fails both before and after, failure sets identical (diff = 0)**.
- Smoke-tested live against opencode.ai upstream: Claude Code→MiniMax→`/messages` (200, thinking block preserved), Codex→DeepSeek→`/responses` (200), OpenAI→Kimi→`/chat/completions` (200), and the guard case Codex→MiniMax→falls back to `/chat/completions` (not `/responses`).

## Note

May touch some of the same files as #3250 (also opencode-go) — that one covers plan-usage detection, this one covers endpoint routing; different logic.

Resolves #3279


decolua:
Thanks @KiMelody for the contribution! Reviewed and merged into master. 🙏

PR #3285, opened 2026-08-13
fix(open-sse): send opencode-cli headers for zen free models to avoid 429

### LINKED ISSUE #3262: Rate limit exceed on first tool call
Using OpenCode Free Provider the two models oc/deepseek-v4-flash-free and oc/mimo-v2.5-free gives Rate limit exceed error on first tool call.
it was tested on 5 different people, with different ip (clean ip with no usage recently), different machines, still the same.

Tested on Agent: Hermes

Tested on the 9router and similar projects like omnirouter and the issue persist.

Tested on the Opencode Cli and Opencode desktop app, its working well, no rate limit error.

### LINKED ISSUE #3270: opencode deepseek v4 flash 429

saeedhosseiinii:
### Problem

Requests to the **OpenCode Free** (`oc`) provider are rate-limited with `429 FreeUsageLimitError` even when quota is available. opencode.ai classifies requests as anonymous unless they carry the same client fingerprint headers the official opencode CLI sends. 9router only sent `x-opencode-client: desktop` and no `x-opencode-session` / `x-opencode-project` / `x-opencode-request` / `User-Agent`, so free-tier requests were throttled after a few calls per day even though direct opencode CLI usage on the same IP was fine.

### Fix

Send the official CLI fingerprint headers in `OpenCodeExecutor.buildHeaders()` while keeping `x-opencode-client: desktop`:

- `x-opencode-session` / `x-opencode-request` – dynamically generated ids (`ses_`/`req_`)
- `x-opencode-project: /opencode`
- `User-Agent: opencode/latest/1.18.18/cli`

Verified that `User-Agent` is required (without it, `429` persists) and that `desktop` is a valid client value.

### Verification

After the fix, `POST /v1/chat/completions` with `model: oc/deepseek-v4-flash-free` returns HTTP 200 with a normal completion; before the fix it returned 429 `FreeUsageLimitError`.

Fixes #3270
Fixes #3262

### DIFF
--- a/open-sse/executors/opencode.js
+++ b/open-sse/executors/opencode.js
@@ -22,11 +22,21 @@ export class OpenCodeExecutor extends BaseExecutor {
   }
 
   buildHeaders() {
+    const randomId = (length) =>
+      Array.from({ length }, () =>
+        "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789".charAt(
+          Math.floor(Math.random() * 62)
+        )
+      ).join("");
     return {
       "Content-Type": "application/json",
       "Authorization": "Bearer public",
       "x-opencode-client": "desktop",
-      "Accept": "text/event-stream"
+      "x-opencode-session": `ses_${randomId(24)}`,
+      "x-opencode-project": "/opencode",
+      "x-opencode-request": `req_${randomId(24)}`,
+      "User-Agent": "opencode/latest/1.18.18/cli",
+      "Accept": "text/event-stream",
     };
   }
 }


saeedhosseiinii:
**Temporary workaround until this PR is merged.**

Save as `patch-9router-429fix.sh`, then run:

```bash
sudo bash patch-9router-429fix.sh
```

Idempotent (re-run is a no-op). Re-run after every `npm update -g 9router` until this PR is merged.

```bash
#!/usr/bin/env bash
# 9router OpenCode Free 429 fix — applies the official opencode-CLI fingerprint
# headers to the installed 9router bundle without waiting for upstream merge.
# Root cause: https://github.com/decolua/9router/pull/3285
#
# Usage: sudo bash patch-9router-429fix.sh
# Idempotent: safe to re-run; no-op when already patched.
set -u

OLD='buildHeaders(){return{"Content-Type":"application/json",Authorization:"Bearer public","x-opencode-client":"desktop",Accept:"text/event-stream"}}'
NEW='buildHeaders(){let G=(n)=>[...Array(n)].map(()=>"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789"[Math.floor(62*Math.random())]).join("");return{"Content-Type":"application/json",Authorization:"Bearer public","x-opencode-client":"desktop","x-opencode-session":"ses_"+G(24),"x-opencode-project":"/opencode","x-opencode-request":"req_"+G(24),"User-Agent":"opencode/latest/1.18.18/cli",Accept:"text/event-stream"}}'

find_chunks_dir() {
  local bin dir p
  bin=$(command -v 9router 2>/dev/null)
  if [ -n "$bin" ]; then
    dir=$(readlink -f "$bin" 2>/dev/null)
    dir=$(dirname "$(dirname "$dir")")
    p="$dir/app/.next-cli-build/server/chunks"
    [ -d "$p" ] && { echo "$p"; return 0; }
  fi
  for p in /usr/local/lib/node_modules/9router/app/.next-cli-build/server/chunks \
           /usr/lib/node_modules/9router/app/.next-cli-build/server/chunks; do
    [ -d "$p" ] && { echo "$p"; return 0; }
  done
  return 1
}

CHUNKS=$(find_chunks_dir) || { echo "ERROR: could not locate 9router bundle (chunks dir)"; exit 1; }
echo "chunks dir: $CHUNKS"

TARGET=""
for f in "$CHUNKS"/*.js; do
  if grep -qF "$OLD" "$f"; then TARGET="$f"; break; fi
done

if [ -z "$TARGET" ]; then
  if grep -rqF '"x-opencode-session"' "$CHUNKS" 2>/dev/null; then
    echo "Already patched — nothing to do."
    exit 0
  fi
  echo "ERROR: opencode transport not found (9router updated? file layout changed?)."
  echo "Re-run after checking the bundle, or wait for upstream PR #3285."
  exit 1
fi

echo "patching: $TARGET"
mkdir -p "$HOME/.9router-backup-429fix"
cp -v "$TARGET" "$HOME/.9router-backup-429fix/"
python3 - "$TARGET" "$OLD" "$NEW" <<'PYEOF'
import sys
path, old, new = sys.argv[1], sys.argv[2], sys.argv[3]
s = open(path, encoding="utf-8").read()
assert s.count(old) == 1, f"expected 1 match, got {s.count(old)}"
open(path, "w", encoding="utf-8").write(s.replace(old, new))
print("patched OK")
PYEOF
node --check "$TARGET" && echo "syntax OK"

echo "restarting 9router..."
if systemctl list-unit-files 2>/dev/null | grep -q '^9router.service'; then
  systemctl restart 9router && echo "restarted via systemd"
else
  echo "No systemd unit found. Restart 9router manually (e.g. restart your tmux/screen/PM2 session)."
fi

echo
echo "Done. The OpenCode Free (oc) provider now sends the CLI fingerprint headers."
echo "Verify: curl -X POST http://127.0.0.1:20128/v1/chat/completions -H 'Authorization: Bearer <key>' -H 'Content-Type: application/json' -d '{\"model\":\"oc/deepseek-v4-flash-free\",\"messages\":[{\"role\":\"user\",\"content\":\"hi\"}]}'"
echo "A backup was saved to $HOME/.9router-backup-429fix/"
```

rafia9005:
thxxxx.. my issues fixed @saeedhosseiinii

vibecoder11200:
Thank you @saeedhosseiinii, it solved my problem. Could you give me your contact information, please? There are a few things I'd like to follow up on.

MuhammadAhmedRiaz:
Thank you @saeedhosseiinii it fixed my problem, now I can enjoy my unlimited coding sessions.

decolua:
Thanks @saeedhosseiinii for the investigation and the PR! 🙏

This turned out to be already fixed on our side by a change that sends the full opencode CLI fingerprint (session/request/project headers + User-Agent, with conversation-stable session ids and passthrough when the downstream client is the real opencode CLI). We verified live that the fingerprint headers resolve the 429 FreeUsageLimitError — your session/request header analysis was exactly right and helped confirm the root cause.

Closing as already fixed, but thank you for the contribution!

wuhanwuyanzu123:
still 429 only work for 24 hours after update the 9router

PR #3287, opened 2026-08-14
fix(usage): read Gemini usageMetadata out of the antigravity response envelope

### LINKED ISSUE #3260: Title: Antigravity: usage always shows IN 0 · OUT 0 despite successful DONE (v0.5.50)
## Problem

All requests routed through the antigravity provider complete successfully (DONE with real latency) but usage is always logged as IN 0 · OUT 0, regardless of model, tool presence, or account.

## Environment
- 9Router version: v0.5.50 (Docker, decolua/9router:latest)
- Provider: antigravity (direct OAuth connection, not MITM/IDE — IDE is never opened)
- 4 rotating Google accounts connected to the antigravity provider
- Models used: gemini-3.6-flash-high, gemini-pro-agent (auto-selected upstream, not manually chosen)

## What I've ruled out
- Not tool-call specific: happens both on requests with 3 TOOL and on requests with zero tool calls (e.g. plain 57 MSG request with no TOOL/THINK fields).
- Not model specific: happens on both gemini-3.6-flash-high and gemini-pro-agent.
- Not account specific: happens across all 4 rotating accounts.
- Not a routing/passthrough issue: ag/gemini-pro-agent correctly maps to antigravity/gemini-pro-agent (issue #1221 appears resolved in this version).

## Log sample

[16:43:22] POST ag/gemini-3.6-flash-high -> antigravity/gemini-3.6-flash-high | FMT: openai->antigravity | JSON | 16 MSG | 3 TOOL | THINK:high | ACC:[redacted]
[16:43:29] DONE 6987ms | IN 0 | OUT 0

[16:57:56] POST ag/gemini-pro-agent -> antigravity/gemini-pro-agent | FMT: openai->antigravity | JSON | 57 MSG | ACC:[redacted]
[16:58:01] DONE 4825ms | IN 0 | OUT 0

Note: the second example has no TOOL or THINK fields at all — still 0/0.

## Additional observation (possibly related, possibly separate)

One of the 4 connected accounts intermittently fails project ID onboarding:

[ProjectId] Onboarding user with tier: standard-tier
[ProjectId] onboardUser attempt 1 failed: onboardUser done but no project_id in response, retrying...
[ProjectId] onboardUser failed after 5 attempts: onboardUser done but no project_id in response
[ProjectId] could not fetch projectId for connection d82f9791

This request still completes and still logs 0/0, so it may not be the root cause of the usage issue, but flagging it in case it's relevant.

## Expected

Usage extraction from the Antigravity/Cloud Code Assist API response should report actual input/output token counts (or a reasonable estimate when the provider omits them, per the documented fallback behavior), the same way it works for other providers.

## Question

Since v0.5.18 shipped "Usage: track cached tokens + correct input/output/cache cost" (#2209), is Antigravity's usage extraction covered by that fix, or does it use a separate code path that might be missing usage parsing entirely?

ntdatt812:
Fixes #3260

## Problem

Every non-streaming request through the antigravity provider logs zero usage, whatever the model, tool count or account:

```
[16:43:22] POST ag/gemini-3.6-flash-high -> antigravity/gemini-3.6-flash-high | FMT: openai->antigravity | JSON | 16 MSG | 3 TOOL | THINK:high
[16:43:29] DONE 6987ms | IN 0 | OUT 0
```

Note the `JSON` marker — these are the non-streaming requests.

## Root cause

Antigravity and gemini-cli wrap their payload in a `{ response: { ... } }` envelope. Two places in the non-streaming path already know this:

- `utils/usageTracking.js` — `chunk.usageMetadata || chunk.response?.usageMetadata`, with the comment *"Antigravity wraps usageMetadata inside response"*
- `handlers/chatCore/nonStreamingHandler.js` — `const response = responseBody.response || responseBody` before reading `usageMetadata`

`extractUsageFromResponse()` in `handlers/chatCore/requestDetail.js` did not. It only tested `responseBody.usageMetadata`, so for an enveloped body it fell through all three format branches and returned `null` — which surfaces as `IN 0 | OUT 0` and as zeroed rows in the usage dashboard.

The streaming path was unaffected, which is why this looks provider-wide rather than endpoint-specific.

## Fix

Read the envelope in the Gemini branch, the same way the two call sites above already do. Top-level metadata keeps priority, and the OpenAI and Claude branches are untouched.

## Scope note

This does not touch the `[ProjectId] onboardUser` failures mentioned at the bottom of the issue — that is a separate concern and, as the reporter notes, those requests complete anyway.

## Tests

New: `tests/unit/antigravity-nonstream-usage-3260.test.js` — enveloped metadata is read, top-level still works and wins when both are present, the OpenAI and Claude shapes are unchanged, and a body with no usage still returns `null`.

The first test fails on `master` with `expected null to deeply equal { prompt_tokens: 1234, … }` — i.e. exactly the `IN 0 | OUT 0` the issue reports — and passes with this patch.

Full suite (`npx vitest run unit translator`) against `master` as the baseline: **1670 → 1675 passed, 91 failed on both sides** — no pass→fail, the 5 new passes are this PR's tests.

### DIFF
--- a/open-sse/handlers/chatCore/requestDetail.js
+++ b/open-sse/handlers/chatCore/requestDetail.js
@@ -44,13 +44,14 @@ export function extractUsageFromResponse(responseBody) {
     };
   }
 
-  // Gemini format
-  if (responseBody.usageMetadata) {
+  // Gemini format. Antigravity / gemini-cli wrap the payload in { response: {...} }.
+  const usageMetadata = responseBody.usageMetadata || responseBody.response?.usageMetadata;
+  if (usageMetadata) {
     return {
-      prompt_tokens: responseBody.usageMetadata.promptTokenCount || 0,
-      completion_tokens: responseBody.usageMetadata.candidatesTokenCount || 0,
-      cached_tokens: responseBody.usageMetadata.cachedContentTokenCount || 0,
-      reasoning_tokens: responseBody.usageMetadata.thoughtsTokenCount || 0
+      prompt_tokens: usageMetadata.promptTokenCount || 0,
+      completion_tokens: usageMetadata.candidatesTokenCount || 0,
+      cached_tokens: usageMetadata.cachedContentTokenCount || 0,
+      reasoning_tokens: usageMetadata.thoughtsTokenCount || 0
     };
   }
 
--- a/tests/unit/antigravity-nonstream-usage-3260.test.js
+++ b/tests/unit/antigravity-nonstream-usage-3260.test.js
@@ -1,0 +1,54 @@
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("@/lib/usageDb.js", () => ({
+  appendRequestLog: vi.fn(async () => {}),
+  saveRequestDetail: vi.fn(async () => {}),
+  saveRequestUsage: vi.fn(async () => {})
+}));
+
+const { extractUsageFromResponse } = await import("../../open-sse/handlers/chatCore/requestDetail.js");
+
+const USAGE_METADATA = {
+  promptTokenCount: 1234,
+  candidatesTokenCount: 56,
+  cachedContentTokenCount: 78,
+  thoughtsTokenCount: 90,
+};
+
+const EXPECTED = {
+  prompt_tokens: 1234,
+  completion_tokens: 56,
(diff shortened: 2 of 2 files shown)


decolua:

PR #3290, opened 2026-08-14
feat(tts): add Fish Audio as a text-to-speech provider

### LINKED ISSUE #2411: Feature: Add "Fish Audio" in Text To Speech
# Feature: Add "Fish Audio" in Text To Speech

## Description

It would be great to add **Fish Audio** as a supported Text-to-Speech provider in 9router.

Fish Audio provides high-quality AI voices with features such as:

- High-quality multilingual TTS
- Voice cloning (Reference Model ID)
- Adjustable speed, volume, and prosody
- Streaming audio generation
- Multiple output formats (MP3, WAV, PCM, Opus)
- Low latency generation
- Multi-speaker dialogue (s2.1-pro-free)

## API Documentation

https://docs.fish.audio/api-reference/endpoint/openapi-v1/text-to-speech

---

## Endpoint

```http
POST https://api.fish.audio/v1/tts
```

### Headers

```http
Authorization: Bearer <FISH_AUDIO_API_KEY>
Content-Type: application/json
model: s2.1-pro-free
```

---

## Minimum Request Example

```json
{
  "text": "Hello! Welcome to Fish Audio.",
  "reference_id": "voice-model-id",
  "format": "mp3"
}
```

---

### Environment Variable

```env
FISH_AUDIO_API_KEY=
```

### Provider Name

```
Fish Audio
```

### Supported Models

- `s2.1-pro-free` (recommended)
- `s1`

Default:

```
s2.1-pro-free
```

---

## Recommended Provider Options

| Option | Type | Description |
|---------|------|-------------|
| text | string | Text to synthesize |
| reference_id | string | Voice model ID |
| temperature | float | Controls voice expressiveness |
| top_p | float | Controls sampling diversity |
| speed | float | Speaking speed |
| volume | float | Volume adjustment |
| format | string | `mp3`, `wav`, `pcm`, `opus` |
| sample_rate | integer | Output sample rate |
| latency | string | `low`, `balanced`, `normal` |

---

## Response

Returns generated audio as a binary stream.

---

## Why this would be useful

Fish Audio has become one of the strongest modern Text-to-Speech providers because it offers:

- Excellent natural voice quality
- Vietnamese language support
- Voice cloning
- Streaming generation
- Low latency
- Competitive pricing
- Simple REST API
- Multiple audio formats

Adding Fish Audio would provide another high-quality TTS backend for 9router users and make it easier to build voice AI applications.

---

## References

- Fish Audio Text-to-Speech API  
  https://docs.fish.audio/api-reference/endpoint/openapi-v1/text-to-speech

- API Base URL

```
https://api.fish.audio
```

- Authentication

```
Authorization: Bearer <API_KEY>
```

ntdatt812:
Closes #2411

## What

Fish Audio as a TTS provider: a registry entry plus one entry in the config-driven `FORMAT_HANDLERS` table.

It is not a `SPECIAL_ADAPTERS` provider — it needs no voice fetcher and no credential quirks — but it does not fit any existing generic format either, because of two things the API does differently:

- **The model id travels in an HTTP `model` header**, not in the JSON body. Every other handler in that file puts the model in the body, the query string, or the path.
- **The voice is a `reference_id`** — a preset or cloned voice model — rather than a `voice` field.

Response is a binary audio stream, so `responseToBase64` handles it as-is.

## Contract

Taken from the [API reference](https://docs.fish.audio/api-reference/endpoint/openapi-v1/text-to-speech), not only from the issue:

| | |
|---|---|
| Endpoint | `POST https://api.fish.audio/v1/tts` |
| Auth | `Authorization: Bearer <key>` |
| Model header | `s1`, `s2-pro`, `s2.1-pro`, `s2.1-pro-free` |
| Body | `{ text, reference_id?, format }` |

The issue listed only `s2.1-pro-free` and `s1`; the docs list four, so all four are registered with `s2.1-pro-free` as the default the handler falls back to.

`format` is pinned to `mp3` in the request. The generic dispatcher calls `synthesizeViaConfig(provider, text, model, credentials)` without options, and `responseFormat` is applied downstream by `createTtsResponse` (it selects JSON-vs-binary framing, not the codec) — so plumbing wav/pcm/opus through would mean changing that signature for every provider. Out of scope here; the other config-driven handlers pin a format the same way.

## Baselines

No snapshot regeneration needed. A media-only provider does not enter `PROVIDERS` or `PROVIDER_MODELS`, so `verify-providers` and `verify-alias` are unchanged by this PR. (`verify-providers` is red on a clean `master` for an unrelated pre-existing `tokenrouter.thinkingFormat` drift, both before and after this branch.)

## Verification

Honest scope: **I have no Fish Audio key**, so no live call was made. The wire shape is asserted against the published contract with a mocked `fetch`.

New: `tests/unit/fish-audio-tts.test.js` (9 cases) — registration as a TTS-only apikey provider; visibility to the dispatcher through `AI_PROVIDERS` (which is where `synthesizeViaConfig` looks it up, not `PROVIDER_MEDIA`); the four model ids; registry id **and alias** uniqueness; then the request shape — model in the header and absent from the body, `reference_id` set when a voice is given and omitted when not, the default model fallback, base64 output, and that an upstream `402` surfaces its `message`.

Full suite (`npx vitest run unit translator`) against `master` as the baseline: **1670 → 1679 passed, 91 failed on both sides** — no pass→fail, the 9 new passes are this PR's tests.

### DIFF
--- a/open-sse/handlers/ttsProviders/genericFormats.js
+++ b/open-sse/handlers/ttsProviders/genericFormats.js
@@ -51,6 +51,25 @@ async function huggingface({ baseUrl, apiKey, text, modelId }) {
   return responseToBase64(res, "wav");
 }
 
+// Fish Audio: model travels in an HTTP header, the voice is a reference_id, returns binary
+async function fishAudio({ baseUrl, apiKey, text, modelId, voiceId }) {
+  const res = await fetch(baseUrl, {
+    method: "POST",
+    headers: {
+      "Content-Type": "application/json",
+      "Authorization": `Bearer ${apiKey}`,
+      "model": modelId || "s2.1-pro-free",
+    },
+    body: JSON.stringify({
+      text,
+      format: "mp3",
+      ...(voiceId ? { reference_id: voiceId } : {}),
+    }),
+  });
+  if (!res.ok) await throwUpstreamError(res);
+  return responseToBase64(res, "mp3");
+}
+
 // Inworld: Basic auth, JSON { audioContent }
 async function inworld({ baseUrl, apiKey, text, modelId, voiceId }) {
   const res = await fetch(baseUrl, {
@@ -166,4 +185,5 @@ export const FORMAT_HANDLERS = {
   tortoise,
   openai: openaiCompat,
   "minimax-tts": minimaxTts,
+  "fish-audio": fishAudio,
 };
--- a/open-sse/providers/registry/fish-audio.js
+++ b/open-sse/providers/registry/fish-audio.js
@@ -1,0 +1,31 @@
+// Fish Audio TTS — the model id travels in an HTTP `model` header rather than the
+// JSON body, and the voice is a reference_id (a cloned or preset voice model).
+export default {
+  id: "fish-audio",
+  alias: "fish",
+  display: {
+    name: "Fish Audio",
+    icon: "record_voice_over",
+    color: "#1E9BF0",
+    textIcon: "FA",
+    website: "https://fish.audio",
+    notice: {
+      apiKeyUrl: "https://fish.audio/app/api-keys/",
+    },
+  },
+  category: "apikey",
+  authType: "apikey",
+  serviceKinds: ["tts"],
+  ttsConfig: {
+    baseUrl: "https://api.fish.audio/v1/tts",
+    authType: "apikey",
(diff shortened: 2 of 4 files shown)


decolua:

PR #3299, opened 2026-08-14
fix(security): require proof that x-9r-real-ip came from the socket (GHSA-pjm4-8fpg-f9p6)

### LINKED ISSUE #3294: Authentication bypass via spoofable x-9r-real-ip header
### Summary
The router decides whether a request is "local" (privileged) from the client-supplied `x-9r-real-ip` header. When started via `next start` / `npm run start` — the path documented for Production mode and VPS/Cloud deployment — the sanitizing wrapper `custom-server.js` is not in the request path, so the header is attacker-controlled.

Adding `x-9r-real-ip: 127.0.0.1` makes any unauthenticated remote client be treated as local and bypass the API-key requirement on `/v1/*`, letting an attacker use the operator's configured (paid/free-tier) provider accounts. No password or API key required.

### Affected Product
| Field | Value |
|---|---|
| Ecosystem | npm |
| Package | `9router` |
| Affected versions | <= 0.5.8 |
| Patched versions | *(pending)* |

### Weaknesses
- CWE-290: Authentication Bypass by Spoofing
- CWE-307: Improper Restriction of Excessive Authentication Attempts
- CWE-346: Origin Validation Error

### Severity (CVSS v3.1)
- **Score:** High (8.6)
- **Vector:** `CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:L/A:H`

### Details

`src/dashboardGuard.js` → `isLocalRequest()` returns `true` from `x-9r-real-ip` / `Host`; `canAccessPublicLlmApi()` short-circuits to allow when local, so `/v1/*` is served with no API key.

The header is only made trustworthy by `custom-server.js` (which deletes and re-stamps it from the TCP socket). This wrapper runs only for the global CLI and Docker deployments. The following documented paths all skip it, leaving the header unstripped and the spoof viable:

- `package.json` → `"start": "next start"`
- README "Production mode"
- `gitbook/content/*/deployment/cloud.md` (`npm run start` + `pm2 start npm -- start`)

**Secondary issue:** `src/lib/auth/loginLimiter.js` → `getClientIp()` also trusts the header → login rate-limit bypass.

### Proof of Concept

Deploy as documented for production/VPS (without `custom-server.js`):

```bash
npm run build
PORT=20140 HOSTNAME=0.0.0.0 npm run start
```

From a remote host (LAN/VPN/tunnel/VPS), using the server's reachable address:

```bash
# 1) Baseline — correctly rejected
curl -i http://172.18.192.1:20140/api/v1/models
# HTTP/1.1 401 Unauthorized — {"error":"API key required for remote API access"}

# 2) Same request + one spoofed header — BYPASS
curl -i -H "x-9r-real-ip: 127.0.0.1" http://172.18.192.1:20140/api/v1/models
# HTTP/1.1 200 OK — {"object":"list","data":[{"id":"alicode/qwen3.5-plus",...}]}
```

One header flips `401 → 200` and returns the full model catalog. The same header makes `POST /api/v1/chat/completions` reach the proxy, so with any provider connected (e.g. the docs' free Kiro/OpenCode) the attacker gets live completions billed to the operator.

Image

### Impact

An attacker who can reach an instance deployed via `npm run start` / VPS guide and exposed beyond localhost gains:

1. **Free use of the operator's provider accounts** via `/v1/*` → financial cost, quota exhaustion, free-tier bans
2. **Disclosure of the configured model/provider catalog**
3. **Login brute-force lockout bypass** via the same trusted header in `loginLimiter.js`

---

**Reporter:** @HK4zCzi (Hồ Việt Khánh)

ntdatt812:
Closes #3294 — GHSA-pjm4-8fpg-f9p6 (High, 8.6).

## The hole

`isLocalRequest()` decides "this caller is local, skip the API key" from headers the client controls:

```js
const realIp = request.headers.get("x-9r-real-ip");
if (realIp) { if (!isLoopbackHostname(realIp)) return false; }
else if (!isLoopbackHostname(request.headers.get("host"))) return false;
```

That is only safe because `custom-server.js` deletes and re-stamps `x-9r-real-ip` from the TCP socket. It is **not** in the request path for `"start": "next start"` — the command README "Production mode" and `gitbook/*/deployment/cloud.md` both tell operators to run.

Two things the report does not mention:

1. The `else` branch is spoofable in exactly the same way, so **`Host: localhost` alone bypasses the gate even without `x-9r-real-ip`**. Fixing only the reported header would have left an equivalent bypass.
2. With `requireLogin=false` — the common single-user setup — the spoof also satisfies `canAccessLocalOnlyRoute`, so `LOCAL_ONLY_PATHS` becomes remotely reachable: `/api/mcp/*` (spawns processes), `/api/tunnel/enable`, `/api/auth/reset-password`. Those are the routes explicitly carved out as too dangerous for remote callers.

`start:bun` (`bun ./.next/standalone/server.js`) is a third unsanitized path. Docker (`CMD ["node", "custom-server.js"]`) and the CLI launcher were never affected.

## The fix

Make the header contract enforceable instead of assumed.

- `custom-server.js` generates a 24-byte per-process secret at boot and stamps it as `x-9r-peer-token` on every request it sanitizes, deleting any client-supplied copy first.
- `hasTrustedPeerHeaders()` (new `src/lib/auth/trustedPeer.js`) is the single place that answers "did the wrapper actually handle this request".
- The guard trusts `x-9r-real-ip` only when the secret matches. Otherwise it falls back to `Host` **only when `NODE_ENV === "development"`** — `next dev` forks its server (`next-dev.js` → `fork(startServerPath, { execArgv })`), so no wrapper can reach it and there is no peer address to use. Everything else fails closed.

The secret is random rather than a boolean flag: `process.env` is inherited by forked children, and a `.env` entry or a leaked flag must not be able to re-enable the trust. `custom-server.js` stamps from the constant it generated, not from `process.env`, so if anything ever clobbers the variable the guard stops matching and fails closed instead of trusting a value someone else chose.

Same gate on `loginLimiter.getClientIp()`, so a spoofed header can no longer rotate the login lockout bucket (the report's secondary finding).

### isLoopbackHostname had to be fixed first

`h.split(":")[0]` only works for IPv4. It reduces `::ffff:127.0.0.1`, `::1` and `[::1]` all to the empty string, so none of them matched `LOOPBACK_HOSTS`. That was latent while `Host` could cover for it; once `x-9r-real-ip` is the only production signal it is fatal, because a dual-stack listener reports a loopback peer as `::ffff:127.0.0.1` — which is exactly what `next start` does. I hit this as a live 401 on my own loopback request before shipping. It now strips brackets, strips the port only when a single colon makes that unambiguous, and unwraps the IPv4-mapped prefix; `::ffff:10.204.111.34` and `2001:db8::1` stay rejected.

### Two build-time details this depends on, both verified in next@16

- `process.env.NODE_ENV` is a DefinePlugin constant — `define-env.js:77` emits `dev ? 'development' : 'production'`. In a production build the development branch becomes `"production" === "development"` and is eliminated outright, so **no `.env` value and no runtime environment can switch the `Host` fallback back on in a production build**. My local `.env` has `NODE_ENV=development` and the built server still refused the `Host` spoof.
- Only `NEXT_PUBLIC_*` is statically inlined (`lib/static-env.js:42`), so `NINEROUTER_PEER_TOKEN` stays a genuine runtime lookup. Confirmed live: the token the wrapper stamps and the value the proxy reads are the same string.

### Keeping the secret out of logs

`chat.js:46` snapshots every client header into `clientRawRequest`, which lands in the request detail the dashboard renders and cloud sync uploads. The header is therefore named `x-9r-peer-token` so `requestDetailsRepo.sanitizeHeaders` strips it under the existing `token` rule — the same protection `x-9r-cli-token` already relies on. A test pins that, so a future refactor of the sanitizer cannot silently expose it. It is also weaker than `x-9r-cli-token` by design: machine-independent and regenerated on every restart.

## Keeping the documented deployments working

Gating alone would make `npm run start` lose local access, so the sanitizer now covers those paths too:

- `"start": "node custom-server.js --port 20127"` — `next start` builds its HTTP server in-process (`next-start.js` → `startServer()`), so the wrapper still sees every request. `custom-server.js` runs the sibling standalone `server.js` when present (Docker/CLI, unchanged) and otherwise dispatches to `next start`.
- `postbuild` copies `custom-server.js` into `.next/standalone/`, and `start:bun` points at it. That also makes `node .next/standalone/custom-server.js` available for anyone following Next's own advice — `next start` has always warned that it does not work with `output: standalone`, which is pre-existing on master and not something this PR changes.
- `build-cli.js` only warned when `custom-server.js` was missing. With this change that would ship a CLI where nothing can be proven local — it would demand an API key for its own dashboard and `/v1` — so it now fails the build.

`npm run start`, `pm2 start npm -- start`, the Docker image and the CLI all keep their current commands — **no documentation changes needed**. `HOSTNAME` behaviour is unchanged: only the standalone server template reads it (`build/utils.js:1088`), and `next start` binds `0.0.0.0` either way.

### Live, against a real production build

`npm run build`, then `npm run start -- --port 20142`, driven from this machine's LAN address so the TCP peer is genuinely not loopback — the reporter's setup:

```
loopback, no key                                 -> 200
LAN, no key                          [baseline]  -> 401
LAN + x-9r-real-ip: 127.0.0.1        [CVE]       -> 401
LAN + Host: localhost                [extra]     -> 401
LAN + both spoofs                                -> 401
LAN + forged x-9r-peer-token                     -> 401
LAN + spoof -> /api/mcp/filesystem/sse           -> 403
```

The first line matters as much as the rest: a genuine local caller still gets in with no API key, so nothing regresses for existing users.

### Unit level

The new tests reproduce the bypass. Reverting only `src/dashboardGuard.js` and `src/lib/auth/loginLimiter.js` to master and re-running `unit/local-request-peer-trust-3294.test.js`:

```
PASS (3) FAIL (5)
1. rejects a spoofed loopback peer IP that carries no trust proof
   AssertionError: expected undefined to be 401
2. rejects a spoofed loopback peer IP carrying a wrong trust token
3. rejects a spoofed loopback Host in production
4. rejects a spoofed loopback peer IP when the wrapper never booted
5. blocks spoofed local-only routes that would otherwise spawn processes
   AssertionError: expected undefined to be 403
```

`undefined` is `NextResponse.next()` — master lets the spoofed request through. All pass with the patch.

`unit/custom-server-peer-headers.test.js` boots a real `http.createServer` through the wrapper and asserts the sanitizing directly: a client-supplied `x-9r-real-ip: 203.0.113.55` comes out as the loopback socket address, a forged `x-9r-peer-token` is replaced, `x-9r-via-proxy` is dropped, and the loopback-reverse-proxy path still adopts the forwarded IP.

`isLocalRequest` is also read by `api/auth/login` and `api/provider-nodes/validate`; both fail in the safe direction — forcing a password change, applying the SSRF guard — rather than locking anyone out.

### Suite

`npx vitest run unit translator`: **1670 to 1698 passed, 91 failed before and after, and the failing sets are byte-identical** (`comm -13` on the sorted fail lists is empty). The 91 are the usual pre-existing reds; I confirmed `unit/security-audit.test.js`'s 13 fail the same way on a clean checkout. `unit/cli-build-artifacts.test.js` and `unit/standalone-assets.test.js` cover the packaging changes — the latter never created a `custom-server.js`, so the new copy step was silently uncovered until this PR added a case for it. `npx eslint` is clean on every changed file.

Six existing `dashboard-guard` cases asserted the pre-fix contract, that `Host: localhost` alone implies local. They now go through a `localRequest()` helper that stamps the peer headers the way `custom-server.js` does — same intent, stated explicitly. The old behaviour is still covered, as a development-only case.

One note for anyone reproducing on Windows: `next build` dies here with `EPERM scandir 'C:\Users\...\Application Data'` unless `HOME`, `USERPROFILE`, `APPDATA` and `LOCALAPPDATA` are redirected the way `cli/scripts/build-cli.js` already does. That is pre-existing — it fails identically on a clean `origin/master` checkout.

### DIFF
--- a/cli/scripts/build-cli.js
+++ b/cli/scripts/build-cli.js
@@ -216,7 +216,9 @@ function buildCliPackage() {
     fs.copyFileSync(customServerSrc, path.join(cliAppDir, "custom-server.js"));
     console.log("✅ Copied custom-server.js\n");
   } else {
-    console.warn("⚠️  custom-server.js not found — server will run without real-IP injection\n");
+    console.error("❌ custom-server.js not found — without it no request can be proven local,");
+    console.error("   so the packaged CLI would demand an API key for its own dashboard and /v1.");
+    process.exit(1);
   }
 
   // Step 3b: Ensure sql.js (pure JS fallback) bundled in app/cli/app/node_modules.
--- a/custom-server.js
+++ b/custom-server.js
@@ -1,8 +1,16 @@
 const http = require("http");
 const path = require("path");
+const fs = require("fs");
+const crypto = require("crypto");
 const { pathToFileURL } = require("url");
 
 const origCreate = http.createServer.bind(http);
+
+// Per-process secret proving x-9r-real-ip was stamped below rather than sent by the client.
+// A bare `next start` / `next dev` never loads this file, so it cannot produce a matching
+// header even though the env var is inherited by child processes.
+const PEER_TRUST_TOKEN = crypto.randomBytes(24).toString("hex");
+process.env.NINEROUTER_PEER_TRUST = PEER_TRUST_TOKEN;
 
 let backgroundRefreshStarted = false;
 
@@ -57,7 +65,9 @@ http.createServer = (...args) => {
     delete req.headers["x-9r-real-ip"];
     delete req.headers["x-forwarded-for"];
     delete req.headers["x-9r-via-proxy"];
+    delete req.headers["x-9r-peer-trust"];
     req.headers["x-9r-real-ip"] = ip;
+    req.headers["x-9r-peer-trust"] = PEER_TRUST_TOKEN;
     if (viaProxy) req.headers["x-9r-via-proxy"] = "1";
     return handler(req, res);
   };
@@ -114,4 +124,15 @@ http.createServer = (...args) => {
   return server;
 };
 
-if (require.main === module) require("./server.js");
+if (require.main === module) {
+  const standalone = path.join(__dirname, "server.js");
+  if (fs.existsSync(standalone)) {
+    require(standalone);
+  } else {
+    // Repo checkout has no standalone build next to us. `next start` builds its HTTP
+    // server in-process, so the wrapper above still sanitizes every request.
+    const nextBin = require.resolve("next/dist/bin/next");
+    process.argv = [process.argv[0], nextBin, "start", ...process.argv.slice(2)];
+    require(nextBin);
+  }
+}
--- a/custom-server.js
+++ b/custom-server.js
@@ -8,9 +8,10 @@ const origCreate = http.createServer.bind(http);
 
 // Per-process secret proving x-9r-real-ip was stamped below rather than sent by the client.
 // A bare `next start` / `next dev` never loads this file, so it cannot produce a matching
-// header even though the env var is inherited by child processes.
-const PEER_TRUST_TOKEN = crypto.randomBytes(24).toString("hex");
-process.env.NINEROUTER_PEER_TRUST = PEER_TRUST_TOKEN;
+// header even though the env var is inherited by child processes. Named like x-9r-cli-token
+// so the request-detail header sanitizer redacts it too.
+const PEER_TOKEN = crypto.randomBytes(24).toString("hex");
+process.env.NINEROUTER_PEER_TOKEN = PEER_TOKEN;
 
 let backgroundRefreshStarted = false;
 
@@ -65,9 +66,9 @@ http.createServer = (...args) => {
     delete req.headers["x-9r-real-ip"];
     delete req.headers["x-forwarded-for"];
     delete req.headers["x-9r-via-proxy"];
-    delete req.headers["x-9r-peer-trust"];
+    delete req.headers["x-9r-peer-token"];
     req.headers["x-9r-real-ip"] = ip;
-    req.headers["x-9r-peer-trust"] = PEER_TRUST_TOKEN;
+    req.headers["x-9r-peer-token"] = PEER_TOKEN;
     if (viaProxy) req.headers["x-9r-via-proxy"] = "1";
     return handler(req, res);
   };
--- a/package.json
+++ b/package.json
@@ -9,10 +9,10 @@
     "build": "next build --webpack",
     "postbuild": "node scripts/copy-standalone-assets.mjs",
     "postbuild:bun": "node scripts/copy-standalone-assets.mjs",
-    "start": "next start --port 20127",
+    "start": "node custom-server.js --port 20127",
     "dev:bun": "bun --bun next dev --webpack --port 20127",
     "build:bun": "bun --bun next build --webpack",
-    "start:bun": "bun ./.next/standalone/server.js",
+    "start:bun": "bun ./.next/standalone/custom-server.js",
     "cli:pack": "npm --prefix cli run pack:cli",
     "cli:publish": "npm --prefix cli run publish:cli"
   },
--- a/scripts/copy-standalone-assets.mjs
+++ b/scripts/copy-standalone-assets.mjs
@@ -29,6 +29,14 @@ export function copyStandaloneAssets({ projectRoot = process.cwd(), distDir = pr
     cpSync(publicSource, publicDestination, { recursive: true, force: true });
     console.log(`[standalone-assets] Copied public assets to ${publicDestination}`);
   }
+
+  // Without it beside server.js the standalone build serves requests unsanitized.
+  const serverWrapperSource = resolve(projectRoot, "custom-server.js");
+  const serverWrapperDestination = resolve(standaloneDir, "custom-server.js");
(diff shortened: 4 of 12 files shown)


decolua:

ntdatt812:
This PR was closed, but the change landed in `main` as [`9225921`](https://github.com/decolua/9router/commit/9225921) — thanks for picking it up.

One small request, entirely at your discretion: [GHSA-5mj8-gf6m-fhw8](https://github.com/decolua/9router/security/advisories/GHSA-5mj8-gf6m-fhw8) currently has no credits recorded. Would you consider adding me as `remediation_developer` on it?

To be clear about what I am and am not asking for: the vulnerability was reported by @HK4zCzi in #3294, and the reporter credit belongs to them. I wrote the fix that requires the address to be proven from the socket rather than read from the client-supplied header. If you'd rather leave credits empty, or if the remediation was reworked before it landed, no problem at all — please just close this out.

PR #3309, opened 2026-08-14
feat(providers): add the Fish Audio brand icon

ntdatt812:
Follow-up to #3290 (Fish Audio TTS, landed as `8af5e752`).

The provider tile renders the `FA` text placeholder instead of a logo, because I shipped the registry entry without an icon asset.

`getProviderIconSrc` in `src/shared/utils/providerIcon.js` resolves `/providers/{id}.png` purely by convention and only falls back to `display.textIcon` once that 404s — so nothing in the registry needed changing, the file was simply missing.

Added `public/providers/fish-audio.png`: the official apple-touch-icon from fish.audio, resized 180 to 128 with lanczos3 to match the other 136 icons under `public/providers/`. 4.8KB, RGBA, well inside the existing 0.9KB to 29.5KB range.

The source already carries its own rounded white backdrop with transparent corners, so it reads on both the light and dark dashboard themes without any recolouring — the same treatment `elevenlabs.png` uses.

`display.textIcon: "FA"` stays in the registry as the 404 fallback, matching every other provider (`cartesia` keeps `"CA"` alongside `cartesia.png`).

`unit/provider-display-split.test.js` and `unit/fish-audio-tts.test.js` pass.


ntdatt812:
Folding this into #3328, which adds this same `fish-audio.png` plus the two Alibaba Cloud entries that show the same text placeholder. Same root cause, no overlapping files, and it comes with a test that walks the registry and requires the icon-less set to be exactly the three generic self-host slots — so one review instead of two. Closing here; nothing is lost.

PR #3321, opened 2026-08-14
fix(opencode): stop zen free-tier 429 — official client fingerprint + manual egress switch

### LINKED ISSUE #3270: opencode deepseek v4 flash 429

qianze0628:
## What & why

OpenCode Free (`oc/…-free` models) returns `429 FreeUsageLimitError` ("Rate limit exceeded") through 9router in heavy agent use (Claude Code / DSH), even after changing the machine's IP, while the same model works in the official OpenCode TUI.

Root cause — two independent problems, both verified against the Zen gateway source (`packages/console/app/src/routes/zen/util/ipRateLimiter.ts`) and by live probing:

1. **Anonymous free quota is keyed per REAL egress IP.** The gateway buckets per-IP daily/lifetime request counts (reset at UTC midnight) and reads the `x-real-ip` header for it. In front of opencode.ai the CDN stamps `x-real-ip` from the real TCP peer and **strips client-supplied values** (verified: a request with a brand-new spoofed IP still 429s). 9router sent no usable fingerprint, so every anonymous request landed in one shared `"unknown"` bucket — a single daily budget for every 9router install worldwide. Minutes of agent traffic drain it for everyone: hence "always 429" no matter your own IP, and "works again after a while / after a restart" (UTC reset / egress change).
2. **Unidentified client fingerprint.** The gateway only unlocks anonymous capacity for the official client UA; 9router sent a bare `opencode`. There was also a session-bleed race under concurrency (`_currentSessionId` shared instance field).

## Changes

- `open-sse/executors/opencode.js`
  - `OPENCODE_UA` → `opencode/latest/1.18.18/cli` (official format; keep in sync with opencode releases)
  - per-request session on `credentials._ocSession` (fixes the concurrency race)
  - best-effort `x-real-ip` forwarding (the CDN normally re-stamps it; helps behind pass-through proxies)
  - `execute` applies the manually chosen egress mode (NO manual-switch automation; errors propagate untouched)
- `open-sse/utils/ocEgress.js` (new) — egress mode (`proxy` ⇄ `direct`, i.e. outbound-proxy node IP ⇄ machine's own public IP), persisted in `<APPDATA>/9router/oc-egress.json`, applied via `NO_PROXY` so only opencode.ai changes path. Manual switching only: `node oc-egress.mjs status|flip|set proxy|direct` (+ `nodes`/`switch <node>` to control the Clash node) — there is deliberately NO automatic rotation
- `tests/unit/opencode-executor.test.js` — regression tests (UA, per-request session, x-real-ip)
- `tests/unit/oc-egress.test.js` — 8 cases (read / flip / cooldown / NO_PROXY apply)

## Result

Each machine gets its own per-IP quota bucket again ("one IP for yourself"). When that IP's budget is exhausted the user switches egress manually (Clash node in the UI, or oc-egress.mjs) — the same conversation continues from the new IP with a fresh daily budget, no restart needed. Verified live: the same session keeps answering 200 while switching across 4 different node egresses.

## Test

(code)
cd tests && npx vitest run unit/opencode-executor.test.js unit/oc-egress.test.js
# 2 files, 15 tests, all pass
(/code)

Closes #3270. Ref #3285.

### DIFF
--- a/open-sse/executors/opencode.js
+++ b/open-sse/executors/opencode.js
@@ -4,7 +4,12 @@ import { PROVIDERS } from "../config/providers.js";
 import { injectReasoningContent } from "../utils/reasoningContentInjector.js";
 import { resolveSessionId } from "../utils/sessionManager.js";
 
-const OPENCODE_UA = "opencode";
+// Official opencode CLI sends "opencode/<version>" (e.g. opencode/1.18.18).
+// OpenCode Zen's free-tier ("-free") models gate anonymous capacity on this
+// User-Agent — a bare "opencode" or any non-opencode UA is still classified
+// as unidentified and gets FreeUsageLimitError/429 immediately. Keep the
+// version in sync with opencode releases when it bumps.
+const OPENCODE_UA = "opencode/latest/1.18.18/cli";
 const MESSAGES_MODELS = new Set();
 
 function generateRequestId() {
@@ -33,11 +38,14 @@ function resolveOpencodeSession(body, credentials) {
 export class OpenCodeExecutor extends BaseExecutor {
   constructor() {
     super("opencode", PROVIDERS.opencode);
-    this._currentSessionId = null;
   }
 
   transformRequest(model, body, stream, credentials) {
-    this._currentSessionId = resolveOpencodeSession(body, credentials);
+    // Stash the resolved session on the per-request credentials object instead
+    // of an instance field: OpenCodeExecutor is a module-level singleton and
+    // concurrent requests used to overwrite _currentSessionId between
+    // transformRequest and buildHeaders, bleeding sessions across requests.
+    if (credentials) credentials._ocSession = resolveOpencodeSession(body, credentials);
     return injectReasoningContent({ provider: this.provider, model, body });
   }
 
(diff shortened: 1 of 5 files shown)


FerryAr (commented):
After applying and testing this PR locally, OpenCode free models still hit `FreeUsageLimitError` (HTTP 429).

When 9router runs on localhost (`localhost:20128`), `custom-server.js` stamps `x-9r-real-ip` as `"127.0.0.1"` (or `"::1"`). Forwarding `"x-real-ip": "127.0.0.1"` upstream causes Zen's rate limiter to place all local users into the same shared `127.0.0.1` bucket, which exhausts immediately.

Filtering out loopback/private IPs so `x-real-ip` is only forwarded when it's a real public IP resolves this:

```javascript
function isPublicIp(ip) {
  if (!ip || typeof ip !== "string") return false;
  const clean = ip.replace(/^::ffff:/, "").trim();
  if (clean === "127.0.0.1" || clean === "::1" || clean === "localhost") return false;
  if (clean.startsWith("10.") || clean.startsWith("192.168.")) return false;
  if (/^172\.(1[6-9]|2[0-9]|3[0-1])\./.test(clean)) return false;
  if (clean.startsWith("fc00:") || clean.startsWith("fe80:")) return false;
  return /^[0-9a-fA-F:.]+$/.test(clean);
}
```

And in `buildHeaders`:
```javascript
const rawIp = lower["x-9r-real-ip"] || lower["x-real-ip"] || "";
const clientIp = isPublicIp(rawIp) ? rawIp : "";
```

qianze0628:
Thanks @FerryAr for the thorough local test — you are right, and the loopback guard is now in (commit 0e7a4ae).

In `buildHeaders` we only forward `x-real-ip` when the peer IP is a real public address; loopback/private/LAN peers (127.0.0.1, ::1, 10.x, 192.168.x, 172.16-31.x, fc00::/fe80::) fall back to the machine's own public IPv4, discovered once via direct https (4.icanhazip.com → ip.sb → ifconfig.me) and cached — so a local-only 9router still gets one dedicated bucket instead of the shared `127.0.0.1` one. Unknown peers send no header at all.

Added regression tests for the loopback/private filtering (17 unit tests pass).

rusak47:
One more thing I think is worth considering regarding OpenCode provider configuration — currently, there is no way to configure a custom API key, only a proxy. However, OpenCode does provide its own API keys.

PR #3328, opened 2026-08-15
test(providers): pin that every provider ships a brand icon

ntdatt812:
Supersedes #3309, which carried the Fish Audio icon alone. Same root cause, same kind of fix, no overlapping files — one review instead of two.

Three providers render a text placeholder instead of a logo on the providers grid: `fish-audio` (`FA`), `alims-intl` (`ALi`, Alibaba Studio) and `alitp-intl` (`ATP`, Alibaba Token Plan).

`getProviderIconSrc` in `src/shared/utils/providerIcon.js` resolves `/providers/{id}.png` purely by convention and only falls back to `display.textIcon` once that 404s — so nothing in the registry or in `ICON_ALIASES` needed changing, the three files were simply missing.

## Where the marks come from

**Fish Audio** — the official apple-touch-icon from fish.audio, resized 180 to 128 with lanczos3. It already carries its own rounded white backdrop with transparent corners, so it reads on both the light and dark dashboard themes without recolouring, the same treatment `elevenlabs.png` uses. Follow-up to #3290 (Fish Audio TTS, landed as `8af5e752`), which shipped the registry entry without the asset.

**Alibaba** — both entries are Alibaba Cloud products and both already carry Alibaba Cloud's `#FF6A00`. The correct mark is the one `alicode.png` already ships: I checked it against the favicon `https://www.alibabacloud.com` links (`img.alicdn.com/tfs/TB1ugg7M9zqK1RjSZPxXXc4tVXa-32-32.png`) and it is the same bracket-cloud glyph in the same orange. So both files are byte-identical copies of `alicode.png`, which is exactly how the repo already handles sibling entries of one brand:

```
81fdaf97…  alicode.png
81fdaf97…  alicode-intl.png     <- already identical before this PR
42f40b6c…  ollama.png
42f40b6c…  ollama-local.png     <- same pattern
```

`alicode` itself points at `bailian.console.aliyun.com`, i.e. Model Studio CN, and uses the plain Alibaba Cloud mark — so `alims-intl` (Model Studio Intl) getting that same mark follows the entry that is already there rather than inventing a convention.

All three are 128x128 RGBA PNGs (3.5KB / 3.5KB / 4.8KB), inside the 0.9KB–29.5KB range of the other 136 icons under `public/providers/`. Every `display.textIcon` stays in place as the 404 fallback, the way `cartesia` keeps `"CA"` alongside `cartesia.png`.

## The three that keep their text badge

With these three added, the whole registry is covered except `selfhosted-embedding`, `selfhosted-stt` and `selfhosted-tts` — and those should stay as they are. They are generic slots, not brands: their own header comments list llama.cpp / vLLM / Infinity / TEI, whisper.cpp / faster-whisper / Speaches, Kokoro-FastAPI / openedai-speech / vLLM-served TTS. `display.website` is the first example in an open list, not a vendor. And llama.cpp and whisper.cpp are both ggml-org, so that avatar would render Embedding and STT identically — strictly worse than the distinct `SE` / `ST` badges they have now.

That is asserted rather than described: the test walks the registry and requires the set of icon-less providers to equal exactly those three, so a provider added later without an icon fails the build, and dropping one of these exceptions is a deliberate edit.

## Verification

`unit/provider-brand-icons.test.js`:

- every provider in `providers/registry/index.js` resolves through `getProviderIconSrc` to a file that exists, except exactly the three documented slots
- the three added ids resolve to the convention path and the files exist
- all three are real 128x128 PNGs (magic bytes + IHDR)
- the Alibaba files are byte-identical to `alicode.png`, and `fish-audio.png` is not — a wrong copy/paste fails
- the `FA` / `ALi` / `ATP` badges and `#FF6A00` stay in the registry as the fallback

I checked the coverage assertion actually bites by hiding `alitp-intl.png` and re-running: 4 of the 5 fail, and the coverage test names the offender.

Full suite (`npx vitest run unit translator`) against this base: **1778 to 1783 passed, 93 failed before and after, failing sets byte-identical** (`comm` empty in both directions). `npx eslint` clean.

No code, registry, or i18n changes — three binary assets and one test.


ntdatt812:
Rescoped: `6efb9790` landed all three icons this PR carried, plus three more for the self-hosted slots, so the assets here were redundant and conflicting on rebase. I have dropped them. What is left is the part that commit did not bring — a test pinning the result.

That gap is worth closing because it fails quietly. A provider added with no icon falls back to a text badge, which renders as ordinary-looking output rather than as an error, so it reaches the dashboard with nobody noticing.

Three assertions, driven from the registry rather than from the folder listing, so a new provider is covered the moment it is registered:

- every provider resolves to an icon file that exists;
- every resolved icon is a 128×128 PNG, the shape the folder already standardised on (139 of 142 files);
- the three Alibaba Cloud endpoints share one mark rather than three separately drawn lookalikes that can drift.

**Two things the check deliberately does not claim**, because I checked and neither is true today: `ollama-search`, `perplexity-agent` and `vercel-ai-gateway` resolve to their parent brand's file rather than to `<id>.png`, and twelve providers carry no `display.textIcon`. My first draft asserted both and was wrong; they look deliberate, so the test does not encode them as rules.

**One finding, reported rather than fixed here.** `nebius.png`, `reka.png` and `siliconflow.png` hold JPEG data under a `.png` name (`ff d8 ff e0 … JFIF`; two still carry Photoshop/Exif blocks). Browsers sniff the body and render them, so nothing looks broken. They are named in an explicit exclusion list rather than covered by a loosened assertion, so the exclusion is a statement about those three files and not a hole a new icon can slip through. I did re-encode them locally with `sharp` — it works and keeps 128×128 — but it costs about +31 KB across the three, and bundling an asset re-encode into a test PR seemed like the wrong trade. Happy to send it separately if you want it.

Verified with `vitest run tests/unit/provider-brand-icons.test.js --config tests/vitest.config.js`: 3 passed. Checked by breaking it — deleting one icon fails all three assertions, and resizing one to 64×64 fails only the shape one.

ntdatt812:
Closing this one myself: it is a regression guard, and I re-ran its premise against current master — every provider in the registry resolves to an icon that exists (`ollama-search` reaches `ollama.png` through the alias table). So it pins an invariant that is not currently broken, which is not worth a slot in the review queue. Happy to bring it back if a provider ever ships without an icon.

PR #3333, opened 2026-08-15
fix(tools): DeepSeek same-name tool dedup + endpoint matrix tests

KiMelody:
## What

Fix: exact same-name tool dedup for DeepSeek models (`toolDeduper.js` + `chatCore.js`), plus the endpoint matrix tests that pin the behavior this PR and #3332 debate.

## Why

DeepSeek upstream rejects duplicate tool names — `400 Tool names must be unique.` — on every endpoint and format, verified live 2026-08-15:

- api.deepseek.com: flash/pro, `/chat/completions` and `/anthropic/v1/messages` → all 400
- opencode.go Console Go: deepseek-v4-flash → 400; minimax-m3 → 200
- LiteLLM gateway (jy/基元律动): flash-0731/pro → 400; glm-5.2/minimax-m2.7/kimi-k2.6 → 200
- api.z.ai GLM official: glm-5.2 → 200

The restriction is DeepSeek's own (identical across three independent gateways; GLM/MiniMax/Kimi accept duplicates everywhere). This also rules out the #3332 "fragile shim" theory — DeepSeek's own `/anthropic/v1/messages` rejects it too.

Dedup is gated by model id (`deepseek-*`, thinking suffix stripped), first definition wins; non-DeepSeek models untouched. E2E: the dup-tool Claude-shaped payload to opencode.go/deepseek-v4-flash went from 400 → 200 through the real gateway.

## Tests

- `tests/unit/tool-deduper.test.js` — 11 cases (MCP rules regression, deepseek gating incl. `(max)` and function-shape, non-deepseek untouched)
- `tests/unit/opencode-go-transport-routing.test.js` — 16-cell offline routing matrix
- `tests/translator/real/opencode-go-deepseek.real.test.js` — live matrix, `RUN_REAL=1`, 8/8

Known red (3 cells): the `(max)` suffix guard-bypass in `providerModels.js` — belongs to #3332's suffix-strip fix; the cells are its regression repro.

## Runs

```bash
npx vitest run tests/unit/tool-deduper.test.js tests/unit/opencode-go-transport-routing.test.js   # 27 pass, 3 known-red
DATA_DIR=<snapshot> RUN_REAL=1 npx vitest run --config tests/vitest.config.js tests/translator/real/opencode-go-deepseek.real.test.js  # 8/8 live
```

Relates to #3278, #3332.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -195,9 +195,10 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     stripContinuityFields(translatedBody);
   }
 
-  // Dedupe duplicate built-in tools when equivalent MCP tools are present (Claude clients only).
-  if (clientTool === "claude" && Array.isArray(translatedBody.tools)) {
-    const { tools: deduped, stripped } = dedupeTools(translatedBody.tools);
+  // Tool normalization: MCP-equivalent built-in dedup (Claude clients) + same-name
+  // dedup for DeepSeek models (upstream rejects duplicate tool names on all endpoints).
+  if (Array.isArray(translatedBody.tools)) {
+    const { tools: deduped, stripped } = dedupeTools(translatedBody.tools, { clientTool, model });
     if (stripped.length > 0) {
       translatedBody.tools = deduped;
       log?.debug?.("TOOLDEDUP", `stripped ${stripped.length}: ${stripped.slice(0, 3).join(", ")}${stripped.length > 3 ? "..." : ""}`);
--- a/open-sse/utils/toolDeduper.js
+++ b/open-sse/utils/toolDeduper.js
@@ -1,6 +1,11 @@
 /**
- * Strip built-in/duplicate tools when equivalent MCP tools are present.
- * Goal: reduce tool definitions token bloat for Claude clients.
+ * Tool normalization before dispatch:
+ * - MCP-equivalent built-in tool dedup (Claude clients only, reduces token bloat).
+ * - Exact same-name tool dedup for DeepSeek models — the DeepSeek upstream rejects
+ *   duplicate tool names with 400 "Tool names must be unique" on every endpoint
+ *   (verified live 2026-08-15 against api.deepseek.com, opencode.go and a LiteLLM
+ *   gateway; GLM/MiniMax/Kimi upstreams accept duplicates). First definition wins,
+ *   tool_choice and message-history references are by name/id so nothing breaks.
  */
 
 const DEDUP_RULES = [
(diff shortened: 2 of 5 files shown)


KunN-21:
Thanks for the evidence package — to be precise about why I kept DeepSeek on `["openai"]`, and what your matrix is missing.

## What I reproduced, not just claimed

Against the live upstream (same 9router, same key), `/zen/go/v1/messages`:

| `/messages` request | Result |
|---|---|
| `deepseek-v4-flash`, single user turn, no tools | 200 |
| `deepseek-v4-flash(max)` (same body) | 200 |
| `deepseek-v4-flash` + two tools with the **same name** (`Bash`/`Bash`) | **400** `{"model":"deepseek-v4-flash"}` |
| `deepseek-v4-flash(max)` + same duplicate tools | **400** |

So `/messages` works for DeepSeek **only on minimal bodies**. A real Claude Code payload
carries tools, tool-use history, system blocks, large content — and routinely **duplicate
tool names** (the same built-in `Bash`, `Read`, `Edit` can appear more than once across
MCP servers). That is the 400 your matrix never exercises: your `BODIES.claude` sends one
user turn with `content: "Reply with the single word: hi"`, no `tools` at all.

## About "docs are stale"

Today's fetch of https://opencode.ai/docs/go/ still lists:

```
DeepSeek V4 Pro   | deepseek-v4-pro   | https://opencode.ai/zen/go/v1/chat/completions
DeepSeek V4 Flash | deepseek-v4-flash | https://opencode.ai/zen/go/v1/chat/completions
```

`/v1/messages` is documented only for MiniMax and Qwen. A "support confirmed" chat reply
plus one minimal 200 is not an endpoint contract; there's no docs/changelog/announcement
that changed it. "Stale" needs a source.

## What your PR should change

Your bug half is real: `findModel` exact-match + `(max)` → null → guard passes, so
`glm-5.2(max)`→`/messages` and `minimax-m3(max)`→`/responses` are wrong routes. That
guard-bypass test is worth keeping. But the matrix assertions should be corrected:

- `getModelSupportedFormats("opencode-go","deepseek-v4-flash")` → `["openai"]` (docs +
  real-payload behavior), not 3 formats.
- Claude client → DeepSeek → `/chat/completions`, not `/messages`.
- Add a cell: duplicate tool names on `/messages` → expect 400 upstream (documents why
  `["openai"]` is the safe declaration); `(max)` must not change the chosen endpoint.

I've cleaned PR #3332 to exactly that (single commit, `["openai"]`, suffix strip, tests),
and removed the unrelated #3321 changes from it. If you'd rather land the guard-bypass
test first, I'm happy to rebase — it's the one part of #3333 I consider correct.

KunN-21:
One more note to make the standard explicit — contributions are welcome, but conclusions must hold against the docs, not just a personal live test.

**The defect, precisely**
- Where: `opencode-go` with a Claude-format client. #3278's transport routing picks `/zen/go/v1/messages` for DeepSeek, and the upstream `/messages` shim rejects real Claude Code bodies with `400 {"model":"deepseek-v4-flash"}`.
- Triggering case: duplicate exact tool names (two `Bash`, two `Read`, …). Reproduced live: same `/messages` request 200 with one tool, 400 with a second same-name tool, regardless of `(max)`.
- Consequence: every genuine Claude Code session (which commonly sends duplicated built-in tools across MCP servers) fails on `ocg/deepseek-v4-flash(max)`. That's the case you saw in the 14:39 log.

**The evidence that governs**
- Official docs, fetched today: https://opencode.ai/docs/go/ — DeepSeek V4 Pro/Flash endpoints column: `https://opencode.ai/zen/go/v1/chat/completions`. `/v1/messages` is documented only for MiniMax and Qwen models.
- A minimal single-turn live matrix (no tools) is not a substitute for that contract; it does not cover the payloads Claude Code actually sends.

**The principle**
Please keep the matrix assertions aligned with the docs: `supportedFormats` must mirror what the documentation endorses, and a new live claim ("X endpoint works") needs a doc/changelog/announcement reference plus a payload representative of real clients — including tools and duplicate names. Test personal success, then confirm it generalizes; don't generalize from one happy-path probe.

I've aligned PR #3332 to exactly this: DeepSeek = `["openai"]`, `(max)` stripped before registry lookup, and tests covering the duplicate-tools case so the regression stays documented. Happy to rebase/merge whichever lands first; the guard-bypass tests in #3333 remain valid and worth keeping.

PR #3337, opened 2026-08-15
feat(opencode-go): show subscription quota on the dashboard

### LINKED ISSUE #3334: Feature Request / Bug: Support Quota Tracking and Display for `opencode-go` Provide
### Description
The subscription quota and limits for the **OpenCode Go** (`opencode-go`) provider connection do not display or render in the `/dashboard/quota` dashboard page. 

Currently, other API-key based providers (like `ollama`) and OAuth-based ones (like `antigravity`) successfully fetch and display progress bars for their weekly, monthly, or rolling quotas. This issue details why it is missing and provides the necessary codebase modifications to support it natively.

---

### Root Cause Analysis

1. **Provider Registry Feature Flags**:
   The registry definition file at `open-sse/providers/registry/opencode-go.js` is missing the `features` block. Without this, the system filters out the connection from quota-related endpoints.
2. **Missing Backend Fetcher & Dispatcher**:
   - `open-sse/services/usage/misc.js` does not have a fetcher to query the OpenCode Go subscription endpoint (`https://opencode.ai/zen/go/v1/usage`).
   - `open-sse/services/usage.js` does not register the `"opencode-go"` provider ID under the `USAGE_HANDLERS` dictionary.
3. **Frontend Quota Display Logic**:
   The frontend React page/component for quotas (`src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js` or the `/dashboard/quota` page bundle) has a `switch (provider)` block. Because `"opencode-go"` is not mapped to the block handling `ollama` or `deepseek`, it falls back to the `default` case. The default case does not pass the `remainingPercentage` property, causing the progress bars to fail to render.

---

#### 1. Add Features Flag to the Registry
In `open-sse/providers/registry/opencode-go.js`, append the `features` property to the exported object:

```diff
     { id: "qwen3.6-plus", name: "Qwen 3.6 Plus", supportedFormats: ["openai", "claude"] },
   ],
+  features: {
+    usage: true,
+    usageApikey: true,
+  },
 };
```

#### 2. Implement the Fetcher Function
In `open-sse/services/usage/misc.js`, implement `getOpencodeGoUsage`:

```javascript
export async function getOpencodeGoUsage(apiKey, proxyOptions = null) {
  if (!apiKey) {
    return { message: "OpenCode Go API key not available." };
  }

  try {
    const response = await proxyAwareFetch("https://opencode.ai/zen/go/v1/usage", {
      headers: {
        Authorization: `Bearer ${apiKey}`,
        Accept: "application/json",
      },
    }, proxyOptions);

    if (response.status === 401 || response.status === 403) {
      return { message: "OpenCode Go API key invalid or expired." };
    }

    if (!response.ok) {
      return { message: `OpenCode Go usage API error (${response.status}).` };
    }

    const data = await response.json();
    const usage = data?.usage || {};
    const quotas = {};

    if (usage.rolling) {
      const rollingPercent = Number(usage.rolling.percent) || 0;
      quotas["Rolling (5h)"] = {
        used: rollingPercent,
        total: 100,
        remainingPercentage: Math.max(0, 100 - rollingPercent),
        resetAt: usage.rolling.resetsAt || null,
        unlimited: false,
      };
    }

    if (usage.weekly) {
      const weeklyPercent = Number(usage.weekly.percent) || 0;
      quotas["Weekly (7d)"] = {
        used: weeklyPercent,
        total: 100,
        remainingPercentage: Math.max(0, 100 - weeklyPercent),
        resetAt: usage.weekly.resetsAt || null,
        unlimited: false,
      };
    }

    if (usage.monthly) {
      const monthlyPercent = Number(usage.monthly.percent) || 0;
      quotas["Monthly (30d)"] = {
        used: monthlyPercent,
        total: 100,
        remainingPercentage: Math.max(0, 100 - monthlyPercent),
        resetAt: usage.monthly.resetsAt || null,
        unlimited: false,
      };
    }

    return { plan: "OpenCode Go", quotas };
  } catch (error) {
    return { message: `OpenCode Go error: ${error.message}` };
  }
}
```

#### 3. Register the Handler in Dispatcher
In `open-sse/services/usage.js`, import and register the handler:

```diff
 import {
   getOllamaUsage,
   getGlmUsage,
   getMinimaxUsage,
   getVercelAiGatewayUsage,
   getCodebuddyUsage,
   getGrokCliUsage,
   getKimiUsage,
   getDeepseekUsage,
   getQoderUsage,
+  getOpencodeGoUsage,
 } from "./usage/misc.js";

 ...

 export const USAGE_HANDLERS = {
   ...
   deepseek: (c) => getDeepseekUsage(c.apiKey, c.proxyOptions),
+  "opencode-go": (c) => getOpencodeGoUsage(c.apiKey, c.proxyOptions),
 };
```

#### 4. Update the Frontend Quota Component
In the frontend dashboard quota rendering component (e.g., `src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js`), group `"opencode-go"` under the same case structure as `"ollama"` or `"deepseek"` so that the `remainingPercentage` progress bar is correctly mapped:

```diff
-case "vercel-ai-gateway":
-case "ollama":
-case "deepseek":
+case "vercel-ai-gateway":
+case "ollama":
+case "deepseek":
+case "opencode-go":
   t.quotas && Object.entries(t.quotas).forEach(([e,t])=>{
     r.push({
       name: e,
       used: t.used||0,
       total: t.total||0,
       resetAt: t.resetAt||null,
       remainingPercentage: t.remainingPercentage
     })
   });
   break;
```

ntdatt812:
Closes #3334.

`opencode-go` connections render nothing on the quota dashboard while `ollama`, `deepseek` and `antigravity` render progress bars. The provider is a paid subscription with real rolling/weekly/monthly windows, so there is something to show.

The diagnosis in #3334 is the reporter's. I checked each of its claims against this codebase and against the upstream route source rather than taking them on trust — three hold, one does not, and the upstream source settled several details the issue could only guess at.

## What the upstream route actually returns

The endpoint is first-party but undocumented. Its implementation is `anomalyco/opencode` → `packages/console/app/src/routes/zen/go/v1/usage.ts` (added in anomalyco/opencode#16513), and it decides the whole contract:

```jsonc
{ "usage": {
    "rolling": { "status": "ok",           "percent": 12,  "resetsAt": "<ISO>" },
    "weekly":  { "status": "ok",           "percent": 47,  "resetsAt": "<ISO>" },
    "monthly": { "status": "rate-limited", "percent": 100, "resetsAt": "<ISO>" }
}}
```

Four things follow from reading it, none of which are in the issue:

- **`percent` is already a floored, clamped `0..100` integer** — `usagePercent: Math.floor(Math.min(100, (usage / limit) * 100))` in `console/core/src/subscription.ts`. So `used` is the percent and `total` is `100`; my clamp is belt-and-braces, not a scale conversion.
- **`resetsAt` is an ISO string**, `new Date(Date.now() + resetInSec * 1000).toISOString()` — handled by the shared `parseResetTime`.
- **403 is not an auth failure.** The route answers `403 EntitlementError "OpenCode Go subscription required."` when the key authenticates but the workspace has no Go plan. Folding 403 into the 401 branch — as the proposed patch does — tells a user with a perfectly good key to go reissue it. They are split here.
- **All three windows are always emitted**, so "skip an absent window" never fires in practice. The guard stays anyway, because the payload shape has already changed once since the route merged; an unreadable window is skipped rather than rendered as a `0%` bar that reads as "quota untouched".

That last point also fixes labels. I originally wrote `Weekly (7d)` / `Monthly (30d)`, copying the issue. Both are wrong: `analyzeWeeklyUsage` resets on a **calendar week** boundary (`getWeekBounds`) and `analyzeMonthlyUsage` on the **subscription anniversary** (`getMonthlyBounds(now, timeSubscribed)`), not rolling spans. The rolling window's length is server-side plan config and never appears in the payload, so `(5h)` was a guess too. Labels are now plain `Rolling` / `Weekly` / `Monthly` — each row already renders its own countdown from `resetAt`, which is the honest version of the same information.

Error bodies are `{type:"error", error:{type, message}}`, so the message is unwrapped rather than echoed as a raw JSON envelope.

## The four claims in the issue

**1. Registry `features` block — confirmed, and it is the actual blocker.** `USAGE_SUPPORTED_PROVIDERS` / `USAGE_APIKEY_PROVIDERS` are derived in `src/shared/constants/providers.js` by filtering the registry on `features.usage` / `features.usageApikey`. Without them `/api/usage/[connectionId]` returns `"Usage not available for this connection"` before any fetch happens, and `ProviderLimits` never calls for the connection.

**2. No fetcher in `misc.js` — confirmed.**

**3. No `USAGE_HANDLERS` entry — confirmed.** `getUsageForProvider` would return `"Usage API not implemented for opencode-go"`.

**4. "The default case does not pass `remainingPercentage`, causing the progress bars to fail to render" — this one is not right.** Both render paths fall back to the same arithmetic:

- `QuotaTable` → `getRemainingPercentage(quota)` → `calculatePercentage(used, total)` when `remainingPercentage` is absent.
- `ProviderLimitCard` recomputes `Math.round(((total - used) / total) * 100)` and does not read `remainingPercentage` at all.

For a `used = percent, total = 100` row both yield `100 - used` — exactly what `remainingPercentage` carries, including at the `used = 0` and `used = 100` edges. I proved it rather than reasoning about it: deleting the `case "opencode-go":` line makes the field-forwarding test fail while **"renders the remaining percentage the provider reported" still passes**. The bars would have rendered correctly without any frontend change.

I kept the frontend hunk anyway — it is one line, it puts the provider with `deepseek`/`ollama`/`kimi` where it belongs, and it keeps the row correct if the shape ever stops being a straight percentage — but it is grouped for intent, not a fix, and the comment and commit message now say so. Note also that the diff in the issue is written against the built bundle (`([e,t])=>{…}`), which is why it names `ProviderLimits/index.js`; the real `switch (provider)` is in `ProviderLimits/utils.js`.

## One more deviation

The proposal hardcodes the URL inside `misc.js`. This repo keeps usage endpoints in `transport.usage.url` and reads them through `U(id)` in `services/usage/shared.js` — that is how `glm`, `glm-cn` and `vercel-ai-gateway` work — so it is declared next to the other transports, with a guard for the case where it is absent.

## What I could not verify

**I have no OpenCode Go subscription, so I have never seen a live 200 body.** The shape above is read off the upstream source and cross-checked against four independent third-party clients that consume the same endpoint, but it is not a captured response. Every failure path therefore degrades to a `{message}` instead of throwing or rendering a wrong bar: unknown shape → "No quota windows reported", non-JSON → "response was not JSON", transport error → the error text. Usage fetchers are display-only and off the routing path, so the worst case of a shape drift is a message where bars would be.

If a maintainer or the reporter can paste one redacted 200 body, I will tighten the parser to it.

## Verification

15 tests in `unit/opencode-go-usage-3334.test.js`, following the existing `deepseek-usage.test.js` structure: the registry flags land in both derived lists, the endpoint comes from the registry, the request is a GET with `Authorization: Bearer`, the three windows map with the remainder derived, labels assert no span, an unreadable window is skipped, an out-of-range percent clamps, an empty payload returns a message, a missing key never calls out, 401 reads as an invalid key, **403 reads as a missing subscription and explicitly not as an invalid key**, an upstream error message is unwrapped from its envelope, and a non-JSON body plus a transport failure both return messages. Two `parseQuotaData` tests close it: one that the field is forwarded, one that the *rendered* percentage is right.

Mutation-checked, since a passing test proves nothing about whether it bites:

| mutation | result |
|---|---|
| drop `features` from the registry | registry test fails — `expected [...] to include 'opencode-go'` |
| fold 403 back into the 401 branch | only the 403 test fails |
| drop `case "opencode-go":` | field test fails, **render test still passes** — the evidence for claim 4 above |

Full suite (`npx vitest run unit translator`): **1808 passed / 93 failed, failing set byte-identical to master's**, `comm` empty in both directions. `npx eslint` reports no issues on the three changed files.

### DIFF
--- a/open-sse/providers/registry/opencode-go.js
+++ b/open-sse/providers/registry/opencode-go.js
@@ -21,6 +21,9 @@ export default {
   transport: {
     baseUrl: "https://opencode.ai/zen/go/v1/chat/completions",
     headers: {},
+    usage: {
+      url: "https://opencode.ai/zen/go/v1/usage",
+    },
   },
   // Multi-endpoint: pick the transport matching the client sourceFormat to skip
   // translation. Guarded per-model by `supportedFormats` (see chatCore) because
@@ -48,4 +51,8 @@ export default {
     { id: "qwen3.7-plus", name: "Qwen 3.7 Plus", supportedFormats: ["openai", "claude"] },
     { id: "qwen3.6-plus", name: "Qwen 3.6 Plus", supportedFormats: ["openai", "claude"] },
   ],
+  features: {
+    usage: true,
+    usageApikey: true,
+  },
 };
--- a/open-sse/providers/registry/opencode-go.js
+++ b/open-sse/providers/registry/opencode-go.js
@@ -21,6 +21,9 @@ export default {
   transport: {
     baseUrl: "https://opencode.ai/zen/go/v1/chat/completions",
     headers: {},
+    usage: {
+      url: "https://opencode.ai/zen/go/v1/usage",
+    },
   },
   // Multi-endpoint: pick the transport matching the client sourceFormat to skip
   // translation. Guarded per-model by `supportedFormats` (see chatCore) because
@@ -46,4 +49,8 @@ export default {
     { id: "qwen3.7-plus", name: "Qwen 3.7 Plus", supportedFormats: ["openai", "claude"] },
     { id: "qwen3.6-plus", name: "Qwen 3.6 Plus", supportedFormats: ["openai", "claude"] },
   ],
+  features: {
+    usage: true,
+    usageApikey: true,
+  },
 };
--- a/open-sse/services/usage.js
+++ b/open-sse/services/usage.js
@@ -21,6 +21,7 @@ import {
   getGlmUsage,
   getVercelAiGatewayUsage,
   getQoderUsage,
+  getOpencodeGoUsage,
 } from "./usage/misc.js";
 
 /**
@@ -54,6 +55,7 @@ const USAGE_HANDLERS = {
   "grok-cli": (c) => getGrokCliUsage(c.accessToken, c.providerSpecificData, c.proxyOptions),
   kimi: (c) => getKimiUsage(c.accessToken, c.apiKey, c.proxyOptions, c.providerSpecificData),
   deepseek: (c) => getDeepseekUsage(c.apiKey, c.proxyOptions),
+  "opencode-go": (c) => getOpencodeGoUsage(c.apiKey, c.proxyOptions),
 };
 
 export async function getUsageForProvider(connection, proxyOptions = null, options = {}) {
--- a/open-sse/services/usage.js
+++ b/open-sse/services/usage.js
@@ -22,6 +22,7 @@ import {
   getOllamaUsage,
   getVercelAiGatewayUsage,
   getQoderUsage,
+  getOpencodeGoUsage,
 } from "./usage/misc.js";
 
 /**
@@ -56,6 +57,7 @@ const USAGE_HANDLERS = {
   kimi: (c) => getKimiUsage(c.accessToken, c.apiKey, c.proxyOptions, c.providerSpecificData),
   deepseek: (c) => getDeepseekUsage(c.apiKey, c.proxyOptions),
   zed: (c) => getZedUsage(c.accessToken, c.providerSpecificData, c.proxyOptions),
+  "opencode-go": (c) => getOpencodeGoUsage(c.apiKey, c.proxyOptions),
 };
 
 export async function getUsageForProvider(connection, proxyOptions = null, options = {}) {
--- a/open-sse/services/usage/misc.js
+++ b/open-sse/services/usage/misc.js
@@ -1,9 +1,9 @@
 /**
- * Misc usage handlers (iFlow, Ollama, GLM, Vercel AI Gateway, Qoder)
+ * Misc usage handlers (iFlow, Ollama, GLM, Vercel AI Gateway, Qoder, OpenCode Go)
  */
 
 import { proxyAwareFetch } from "../../utils/proxyFetch.js";
-import { U } from "./shared.js";
+import { U, parseResetTime } from "./shared.js";
 
 export { getGlmUsage } from "./glm.js";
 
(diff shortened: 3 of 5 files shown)


ntdatt812:
Rebased onto `master` (`90b52e06`). Two things changed in the process, both worth naming:

`getGlmUsage` moved out of `misc.js` into `open-sse/services/usage/glm.js` while this sat open. This branch had been carrying a one-line drive-by on it (accepting 403 alongside 401 as an invalid key), which is unrelated to opencode-go and no longer belongs in this file. I dropped it rather than following it across the move — happy to send it separately if you want it, but it should not ride on a feature PR.

What remains is only the opencode-go path: the registry features block, the `USAGE_HANDLERS` entry, the fetcher, and the `parseQuotaData` case. `tests/unit/opencode-go-usage-3334.test.js` — 15 passed on the rebased branch.

IEatCodeDaily:
Validated against a live OpenCode Go subscription on 9router v0.5.59. The connection appears in GET /api/usage/providers as `opencode-go`, but GET /api/usage/<connection-id> returns exactly `{"message":"Usage not available for this connection"}`. This confirms the registry/dispatch blocker addressed by this PR. I can test a patched build against the live account and provide a redacted 200 response shape once available.

ntdatt812:
Thank you — that is the half I could not get. I have no OpenCode Go subscription,
so the response shape in this PR is inferred, and your `{"message":"Usage not
available for this connection"}` is the dispatch blocker reproduced exactly.

Yes please to the patched-build test. Here is what the parser currently assumes,
so you know what to look for in the redacted payload:

```jsonc
{
  "usage": {
    "rolling": { "percent": 42, "resetsAt": "…" },
    "weekly":  { "percent": 17, "resetsAt": "…" },
    "monthly": { "percent":  3, "resetsAt": "…" }
  }
}
```

Four things decide whether it renders correctly, and any one of them being
different is a bug I would rather find from your payload than from a user:

1. **The top-level key is `usage`.** If it is `limits`, `quotas` or the windows
   sit at the root, every bar disappears — unknown windows are skipped rather
   than drawn at 0%, so the failure is silent.
2. **The window keys are `rolling` / `weekly` / `monthly`.** Any other naming
   skips the same way.
3. **`percent` is 0–100, not a 0–1 ratio.** A ratio would render 100% used as a
   1% bar, which reads as "quota untouched" — the worst possible wrong answer
   here. (Ollama's endpoint does use 0–1, which is why this is worth pinning.)
4. **`resetsAt`** — `parseResetTime` accepts an ISO string, a `Date`, or a Unix
   timestamp in seconds or milliseconds, so most shapes are fine; I would just
   like to know which, and whether it is present on all three windows or only
   some.

Percent values and reset timestamps can be blanked; only the key names, the
nesting, and the *type* of each value matter.

If you can also capture a 403 body from an account with a key but no Go
subscription, that would pin the other branch: this PR treats 403 as an
entitlement error rather than an invalid key, and surfaces
`error.message` from `{type:"error", error:{type, message}}`. If the real
envelope differs, the user gets the raw text truncated to 200 characters
instead of a useful sentence.

Branch is `feat/opencode-go-quota-3334` on `ntdatt812/9router`, rebased onto
`master` at `90b52e06`.

ntdatt812:

PR #3346, opened 2026-08-15
fix(kiro/cloudflare): dedup catalogs, request-scoped 400s, live model import + usage

askidmobile:
## Summary

Batch of fixes across Kiro and Cloudflare providers.

### Kiro — model catalog dedup & error handling
- **Dedup catalogs**: removed synthetic `-thinking`/`-agentic`/`-thinking-agentic` variants from `buildVariants` (kiroModels.js) and the static registry (kiro.js). Thinking/agentic behavior is driven by request params + translator suffix handling, not duplicated model slots. Fixes the `auto-thinking` `REQUEST_BODY_INVALID` 400.
- **Drop experimental GPT-5.6 tiers** (sol/terra/luna) + stale `claude-opus-4.7/4.5`, `claude-sonnet-5` from static catalogs — experimental rollouts arrive via the live `ListAvailableModels` catalog; accounts outside the rollout got `INVALID_MODEL_ID`. Added base `auto` to the static registry.
- **Request-scoped 400s**: `improperly formed request` and `invalid model id` now `shouldFallback: false, cooldownMs: 0` in `ERROR_RULES` — no account rotation/lockout for deterministic body/model errors (same body fails on every account). `checkFallbackError` flattened to dedupe the two identical `shouldFallback === false` branches.
- `capabilities.js`: kept base-id entries only (translator strips legacy suffixes).
- `tests/translator/registerAll.js`: removed imports of translators deleted by the commandcode migration (revives kiro/claude direct-route suites).

### Antigravity
- Disambiguated duplicate display name "Gemini 3.5 Flash (High)" (`gemini-3-flash-agent` MITM slot → "Gemini 3.5 Flash Agent (High)"). Tier suffixes are real upstream models with separate quotas — not deduped.

### Cloudflare — live model import + usage
- **Live model import**: `deriveModelsEndpoint` skipped cloudflare (baseUrl has `{accountId}` placeholder) → no Import button. Added `customResolver` hitting `GET /accounts/{accountId}/ai/models/search`.
- **Correct field mapping**: Cloudflare returns `{id: <uuid>, name: "@cf/..."}` — `name` is the real inference id. Maps `id = m.name`, pretty-prints display name from the path segment, filters to chat/image (drops embeddings/TTS/STT/classification/translation/guard/reranker), derives kind from name keywords.
- **Daily neuron usage**: Cloudflare has no quota API; added a usage handler querying the GraphQL analytics endpoint (`aiWorkersAiInvocationsAdaptiveGroups`) for today's neuron sum, surfaced as a best-effort daily quota (10k free-tier neurons, resets UTC midnight). Enabled `features.usage`.

### Tests
- Updated kiro-model-slots + capabilities tests for the cleaned catalogs.
- No regressions: 84 baseline fails unchanged before/after.
- Verified live on a deployment: Kiro live catalog (9 models, no GPT-5.6), Cloudflare models (40 chat/image, correct `@cf/` ids), Cloudflare usage endpoint returns neuron quota.


askidmobile:
Post-review follow-up (9696b804): flattened duplicated `shouldFallback === false` early-returns in `checkFallbackError` into one branch; fixed stale `buildVariants` docblock. Verified `checkFallbackError` semantics: 'improperly formed request' → no fallback/cooldown; 429 backoff, 401/404 long cooldown, transient 30s default all unchanged. Legacy `-thinking` model ids from old clients still resolve via translator suffix stripping (back-compat), so no hard breakage for saved selections.

Issue #3, opened 2026-01-10
How to use raptor mini ?


goguma-icedtea:
How to use raptor mini ?

Continue config isn't using jsom file

decolua:
Sorry for the delay — too many issues to handle in time. Auto-closing issues inactive for more than 10 days. Feel free to reopen if this is still relevant. Thanks for your understanding!

decolua:
Sorry for the delay — too many issues to handle in time. Auto-closing issues inactive for more than 10 days. Feel free to reopen if this is still relevant. Thanks for your understanding!

PR #3347, opened 2026-08-15
Improvements: opencode/Hermes QOL - usable /v1/models listing, bare-name resolution, usage pipeline fixes

CyrixJD115:
## Summary

Three fixes for the usage/activity pipeline and the /v1/models catalog,
verified on a live deployment.

## 1 — Combo requests logged under the wrong name

Combos resolve to their first leg before execution, and the usage write path
only ever sees the resolved model — the requested name is dropped. Requests
through combos were filed under the leg model, so filtering activity by the
model actually called returns nothing.

Fix: carry the requested model name through to the usage row (stored in
usageHistory.meta, no schema change) and render it as
`RequestedModel -> resolved-model` in the activity lists. Plain rows unchanged.

## 2 — Aborted streams leave no usage row

Usage is only persisted at stream completion. Clients that close the SSE early
(common with agent tool-call loops) never deliver the final usage chunk, so
those requests vanished from the logs entirely — in one observed session,
15 of 18 requests were missing.

Fix: on mid-stream disconnect, persist a 0-token row with status "aborted".
Completed streams still write exactly one normal row (one-shot disconnect flag).

## 3 — /v1/models omits custom models on connection-less providers

The listing loop is driven by active provider connections. noAuth providers can
never have a connection row, so they never enter the loop — their custom models
route fine but are invisible in the listing.

The tricky part: requests to these models still succeed — the server resolves
and proxies them regardless (passthrough routing). So the model "works" when
called directly, which masks the bug: it only surfaces in clients that perform
model-listing validation rather than just attempting the completion. Hermes,
for example, pre-validates a switch against /v1/models and warns (or suggests
a "similar" listed model) even though a completion call would succeed —
leading users to think the client is broken when the catalog is.

Client before the fix:
```
  Model switched to oc/deepseek-v4-flash-free
  Provider: 9router
  Context: 1,000,000 tokens
  Warning: Model oc/deepseek-v4-flash-free was not found in this provider's
  model listing.
  Similar models: bzl/deepseek-v4-flash
```

Client after the fix — same switch, warning gone:
```
  Model switched to oc/deepseek-v4-flash-free
  Provider: 9router
  Context: 1,000,000 tokens
```

Fix: after the connection-driven listing is built, backfill custom models for
any provider alias not already covered, then dedupe. Generic — covers noAuth
providers, alias mismatches, and future connection-less providers.

Before/after on live instance: 15 -> 18 listed models, previously invisible
entries now present; client switches cleanly with no warning.

## Testing

7 files, +101/-24. No changes to routing, translators, or executors.

- Live-data validation (production DB copy): 9/9 pass
- Focused usage tests: 21/21 pass
- Full unit suite vs clean upstream worktree: identical failure sets, zero
  new failures
- End-to-end on live deployment: all three behaviors confirmed via API + DB

## AI-assisted changes

These changes and their validation were produced with GLM 5.2 (Z.AI) driving
the diagnosis and test runs — live-data verification (9/9), focused usage
tests (21/21), and a full unit-suite regression comparison against a clean
upstream worktree (identical failure sets, zero new failures).

Treat this PR like any other: the AI ran the tests, but it can be wrong.
Review the diff, re-run the suite, and verify the behaviors on your own
instance before merging.

## Notes

Abort rows intentionally fire on any mid-stream disconnect — usage genuinely
happened upstream. Separate pre-existing issue (not addressed here): usage
extraction can yield IN 0 / OUT 0 on some completed streams.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -11,7 +11,7 @@ import { PROVIDERS } from "../config/providers.js";
 import { createErrorResult, parseUpstreamError, formatProviderError } from "../utils/error.js";
 import { HTTP_STATUS, TOKEN_SAVER_HEADER } from "../config/runtimeConfig.js";
 import { handleBypassRequest } from "../utils/bypassHandler.js";
-import { trackPendingRequest, appendRequestLog, saveRequestDetail } from "@/lib/usageDb.js";
+import { trackPendingRequest, appendRequestLog, saveRequestDetail, saveRequestUsage } from "@/lib/usageDb.js";
 import { getExecutor } from "../executors/index.js";
 import { supportsGrokCliReasoningEffort } from "../config/grokCli.js";
 import { buildRequestDetail, extractRequestConfig } from "./chatCore/requestDetail.js";
@@ -299,6 +299,21 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   const streamController = createStreamController({
     onDisconnect: (reason) => {
       trackPendingRequest(model, provider, connectionId, false);
+      // Client aborted mid-stream (e.g. tool-call loops that close the SSE as
+      // soon as the tool_call delta arrives). The final usage chunk never
+      // arrives, so no row would be persisted — record the abort with the
+      // requested model name so the activity log still shows the request.
+      saveRequestUsage({
+        provider: provider || "unknown",
+        model: model || "unknown",
+        tokens: { prompt_tokens: 0, completion_tokens: 0 },
+        timestamp: new Date().toISOString(),
+        connectionId: connectionId || undefined,
+        apiKey: apiKey || undefined,
+        endpoint: clientRawRequest?.endpoint || null,
+        requestedModel: clientRawRequest?.body?.model || undefined,
+        status: "aborted",
+      }).catch(() => { });
       if (onDisconnect) onDisconnect(reason);
     },
     onError: () => trackPendingRequest(model, provider, connectionId, false),
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -370,6 +370,15 @@ export async function handleNonStreamingResponse({ providerResponse, provider, m
 
   reqLogger.logConvertedResponse(translatedResponse);
 
+  // Echo the exact model the client requested instead of the upstream id.
+  // Passthrough providers (opencode free tier) return the bare resolved model
+  // with the provider prefix stripped; clients that trust the echo re-send the
+  // bare name, which then mis-routes on the next hop. Rewriting keeps the
+  // round-trip name stable (OpenRouter-style proxy echo).
+  if (body?.model && translatedResponse && typeof translatedResponse === "object" && !Array.isArray(translatedResponse)) {
+    translatedResponse.model = body.model;
+  }
+
   const totalLatency = Date.now() - requestStartTime;
   saveRequestDetail(buildRequestDetail({
     provider, model, connectionId,
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -4,6 +4,7 @@ import { fromOpenAIFinish } from "../../translator/concerns/finishReason.js";
 import { ollamaBodyToOpenAI } from "../../translator/response/ollama-to-openai.js";
 import { addBufferToUsage, filterUsageForFormat } from "../../utils/usageTracking.js";
 import { createErrorResult } from "../../utils/error.js";
+import { canonicalEchoModel } from "../../services/model.js";
 import { HTTP_STATUS } from "../../config/runtimeConfig.js";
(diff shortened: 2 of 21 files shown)


CyrixJD115:
I did further tests myself, manually, and for some reason the issue keeps arising with provider-less models, even though I thought the fix was already done. I did the test and it was successful, but after a bit, the same issue comes back 😤 Gonna keep working on this. This issue usually happens on clients that do model listing validations, like Hermes for example.

CyrixJD115:
I closed this PR because something isnt right with OpenCode free models. I noticed it on my setup, my friend’s setup, and another dev partner’s setup, something is screwing up. And I just noticed that even within OpenCode, some models are having issues, except for a few

CyrixJD115:
# Summary

QOL pass to make 9router play nicer with opencode and Hermes Agent. The main
bits: `/v1/models` on a fresh install shows only models that actually work
instead of the whole static catalog, bare model names resolve to the right
provider, and the Recent Requests card stops flip-flopping between prefixed and
bare names. It also carries through the earlier usage/activity pipeline fixes
and some test cleanup.

## 1 — /v1/models shows what's usable, not everything

Before, a fresh install with zero connections dumped the full static catalog —
698 models, most of which need an API key and fail on call. Hermes auto-detect
offered all 698. Now:

- DB healthy but empty: only connection-less `noAuth` providers
  (`mmf/mimo-auto`), custom models, and the live opencode catalog (`oc/*`).
  API-key providers are dropped.
- DB unavailable: keep the full static dump as a degraded fallback — stale but
  complete beats empty.

On the test container that's 698 → 8 working models.

## 2 — Bare model names resolve to their real owner

Clients like Hermes send bare names. Previously `deepseek-v4-flash-free`
blind-routed to openrouter and `mimo-v2.5-free` to openai via prefix inference —
both reject. Resolution order is now: custom models → user aliases → static
registry → live opencode catalog → prefix inference as a last resort. When the
catalog is unreachable, `-free` names (and `big-pickle`) stay owned by opencode
instead of falling through.

## 3 — Echo canonicalization

Bare requests resolved to a connection-less catalog provider now echo back the
listing form (`oc/big-pickle`), so clients that validate the echoed model
against `/v1/models` don't warn or re-send a name 9router can't route. Prefixed
requests keep their exact form.

## 4 — Recent Requests: stable display + Model ID | Model Name toggle

The card flip-flopped between `oc/big-pickle` and `big-pickle` (and showed
`oc/big-pickle → big-pickle` arrows) because two SSE producers built rows
differently. Both now share one row builder that keeps the resolved model,
requested model, and provider as separate fields, with dedupe on the resolved
model. The card shows the bare name by default, with a sliding toggle to switch
to the full id.

## 5 — Carried usage/activity fixes

- **Combo requests log the name the client called** — it's carried in
  `usageHistory.meta` (no schema change).
- **Aborted streams leave a usage row** — mid-stream disconnects now persist a
  0-token row with status `aborted` instead of vanishing (one observed session
  was missing 15 of 18 requests). Completed streams still write exactly one row.
- **Custom models on connection-less providers are backfilled** into `/v1/models`
  after the main listing and deduped — they never entered the connection-driven
  loop, so clients validating against the listing couldn't see them. Before,
  Hermes warned:

  ```
  Model switched to oc/deepseek-v4-flash-free
  Provider: 9router
  Context: 1,000,000 tokens
  Warning: Model oc/deepseek-v4-flash-free was not found in this provider's
  model listing.
  Similar models: bzl/deepseek-v4-flash
  ```

  After — same switch, no warning.

## 6 — Tests

- 13 new regression cases covering the fresh-install listing, bare-name
  resolution, canonical echo, and the recent-requests row shape.
- The mimo-free live endpoint test (hits real upstream) was failing in the
  default suite; gated behind `RUN_LIVE=1` per the `antigravity-cache` /
  `rtk.e2e` convention.

# Testing

- Full unit suite vs the pre-change worktree: identical failure sets, only the
  live test is now skipped (95 → 94).
- Fresh container: `/v1/models` lists exactly `mmf/mimo-auto` + `oc/*` (8
  models, no static dump).
- Live API: recent-requests rows carry `model` + `requestedModel` separately,
  no arrow joins.

# AI-assisted changes

Done with an AI agent driving the diagnosis, test runs, and deployments. The AI
ran the tests but can be wrong — review the diff, re-run the suite, and verify
the behaviors on your own instance before merging.

tmaarcxs:
Confirming this still reproduces on 0.5.65 (image 2026-09-03, `:latest` digest `sha256:3eb10446…`).

- `GET /api/models` (925 entries) includes `oc/muse-spark-1.3-contributor-free` (OpenCode Free / Zen, no-auth passthrough)
- `GET /v1/models` returns 146 models, no `oc/` prefix at all (checked 3 consecutive fetches)
- `POST /v1/chat/completions` with `{"model": "oc/muse-spark-1.3-contributor-free", ...}` through 9router returns HTTP 200 — routing works, only the listing is missing
- Hermes pre-validates `/model` switches against `/v1/models`, so it rejects the model even though a direct call would succeed — exactly the masking behavior described in fix #3 here

Same root cause as #1046 (open since May 12). Happy to re-test after merge.

PR #3352, opened 2026-08-15
feat(backoff): make 429 cooldown schedule configurable

### LINKED ISSUE #3343: Feature request: make 429 backoff configurable
When a provider/model returns 429, 9router applies a fixed exponential backoff before the account+model lock expires. The schedule is hardcoded in `open-sse/config/errorConfig.js`:

- `BACKOFF_CONFIG`: `base: 2000`, `max: 5 * 60 * 1000`, `maxLevel: 15` (2s → 4s → 8s… capped at 5 min)
- Applied via `ERROR_RULES` (`{ status: 429, backoff: true }` and the text rules "rate limit" / "too many requests" / "quota exceeded" / "capacity" / "overloaded") and stored per connection in SQLite (`open-sse/services/accountFallback.js`), reset on success

Different providers reset quota at very different rates (seconds to minutes), so a fixed schedule means either retrying too aggressively or keeping accounts locked far longer than needed.

**Proposed change**
Expose the 429 backoff schedule via environment variables with the current values as defaults, e.g.:

- `BACKOFF_BASE_MS` (default `2000`)
- `BACKOFF_MAX_MS` (default `300000`)
- `BACKOFF_MAX_LEVEL` (default `15`)

Consumed in `open-sse/config/errorConfig.js`, documented in `.env.example` / DOCKER.md. Optionally allow disabling backoff entirely (e.g. `BACKOFF_DISABLED=true`) so 429'd models are locked for a short fixed period instead.

**Out of scope (follow-ups)**
- Per-provider backoff overrides
- Honoring provider `Retry-After` / `resets_at` headers for cooldown (only Antigravity executor honors them today, capped at 10 s)

Jordannst:
## Summary

429 fallback locks used one fixed schedule for every provider: 2s, 4s, 8s, capped at 5 minutes. That works for some quota windows, but is too aggressive for others and unnecessarily long for providers that recover faster.

This adds three optional environment variables for tuning that account/model lock schedule:

- `BACKOFF_BASE_MS`  first cooldown delay (default: `2000`)
- `BACKOFF_MAX_MS`  maximum cooldown delay (default: `300000`)
- `BACKOFF_MAX_LEVEL`  maximum stored backoff level (default: `15`)

Each setting can be changed independently. Invalid values use that setting's existing default. A cap below the base delay is rejected as a contradictory configuration and retains the complete default schedule.

The default behavior is unchanged when no variables are set.

## Scope

- Applied only to the existing account/model lock backoff used after rate-limit fallback.
- Documented in `.env.example` and `DOCKER.md`.
- Deliberately does not add a global `BACKOFF_DISABLED` switch, provider-specific overrides, or alter provider `Retry-After` handling.

## Verification

- Added regression coverage for defaults, custom status/text-rule cooldowns, max-level/cap behavior, independent overrides, malformed values, contradictory values, and persisted account state.
- Focused suite: `25` tests passed across the backoff, account-reset, Kiro error-rule, and executor-retry suites.
- `eslint` passed for changed JavaScript files.
- `git diff --check` passed.

Closes #3343

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -10,6 +10,14 @@ DATA_DIR=/var/lib/9router
 PORT=20128
 NODE_ENV=production
 
+# Optional 429 account/model lock backoff. Defaults retain the current schedule:
+# 2s, 4s, 8s... capped at 5 minutes, with at most 15 backoff levels.
+# Each value is optional and must be a positive integer. Invalid values use that
+# knob's default; a max below the base is rejected as a complete schedule.
+# BACKOFF_BASE_MS=2000
+# BACKOFF_MAX_MS=300000
+# BACKOFF_MAX_LEVEL=15
+
 # Recommended security and ops variables
 API_KEY_SECRET=endpoint-proxy-api-key-secret
 MACHINE_ID_SALT=endpoint-proxy-salt
--- a/DOCKER.md
+++ b/DOCKER.md
@@ -64,6 +64,25 @@ docker run -d \
   decolua/9router:latest
 ```
 
+### Configure 429 account backoff
+
+When an upstream rate limit is detected, 9Router temporarily locks the affected
+account/model using exponential backoff. The default is unchanged: `2s`, `4s`,
+`8s` and so on, capped at `5 minutes` for up to `15` levels. Operators can tune
+that schedule for providers with different quota-reset windows:
+
+```bash
+-e BACKOFF_BASE_MS=2000 \
+-e BACKOFF_MAX_MS=300000 \
+-e BACKOFF_MAX_LEVEL=15
+```
+
+Each value is optional and must be a positive integer. Malformed values use
+that knob's default; a cap below the base is rejected as a contradictory
+schedule and uses the unchanged defaults. These settings govern account/model
+locks after rate-limit fallback, not provider-specific retry mechanisms or
+provider `Retry-After` hints.
+
 ## Optional Headroom sidecar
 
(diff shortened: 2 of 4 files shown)


GuoZhuang-cn:
+1 — and could the same treatment extend to the **fixed (non-backoff) cooldowns** as well?

In v0.5.69 the service-unavailable path is still hardcoded in the error-rules module (`chunks/3662.js`): 401/402/403/404 and "no credentials" / "improperly formed request" lock the account/model with `cooldownMs: 120000`, but the *default* for everything else — notably **500/503 from new-api-style upstreams** (`"No available channel for model X under group auto"`, `"分组 auto 下模型 X 的可用渠道不存在"`) — is a fixed `g = 30000` (30 s), with `h = 1800000` as the heavier cap.

Practical impact in my deployment (21 providers, a 6-member and a 10-member combo, each on fallback strategy): a Pro member whose upstream key was disabled (`401 → 503` wrapper, "API key is disabled") and another whose channel was dead (`500`) were still re-probed by every combo request every ~30 s for hours. Under load the whole combo pays a wasted upstream RTT per request, per dead member — exactly the "asking each model in turn even though it is locked" experience users report.

`BACKOFF_BASE_MS / MAX / MAX_LEVEL` solves the 429 schedule; a sibling knob like `DEFAULT_ERROR_COOLDOWN_MS` (default 30000) — or per-status overrides — would let us hold broken channels out for 120 s+ without patching minified chunks on every release (this project ships ~1 release/day, so chunk patches do not survive).

Happy to help test.

Issue #1982, opened 2026-06-22
Server OOM at 6 GB heap under concurrent multi-provider load


kwanter:
## Bug

Server process dies with V8 heap OOM under concurrent load. Supervisor restarts it, it immediately climbs again, and the cycle repeats.

## 9Router version

0.4.80 (installed globally via npm)

## Environment

- Node: v24.15.0 (nvm)
- OS: macOS 15.x, 16 GB RAM, Apple Silicon
- No container/cgroup limits

## Crashing command / launch path

cli.js line 557 hard-codes the heap cap:

```js
const child = spawn(RUNTIME, ["--max-old-space-size=6144", serverPath], { ... })
```

## Crash log (abridged)

```
FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
...
7: v8::internal::FactoryBase<>::NewRawTwoByteString(int, ...)
8: v8::internal::String::SlowFlatten(...)
9: v8::String::WriteUtf8(...)
10: node::StringBytes::Write(...)
11: node::Buffer::SlowWriteString<...>(...)
...
22: node::inspector::InspectorConsoleCall(...)
26: Builtins_AsyncFunctionAwaitResolveClosure
46: node::http_parser::Parser::OnStreamRead(...)
```

The heap is already capped at 6144 MB, so this is **not** a missing-cap issue. It dies right at the ceiling while stringifying a large two-byte string (crash frames 7-11) during an active HTTP stream.

## Reproduction

1. Start 9Router normally (`9router start`).
2. Send ~6 concurrent requests to different providers (claude-sonnet-4.5, deepseek-v4-pro, gpt-5.5, glm-5.1, MiniMax-M2.7, minimax-m2.5-free), each returning 70-110K tokens.
3. Heap climbs past 5.8-5.9 GB and V8 aborts.

## Evidence on disk

```
-rw-r--r--  24.4M  ~/.9router/request-details.json
-rw-r--r--  835.6K  ~/.9router/usage.json
-rw-r--r--  15.0K  ~/.9router/log.txt    (644 entries, many 200 OK / PENDING)
```

A 24 MB `request-details.json` plus in-flight 100K-token response buffers is consistent with request/response bodies being retained past the point they are needed. The `JsonStringify` frame in the crash trace suggests the server is serializing a very large object (likely the accumulated request/response payload) right before the allocation failure.

## Suggestions

[...]

PR #3368, opened 2026-08-16
fix(cli): stop the hard-coded heap cap from overriding the operator

### LINKED ISSUE #3365: [Bug] Hard-coded `--max-old-space-size=6144` in cli.js makes cgroup/low-memory deployments OOM-loop
## Related

- #1982 — same hard-coded heap cap, desktop 16GB environment
- #1245 — severe memory leak over days of uptime
- #2472 — request log bloat causing high-frequency OOM

## Bug

`cli.js` (line 615 in v0.5.55, line 557 in v0.4.80 per #1982) hard-codes the V8 heap cap for the spawned `next-server` child process:

```js
const child = spawn(RUNTIME, ["--dns-result-order=ipv4first", "--max-old-space-size=6144", serverPath], {
```

The parent `cli.js` process respects whatever node flags the launcher passes (e.g. a systemd `ExecStart` with `--max-old-space-size=384`), but the child — which is the process that actually serves HTTP — always starts with a 6 GB cap. This makes any host-level memory limit (systemd `MemoryMax`, Docker `--memory`, k8s limits) unenforceable by design:

1. The child's V8 believes it has 6 GB, so GC thresholds sit near 6 GB and the heap only grows.
2. Under traffic, RSS climbs to the cgroup ceiling and the kernel OOM-kills `next-server` mid-stream.
3. In-flight streaming responses die with the connection → clients see "Connection lost mid-response".
4. Supervisor restarts, heap climbs again. Repeat.

## Environment

- 9Router v0.5.55 (installed globally via npm, auto-updated 2026-08-15)
- Node v26.7.0, Ubuntu 24.04, x86_64
- 1.6 GB total RAM, `systemd` service with `MemoryMax=768M`

## Observation

- Idle: cgroup RSS ~90 MB, stable for hours.
- Under AI traffic: RSS climbs 92 MB → 485 MB in ~6 minutes and does not come back down (GC effectively never runs because the 485 MB heap is ~8% of the 6 GB cap).
- 64 restart segments since 2026-08-14, each segment peaking in a tight 478–485 MB band — i.e. "grow until cgroup kill", not an unbounded leak.
- dmesg confirms cgroup OOM kills of `next-server` (anon-rss ~485 MB at kill time).

## Suggested fix

1. **Make the heap cap configurable** (this was already requested in #1982): env var (e.g. `NINEROUTER_NODE_OPTIONS`) or CLI flag, transparently appended to the child spawn args. A sensible default for server deployments would be derived from available RAM, not a fixed 6 GB.
2. At minimum, respect a `NODE_OPTIONS` passthrough so operators can override without patching the installed `cli.js`.
3. Long-term: audit where request/response bodies are retained in the heap (see #1982's `request-details.json` evidence and #2472's log-bloat finding) — with a correct heap cap, GC can reclaim streaming buffers; without that audit, a bigger cap only delays the OOM.

## Workaround (for other low-memory operators)

Patch the installed file until a configurable cap ships:

```bash
sed -i 's/--max-old-space-size=6144/--max-old-space-size=448/' \n  "$(npm root -g)/9router/cli.js"
```

Note: any `npm update` of 9router reverts this patch — please add a log line or banner when a custom node-flags file is missing, so patched deployments notice the regression.

ntdatt812:
Closes #3365. Also the ask in #1982.

`cli/cli.js:615` spawns the next-server child with a fixed heap cap:

```js
spawn(RUNTIME, ["--dns-result-order=ipv4first", "--max-old-space-size=6144", serverPath], …)
```

## Why that cannot be worked around from outside

Node reads `NODE_OPTIONS` **before** command-line flags and lets the command line win. So an operator who sets a lower cap gets it silently discarded. Measured, not assumed:

```console
$ NODE_OPTIONS="--max-old-space-size=256" node --max-old-space-size=4096 -e "…heap_size_limit…"
4288 MB          # the flag wins
$ NODE_OPTIONS="--max-old-space-size=256" node -e "…heap_size_limit…"
448 MB           # NODE_OPTIONS alone works fine
```

That is the whole bug. On a host with a cgroup limit — systemd `MemoryMax`, `docker --memory`, k8s — the child runs believing it has 6 GB. GC thresholds sit near 6 GB, so a 485 MB heap looks like 8% utilisation and collection never becomes urgent; RSS climbs to the cgroup ceiling and the kernel OOM-kills `next-server` instead. In-flight streaming responses die with the process, the supervisor restarts it, and the cycle repeats — which matches the reporter's 64 restart segments all peaking in the same 478–485 MB band.

## What changes

The default is **unchanged at 6144** for the desktop case it was raised for. It just stops overriding the operator:

| configuration | flags passed to the child |
|---|---|
| nothing set | `--max-old-space-size=6144` |
| `NINEROUTER_MAX_OLD_SPACE_SIZE=384` | `--max-old-space-size=384` |
| `NINEROUTER_MAX_OLD_SPACE_SIZE=0` | *(none — node sizes the heap itself)* |
| `NODE_OPTIONS=--max-old-space-size=384` | *(none — NODE_OPTIONS is left to apply)* |
| `NINEROUTER_MAX_OLD_SPACE_SIZE=abc` | `--max-old-space-size=6144` + a warning |

Both mechanisms are covered because they serve different users: `NODE_OPTIONS` is what a systemd unit or `docker run -e` already uses, while the dedicated variable is reachable for someone running the tray build or `npx 9router` who has no convenient place to set node flags.

A junk value keeps the default and warns rather than silently leaving the heap uncapped — losing a safety cap to a typo is the wrong direction to fail in.

Verified end to end against a real child process, not just the resolver:

```
default                                ["--max-old-space-size=6144"]  => heap limit 6336 MB
NINEROUTER_MAX_OLD_SPACE_SIZE=384      ["--max-old-space-size=384"]   => heap limit  576 MB
NODE_OPTIONS=--max-old-space-size=384  []                             => heap limit  576 MB
```

The resolver lives in `cli/hooks/nodeFlags.js` next to `sqliteRuntime.js`/`trayRuntime.js`, which is both the existing convention for CLI runtime helpers and what makes it testable — `cli.js` itself runs on require. `hooks/` is already in the package's `files`, so it ships.

## What I did not do

The issue's first suggestion is to derive the default from available RAM. I left it out on purpose: `os.totalmem()` reports the **host's** memory, not the cgroup limit, so under `docker --memory` or a k8s limit it would still report the wrong number — it would not help the case that motivates this report, while changing behaviour for every existing install. Reading `/sys/fs/cgroup/memory.max` would work but is Linux-only and I have no containerised host here to verify it on. Worth doing as a follow-up, with the caveat that an explicit setting must still win.

The third suggestion — auditing where request/response bodies are retained (#1982, #2472) — is a separate piece of work. As the reporter says, a correct cap lets GC reclaim; it does not remove whatever is retaining.

## Verification

8 tests in `unit/cli-heap-flags-3365.test.js`: the 6144 default is preserved when nothing is set; an explicit cap is honoured (including surrounding whitespace); `0` emits no flag; a `--max-old-space-size` in `NODE_OPTIONS` suppresses ours, including the underscore spelling node also accepts and when mixed with other flags; an unrelated `NODE_OPTIONS` does not; a lookalike flag (`--max-old-space-size-hint`) does not false-positive; the dedicated variable beats `NODE_OPTIONS`; junk falls back to the default and warns; an empty/whitespace value counts as unset and warns about nothing.

Mutation-checked — deleting the `NODE_OPTIONS` passthrough fails exactly the "stands aside when NODE_OPTIONS already caps the heap" test.

`node --check cli/cli.js` passes. `npx eslint` reports no issues on the changed files.

Full suite (`npx vitest run unit translator`): 1808 passed / 95 failed. Two runs differed from master's failing set only in the network-dependent `unit/xai-oauth-service.test.js` timeouts (and, in one run, two known order-dependent DB tests) — these flip between runs on an unmodified tree too. Nothing under `tests/` imports `cli/` except the new test, so this change cannot reach them.

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -88,6 +88,20 @@ That's it! Start coding with FREE AI models.
 
 **Dashboard**: `http://localhost:20128/dashboard`
 
+### Memory limit
+
+The server process starts with a 6 GB V8 heap cap. On a memory-limited host
+(systemd `MemoryMax`, `docker --memory`, k8s limits) lower it so the garbage
+collector feels the limit before the kernel does:
+
+```bash
+NINEROUTER_MAX_OLD_SPACE_SIZE=384 9router   # cap the heap at 384 MB
+NINEROUTER_MAX_OLD_SPACE_SIZE=0 9router     # no cap — let node size it
+```
+
+`NODE_OPTIONS=--max-old-space-size=…` is honored too, and takes effect only
+because 9Router stops passing its own default when you set one.
+
 ---
 
 ## 🛠️ Supported CLI Tools
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -64,6 +64,7 @@ function createSpinner(text) {
 
 const pkg = require("./package.json");
 const { ensureSqliteRuntime, buildEnvWithRuntime } = require("./hooks/sqliteRuntime");
+const { resolveHeapFlags } = require("./hooks/nodeFlags");
 const { ensureTrayRuntime } = require("./hooks/trayRuntime");
 const args = process.argv.slice(2);
 
@@ -612,7 +613,7 @@ function startServer(updatePromise) {
   function spawnServer() {
     serverStartTime = Date.now();
     crashLog = [];
-    const child = spawn(RUNTIME, ["--dns-result-order=ipv4first", "--max-old-space-size=6144", serverPath], {
+    const child = spawn(RUNTIME, ["--dns-result-order=ipv4first", ...resolveHeapFlags(process.env), serverPath], {
       cwd: standaloneDir,
       stdio: showLog ? "inherit" : ["ignore", "ignore", "pipe"],
       detached: true,
--- a/cli/hooks/nodeFlags.js
+++ b/cli/hooks/nodeFlags.js
@@ -1,0 +1,47 @@
+/**
+ * Node flags for the spawned next-server child.
+ *
+ * Node reads NODE_OPTIONS first and lets command-line flags win, so a
+ * hard-coded --max-old-space-size on the spawn line silently overrides
+ * whatever the operator configured. Where a cgroup limit applies (systemd
+ * MemoryMax, docker --memory, k8s), the child then runs believing it has
+ * several GB of heap: GC never feels pressure, RSS climbs to the ceiling, and
+ * the kernel OOM-kills next-server mid-stream — taking in-flight streaming
+ * responses with it (#3365).
+ *
+ * The default stays as it was, for the desktop case it was raised for. It just
+ * steps aside once the operator has said what they want.
+ */
+
+const DEFAULT_MAX_OLD_SPACE_MB = 6144;
+
+// Node accepts the underscore spelling of V8 flags too (--max_old_space_size).
+const HEAP_FLAG_PATTERN = /(^|\s)--max[-_]old[-_]space[-_]size(=|\s|$)/;
+
+/**
+ * @param {NodeJS.ProcessEnv} env
+ * @returns {string[]} heap flags to pass to the child, possibly empty
+ */
+function resolveHeapFlags(env = process.env) {
+  const explicit = String(env.NINEROUTER_MAX_OLD_SPACE_SIZE ?? "").trim();
+  if (explicit) {
+    // 0 hands the decision back to node, which sizes the heap from the memory
+    // it can actually see — the right answer inside a container.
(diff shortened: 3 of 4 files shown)


ntdatt812:
Superseded by #3881, which carries this commit unchanged, rebased onto current `master` and grouped with the other fixes in the same area. Closing here so the two do not sit in the queue as duplicates — reopen if you would rather review it on its own.

PR #3369, opened 2026-08-16
fix(translator): recover a tool result that arrived without an id

### LINKED ISSUE #3362: Claude tool-result requests intermittently fail with missing tool_use_id (v0.5.55)
# Claude tool-result requests intermittently fail with missing `tool_use_id`

## Environment
- 9Router: v0.5.55 (`decolua/9router`)
- API: OpenAI-compatible `/v1/chat/completions`
- Client: OpenClaw and Hermes
- Model: `cc/claude-sonnet-5`

## Summary
Requests that contain prior Claude tool calls/results intermittently fail with HTTP 400:

```text
messages.0.content.0.tool_result.tool_use_id: Field required
```

The same route succeeds for plain conversational requests and some tool requests. On failure, OpenClaw falls back to Codex, causing unexpected Codex usage and breaking continuity.

## Observed evidence
OpenClaw live logs against `http://192.168.1.11:20128/v1/chat/completions`:

```text
provider=9router model=cc/claude-sonnet-5 status=400
[claude/claude-sonnet-5] [400]: messages.0.content.0.tool_result.tool_use_id: Field required
```

The error has also appeared at `messages.6.content.0.tool_result.tool_use_id`. Other requests to the same model returned HTTP 200.

## Likely cause
The OpenAI-compatible to Anthropic/message-content conversion path is dropping, renaming, or failing to map the original tool-call ID when serializing a `tool_result`. The upstream Anthropic-compatible payload consequently contains a tool result without a valid `tool_use_id`.

## Requested fix
1. Preserve the original tool-call ID end-to-end when converting assistant tool calls and subsequent tool results.
2. Validate that every emitted `tool_result` has a non-empty `tool_use_id`.
3. Validate that each ID matches a preceding assistant tool call in the same request.
4. Return a clear conversion error rather than emitting a malformed upstream request if repair is impossible.
5. Add regression tests covering multi-turn OpenAI-compatible tool calls, multiple tool calls in one assistant turn, streamed responses, and mixed text/tool content.
6. Add structured debug logging (with content and secrets redacted) showing tool-call/result ID mapping.

## Impact
- Claude tool-using turns fail intermittently.
- OpenClaw/Hermes invoke fallback models unnecessarily.
- Users experience generic agent errors and unexpected Codex token consumption.
- Any client using the OpenAI-compatible endpoint with Claude tool history may be affected.

## Reproduction shape
1. Send a request to `/v1/chat/completions` using `cc/claude-sonnet-5` with tools enabled.
2. Receive an assistant tool call.
3. Execute the tool.
4. Send the assistant/tool result history back in the next request.
5. Observe intermittent HTTP 400 indicating missing `tool_use_id`.

ntdatt812:
Closes #3362.

Requests carrying Claude tool history intermittently 400 upstream:

```
messages.0.content.0.tool_result.tool_use_id: Field required
```

The reporter's read — "the conversion path is dropping or failing to map the tool-call ID" — is correct. Here is the exact line.

## Root cause

`ensureToolCallIds` in `open-sse/translator/concerns/toolCall.js` repairs an assistant `tool_call` whose id is missing *or* malformed:

```js
if (!tc.id || !TOOL_ID_PATTERN.test(tc.id)) { … }   // handles absent
```

but both tool-**result** branches are guarded on the id already being truthy:

```js
if (msg.role === "tool" && msg.tool_call_id && !TOOL_ID_PATTERN.test(msg.tool_call_id)) { … }
if (block.type === "tool_result" && block.tool_use_id && !TOOL_ID_PATTERN.test(block.tool_use_id)) { … }
```

A result that arrives with **no id at all** therefore skips validation entirely, and `openai-to-claude.js:207` copies it straight through:

```js
blocks.push({ type: CLAUDE_BLOCK.TOOL_RESULT, tool_use_id: msg.tool_call_id, content: msg.content });
```

`tool_use_id: undefined` — exactly the field Anthropic reports as missing. The producer side of the id was hardened; the consumer side was not.

That also explains "intermittent": it depends entirely on whether the client dropped `tool_call_id` on that particular turn, which is why the same route succeeds for plain conversation and for tool turns where the id survived, and why the error appears at varying `messages.N`.

There is a second-order effect worth noting. `hasToolResults()` matches on the same field:

```js
if (msg.role === "tool" && msg.tool_call_id) return toolCallIds.includes(msg.tool_call_id);
```

With the id absent this returns false, so `fixMissingToolResponses` concluded the result was missing and spliced an empty duplicate in next to the real one.

## The fix

Sanitizing cannot invent an id, and generating a random one only trades one 400 for another — Anthropic equally rejects a `tool_result` whose id matches no `tool_use`. So the id is **recovered by pairing**: a result with no usable id takes the oldest tool call from the preceding assistant turn that nothing has answered yet.

That is the id upstream is expecting, and it holds up under the shapes the issue asks about:

- **Parallel tool calls** — results are paired in order.
- **Out-of-order results** — an id claimed by an explicit result is removed from the open set, so it is never handed to a second result.
- **Multi-turn** — an assistant turn replaces the open set, so a later result cannot pick up a stale id from an earlier turn.
- **Claude-shaped history** — `tool_result` blocks inside user content are repaired the same way as OpenAI `role: "tool"` messages.

Unchanged: a well-formed id passes through untouched, and a malformed one is still sanitized exactly as before.

When there is genuinely nothing to pair with — a result following no tool call at all — the id is generated rather than left absent. That request may still be rejected as unmatched, but a well-formed request that *might* be accepted beats one that is guaranteed to fail, and it is the same deterministic `generateToolCallId` the malformed branch already used.

The repair sits in the shared normalization pass that runs before translation, so every target consuming these ids benefits — claude, gemini, kiro, cursor — not just `openai-to-claude`. It also runs before `fixMissingToolResponses`, which is what stops the duplicate-insert above.

## On the rest of the requested fix list

Points 1–3 and 5 are covered. Point 4 (fail the conversion instead of emitting a malformed request) I did not implement: repair now succeeds for every shape that has a tool call to pair with, and turning the residual case into a hard error would break requests that currently work. Point 6 (structured id-mapping debug logging) is a separate change — this one is deliberately confined to the defect so it is easy to review and to revert.

## Verification

10 tests in `unit/tool-result-missing-id-3362.test.js`: the reporter's shape (assistant tool_call → `role: "tool"` with no id) resolves to the right id; parallel results pair in order; an id already claimed is not reissued, in both arrival orders; the Claude `tool_result` block shape is repaired; a valid id is left alone; a malformed id is still sanitized and still matches its call; an orphan result gets a well-formed id rather than `undefined`; ids are not reused across assistant turns; the duplicate empty result is no longer inserted; and an end-to-end pass through `openaiToClaudeRequest` shows a real `tool_use_id` on the emitted block.

Mutation-checked:

| mutation | result |
|---|---|
| restore the old truthy guard (return the raw id) | 8 of 10 fail |
| stop removing a claimed id from the open set | the reissue test fails |

The second mutation initially passed, which showed the test was too weak — the case only bites when the explicit result claims the *first* id. The test now covers both arrival orders.

Full suite (`npx vitest run unit translator`): **1809 passed / 95 failed**; against master's failing set the only difference is the network-dependent `unit/xai-oauth-service.test.js` timeouts, which flip between runs on an unmodified tree. The translator golden snapshots pass unchanged, which is the check that matters most here — this function runs on every request, whatever the target format. `npx eslint` reports no issues on the changed files.

### DIFF
--- a/open-sse/translator/concerns/toolCall.js
+++ b/open-sse/translator/concerns/toolCall.js
@@ -21,9 +21,37 @@ function sanitizeToolId(id) {
   return sanitized.length > 0 ? sanitized : null;
 }
 
+// Resolve the id a tool result should carry.
+//
+// A missing id cannot be sanitized into existence, and Anthropic rejects the
+// whole request when one is absent ("tool_result.tool_use_id: Field required").
+// Pair it instead with the oldest tool call from the preceding assistant turn
+// that nothing has answered yet — that is the id upstream is expecting, and
+// OpenAI-format clients that drop tool_call_id on a turn are the reason this
+// path exists (#3362). Only if there is nothing to pair with do we generate an
+// id, which at least keeps the request well-formed.
+function resolveToolResultId(rawId, unanswered, makeFallbackId) {
+  const usable =
+    typeof rawId === "string" && rawId
+      ? (TOOL_ID_PATTERN.test(rawId) ? rawId : sanitizeToolId(rawId))
+      : null;
+
+  if (usable) {
+    const at = unanswered.indexOf(usable);
+    if (at >= 0) unanswered.splice(at, 1);
+    return usable;
+  }
+
+  return unanswered.shift() || makeFallbackId();
+}
+
 // Ensure all tool_calls have valid id field and arguments is string (some providers require it)
 export function ensureToolCallIds(body) {
   if (!body.messages || !Array.isArray(body.messages)) return body;
+
+  // Tool call ids from the most recent assistant turn that no result has
+  // claimed yet, oldest first.
+  let unanswered = [];
 
   for (let i = 0; i < body.messages.length; i++) {
     const msg = body.messages[i];
@@ -45,13 +73,8 @@ export function ensureToolCallIds(body) {
       }
     }
 
-    // Validate tool_call_id in tool messages (role: "tool")
-    if (msg.role === "tool" && msg.tool_call_id && !TOOL_ID_PATTERN.test(msg.tool_call_id)) {
-      const sanitized = sanitizeToolId(msg.tool_call_id);
-      msg.tool_call_id = sanitized || generateToolCallId(i, 0);
-    }
-
-    // Also validate tool_use blocks in content (Claude format)
+    // Validate tool_use blocks in content (Claude format) before the ids are
+    // read back out below.
     if (Array.isArray(msg.content)) {
       for (let k = 0; k < msg.content.length; k++) {
         const block = msg.content[k];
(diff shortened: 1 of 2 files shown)


ntdatt812:

Issue #3293, opened 2026-08-14
Server-Side Request Forgery (SSRF) in /api/provider-nodes/validate via bypassable URL guard


HK4zCzi:
### Summary
The dashboard endpoint `POST /api/provider-nodes/validate` takes a user-controlled `baseUrl` and makes server-side HTTP requests to it. The SSRF protection is ineffective in two ways: (1) it is skipped entirely when the caller is treated as "local", and (2) for remote callers the `assertPublicUrl` denylist is string-based, does not resolve DNS, and does not normalize IPv4-mapped IPv6, so `http://[::ffff:7f00:1]` (= `127.0.0.1`) and `http://[::ffff:169.254.169.254]` reach internal hosts.

This allows an attacker to make the server issue requests to internal-only addresses (internal services, link-local cloud metadata, loopback). With validation `type: "custom-embedding"` there is bounded response disclosure: the first 200 bytes of the upstream response body are reflected back to the caller, but only when the upstream returns a non-2xx status that is not 401/403. With `type: "openai"`/`"anthropic-compatible"` no body is reflected, but the endpoint is still usable as a blind SSRF for internal port-scanning (status/timing/error-class differences). The attacker-supplied `apiKey` is forwarded to the internal target as an `Authorization: Bearer` header.

### Affected Product
| Field | Value |
|---|---|
| Ecosystem | npm |
| Package | `9router` |
| Affected versions | <= 0.5.8 |
| Patched versions | *(pending)* |

### Weaknesses
- CWE-184: Incomplete List of Disallowed Inputs
- CWE-200: Exposure of Sensitive Information to an Unauthorized Actor
- CWE-918: Server-Side Request Forgery (SSRF)

### Severity (CVSS v3.1)
- **Score:** Critical (9.1)
- **Vector:** `CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:C/C:H/I:L/A:L`

### Details

**File:** `src/app/api/provider-nodes/validate/route.js`

**1. Guard is skipped for "local" callers**, and "local" is derived from request-controlled signals (Host / `x-9r-real-ip`):
```js
// ~line 70
if (!isLocalRequest(request)) {
  try { assertPublicUrl(baseUrl); }
  catch { return NextResponse.json({ error: "URL not allowed" }, { status: 400 }); }
}
```
Any request considered "local" skips the guard entirely and the server will fetch any `baseUrl`.

[...]

PR #3370, opened 2026-08-16
fix(ssrf): judge IPv4-mapped IPv6 by address, not by spelling

ntdatt812:
Addresses the SSRF filter bypass in #3293.

**Read #3313 first — it was opened two days before this one and I missed it.** It fixes the same bypass and goes further. What is below is what I found independently; the part still worth taking from this PR is at the end.

## The bypass is real, and the existing fix is dead code

`isBlockedIpv6` already had a mapped-address check, added for GHSA-hj98-rc6w-m8cw:

```js
const v4Mapped = h.match(/^::ffff:(\d+\.\d+\.\d+\.\d+)$/);
if (v4Mapped) return isBlockedIpv4(v4Mapped[1]);
```

It can never match. WHATWG URL canonicalizes an embedded dotted quad to hextets **before** the guard sees the hostname:

```js
new URL("http://[::ffff:127.0.0.1]/").hostname       // "[::ffff:7f00:1]"
new URL("http://[::ffff:169.254.169.254]/").hostname // "[::ffff:a9fe:a9fe]"
```

The pattern was written against a string `assertPublicUrl` is never handed. Running master's guard directly:

```
http://[::ffff:127.0.0.1]/         ALLOWED
http://[::ffff:7f00:1]/            ALLOWED
http://[::ffff:169.254.169.254]/   ALLOWED
http://[::ffff:10.0.0.1]/          ALLOWED
http://[::ffff:192.168.1.1]/       ALLOWED
http://[::1]/                      blocked
http://127.0.0.1/                  blocked
```

And it is not merely a parsing curiosity — the address connects:

```
$ node -e "http server on 127.0.0.1:19998 returning 500 + a body"
[::ffff:7f00:1]    -> HTTP 500 | body: INTERNAL-ONLY-SECRET token=abc123
127.0.0.1          -> HTTP 500 | body: INTERNAL-ONLY-SECRET token=abc123
```

A 500 is exactly the case the `custom-embedding` branch reflects (non-2xx, not 401/403), so the first 200 bytes of an internal service's response come back to the caller.

**This matters beyond the fix itself: GHSA-hj98-rc6w-m8cw is recorded as fixed, and it was not.** Whichever PR lands, that advisory needs revisiting.

## The report's other claim no longer applies

The first point — that "local" is derived from request-controlled signals, so the guard is skipped — described the state before `92259214` (GHSA-pjm4-8fpg-f9p6), which shipped in v0.5.55. `isLocalRequest` now requires `isLoopbackPeer`, backed by the per-process token `custom-server.js` stamps on sanitized requests, so a spoofed `x-9r-real-ip` no longer buys locality. That skip is now a real trust decision rather than an attacker-controlled one, so I left it alone.

## Checked and clean

Worth recording, because these are the usual suspects and the URL parser already handles all of them:

```
http://127.1/             -> hostname 127.0.0.1   blocked
http://2130706433/        -> hostname 127.0.0.1   blocked
http://0177.0.0.1/        -> hostname 127.0.0.1   blocked
http://0x7f000001/        -> hostname 127.0.0.1   blocked
http://[0:0:0:0:0:0:0:1]/ -> hostname [::1]       blocked
```

Decimal, octal and hex IPv4 notation are not bypasses here.

## The fix

Parse the IPv6 literal to its 16 bytes and judge the address instead of matching text. `::ffff:a.b.c.d` and the v4-compatible `::a.b.c.d` go to the IPv4 range check, since that is the address the OS connects to. That subsumes the special cases: `::` and `::1` come out as `0.0.0.0` and `0.0.0.1`, both inside the existing `0.0.0.0/8` rule, so the spelling of the loopback literal stops mattering. A zone id (`fe80::1%eth0`) is stripped before parsing.

Two ranges widen correctly as a result: `fe80::/10` spans fe80–febf (the old test matched the literal prefix `"fe80:"` and missed `fe90::`, `feb0::`), and `fc00::/7` is checked on the address rather than an `"fc"`/`"fd"` string prefix.

A second commit widens the reserved list — CGNAT (100.64/10), 192.0.0/24, benchmarking (198.18/15), multicast (224/4), reserved (240/4, which covers 255.255.255.255) — and rejects non-http(s) schemes, since `assertPublicUrl` accepted `file://` and `gopher://` despite its name. One of the three call sites guarded the scheme itself; the guard should not depend on that. A trailing dot (`localhost.`) now normalizes away.

Public IPv6 is unaffected: Cloudflare's `2606:4700:4700::1111` and Google's `2001:4860:4860::8888` still pass.

## How this compares to #3313

#3313 adds a DNS-aware undici dispatcher that validates every resolved address at the socket, which also closes rebinding and redirect hops — strictly more than this PR does. I applied its source to master and ran **my 15 tests against it: all pass.** I went looking for a weakness in its address handling and did not find one; the scoped-literal case I expected to slip through (`fe80::1%eth0`) is blocked there too.

**So on the overlapping ground, take #3313.** Two things here do not overlap:

**1. The dead-code finding above.** #3313 replaces `isBlockedIpv6` wholesale without noting that the mapped check already present could never fire, so the advisory record stays wrong unless someone says it.

**2. Sink coverage.** Three call sites share this guard, and #3313 hardens one:

| sink | does 9router open the socket? | after #3313 |
|---|---|---|
| `POST /api/provider-nodes/validate` | yes | DNS-aware ✅ |
| web-search `baseUrl` override → `open-sse/handlers/search/index.js:103` | yes, plain `fetch` | string check only ⚠️ |
| `/v1/web/fetch` (`src/sse/handlers/fetch.js`) | **no** — the URL is passed as a body field to Firecrawl / Jina / Tavily / Exa | string check is the right level ✅ |

The third row is worth stating so nobody "hardens" it by reflex: there is no socket of ours to pin, so DNS validation would buy nothing there. The second is a real remaining gap and the natural follow-up once `fetchPublicUrl` from #3313 exists — I deliberately did not reimplement that dispatcher here rather than submit a second copy of someone else's design.

Note also that pinning a resolved IP (the approach already in `open-sse/translator/concerns/image.js`) is **not** a drop-in for the validate route: pinning plus following a redirect to a different host would connect to the wrong address, so #3313's per-connect lookup hook is the correct shape for a redirect-following caller. That is why this PR does not carry one.

## Still open either way

- **`fetchOidcDiscovery`** (`POST /api/auth/oidc/test`) has no guard at all. I did not add one: an enterprise OIDC issuer on an internal network is a legitimate configuration, so blanket-blocking private addresses could break working setups. That needs the maintainer's intent, not my guess.

## Verification

15 tests in `unit/ssrf-guard-ipv6-3293.test.js`: both spellings of a mapped loopback and of metadata, mapped RFC1918, the v4-compatible form, every spelling of `::1`, `fc00::/7`, the full `fe80::/10` range, a zone id, public IPv6 still allowed, the pre-existing IPv4 and hostname rules still blocking, the newly-added reserved blocks, non-http schemes, and a trailing dot. One test pins the canonicalization itself (`new URL(...).hostname === "[::ffff:7f00:1]"`), since that is the fact the old check got wrong.

**5 of them fail on unpatched master** — the tests reproduce the vulnerability, they do not merely describe it.

Full suite (`npx vitest run unit translator`): 1806 passed / 95 failed; against master's failing set the only difference is the network-dependent `unit/xai-oauth-service.test.js` timeouts, which flip between runs on an unmodified tree. `npx eslint` reports no issues on the changed files.

If a maintainer prefers, I am happy to redo this privately through a draft security advisory instead of a public PR — the bypass is already public in #3293, which is why I posted it here.

### DIFF
--- a/src/shared/utils/ssrfGuard.js
+++ b/src/shared/utils/ssrfGuard.js
@@ -21,10 +21,15 @@ function ipv4ToInt(host) {
 const BLOCKED_V4_RANGES = [
   [ipv4ToInt("0.0.0.0"), 8],
   [ipv4ToInt("10.0.0.0"), 8],
+  [ipv4ToInt("100.64.0.0"), 10],      // CGNAT
   [ipv4ToInt("127.0.0.0"), 8],
-  [ipv4ToInt("169.254.0.0"), 16],
+  [ipv4ToInt("169.254.0.0"), 16],     // link-local + cloud metadata
   [ipv4ToInt("172.16.0.0"), 12],
+  [ipv4ToInt("192.0.0.0"), 24],       // IETF protocol assignments
   [ipv4ToInt("192.168.0.0"), 16],
+  [ipv4ToInt("198.18.0.0"), 15],      // benchmarking
+  [ipv4ToInt("224.0.0.0"), 4],        // multicast
+  [ipv4ToInt("240.0.0.0"), 4],        // reserved, incl. 255.255.255.255
 ];
 
 function isBlockedIpv4(host) {
@@ -92,10 +97,18 @@ function isBlockedIpv6(host) {
 }
 
 // Throw if URL targets a non-public host. Caller should map to 400.
+//
+// This is a string check: it cannot see where a hostname resolves to. A caller
+// that then opens the connection itself needs a DNS-aware layer on top —
+// see the note on #3293.
 export function assertPublicUrl(rawUrl) {
   const parsed = new URL(rawUrl);
-  const host = parsed.hostname.toLowerCase();
+  if (parsed.protocol !== "http:" && parsed.protocol !== "https:") {
+    throw new Error("Blocked URL: unsupported protocol");
+  }
+  const host = parsed.hostname.toLowerCase().replace(/\.$/, "");
 
+  if (!host) throw new Error("Blocked URL: missing host");
   if (BLOCKED_HOSTNAMES.has(host)) throw new Error("Blocked URL: internal host");
   if (BLOCKED_SUFFIXES.some((s) => host.endsWith(s))) throw new Error("Blocked URL: internal host");
   if (isBlockedIpv4(host)) throw new Error("Blocked URL: private IP");
--- a/src/shared/utils/ssrfGuard.js
+++ b/src/shared/utils/ssrfGuard.js
@@ -36,12 +36,59 @@ function isBlockedIpv4(host) {
   });
 }
 
+// Expand an IPv6 literal to its 16 bytes, or null if it is not one.
+//
+// Matching on the text is not enough: WHATWG URL canonicalizes an embedded
+// dotted quad to hextets, so `new URL("http://[::ffff:127.0.0.1]")` reports its
+// hostname as `[::ffff:7f00:1]`. A pattern written against the dotted spelling
+// therefore never matches anything that came out of the parser. Normalize to
+// bytes once and judge the address itself.
+function ipv6ToBytes(host) {
+  let text = host;
+
+  // A trailing dotted quad (::ffff:127.0.0.1) is just another spelling of the
+  // two hextets it encodes — rewrite it so one parser covers both.
+  const dotted = text.match(/^(.*:)(\d{1,3}(?:\.\d{1,3}){3})$/);
+  if (dotted) {
+    const value = ipv4ToInt(dotted[2]);
+    if (value === null) return null;
+    text = `${dotted[1]}${(value >>> 16).toString(16)}:${(value & 0xffff).toString(16)}`;
+  }
+
+  const halves = text.split("::");
+  if (halves.length > 2) return null;
+  const head = halves[0] ? halves[0].split(":") : [];
+  const tail = halves.length === 2 && halves[1] ? halves[1].split(":") : [];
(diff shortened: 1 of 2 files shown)


ntdatt812:
Closing in favour of #3313, which was open two days before I started this and covers the same ground more thoroughly.

We found the same root cause independently — the guard's mapped-address check could never fire, because WHATWG URL rewrites the dotted quad to hextets before the guard sees the host (`new URL("http://[::ffff:127.0.0.1]/").hostname === "[::ffff:7f00:1]"`). But #3313 also blocks `fec0::/10`, `ff00::/8`, `2001:db8::/32` and the TEST-NET ranges, and adds the per-connect lookup hook, which this PR deliberately left out of scope. There is no reason to review two patches to one file.

I applied #3313 to master and ran the 15 cases from this PR's test file against it — all pass, including the scoped literal `fe80::1%eth0` I went looking for as a weakness and didn't find.

The findings that don't disappear with this PR are in my comment on #3313: the dead mapped-address check was added for GHSA-hj98-rc6w-m8cw, so that advisory is recorded as fixed and was not; decimal/octal/hex IPv4 forms are not bypasses; and the web-search `baseUrl` override at `open-sse/handlers/search/index.js:103` is still on the plain string guard while opening its own socket. I'll send that last one as a follow-up on top of #3313 once it lands.

PR #3393, opened 2026-08-17
fix(dashboard): prevent UI crash when API key is less than 8 characters

fasilu:
### Problem
In the provider connection cards (`TtsExampleCard`, `GenericExampleCard`, `SttExampleCard`), the dashboard masks API keys by showing the first 8 characters and repeating a mask character (`•`) for the remaining length:
`"•".repeat(apiKey.length - 8)`

If a developer enters a short mock key or a local testing key under 8 characters (e.g., `12345`), the calculation returns a negative number (e.g., `-3`), throwing a `RangeError: Invalid count value` in JavaScript. This crashes the entire dashboard UI render phase.

### Solution
Wrap the repeat count in `Math.max(0, ...)` to ensure the masking repeat count is never negative:
`"•".repeat(Math.min(20, Math.max(0, apiKey.length - 8)))`

### Impact
- Fixes the dashboard UI page crash when dealing with short API keys.
- Improves UI robustness for local test connections and custom configurations.

### Testing
- Verified that the test suite runs successfully and the changes introduce no new regressions.


decolua:
Thanks @fasilu for the contribution! Reviewed and merged into master. 🙏

PR #3395, opened 2026-08-17
feat(open-sse): support dynamic image aspect ratios for antigravity image generation

fasilu:
### Problem
Gemini/Antigravity image generation requests lacked support for dynamic aspect ratios mapping from standard request dimensions.

### Solution
- Intercept non-image model calls in the Antigravity image handler to map them to `gemini-3.1-flash-image`.
- Extract `body.size` parameters and resolve them to standard aspect ratios (e.g. `16:9` -> suffix `-16x9`).
- Append the aspect ratio suffix to the target model string to request the correct dimensions from the upstream provider.

### Impact
- Enables full support for requesting different aspect ratios during image generation on Antigravity models.

### Testing
- Verified that the test suite runs successfully and the changes introduce no new regressions.

### DIFF
--- a/open-sse/handlers/imageProviders/antigravity.js
+++ b/open-sse/handlers/imageProviders/antigravity.js
@@ -1,6 +1,6 @@
 // Antigravity image adapter - delegates to the executor for correct request
 // envelope (project, model, requestType, sessionId) and auth headers.
-import { nowSec } from "./_base.js";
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @fasilu for the contribution! Reviewed and merged into master. 🙏

PR #3405, opened 2026-08-18
fix(executor): handle CommandCode in-stream errors for combo and account fallback

berthojoris:
## Summary
- Add early inspection (`inspectAndWrapCommandCodeResponse`) to CommandCode streaming responses to detect in-stream errors before streaming tokens downstream.
- Map in-stream CommandCode errors (e.g. overloaded / 503, rate-limited / 429) to proper non-200 HTTP responses.
- Enables combo and account fallback mechanisms to properly trigger when upstream CommandCode returns an error payload inside an HTTP 200 stream.
- Replays buffered chunks losslessly for normal non-error streams into OpenAI SSE format.
- Add `parseError` implementation to `CommandCodeExecutor` and helper `parseCommandCodeError`.

## Test Plan
- [x] `tests/unit/commandcode-to-openai.test.js` (9 tests passed)
- [x] `tests/unit/commandcode-executor.test.js` (7 tests passed)
- [x] Baseline verifications passed:
  - `verify-alias.mjs`
  - `verify-oauth-urls.mjs`
  - `verify-providers.mjs`

### DIFF
--- a/open-sse/executors/commandcode.js
+++ b/open-sse/executors/commandcode.js
@@ -42,12 +42,235 @@ export class CommandCodeExecutor extends BaseExecutor {
   async execute(opts) {
     const result = await super.execute(opts);
     if (!result?.response?.ok || !result.response.body) return result;
-    result.response = wrapNdjsonAsOpenAISse(result.response, opts.model);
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @berthojoris for the contribution! Reviewed and merged into master. 🙏

PR #3417, opened 2026-08-19
feat(headroom): configure compression timeout

snower:
Add a `headroomTimeoutMs` configuration field, which sets the request timeout when sending a request to Headroom/v1/compress.

A fixed value of 3000 milliseconds may result in frequent timeouts in scenarios with high request volume or insufficient machine performance. Frequent timeouts may lead to inconsistent data submitted to the LLM, affecting caching performance. This timeout should be customizable to meet specific needs.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -57,7 +57,7 @@ export function stripContinuityFields(body) {
   return body;
 }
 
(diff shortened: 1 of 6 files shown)


decolua:
Thanks @snower for the contribution! Reviewed and merged into master. 🙏

PR #3419, opened 2026-08-19
fix(minimax): preserve images on matched OpenAI transport

### LINKED ISSUE #3418: MiniMax-M3 ignores images from OpenAI clients because matched transport format loses to targetFormat
## Summary

OpenAI-format clients sending image content to `minimax-cn/MiniMax-M3` (and `minimax/MiniMax-M3`) receive HTTP 200, but MiniMax responds as if no image was attached.

9Router correctly detects M3 as vision-capable and correctly selects the OpenAI runtime transport, but then translates the body to Claude format because the model-level `targetFormat: "claude"` has higher precedence than the matched transport format. The translated Claude body is posted to MiniMax's OpenAI `/v1/chat/completions` endpoint, so the image is silently ignored.

## Reproduction

Send an OpenAI multimodal request through 9Router:

```json
{
  "model": "minimax-cn/MiniMax-M3",
  "messages": [{
    "role": "user",
    "content": [
      {"type": "text", "text": "Read the exact text in this image."},
    ]
  }],
  "stream": false
}
```

Observed via 9Router: MiniMax says `I don't see any image attached`.

Control tests with the same image and key against MiniMax directly both work:

- `POST https://api.minimaxi.com/v1/chat/completions` with OpenAI `image_url` content: image recognized.
- `POST https://api.minimaxi.com/anthropic/v1/messages` with Anthropic `image` content: image recognized.

This rules out the image, credentials, MiniMax-M3 vision support, and both upstream APIs.

## Root cause

`open-sse/handlers/chatCore.js` currently resolves:

```js
const targetFormat = modelTargetFormat || useTransport?.format || getTargetFormat(...);
```

For an OpenAI M3 request:

- `sourceFormat = "openai"`
- `useTransport.format = "openai"`
- `modelTargetFormat = "claude"`
- selected endpoint = OpenAI `/v1/chat/completions`
- translated body format = Claude

The endpoint and body wire format diverge.

## Expected behavior

A source-format-matched, model-supported runtime transport should take precedence:

```js
const targetFormat = useTransport?.format || modelTargetFormat || getTargetFormat(...);
```

Then OpenAI images stay OpenAI `image_url` blocks on the OpenAI transport, while `targetFormat` remains the fallback when there is no compatible transport.

## Related work

- #2435 reports the same transport/body mismatch as a tool-schema 400.
- #2525 contains a broader fix but is currently conflicting/stale and has no CI checks. This issue isolates the still-reproducible vision data-loss path on current `master`.

turingcat:
## Problem

Hermes and other OpenAI-format clients can send a valid `image_url` block to `minimax-cn/MiniMax-M3` and receive HTTP 200, but M3 responds as if no image was attached.

9Router selects MiniMax's OpenAI runtime transport, then lets M3's model-level `targetFormat: "claude"` override that matched transport format. The resulting Claude-shaped body is posted to `/v1/chat/completions`, silently losing the image at the upstream boundary.

Fixes #3418. Also addresses the transport/body mismatch behind #2435 for current `master`.

## Root cause

Before:

```js
const targetFormat = modelTargetFormat || useTransport?.format || getTargetFormat(...);
```

For an OpenAI M3 vision request this combines:

- OpenAI endpoint selected by `runtimeTransport`
- Claude request body selected by `modelTargetFormat`

## Fix

Prefer a source-format-matched, model-supported runtime transport over the model fallback:

```js
const targetFormat = useTransport?.format || modelTargetFormat || getTargetFormat(...);
```

The existing `supportedFormats` guard remains authoritative. If the model does not support the client's source format, `useTransport` is null and the model-level target still wins.

## Regression test

Adds a chat-core test that verifies an OpenAI `image_url` block for `minimax-cn/MiniMax-M3`:

- resolves OpenAI -> OpenAI,
- reaches the executor unchanged,
- uses the OpenAI runtime transport,
- is not converted into a Claude `system`/message body.

The test fails on pre-fix `master` (`targetFormat` resolves to `claude`) and passes with this change.

## Verification

- `npx vitest run unit/minimax-transport-target-format.test.js unit/provider-models-minimax-m3.test.js unit/modality-strip.test.js unit/hermes-vision-detection.test.js --reporter=verbose` — 22/22 passed
- `npx eslint open-sse/handlers/chatCore.js tests/unit/minimax-transport-target-format.test.js` — passed
- `npm run build` — passed
- Direct controls using the same PNG and MiniMax CN credential succeeded on both MiniMax OpenAI and Anthropic APIs; the current installed 9Router reproduced the failure only through its routing layer.

## Scope

This intentionally does not include the reasoning-stream sanitization from stale/conflicting #2525. It isolates the transport precedence bug and image data loss with a current-master regression test.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -90,7 +90,12 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   // differ — kimi/glm only do /chat/completions). Undeclared models keep the
   // upstream default (use the transport), preserving behavior for glm/deepseek/...
   const useTransport = (!modelSupportedFormats || modelSupportedFormats.includes(sourceFormat)) ? runtimeTransport : null;
-  const targetFormat = modelTargetFormat || useTransport?.format || getTargetFormat(provider, credentials);
+  // A source-format-matched endpoint keeps the request lossless. Prefer it
+  // over a model-level targetFormat, which is only the fallback for clients
+  // whose wire format has no supported transport (for example MiniMax-M3:
+  // OpenAI clients should stay on /chat/completions; other clients can fall
+  // back to its declared Claude target).
+  const targetFormat = useTransport?.format || modelTargetFormat || getTargetFormat(provider, credentials);
   if (useTransport && credentials) credentials.runtimeTransport = useTransport;
   const stripList = getModelStrip(alias, model);
   const upstreamModel = getModelUpstreamId(alias, model);
--- a/tests/unit/minimax-transport-target-format.test.js
+++ b/tests/unit/minimax-transport-target-format.test.js
@@ -1,0 +1,189 @@
+/**
+ * Multi-transport providers must keep the request body and selected endpoint on
+ * the same wire format. MiniMax-M3 declares a Claude target for compatibility,
+ * but an OpenAI client should use MiniMax's matching OpenAI transport without
+ * an OpenAI -> Claude translation.
+ * Regression: https://github.com/decolua/9router/issues/3418
+ */
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const {
+  executeMock,
+  translateRequestMock,
+  handleNonStreamingResponseMock,
+} = vi.hoisted(() => ({
+  executeMock: vi.fn(),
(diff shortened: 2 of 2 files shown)


decolua:
Thanks @turingcat for the contribution! Reviewed and merged into master. 🙏

PR #3423, opened 2026-08-20
feat(open-sse): Revise Qwen3.8 pricing and add Meta Muse patterns and capabilities

Bonbocat:
Also added Step 3.7 multimodal


chen842081173:
Still missing in v0.5.65 (npm, released 2026-09-03): the built-in patterns only cover `*qwen*` / `*qwen3.5*` / `*qwen3.6*` / `*qwen3.7*` — no `*qwen3.8*` and no `*qwen38*`.

Real-world impact (self-hosted vLLM-Ascend, model id `Qwen3.8-27B`): any image request is **silently rerouted to a free fallback VLM** by the capacity adapter — the response `model` field changes to something else entirely, no error, nothing in logs. I only found it by diffing npm package `chunks/` between versions. Current workaround is patching the built chunk file after every `npm update`, which is not sustainable.

Please merge — this fixes #3568 as well, and the pattern gap affects every new Qwen minor release until it lands.

Bonbocat:
Resolved conflict:
My edit is more comprehensive (video input, etc) than upstream, prefer this edit (also added max output for Muse Spark while Muse Glimmer have conflicting information, it's left empty)

PR #3433, opened 2026-08-20
fix(responses): preserve usage in completed events

### LINKED ISSUE #3432: responsesTransformer drops usage: /v1/responses response.completed never includes usage object
## Summary

`/v1/responses` (Responses API) responses never include a `usage` object in the final `response.completed` event. Clients using the Responses API (e.g. pi-coding-agent / PiDeck) therefore see all-zero token counts and cannot report cache hit rates or cost.

## Root cause

In `open-sse/transformer/responsesTransformer.js`, `sendCompleted()` hardcodes the `response.completed` payload without usage:

```js
emit(controller, "response.completed", {
  type: "response.completed",
  response: {
    id: state.responseId,
    object: "response",
    created_at: state.created,
    status: "completed",
    background: false,
    error: null
  }
});
```

The transformer consumes upstream `chat.completions` SSE chunks (`parsed.usage`) but never reads `usage`. `grep -n usage open-sse/transformer/responsesTransformer.js` → 0 hits.

Meanwhile `open-sse/translator/response/openai-responses.js` (the reverse direction, Responses→Chat) DOES read `data.response.usage` (lines 504-513), so the gateway is clearly capable of propagating usage when it exists — it just never receives it on the outbound path.

`open-sse/transformer/streamToJsonConverter.js` also expects `parsed.response?.usage` (lines 32-35) but the upstream events never carry it, so non-streaming JSON responses are zeros too.

## Evidence

Upstream `POST /v1/chat/completions` (streaming) DOES return usage in the final chunk (like other OpenAI-compatible providers):

```
data: {"choices":[],"usage":{"prompt_tokens":884,"completion_tokens":37,"total_tokens":921,"prompt_tokens_details":{"cached_tokens":256}}}
```

But `POST /v1/responses` returns only:

```
event: response.completed
data: {"type":"response.completed","response":{"id":"resp_...","object":"response","created_at":...,"status":"completed","background":false,"error":null},"sequence_number":1}
```

No `usage` in any event.

## Expected

`response.completed.response.usage` should mirror the standard Responses API shape, e.g.:

```json
{
  "input_tokens": 884,
  "output_tokens": 37,
  "total_tokens": 921,
  "input_tokens_details": { "cached_tokens": 256 }
}
```

## Suggested fix

Capture `parsed.usage` from the last upstream chunk in the transformer state, then include it in `sendCompleted()`. ~10 lines. The upstream usage chunk with `choices: []` and `finish_reason` absent is the standard marker (cf. how `openai-responses.js` already handles `input_tokens_details.cached_tokens`).

cmondragon023:
## Summary

- capture usage-only Chat Completions SSE chunks before skipping empty choices
- map prompt, completion, total, cached, and reasoning token counts to the Responses API usage shape
- defer response.completed until stream flush so a trailing usage chunk is included
- add regression coverage for streaming and forced non-streaming Responses output

Fixes #3432

## Testing

- 2 focused regression tests passed
- 39 Responses-related tests passed
- node syntax checks passed for both changed files
- git diff --check passed

The full test command was also attempted: 1,018 tests passed and 101 skipped, but the checkout-wide run is not clean in this environment because multiple unrelated suites require missing optional/root dependencies, cloud fixtures, OS-specific snapshots, or localhost binding unavailable in the sandbox.

## AI assistance

The implementation and tests were prepared with Codex assistance and verified locally with the commands above.

### DIFF
--- a/open-sse/transformer/responsesTransformer.js
+++ b/open-sse/transformer/responsesTransformer.js
@@ -10,8 +10,8 @@ import path from "path";
 function toResponsesUsage(usage) {
   if (!usage || typeof usage !== "object") return null;
 
-  const inputTokens = Number.isFinite(usage.prompt_tokens) ? usage.prompt_tokens : 0;
-  const outputTokens = Number.isFinite(usage.completion_tokens) ? usage.completion_tokens : 0;
+  const inputTokens = [usage.input_tokens, usage.prompt_tokens].find(Number.isFinite) ?? 0;
+  const outputTokens = [usage.output_tokens, usage.completion_tokens].find(Number.isFinite) ?? 0;
   const responseUsage = {
     input_tokens: inputTokens,
     output_tokens: outputTokens,
@@ -19,8 +19,14 @@ function toResponsesUsage(usage) {
       ? usage.total_tokens
       : inputTokens + outputTokens
   };
-  const cachedTokens = usage.prompt_tokens_details?.cached_tokens;
-  const reasoningTokens = usage.completion_tokens_details?.reasoning_tokens;
+  const cachedTokens = [
+    usage.input_tokens_details?.cached_tokens,
+    usage.prompt_tokens_details?.cached_tokens
+  ].find(Number.isFinite);
+  const reasoningTokens = [
+    usage.output_tokens_details?.reasoning_tokens,
+    usage.completion_tokens_details?.reasoning_tokens
+  ].find(Number.isFinite);
 
   if (Number.isFinite(cachedTokens)) {
     responseUsage.input_tokens_details = { cached_tokens: cachedTokens };
--- a/open-sse/transformer/responsesTransformer.js
+++ b/open-sse/transformer/responsesTransformer.js
@@ -7,6 +7,31 @@
 import fs from "fs";
 import path from "path";
 
+function toResponsesUsage(usage) {
+  if (!usage || typeof usage !== "object") return null;
+
+  const inputTokens = Number.isFinite(usage.prompt_tokens) ? usage.prompt_tokens : 0;
+  const outputTokens = Number.isFinite(usage.completion_tokens) ? usage.completion_tokens : 0;
+  const responseUsage = {
+    input_tokens: inputTokens,
+    output_tokens: outputTokens,
+    total_tokens: Number.isFinite(usage.total_tokens)
+      ? usage.total_tokens
+      : inputTokens + outputTokens
+  };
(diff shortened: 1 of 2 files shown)


ntdat812:
I opened #3434 for the same issue 11 minutes after this one without spotting it — my mistake. I have closed mine in favour of this PR, which is first and maps `reasoning_tokens`, which mine did not.

Two things mine covered that this branch does not, offered as suggestions rather than requests. Happy to send either as a PR against your branch if useful.

**1. `input_tokens` fallback.** `toResponsesUsage()` reads only `prompt_tokens` / `completion_tokens`. When the upstream is itself Responses-native, the usage object already arrives in Responses shape and both come back `undefined`, so the mapper emits zeros:

```js
const inputTokens = Number.isFinite(usage.prompt_tokens) ? usage.prompt_tokens : 0;
```

Accepting either spelling keeps that case honest:

```js
const inputTokens = usage.input_tokens ?? usage.prompt_tokens ?? 0;
const outputTokens = usage.output_tokens ?? usage.completion_tokens ?? 0;
```

Same for `input_tokens_details.cached_tokens`, which some upstreams send instead of `prompt_tokens_details.cached_tokens`.

**2. Three regression cases.** Your two cover the happy path and the non-streaming body. The ones I would add, because each corresponds to a way the fix can silently half-work:

- **usage arriving after `finish_reason`** — this is the one that caught an incomplete version of my own fix. Capturing the usage-only chunk is not sufficient on its own: if `response.completed` is still emitted from the `finish_reason` branch, the trailing usage chunk arrives too late and is dropped. Your PR defers to flush and so handles it, but nothing currently fails if that deferral is reverted.
- **no cached tokens** — asserts `input_tokens_details` is absent rather than `{cached_tokens: 0}`, so the optional block cannot regress into always emitting.
- **upstream sends no usage at all** — asserts the stream still completes and simply omits `usage`, rather than emitting `usage: null` or stalling.

Test bodies are in the closed #3434 (`tests/unit/responses-usage-3432.test.js`) if you want to lift them; consider them yours, no attribution needed.

cmondragon023:
Thanks for the careful comparison. The latest push adds the Responses-native `input_tokens` / `output_tokens` and token-details fallbacks, plus explicit coverage for usage after `finish_reason`, no cached details, and no usage chunk. The focused usage suite now has 5 passing cases.

ntdat812:
Great — thanks for folding both in. That covers everything I had; nothing further from me on this one.

PR #3434, opened 2026-08-20
fix(responses): propagate upstream usage into response.completed (#3432)

### LINKED ISSUE #3432: responsesTransformer drops usage: /v1/responses response.completed never includes usage object

ntdat812:
Fixes #3432.

## Problem

`/v1/responses` never emitted a `usage` object on `response.completed`, so Responses API clients
(pi-coding-agent / PiDeck in the report) read all-zero token counts and could not report cost or
cache-hit rate. `streamToJsonConverter.js` already expects `parsed.response?.usage`, so non-streaming
JSON came back zeroed for the same reason.

## Root cause

Two things dropped it, and fixing only the first is not enough.

**1. The usage chunk is the one the loop discards.** With `include_usage` the upstream sends usage in
a trailing chunk whose `choices` array is empty:

```
data: {"choices":[],"usage":{"prompt_tokens":884,"completion_tokens":37,...}}
```

`if (!parsed.choices?.length) continue;` skips it before any usage handling could run — which is why
`grep -n usage responsesTransformer.js` returned nothing to hook into.

**2. That chunk arrives after finish_reason.** `sendCompleted()` fired from the `finish_reason`
branch, so even once usage is captured, the completed event had already been emitted and
`completedSent` latched.

## Fix

- Read `parsed.usage` into state **above** the `choices` guard.
- From the `finish_reason` branch, emit `response.completed` only when usage is already in hand.
  Otherwise leave it to `flush()`, which runs as soon as the upstream stream ends. A stalled upstream
  cannot hang here — that path is covered by the aborted-terminal synthesis in
  `open-sse/utils/responsesStreamHelpers.js` (`buildAbortedResponsesTerminalBytes`), which is
  independent of this transformer.
- Map to the Responses shape (`input_tokens` / `output_tokens` / `total_tokens` /
  `input_tokens_details.cached_tokens`), accepting either chat.completions or Responses field names —
  mirroring how the reverse direction reads it in `translator/response/openai-responses.js`.

Item-close events (`output_item.done`, reasoning, tool calls) still fire at `finish_reason` exactly
as before; only the terminal event moves, and only when usage has not arrived yet.

## Compatibility

A stream whose upstream reports no usage still emits `response.completed` with no `usage` key — the
field is spread in conditionally, so no client sees a new zero-filled object where there was
previously nothing. Pinned by a test.

## Tests

`tests/unit/responses-usage-3432.test.js` (new, 5 cases): usage before finish, usage after finish
(the real ordering), no cached tokens, an upstream already using Responses field names, and the
no-usage case. It fails **4/5 on master** and passes 5/5 here.

```
npx vitest run unit/responses-usage-3432.test.js                     5/5 pass
+ responses-abort-terminal, openai-responses-terminal-event,
  openai-responses-nonstream, openai-responses-multiturn,
  openai-responses-custom-tools, openai-responses-empty-toolcalls,
  headroom-responses-format, responses-prompt-cache-key-3216,
  github-responses-routing                                          42/42 pass (10 files)
npx eslint <both changed files>                                     clean
```

Full suite for comparison, same machine:

```
master                    93 failed | 1776 passed  (1945)
this branch               93 failed | 1781 passed  (1950)
```

Identical failure set — the 93 are pre-existing here (saml, db-benchmark, db-concurrent,
embeddings.cloud, kimchi, GOLDEN buildUrl, …), unrelated to this change. The delta is exactly the 5
tests this PR adds.

### DIFF
--- a/open-sse/transformer/responsesTransformer.js
+++ b/open-sse/transformer/responsesTransformer.js
@@ -73,7 +73,8 @@ export function createResponsesApiTransformStream(logger = null) {
     funcArgsDone: {},
     funcItemDone: {},
     buffer: "",
-    completedSent: false
+    completedSent: false,
+    usage: null
   };
 
   const encoder = new TextEncoder();
@@ -222,9 +223,28 @@ export function createResponsesApiTransformStream(logger = null) {
     }
   };
 
+  // chat.completions usage -> Responses API shape. Tolerates upstreams that
+  // already speak the Responses names, mirroring the reverse direction in
+  // translator/response/openai-responses.js.
+  const toResponsesUsage = (usage) => {
+    if (!usage || typeof usage !== "object") return null;
+    const inputTokens = usage.input_tokens ?? usage.prompt_tokens ?? 0;
+    const outputTokens = usage.output_tokens ?? usage.completion_tokens ?? 0;
+    const cachedTokens =
+      usage.input_tokens_details?.cached_tokens ?? usage.prompt_tokens_details?.cached_tokens ?? 0;
+    const result = {
+      input_tokens: inputTokens,
+      output_tokens: outputTokens,
+      total_tokens: usage.total_tokens ?? inputTokens + outputTokens
+    };
+    if (cachedTokens) result.input_tokens_details = { cached_tokens: cachedTokens };
+    return result;
+  };
+
   const sendCompleted = (controller) => {
     if (!state.completedSent) {
       state.completedSent = true;
+      const usage = toResponsesUsage(state.usage);
       emit(controller, "response.completed", {
         type: "response.completed",
         response: {
@@ -233,7 +253,8 @@ export function createResponsesApiTransformStream(logger = null) {
           created_at: state.created,
           status: "completed",
           background: false,
-          error: null
+          error: null,
+          ...(usage ? { usage } : {})
         }
       });
     }
@@ -263,6 +284,10 @@ export function createResponsesApiTransformStream(logger = null) {
         } catch {
           continue;
         }
+
+        // The final chat.completions chunk carries usage with an empty choices
+        // array, so this has to be read before the guard below drops it.
+        if (parsed.usage) state.usage = parsed.usage;
 
         if (!parsed.choices?.length) continue;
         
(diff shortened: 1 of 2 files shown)


ntdat812:
Closing this in favour of #3433, which was opened 11 minutes earlier and fixes the same defect the same way — capture the usage-only chunk before the empty-`choices` guard, and defer `response.completed` to flush so a trailing usage chunk still lands.

I missed it when I checked for duplicates; the two PRs were opened 14:31 and 14:42 the same day. Apologies for the extra review load.

#3433 is also more complete than this branch in one respect: it maps `completion_tokens_details.reasoning_tokens` to `output_tokens_details.reasoning_tokens`, which this PR does not.

I have left two notes on #3433 with the only things this branch covers that it does not — an `input_tokens` fallback and three extra regression cases — so nothing here is lost if you merge that one.

PR #3440, opened 2026-08-20
fix(codex): stop the prewarm turn coming back as a blank reply

some-du6e:
Codex was flakey against the gateway — sometimes a message just came back blank, sometimes it worked fine.

The cause turned out to be our own prewarm handling. Codex ≥0.147 opens a second websocket at session start tagged `request_kind: "prewarm"` and sends the full turn payload down it, so the upstream prompt cache is warm by the time the user hits enter. The bridge short-circuited that: it invented a `prewarm_<hex>` id and sent back `response.created` + `response.completed` with zero usage, routing nothing. The client doesn't discard that — it adopts it as the turn's completed response. So if you typed while that socket was still the current one, you got an empty bubble and the request never reached a provider at all.

Evidence from a live session: three turns today ended with `token_count in=0 out=0` in under a second, with no matching entry in the gateway log, and the client trace recorded `last_model_response_id="prewarm_fdecd2d22b9a300bb828bb1e"` — our synthetic id. Turns sent minutes later, after the prewarm socket had gone stale, routed and answered normally.

This drops the local stub and routes a prewarm connection through `bridgeCodexTurn` like any other turn, which is both the native behaviour and the thing that actually warms the cache the client is trying to warm. A one-line log records when a prewarm is routed so it can be confirmed from the container logs.

Tradeoff: one real model call per session start, billed to the account serving it. That is what ChatGPT itself does for the same request, and it is the cost of the prewarm being real rather than a lie.

Made by Opus 5 using Claude Code


some-du6e:
Opened against the wrong repo by mistake — refiling on my own fork.

PR #3454, opened 2026-08-21
fix(translator): sanitize 'opencode' mentions in system prompts for Gemini/Antigravity

AhooraZen:
## Summary
- Adds `sanitizeSystemPrompt()` in `open-sse/translator/request/openai-to-gemini.js` that replaces client-identifying `opencode` / `OpenCode` / `OPENCODE` strings with `antigravity` equivalents (case-preserving) inside **system prompt text only**.
- Applied at the 3 system-prompt injection points:
  - `openaiToGeminiBase` -> `systemInstruction.parts[].text`
  - `wrapInCloudCodeEnvelopeForClaude` -> array-form `claudeRequest.system` blocks
  - `wrapInCloudCodeEnvelopeForClaude` -> string-form `claudeRequest.system`

## Why
Gemini/Antigravity backends can reject or degrade requests whose system prompt identifies the calling CLI as a different vendor's client. Message contents and tool definitions are left untouched - this only rewrites the word inside system instructions.

## Testing
- `node --check` passes.
- Verified locally against built chunks (`app/.next-cli-build/server/chunks/8499.js`) with equivalent string-level patch; sanitizer output: `You are OpenCode...` -> `You are Antigravity...`, `opencode.ai` -> `antigravity.ai`, `OPENCODE` -> `ANTIGRAVITY`.

### DIFF
--- a/open-sse/translator/request/openai-to-gemini.js
+++ b/open-sse/translator/request/openai-to-gemini.js
@@ -45,6 +45,18 @@ function normalizeGeminiContents(contents) {
   return out;
 }
 
+// Replace client-identifying "opencode" mentions inside system prompts with
+// provider-neutral naming so Gemini/Antigravity backends don't reject
+// requests carrying another client's branding.
+function sanitizeSystemPrompt(text) {
+  if (typeof text !== "string") return text;
+  return text.replace(/opencode/gi, (match) => {
+    if (match === "OpenCode") return "Antigravity";
+    if (match === "OPENCODE") return "ANTIGRAVITY";
+    return "antigravity";
+  });
+}
+
 // Core: Convert OpenAI request to Gemini format (base for all variants)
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @AhooraZen for the contribution! Reviewed and merged into master. 🙏

One adjustment on merge: the sanitizer was moved from `open-sse/translator/request/openai-to-gemini.js` into the Antigravity executor. That translator is shared by five targets (gemini, gemini-cli, antigravity, vertex, zed), so rewriting there would also alter prompts for plain Gemini and Vertex requests, which don't need it.

It now lives as a config-driven rule table (`ANTIGRAVITY_PROMPT_REWRITES` in `appConstants.js`) alongside the existing Zed/Claude prompt rewrite, applied in `executors/antigravity.js`. Same case-preserving opencode → antigravity mapping you wrote, and it additionally covers the single-system-message branch that the original patch didn't reach.

PR #3456, opened 2026-08-21
feat(search+dashboard): ollama-search/zai-search providers, unified model cards, combo cooldown timers

askidmobile:
## Summary

Three related improvements, cherry-picked from our fork:

### 1. Web search providers: ollama-search + zai-search
- Register `ollama-search` (POST `ollama.com/api/web_search`) and `zai-search` (POST `api.z.ai/api/mcp/web_search_prime/mcp`) as search providers with request builders + response normalizers.
- New `credentialFallback` registry field: search providers reuse the API key of a related chat provider (`ollama-search -> ollama`, `zai-search -> zai`) when they have no own connection — one key serves both chat and search.
- New Claude-compatible `zai` LLM provider so a Z.AI Coding key has a home; dashboard reflects inherited keys on search provider cards ("Reuses N active keys from X").
- Logos for all three providers.

### 2. Combo cards: effective capabilities + cooldown timers
- Third row on each combo card shows the combined caps exactly as `/v1/models` advertises them (booleans AND across members, context/output MIN), driven by the same `getConservativeComboCapabilities` and honoring per-provider capability overrides.
- Per-model cooldown chips (orange, live countdown) from `/api/models/availability`; when every member is cooling down a red banner shows when the combo becomes usable.

### 3. Unified model cards across provider pages
- Same 3-row anatomy everywhere (name / routed id + copy / caps + ctx-out-price meta + CONFIGURED badge) with a right-hand action column (test / settings / delete).
- Settings icon opens the shared EditModelModal, which now has an "Import from models.dev" button for the model's provider.
- noAuth providers (e.g. OpenCode free): the models route accepts a registry provider id when the entry is noAuth and has a derivable models endpoint — Import Models works without any connection.
- Cards display effective (overridden) caps, trying both alias- and id-keyed overrides.

### Bonus fixes included
- Rate-limit reset parsing: GLM/Z.AI "reset at <datetime>", Retry-After, etc. now produce precise cooldowns (cap raised 30m to 6h to match its own comment).
- TokenRouter free-tier models: clamp reasoning_effort to per-model allowed set (qwen3.8-max-free: low/medium/xhigh).
- Alibaba Token Plan (alitp-intl): quota card from local counters against the documented 10k credits/7-day plan, models.dev mapping to `alibaba-token-plan`, correct qwen3.8 caps pattern.

## Testing

- Full unit suite unchanged vs baseline (84 known fails / 1466 pass).
- Live-verified on a production deployment: credential fallback reaches upstream, reset-time locks (6h), TokenRouter clamp (THINK:xhigh on wire), opencode noAuth import (64 models), alitp-intl quota payload + models.dev import (25 cap sets).

## Notes

- Conflicts with upstream's own glm-5.2/5.3 capability patterns were resolved in favor of upstream's entries.


decolua:
Thanks @askidmobile for the contribution! 🙏

I've merged the **web search** part of this PR into master: `ollama-search`, the Z.AI `web_search_prime` endpoint, and the `credentialFallback` mechanism. The Z.AI search ended up folded into the existing `glm` provider rather than a separate `zai-search` entry — credentials already live on the `glm` connection, and every other provider that does both chat and search (antigravity, kimi, xai, gemini) declares `webSearch` on the provider itself. The request builder and normalizers are essentially yours.

I did not take the rest of the PR. Reviewed against the current tree, several imports have no corresponding file or export on the branch:

- `open-sse/services/usage/cloudflare.js` — imported by `open-sse/services/usage.js`; the module fails to load, which breaks every usage route
- `src/shared/hooks/usePricing`, `src/shared/utils/modelMeta`, `./ImportModelsModal` — imported by the dashboard, not present
- API routes `/api/models-dev/import` and `/api/models/caps` — fetched by `EditModelModal`, not present
- `getCapsOverrides` — imported from `@/lib/db/index.js`, not exported there
- `getRegistryEntry` / `deriveModelsEndpoint` — used in `src/app/api/providers/[id]/models/route.js` without being imported or defined
- `useModelCaps` returns `overrides`, which is never declared in the hook
- `getOpenRouterUsage` reads `totalCredits` outside the block that declares it

It looks like the cherry-pick left behind files that exist only in your fork. There's real value in the rest — the stream fixes (duplicate finish chunks, `delta.reasoning` normalization) and the quota work especially — but as one 2200-line PR spanning eight concerns it's hard to land. Separate PRs, one concern each with the missing files included, would be much easier to review and merge.

Closing this one since the search feature is in. Thanks again!

askidmobile:
Thanks for merging the search part — and agreed, folding the Z.AI search into `glm` fits better than a separate entry.

On the "missing files": they're not gone — they all live in our three still-open PRs (#3032, #3345, #3346, opened Aug 14–15). #3456 was cherry-picked on top of that stack, so its commits reference files those older PRs introduce (`usage/cloudflare.js`, the modelsDev lib, `ImportModelsModal`, `/api/models-dev/*`, `/api/models/caps`, `capsRepo.getCapsOverrides`, `deriveModelsEndpoint`, the `useModelCaps` overrides). Reviewed against bare master, the imports dangle — that's on us for stacking. The `totalCredits` scoping and `overrides` issues are real bugs from that pick; both are fixed in the series below.

Fair point on size too. We're re-cutting everything as one-concern, self-contained PRs off current master — each includes its own files and passes the test suite at the master baseline. The two you flagged as most valuable are up first:

- #3611 — stream fixes: dedup finish chunks/`[DONE]`, `delta.reasoning` → `reasoning_content`, newline before `[DONE]`, absent `stream` field means non-streaming, abort+fallback on `native_finish_reason` errors
- #3612 — rate-limit reset parsing: provider-reported resets (GLM "reset at …", Retry-After), 6h cap

Next in the queue: quota work (ollama exact reset dates, openrouter quota card/balance/402 lock, cloudflare neuron usage, alitp-intl plan) → models.dev metadata + caps (the home of most of those "missing" files) → kiro/cloudflare catalogs → UI-configured models → dashboard/combo cards last (depends on models.dev).

For #3032/#3345/#3346 — want us to close them as superseded by this series, or leave them open until the replacements land?

PR #3465, opened 2026-08-22
fix(stream): fallback to next combo account on empty/null content responses

h26v:
## Summary
- Upstream providers occasionally return HTTP 200 with no usable content (null message, no `tool_calls`, no reasoning) instead of an error. Both the non-streaming and streaming chat paths treated any 2xx as unconditional success, so combo/account fallback never triggered and the same silently-broken account kept getting retried.
- Reuses the existing `markAccountUnavailable` / `modelLock_${model}` cooldown machinery instead of adding new state.
- Non-streaming: content is checked before the handler returns success; empty content is now turned into a 502 that flows through the standard fallback path.
- Streaming: content is not known until the stream completes, so emptiness is detected in `onStreamComplete` and the account is locked for the *next* request via a new `onEmptyStream` callback (7 minute cooldown, matches the existing cooldown pattern in `errorConfig.js`).
- Heuristic checks `tool_calls`/`content`/`reasoning_content` (non-streaming) and `completion_tokens`/`output_tokens` specifically (streaming) so legitimate tool-call-only or reasoning-only responses are not false-flagged as empty.

## Test plan
- [x] `npx vitest run` on the full suite no regressions introduced (verified via `git stash`/`git stash pop` clean-checkout comparison against the pre-existing failure baseline)
- [x] Maintainer smoke test against a real flaky provider (this was fixed based on local reproduction with a 9Router gateway seeing intermittent null-content 200s from a provider)

### DIFF
--- a/open-sse/config/errorConfig.js
+++ b/open-sse/config/errorConfig.js
@@ -38,6 +38,13 @@ export const BACKOFF_CONFIG = {
 // Default cooldown for transient/unknown errors
 export const TRANSIENT_COOLDOWN_MS = 30 * 1000;
 
+// Cooldown after a provider returns HTTP 200 with no usable content (null/empty
+// message, no tool_calls, no reasoning). Upstream reports success but produced
+// nothing — lock the model on this account for a few minutes so the combo/account
+// fallback loop skips straight to the next candidate instead of retrying a backend
+// that just failed silently, then automatically retries it once the lock expires.
+export const EMPTY_CONTENT_COOLDOWN_MS = 7 * 60 * 1000;
+
 // Hard cap for provider-reported rate limit cooldown (e.g. codex resets_at can be 5-6h)
 export const MAX_RATE_LIMIT_COOLDOWN_MS = 30 * 60 * 1000;
 
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -57,7 +57,7 @@ export function stripContinuityFields(body) {
   return body;
 }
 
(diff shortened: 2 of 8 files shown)


h26v:
Deployed this fix to my local 9Router instance (v0.5.55) and started a live smoke test: resuming a ~30-sample malware-analysis pipeline (HyperAgent) that drives real chat completions through this gateway across many providers/combos over several hours.

Will update this thread with results, specifically watching for recurrence of the HTTP-200-but-empty-content symptom (0 input/0 output tokens on a fresh, non-cached run) that this PR addresses.

h26v:
## Update: found and fixed 3 more bugs while smoke-testing this

During the smoke test I also happened to be adding a new custom `openai-compatible` connection (`apiType: "responses"`) to my own instance, and every non-streaming Claude/OpenAI-chat request against it started returning `502 Empty response content` â€” a regression surface this PR exposed (the previous behavior silently returned the empty response as "success").

Root-caused with a live upstream (`beta.singularityapi.tech`, a third-party Responses-API-compatible backend) â€” **not a false positive from this PRs empty-check**, but three real, separate pre-existing bugs the empty-check happened to make visible:

1. `translateNonStreamingResponse` never had a branch for "provider responded in Responses-API shape (`output: []`), client wants Chat/Claude format" â€” fell through untouched, producing an unusable body.
2. `handleNonStreamingResponse` only knows a provider forces SSE via the static `PROVIDERS[provider].forceStream` registry flag. A dynamically-added custom connection isnt in that registry, so when its upstream forced SSE anyway (despite `stream:false`), the response fell into the chat-only `parseSSEToOpenAIResponse` parser, which cant read Responses-style SSE events.
3. `convertResponsesStreamToJson` (the correct Responses-aware converter, already used elsewhere) required both an `event:` line and a `data:` line per SSE message. This upstream sends **data-only SSE** (`type` embedded in the JSON payload, no `event:` line) â€” a legal, common variant â€” which was silently dropping every message.

All three fixed in the last 2 commits, each with regression tests. Verified end-to-end against the live connection: `502` before â†’ `200` with real content after, for both the direct repro and through 9Router.

Streaming was unaffected throughout (confirmed separately) â€” this only hit the non-streaming JSON path.

## Smoke test status
Resumed the ~30-sample pipeline against the patched build (deployed locally, not published). 9/30 samples through so far, most completing normally. Two samples hit a dropped connection / retry exhaustion â€” expected collateral from me restarting the local service several times mid-run while iterating on the fixes above, not a fallback-logic issue; will retry those two once the run finishes. No recurrence of the original null/empty-content symptom this PR targets.

PR #3466, opened 2026-08-22
fix(claude): decloak tool names in same-format streaming passthrough

huohua-dev:
## Problem

When a Claude-format client (e.g. any `/v1/messages` consumer) talks to an OAuth-cloaked Claude provider (`cloakToolsOnOAuth`), **every tool call fails on the client side** with errors like:

```
Error: unknown tool "run_code_ide"
```

Root cause: `translateRequest()` cloaks client tool names with `CLAUDE_TOOL_SUFFIX` (anti-ban) whenever the provider is flagged `cloakToolsOnOAuth` — even when the request needs **no format conversion** (claude → claude). The two response paths are asymmetric:

- **Non-streaming**: `handleNonStreamingResponse` calls `decloakToolNames()` → names restored ✅
- **Streaming**: `translateResponse()` short-circuits on `sourceFormat === targetFormat` and returns chunks **untouched** → the client receives `tool_use` blocks still carrying the cloaked name (`run_code_ide`) and rejects them ❌

## Reproduction

Against a Claude OAuth connection, send the same `/v1/messages` request with one tool named `run_code`:

- `"stream": false` → response contains `tool_use` named `run_code` ✅
- `"stream": true` → `content_block_start` event contains `tool_use` named `run_code_ide` ❌

Any tool name is affected (`mytool` → `mytool_ide`), confirming it is the cloak path, not model behavior.

## Fix

Add `decloakStreamChunk()` — the streaming counterpart of `decloakToolNames()` — and call it on `translateResponse()`'s same-format branch:

```js
if (sourceFormat === targetFormat) {
  return [decloakStreamChunk(chunk, state?.toolNameMap)];
}
```

In a Claude SSE stream a tool name appears exactly once per call (the `content_block_start` event of a `tool_use` block), so that single event is the only thing that needs rewriting. Names not present in the cloak map — e.g. a CC decoy tool the model called anyway — pass through unchanged, matching `decloakToolNames()` semantics. The `state.toolNameMap` plumbing already existed (`chatCore` threads `_toolNameMap` into the stream state); it was simply never consulted on this path.

Also fixes two stale comments that still described the suffix as `_cc` (it is `_ide` via `CLAUDE_TOOL_SUFFIX`).

## Tests

- `tests/unit/claude-cloaking.test.js`: 5 new cases for `decloakStreamChunk()` (restore, no-mutation, unknown names, non-tool events, null/map-less tolerance)
- `tests/translator/claude-claude-stream-decloak.test.js`: 4 regression cases for the claude→claude passthrough
- Both new suites **fail without the fix** (verified by stashing the source change) and pass with it; full-suite failure set is otherwise identical to master (pre-existing unrelated failures only).

### DIFF
--- a/open-sse/translator/index.js
+++ b/open-sse/translator/index.js
@@ -1,7 +1,7 @@
 import { FORMATS } from "./formats.js";
 import { ensureToolCallIds, fixMissingToolResponses } from "./concerns/toolCall.js";
 import { prepareClaudeRequest } from "./formats/claude.js";
-import { cloakClaudeTools } from "../utils/claudeCloaking.js";
+import { cloakClaudeTools, decloakStreamChunk } from "../utils/claudeCloaking.js";
 import { filterToOpenAIFormat } from "./formats/openai.js";
 import { normalizeThinkingConfig } from "../services/provider.js";
 import { applyThinking, captureThinking } from "./concerns/thinkingUnified.js";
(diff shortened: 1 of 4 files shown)


decolua:
Thanks @huohua-dev for the contribution! Reviewed and merged into master. 🙏

PR #3479, opened 2026-08-23
fix(translator): zai thinkingFormat sends reasoning.effort object

### LINKED ISSUE #2721: zai thinkingFormat drops reasoning_effort — GLM-5.2 always runs at max
## Bug

For models using the `zai` thinkingFormat (e.g. `glm-5.2`, `glm-4.7`, `glm-4`, matched via the `*glm-5*` / `*glm-4*` / `*glm*` patterns), the client-selected thinking level is silently dropped before reaching the upstream. GLM-5.2 always runs at its default (`max`), regardless of the `reasoning_effort` the client sent.

## Root cause

`open-sse/translator/concerns/thinkingUnified.js`, `zai` case (line 261):

```js
case "zai": {
  // Z.ai ignores thinking.disabled → must use enable_thinking:false to turn off.
  if (none && canDisable) { body.enable_thinking = false; delete body.thinking; break; }
  body.thinking = { type: "enabled" };   // ← only the on/off toggle, no level
  break;
}
```

It sends only `thinking: {type:"enabled"}` and never sets `reasoning_effort`. Every other thinking format in this switch preserves the level:

- `qwen` → `thinking_budget`
- `deepseek` → `reasoning_effort` (high/max)
- `kimi` → `reasoning_effort` (low/high/max via `toKimiReasoningEffort`)
- `hunyuan` → `budget_tokens`
- `step` → `reasoning_effort`

Only `zai` discards it.

## Authoritative GLM-5.2 behavior

From the official z.ai migration guide (https://docs.z.ai/guides/overview/migrate-to-glm-new):

> Deep thinking: Enable or disable `thinking={ type: \"enabled\" }` as needed for complex reasoning/coding
> Control reasoning effort: Configure `reasoning_effort` to decide between `high` (enhanced reasoning) or `max` (deep reasoning, default)

So GLM-5.2 uses **two** fields: `thinking:{type:"enabled"}` for on/off (9router does this correctly), and `reasoning_effort` for the level (`high` | `max`, default `max`). Omitting `reasoning_effort` → upstream defaults to `max`.

## Repro

```bash
curl -s http://localhost:20128/v1/chat/completions \
  -d "{\"model\":\"glm-5.2\",\"stream\":false,\"max_tokens\":24,\"reasoning_effort\":\"high\",\"messages\":[{\"role\":\"user\",\"content\":\"say ok\"}]}"
```

Response shows `reasoning_tokens` consumed regardless of the `reasoning_effort` value sent — sending `high`, `low`, `medium`, `xhigh` all produce the same upstream behavior (GLM running at `max`), because `reasoning_effort` is never forwarded.

Verified on v0.5.35 and current master (`0513bf3`) — the `zai` case is unchanged.

## Suggested fix

GLM-5.2 only accepts `high` | `max` (plus `enable_thinking=false` to disable). Map the unified level onto those two values:

```js
case "zai": {
  if (none && canDisable) { body.enable_thinking = false; delete body.thinking; break; }
  body.thinking = { type: "enabled" };
  const lvl = toLevel(eff);
  // GLM only accepts high|max; default max. minimal/low/medium → high, high → high, xhigh/max → max.
  body.reasoning_effort = (lvl === "high") ? "high" : "max";
  break;
}
```

(`auto` and unset → `toLevel` returns null → falls to `max`, which matches GLM's default.)

## Scope

Affects every model matched by the `*glm-5*`, `*glm-4.7*`, `*glm-4*`, `*glm*` patterns (all `thinkingFormat: "zai"`), when routed through an OpenAI-compatible upstream that honors `reasoning_effort`. Not provider-specific to z.ai direct — also affects OpenAI-compatible gateways fronting GLM (e.g. aliyun tokenplan, codebuddy-cn uses `openai` format so unaffected).

Spoon94:
## Summary
- Z.ai / GLM-5.2 / GLM-5.3 ignore `thinking:{type:\"enabled\"}` and require a nested `reasoning:{effort}` object. The zai branch previously only set `thinking` and dropped the client's `reasoning_effort`, so every GLM-5.x request ran at the model's default (max).
- Check `reasoning_effort` / `reasoning.effort` before the `thinking` object in `extractThinking` so a client-supplied effort is not overwritten by `thinking:{type:\"enabled\"}` mapping to `mode:auto`.

Fixes #2721

## Test plan
- [ ] Send `reasoning_effort: \"high\"` to a glm-5.2 / glm-5.3 combo; confirm 9router console shows `THINK:high` (not `THINK:auto`)
- [ ] Confirm the upstream request body contains both `thinking:{type:\"enabled\"}` and `reasoning:{effort:\"high\"}`
- [ ] Confirm `reasoning_effort: \"minimal\"` / `\"low\"` / `\"medium\"` map to the corresponding `reasoning.effort` value
- [ ] Confirm `reasoning_effort: \"xhigh\"` / `\"max\"` map to `reasoning.effort: \"high\"` (zai's max)
- [ ] Confirm `thinking:{type:\"disabled\"}` still disables thinking when `thinkingCanDisable` is true

### DIFF
--- a/open-sse/translator/concerns/thinkingUnified.js
+++ b/open-sse/translator/concerns/thinkingUnified.js
@@ -58,6 +58,15 @@ export function extractThinking(body) {
     return { mode: "level", level: e };
   }
 
+  // OpenAI chat / Responses shape — check effort first (zai sends both thinking object and reasoning.effort)
+  const effort = body.reasoning_effort ?? (typeof body.reasoning === "object" ? body.reasoning?.effort : null);
+  if (typeof effort === "string" && effort) {
+    const e = effort.toLowerCase();
+    if (e === "none" || e === "off") return { mode: "none" };
+    if (e === "auto") return { mode: "auto" };
+    return { mode: "level", level: e };
+  }
+
   // Claude shape
   const t = body.thinking;
   if (t && typeof t === "object") {
@@ -67,15 +76,6 @@ export function extractThinking(body) {
       if (Number.isFinite(budget) && budget > 0) return { mode: "budget", budget };
       return { mode: "auto" };
     }
-  }
-
-  // OpenAI chat / Responses shape
-  const effort = body.reasoning_effort ?? (typeof body.reasoning === "object" ? body.reasoning?.effort : null);
-  if (typeof effort === "string" && effort) {
-    const e = effort.toLowerCase();
-    if (e === "none" || e === "off") return { mode: "none" };
-    if (e === "auto") return { mode: "auto" };
-    return { mode: "level", level: e };
   }
 
   // Gemini shape (top-level, generationConfig, or request envelope)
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @Spoon94 for the contribution! Reviewed and merged into master. 🙏

Note: adjusted the reasoning_effort mapping slightly before merging — GLM-5.3 only accepts `low`/`high`/`max` per z.ai's docs (any other value errors), and the previous mapping never sent `low` (it collapsed everything but `high` to `max`). Also gated the field to GLM-5.2+ only, since older GLM (4.x, 5.0, 5.1, 5-turbo, 5v-turbo) do not read it.

PR #3484, opened 2026-08-23
fix(tools): default Claude tool type when missing

### LINKED ISSUE #2195: [Bug] HTTP 400 from strict Anthropic-compatible gateways (MiniMax) when Claude tools omit `type` field

warelik:
Fixes #2195

## Summary

Strict Anthropic-compatible gateways reject Claude-format requests with HTTP 400 when entries in `tools` omit the required `type` discriminator. This PR normalizes each missing or falsy `tools[].type` to `"custom"` before dispatch when the final request format is Claude.

This is a general Claude schema fix, not a MiniMax-M3 transport fix. It does not change or address the transport behavior tracked by #2525/#2533.

PR #2196 was reviewed and accepted, but it was closed without being merged; `defaultClaudeToolType` is absent from current `master` (`699edac3`). This PR is a fresh replacement from current `master`; #2196 is not reopened.

## Changes

- Export `defaultClaudeToolType(tools)` from `open-sse/translator/concerns/toolCall.js`.
- Add `type: "custom"` using `{ ...tool, type: "custom" }` only when `type` is falsy.
- Return tools with truthy `type` by reference without cloning.
- Apply normalization to Claude-format final bodies after the TTS special case and before token-saver processing.
- Add `open-sse/utils/claudeToolTypeSelfCheck.mjs` with no-framework assertions for missing, falsy, truthy, mixed, non-array, empty-array, non-mutation, and null-entry cases.

## Verification

```text
ok - Tool without type property defaults to custom
ok - Tool with type:null defaults to custom
ok - Tool with type:undefined defaults to custom
ok - Tool with type:empty-string defaults to custom
ok - Built-in tool (computer_use) passed through
ok - Tool already with type:custom passed through
ok - Mixed: built-in kept, function tool defaulted
ok - Non-array input returned unchanged
ok - Empty array returns empty array
ok - Original tools not mutated by reference
ok - Array with null entry defaults to custom

11/11 checks passed
```

Run with `node open-sse/utils/claudeToolTypeSelfCheck.mjs`.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -28,6 +28,7 @@ import { compressWithPxpipe } from "../rtk/pxpipe.js";
 import { getCapabilitiesForModel } from "../providers/capabilities.js";
 import { stripUnsupportedModalities } from "../translator/concerns/modality.js";
 import { prefetchRemoteImages } from "../translator/concerns/prefetch.js";
+import { defaultClaudeToolType } from "../translator/concerns/toolCall.js";
 import { resolveSessionId } from "../utils/sessionManager.js";
 
 /**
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @warelik for the contribution! Reviewed and merged into master. 🙏

PR #3491, opened 2026-08-23
fix(rtk): make system prompt injection format-safe and idempotent

### LINKED ISSUE #3202: CAVEMAN injects `input_text` into array-content system message → StepFun 400 "Unrecognized chat message"

KunN-21:
## Summary  - make Caveman/Ponytail system injection match each target wire format - preserve Responses non-message items and Kiro mirrored system-prefix semantics - make injection exact-idempotent and fail-open across retries and malformed bodies  ## Behavior  - Chat arrays append `text`; Responses arrays append `input_text` and create typed message items - Claude inserts before the final cache-control block; Gemini preserves snake/camel wrappers - Kiro updates both `systemPrompt` and its mirrored first-user prefix atomically - full-prompt delimiter-aware dedup allows distinct prompts sharing long prefixes - saver order and `X-9Router-Token-Saver: off` behavior stay unchanged  Supersedes the injection behavior covered separately by #2443, #2657, #3204, #3245, and #3361. This PR does not close those PRs automatically.  ## Test plan  - 47 focused system-injection tests - 165 targeted tests across RTK, Kiro, Caveman, and translator coverage - `verify-no-regression.mjs`: 0 new failures - ESLint changed files - GitNexus impact acknowledged as HIGH before edit; final diff maps to 2 files and 22 symbols  ## Scope  Exactly two files:  - `open-sse/rtk/systemInject.js` - `tests/unit/system-inject.test.js` 

Fixes #3202.

### DIFF
--- a/open-sse/rtk/systemInject.js
+++ b/open-sse/rtk/systemInject.js
@@ -3,96 +3,335 @@
 // native-passthrough flows. Used by caveman.js and ponytail.js.
 
 import { FORMATS } from "../translator/formats.js";
+import { OPENAI_BLOCK, CLAUDE_BLOCK, RESPONSES_ITEM } from "../translator/schema/blocks.js";
+import { ROLE } from "../translator/schema/roles.js";
 
 const SEP = "\n\n";
 
 export function injectSystemPrompt(body, format, prompt) {
-  if (!body || !prompt) return;
+  try {
+    if (!body || !prompt) return;
+    if (typeof body !== "object") return;
 
-  switch (format) {
-    case FORMATS.CLAUDE:
-      injectClaudeSystem(body, prompt);
+    // Kiro wire shape is unique (conversationState/systemPrompt) — handle directly.
+    if (isKiroBody(body) || format === FORMATS.KIRO) {
+      injectKiroSystem(body, prompt);
       return;
-    case FORMATS.GEMINI:
-    case FORMATS.GEMINI_CLI:
-    case FORMATS.VERTEX:
-    case FORMATS.ANTIGRAVITY:
-      // Antigravity wraps Gemini shape in body.request → injectGeminiSystem handles it
-      injectGeminiSystem(body, prompt);
-      return;
-    default:
-      // OpenAI and OpenAI-shaped formats (responses/codex/cursor/kiro/ollama)
-      injectMessagesSystem(body, prompt);
-  }
-}
-
-// OpenAI-shaped: messages[] (chat) or input[] (responses) or instructions (responses string)
-function injectMessagesSystem(body, prompt) {
-  // OpenAI Responses API: top-level string field
-  if (typeof body.instructions === "string") {
-    body.instructions = body.instructions
-      ? `${body.instructions}${SEP}${prompt}`
-      : prompt;
-    return;
-  }
-
-  const arr = Array.isArray(body.messages) ? body.messages
-    : Array.isArray(body.input) ? body.input
-    : null;
-  if (!arr) return;
-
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @KunN-21 for the contribution! Reviewed and merged into master. 🙏

One adjustment was made during review: the wire-shape dispatch matched `messages[]` before the format label, so Claude bodies (which carry both `system` and `messages[]`) fell into the Chat branch and got a `{role: "system"}` turn unshifted at index 0. The Anthropic Messages API has no `system` input role — system content belongs in the top-level `system` field — so this would have returned a 400 whenever Caveman/Ponytail ran against a Claude target.

Fix: `FORMATS.CLAUDE` and the Gemini family are now dispatched by format label before the shape sniff, and the trailing `switch (format)` was dropped as dead code. Two regression tests cover a Claude body carrying `messages[]`, both with and without an existing `system` field.

Everything else landed as submitted — Responses typing, Kiro atomic mirroring, delimiter-aware dedup, and the fail-open guards. 49 tests pass, ESLint clean, no new failures against the baseline.

Issue #2382, opened 2026-07-05
Bug: Non-streaming OpenAI requests return 0 tokens and missing from Usage Overview (due to forceStream)


letinhminh:
### Description
When making non-streaming chat completion requests using the `openai` provider, the API returns the completion successfully, but token usage is recorded as `0` in **Usage / Details** and is entirely missing from **Usage / Overview**.

### Root Cause
1. The `openai` provider has `forceStream: true` by default, which converts all non-streaming requests into streaming requests.
2. During this conversion, `transformRequest` in the base `openai` class (`s` in `src/shared/services/...` / `chunks/318.js`) does not add `stream: true` to the request body sent upstream, only the `Accept: text/event-stream` header.
3. Furthermore, it does not inject `stream_options: { include_usage: true }` for streaming requests.
4. As a result, the upstream OpenAI API either ignores the stream (returning a standard non-streaming JSON without stream chunks) or fails if `stream_options` is added without `stream: true` (yielding HTTP 400).
5. When the response is parsed, token usage falls back to `0`. 
6. Since 9Router filters out requests with `0` tokens from `usageHistory` (`promptTokens === 0 && completionTokens === 0`), these requests never show up in the **Usage / Overview** dashboard.

### Steps to Reproduce
Make a non-streaming POST request with a model routed via `openai` provider (e.g., using n8n or `curl`):
```bash
curl -X POST http://localhost:20128/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer <YOUR_API_KEY>" \
  -d '{"model": "openai", "messages": [{"role": "user", "content": "say hello"}]}'
```
* **Expected**: Valid tokens under `usage` object in response, logged in both Usage Overview and Usage Details.
* **Actual**: Response returns `usage: {"prompt_tokens":0,"completion_tokens":0}`, logged in Details as 0 tokens, missing from Overview.

### Proposed Fix
We can patch `transformRequest` in the base `openai` class (in `src/shared/services/providers/openai` or `chunks/318.js` depending on your source structure) to ensure `stream: true` and `stream_options: { include_usage: true }` are correctly set when `stream` parameter is true:

```javascript
transformRequest(model, body, stream, credentials) {
  let reqBody = body && "object" == typeof body ? {...body} : body;
  
  // Ensure stream: true and stream_options are set when 9Router forces streaming
  if (stream && reqBody && "object" == typeof reqBody) {
    reqBody.stream = true;
[...]

PR #3495, opened 2026-08-23
fix(sse): return Anthropic Message when Claude client hits forceStream provider

nitsuah:
## Summary

- **Bug**: Claude Code SDK gets "API returned an empty or malformed response (HTTP 200) — body is JSON but not a Message" on the non-streaming retry of a streaming request
- **Root cause**: `handleForcedSSEToJson` in `sseToJsonHandler.js` parses forced SSE into OpenAI `chat.completion` format but returned it directly to `FORMATS.CLAUDE` clients without converting it to an Anthropic `Message`
- **Affected providers**: any with `forceStream: true` — `openai`, `zed`, `codebuddy-cn`, `codebuddy-intl` (standard path) and `codex`, `grok-cli` (Codex/Responses API path)

### `open-sse/handlers/chatCore/sseToJsonHandler.js`

Added `fromOpenAIFinish` import and inlined `openAICompletionToClaudeMessage` (mirrors the same function in `nonStreamingHandler.js`; inlined to avoid the existing circular import since `nonStreamingHandler` already imports `parseSSEToOpenAIResponse` from this module).

**Standard Chat Completions SSE path** (openai, zed, codebuddy-*):
- Before: `finalBody = parsed` (raw OpenAI `chat.completion`) for all non-Responses clients
- After: when `sourceFormat === FORMATS.CLAUDE`, converts to `{ type:"message", content:[...], stop_reason:..., ... }` via `openAICompletionToClaudeMessage`

**Codex/Responses API path** (codex, grok-cli):
- Added `sourceFormat === FORMATS.CLAUDE` branch that builds an Anthropic Message directly from Responses API output items, including `thinking` blocks from reasoning items and `tool_use` blocks from function_call items

### `tests/unit/openai-responses-nonstream.test.js`

Two new tests added to the existing forced-SSE suite:
- Verifies `handleForcedSSEToJson` returns `type:"message"` with `tool_use` content block and `stop_reason:"tool_use"` for a Claude-format client against a forceStream provider
- Verifies text content path produces proper text content block and `stop_reason:"end_turn"`

## Failure scenario (before fix)

1. Claude Code makes `stream:true` request → 9router → forceStream provider
2. Stream returns 0 SSE events → `StreamNoEventsError`
3. SDK retries with `stream:false`
4. `handleForcedSSEToJson` called → parses SSE into OpenAI format → **returns `chat.completion` body**
5. Anthropic SDK: "JSON but not a Message" (HTTP 200, no `type:"message"`)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/open-sse/handlers/chatCore/sseToJsonHandler.js
+++ b/open-sse/handlers/chatCore/sseToJsonHandler.js
@@ -5,6 +5,7 @@ import { FORMATS } from "../../translator/formats.js";
 import { PROVIDERS } from "../../config/providers.js";
 import { buildRequestDetail, extractRequestConfig, saveUsageStats, formatDoneLine } from "./requestDetail.js";
 import { ROLE, RESPONSES_ITEM } from "../../translator/schema/index.js";
+import { fromOpenAIFinish } from "../../translator/concerns/finishReason.js";
 
 // Responses-API providers (e.g. codex) may emit SSE without content-type + use Responses output shape
 const isResponsesProvider = (p) => PROVIDERS[p]?.format === FORMATS.OPENAI_RESPONSES;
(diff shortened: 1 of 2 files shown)


nitsuah:
Triage on 2026-08-31: this PR head is not in current master. The matching non-PR tracker already exists as #2382, so no duplicate was created. Please use #2382 for forced-stream conversion contract tests covering content, tool use, errors, usage, and cancellation.

nitsuah:
@decolua friendly bump on this as i dont think its been addressed.

nitsuah:
Brought up to date with `master` (`39e36d3d`, v0.5.86; was 192 commits behind) via a merge commit, no rebase. It merged cleanly and the change is still needed: master's `sseToJsonHandler` doesn't cover this case.

Verified in Docker (`node:22`, `CI=true`): `openai-responses-nonstream.test.js` passes. The only failures in related suites are the two `force-stream-config` cases, which fail identically on master (known baseline).

PR #3496, opened 2026-08-24
fix(auth): prevent IP spoofing in brute-force protection (GHSA-32gc, GHSA-5mj8, GHSA-7cfm)

nitsuah:
## Summary
Prevents attackers from spoofing their IP via X-Forwarded-For or Host headers to bypass rate limiting or access local-only endpoints.

## Changes
- src/lib/auth/loginLimiter.js: getClientIp() now requires hasTrustedPeerHeaders() before trusting X-Forwarded-For headers
- src/dashboardGuard.js: isLoopbackPeer() no longer falls back to Host header; isLocalRequest() requires valid peer token + loopback IP + origin validation

## GHSA References
- GHSA-32gc: IP spoofing via X-Forwarded-For
- GHSA-5mj8: Host header spoofing for local-only access
- GHSA-7cfm: Rate limit bypass via header spoofing

### DIFF
--- a/src/dashboardGuard.js
+++ b/src/dashboardGuard.js
@@ -106,14 +106,12 @@ function isLoopbackHostname(h) {
 }
 
 function isLoopbackPeer(request) {
+  // Only trust peer IP when custom-server.js has validated it via peer token.
+  // Host header is spoofable and cannot be used for security decisions.
   if (hasTrustedPeerHeaders(request)) {
     return isLoopbackHostname(request.headers.get("x-9r-real-ip"));
   }
-  // Bare `next dev` forks its server, so the wrapper never loads and no peer address
-  // reaches us. Host is spoofable, so this stays confined to development.
-  if (process.env.NODE_ENV === "development") {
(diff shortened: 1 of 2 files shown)


nitsuah:
Triage on 2026-08-31: this PR head is not in current master and no non-PR tracker existed. Trusted-proxy topology, forwarding-header edge cases, and consistent client-IP resolution are tracked in #3676.

nitsuah:

nitsuah:
Superseded by #4286, which re-audits this against current master (`39e36d3d`) rather than rebasing it.

Current master already trusts `x-9r-real-ip` only when custom-server's per-process peer token vouches for it (GHSA-pjm4), so the core of this PR is covered. The remaining gap (tunnel clients choosing their lockout bucket via `X-Real-IP` / leftmost XFF) is fixed in #4286. The `next dev` Host-fallback removal here isn't carried forward, since master deliberately limits that fallback to development.

Closing to keep the queue clean. Happy to reopen if you'd prefer this version.

PR #3499, opened 2026-08-24
fix(security): extend PROTECTED_SETTING_KEYS to prevent mass assignment (GHSA-vmjq)

nitsuah:
## Summary
Extended PROTECTED_SETTING_KEYS from 14 to 23 keys to prevent attackers from overwriting critical security settings via mass assignment.

## Changes
- src/app/api/settings/route.js: Extended PROTECTED_SETTING_KEYS with 9 new protected keys for SSO, proxy, auth, and security config

## GHSA References
- GHSA-vmjq: Mass assignment vulnerability in settings endpoint

### DIFF
--- a/src/app/api/settings/route.js
+++ b/src/app/api/settings/route.js
@@ -12,7 +12,31 @@ const SETTINGS_RESPONSE_HEADERS = {
 };
 
 // Secrets must never be mass-assigned from request body (CWE-915)
-const PROTECTED_SETTING_KEYS = ["password", "mitmSudoEncrypted"];
+// Security-critical settings that must be explicitly handled, not mass-assigned
+const PROTECTED_SETTING_KEYS = [
+  "password",
+  "mitmSudoEncrypted",
+  "requireLogin",
+  "requireApiKey",
+  "authMode",
+  "ssoType",
+  "oidcIssuerUrl",
+  "oidcClientId",
+  "oidcClientSecret",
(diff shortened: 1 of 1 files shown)


nitsuah:
Triage on 2026-08-31: this PR head is not in current master and no non-PR tracker existed. Protected-key coverage, aliases/nested payloads, and future-proof sensitive-setting metadata are tracked in #3673.

nitsuah:

nitsuah:
Status after re-auditing against current master in #4286: leaving this open as a design question rather than closing it.

The advisory's exploit chain is mitigated on master (remote logins on the default password no longer get a session). But any authenticated session can still set `requireLogin: false` in one PATCH. This PR's approach of stripping those keys would break the Profile and Endpoint toggles that legitimately PATCH them. The better fix is probably a password re-prompt for security-downgrade settings (`requireLogin`, `requireApiKey`, `tunnelDashboardAccess`, `authMode`/SSO), mirroring the DB export re-auth. That needs UI work, so I'll rework this PR in that direction unless you'd prefer otherwise.

nitsuah:
Reworked as described in my earlier comment: the key-stripping approach is replaced with password re-auth for security-downgrade settings, brought up to date with current master, plus a fix for \currentPassword\ being persisted in plaintext. Details and tests are in the updated description. This now also closes #3673.

PR #3501, opened 2026-08-24
fix(security): require explicit JWT_SECRET env var, remove auto-generated fallback (GHSA-jphh)

nitsuah:
## Summary
Application now fails fast at startup if JWT_SECRET is not configured, removing the auto-generated secret file fallback which could lead to weak or predictable secrets.

## Changes
- src/lib/auth/dashboardSession.js: loadJwtSecret() now throws if JWT_SECRET is not set in environment

## GHSA References
- GHSA-jphh: Weak/predictable JWT secret from auto-generation fallback

### DIFF
--- a/src/lib/auth/dashboardSession.js
+++ b/src/lib/auth/dashboardSession.js
@@ -2,22 +2,17 @@ import { SignJWT, jwtVerify } from "jose";
 import bcrypt from "bcryptjs";
 import fs from "node:fs";
 import path from "node:path";
-import crypto from "node:crypto";
 import { DATA_DIR } from "@/lib/dataDir";
 import { getSettings } from "@/lib/localDb";
 
 const DEFAULT_PASSWORD = "123456";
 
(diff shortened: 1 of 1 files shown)


nitsuah:
Triage on 2026-08-31: this PR head is not in current master and no non-PR tracker existed. Missing-secret startup behavior, deployment documentation, and key-rotation expectations are tracked in #3671.

nitsuah:

nitsuah:
Superseded by #4286, which re-audits this against current master (`39e36d3d`) rather than rebasing it.

Master replaced the hardcoded fallback with a random secret persisted at `DATA_DIR/jwt-secret` (0600), so requiring `JWT_SECRET` would break default npm/CLI installs. The remaining issue (the published `.env.example` value being used as a real key) is fixed in #4286.

Closing to keep the queue clean. Happy to reopen if you'd prefer this version.

Issue #2302, opened 2026-07-02
API Error 400 for Claude Opus 4.6 (Antigravity)


TrygerZ:
API Error: 400 [antigravity/claude-opus-4-6-thinking] [400]: {
    "error": {                  
      "code": 400,
      "message": "{\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"This model does not support assistant message 
  prefill. The conversation must end with a user message.\"},\"request_id\":\"req_vrtx_011CcdJSUCe3x2qyowJ7cTQc\"}",
      "status": "INVALID_ARGUMENT"
    }
  }

I don't know if this is my mistake or the mistake of Antigravity. But when I want to use Claude Opus 4.6 (Antigravity) in Claude Code, this error message comes out.

PR #3506, opened 2026-08-24
fix(claude): normalize trailing assistant prefill for Claude targets

anhtran-ai:
## Summary

- normalize a trailing assistant turn before requests reach Claude-format targets
- apply one shared policy to both the translated path and native Claude passthrough
- keep the previous request body when `x-9router-assistant-prefill: preserve` is sent

## Problem

Anthropic-compatible Messages endpoints require a conversation to end on a user turn. Clients that resume from a partial assistant response send a trailing assistant message, and a trailing assistant `tool_use` block can arrive with no matching `tool_result`. Both shapes are rejected with HTTP 400 before reaching the model.

## Behavior

- trailing assistant text gains a user continuation turn
- trailing assistant `tool_use` blocks gain matching error `tool_result` blocks, ids preserved
- `tool_use` without an id degrades to the text/empty branch instead of emitting an undefined `tool_use_id`
- empty or invalid-signature-thinking-only trailing assistant content is dropped
- final user turns are untouched

The policy runs after tool-use ordering is fixed and before the thinking gate is computed, so the thinking gate sees the normalized turn.

## Verification

- new focused suite `tests/translator/assistant-prefill-policy.test.js`: 10/10 pass, covering both paths
- pre-fix reproduction: with sources reverted to base, the same suite fails 6/10 with the expected shapes
- focused ESLint on all four changed files: clean

The following failures reproduce byte-identically on unmodified `master`, so they are pre-existing and not introduced here:

- `tests/translator/bugs-toClaude-context.test.js`: `assistant reasoning_content becomes a thinking block`
- `tests/unit/translator-helpers-edge.test.js`: `hoists mid-conversation system messages into top-level system`
- `tests/unit/force-stream-config.test.js`: the `force-stream` mock around `open-sse/handlers/chatCore.js`
- `npm run build`: route configuration/page data collection for `/api/cli-tools/antigravity-mitm`

## Compatibility notes

- `x-9router-assistant-prefill` is a new public request header, alongside the documented `X-9Router-Token-Saver`. Happy to add a `README.md` line for it if you want it documented in this PR.
- header lookup accepts a `Headers` instance or a plain object, is case-insensitive, and handles array values.
- passthrough rebuilds `messages` before the policy runs, so the caller's body is not mutated across account-fallback attempts.
- a request whose only message is an empty assistant turn normalizes to an empty `messages` array and receives the endpoint's own validation error; that matches the existing empty-content filtering outcome and is not a regression.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -180,7 +180,9 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
       }
     }
     // Normalize newer Cowork/CC beta shapes (adaptive thinking, mid-conversation system) the API rejects
-    if (clientTool === "claude") normalizeClaudePassthrough(translatedBody, translatedBody.model);
+    if (clientTool === "claude") {
+      normalizeClaudePassthrough(translatedBody, translatedBody.model, clientRawRequest?.headers || null);
+    }
   } else {
     translatedBody = translateRequest(sourceFormat, targetFormat, upstreamModel, body, stream, credentials, provider, reqLogger, stripList, connectionId, clientTool);
     if (!translatedBody) {
--- a/open-sse/translator/concerns/assistantPrefillPolicy.js
+++ b/open-sse/translator/concerns/assistantPrefillPolicy.js
@@ -1,0 +1,56 @@
+import { CLAUDE_BLOCK, ROLE } from "../schema/index.js";
+
+const ASSISTANT_CONTINUATION_PROMPT = "Continue from the assistant response above without repeating it.";
+const INCOMPLETE_TOOL_RESULT = "Tool execution was not completed before this request continued.";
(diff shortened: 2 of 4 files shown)


anhtran-ai:
Cross-reference: #3936 supersedes this PR for the case #2302 actually reports.

This PR normalizes the trailing assistant turn at the `chatCore` / claude-format seam for Claude targets broadly; it has since gone `CONFLICTING` against `master`. #3936 re-derives the same policy against current `master` (`eb712ca8`), scoped to the Antigravity→Vertex request path where the "This model does not support assistant message prefill" 400 is reproduced, and adds 13 unit tests with a before/after baseline showing zero regressions.

Maintainers: prefer #3936. I'll close this one on request, or rebase it if the broader seam is wanted as a follow-up.

PR #3524, opened 2026-08-24
fix(codex-settings): refuse to overwrite a config that could not be read

ntdat812:
## Problem

Both write paths in `codex-settings/route.js` merged their fields into whatever they
managed to read, with an empty catch in between:

```js
let authData = {};
try {
  const existingAuth = await fs.readFile(authPath, "utf-8");
  authData = JSON.parse(existingAuth);
} catch { /* No existing auth */ }

// Force apikey mode (keep existing tokens untouched for ChatGPT login reuse)
authData.OPENAI_API_KEY = apiKey;
authData.auth_mode = "apikey";
await fs.writeFile(authPath, JSON.stringify(authData, null, 2));
```

A malformed file, a permission error, or a directory in the way is therefore
indistinguishable from *"no file"* — and the write then **replaces** the user's file
with just the two fields being set.

- `~/.codex/auth.json` — discards the ChatGPT OAuth tokens that the comment on the
  very next line promises to keep. The user is silently logged out of their
  subscription and has to re-authenticate.
- `~/.codex/config.toml` (same pattern at the top of the handler) — discards every
  model provider, MCP server and approval policy the user had.

Neither case reports anything: the operation returns `success: true`.

## Change

`readExistingConfig(filePath, parse)` returns `null` **only** for `ENOENT` — the
legitimate start-fresh case — and throws for everything else, naming the file:

```
/home/u/.codex/auth.json exists but could not be parsed (Unexpected end of JSON input); refusing to overwrite it
```

The route surfaces that one message instead of the generic 500, so the user learns
which file to fix; every other failure stays generic.

The **reset** path is deliberately untouched: it already skips its write when the
file cannot be read, so it never clobbered anything.

## Tests

`tests/unit/cli-tools-refuse-to-clobber.test.js` (new, 7 cases) against real files in
a temp dir: absent → `null`; valid → parsed; malformed → throws **and the file is
still byte-identical afterwards**; the error names the file and the parser's reason;
a non-ENOENT read failure (a directory where the file should be) propagates; a
parser returning `undefined` is not confused with "absent"; plus a source assertion
that both codex write paths use the helper and the two empty catches are gone.

## Not in this PR

The same swallow-then-write shape exists in `copilot-settings/route.js:77-81`
(`catch { /* No existing config */ }` before writing the provider array back, which
would drop the user's other entries). Same fix, different file and different failure
story — happy to send it separately, or fold it in here if you prefer one pass.


ntdat812:
Rebased onto current `master` and **narrowed**, because part of it landed differently upstream.

When this was written, both write paths in `codex-settings/route.js` merged fields into whatever they managed to read and wrote the result back, with an empty `catch` in between — so a malformed file, a permission error or a directory in the way was indistinguishable from "no file", and the write replaced the user's file with just the fields being set.

`master` has since **removed the `auth.json` write entirely** from this route (custom providers carry the key as a static header instead), which takes the ChatGPT-OAuth-token half of the report with it. I took `master`'s deletion rather than reinstating that block: what is left is the `config.toml` path, where an unreadable file still discards every provider, MCP server and approval policy the user had.

So this PR is now: `readExistingConfig(configPath, …)` on the one remaining write path, plus the helper it needs. `readExistingConfig` returns `null` only for ENOENT — the legitimate start-fresh case — and throws for everything else, naming the file so the operator can fix it.

The test file follows the same narrowing: it still covers the helper directly (ENOENT vs malformed vs permission vs a directory in the way, on JSON and on TOML), and its source-level assertion now checks the config path only — plus one assertion that the `auth.json` empty-catch stays gone, so the pattern cannot quietly come back.

PR #3528, opened 2026-08-24
fix(chat): treat a request without a stream key as non-streaming (#3492)

### LINKED ISSUE #3492: Non-streaming /v1/chat/completions returns text/event-stream and appends data: [DONE], breaking JSON clients
# Non-streaming `/v1/chat/completions` returns `text/event-stream` and appends `data: [DONE]`, breaking JSON clients

## Summary

When a client calls `POST /v1/chat/completions` **without** a `stream` field, the
server responds with `Content-Type: text/event-stream` and appends a trailing
`data: [DONE]` to the body.

The body itself is already a complete, correct non-streaming `chat.completion`
object — it is *not* chunked. Only the framing is wrong: the wrong content-type
header, plus an SSE terminator glued onto valid JSON.

Because of that trailing terminator, the response body is no longer parseable
JSON, so any OpenAI-compatible client that does `JSON.parse(await res.text())`
fails.

Per the OpenAI API, `stream` defaults to `false`, so a request that omits it is
a non-streaming request and must receive `application/json`.

## Reproduction

Identical requests, differing only by the presence of the `stream` key:

```bash
# 1. BROKEN — no "stream" key (this is what the Vercel AI SDK sends)
curl -si $BASE_URL/v1/chat/completions \
  -H "Authorization: Bearer $API_KEY" \
  -d '{"model":"<any-model>","max_tokens":30,
       "messages":[{"role":"user","content":"say OK"}]}'

# 2. OK — "stream": false explicitly
curl -si $BASE_URL/v1/chat/completions \
  -H "Authorization: Bearer $API_KEY" \
  -d '{"model":"<any-model>","stream":false,"max_tokens":30,
       "messages":[{"role":"user","content":"say OK"}]}'
```

## Actual

Request 1:

```
content-type: text/event-stream

{"id":"2026...","object":"chat.completion","created":...,"model":"...",
 "choices":[{"index":0,"finish_reason":"stop","message":{...}}],
 "usage":{...},"cost":"0"}data: [DONE]
```

Note there is no `data: ` prefix on the completion object and no chunking — the
payload is a plain `chat.completion`, with `data: [DONE]` concatenated directly
onto it.

Strict JSON parsers reject this:

```
JSON.parse → SyntaxError: Unexpected non-whitespace character after JSON at position 455
Python json.load → Extra data: line 1 column 456 (char 455)
```

Request 2 (with `"stream": false`) returns `content-type: application/json` and a
body that parses cleanly.

## Expected

A request where `stream` is absent or `false` should return:

- the `chat.completion` object with **no** `data: [DONE]` appended

Requests with `"stream": true` should keep the current SSE behaviour — that path
is correct and should not change.

## Scope

Not specific to any one model or to combo models. Reproduced on individual
models and on a combo alike; every model tested returned
`content-type: text/event-stream` with `[DONE]` appended when `stream` was
omitted.

## Client impact

This is not a hypothetical strictness issue. The Vercel AI SDK's
OpenAI-compatible provider (`@ai-sdk/openai-compatible`, also `@ai-sdk/openai`)
**never sends `stream: false`** — its `doGenerate` omits the `stream` key
entirely, and only `doStream` sends `stream: true`. Verified by inspecting the
shipped bundles: zero occurrences of `stream: false`, and `stream: true` only on
the streaming code path.

So every non-streaming call from any Vercel-AI-SDK-based application fails
against this server, while streaming clients (typical chat UIs) work fine. That
asymmetry makes the bug easy to miss: interactive use looks healthy, and only
programmatic `generateText` / `generateObject` callers break.

## Suggested fix

Branch on the requested mode before writing the response, rather than always
taking the SSE path:

```ts
const body = await req.json()

if (body.stream !== true) {
  return Response.json(completion)   // application/json, no [DONE] terminator
}

// stream === true → existing SSE path, unchanged
```

Since the handler already produces a complete `chat.completion` object for the
non-streaming case, this should be a change to how the response is returned
rather than to any request-handling logic.

## Environment

- Server: self-hosted, Next.js App Router behind nginx
- Client that surfaced it: `@ai-sdk/openai-compatible` ^2.0.41 / `ai` ^6.0.168
- Also reproducible with plain `curl` + any strict JSON parser (see above)

ntdatt812:
Fixes #3492.

## Cause

`chatCore.js` resolved the response framing with:

```js
const clientRequestedStreaming = body.stream === true || sourceFormat === FORMATS.ANTIGRAVITY || ...;
const providerRequiresStreaming = PROVIDERS[provider]?.forceStream === true;
```

The OpenAI API defines `stream` with a default of **false**, so a body that omits the key is a non-streaming request. `!== false` reads an absent key as *stream*, so the handler took the SSE branch for a payload it had already built as a complete `chat.completion` object — `text/event-stream` on the wire, plus `data: [DONE]` concatenated onto valid JSON, exactly as the issue reports.

The predicate that was actually wanted is on the line above: `clientRequestedStreaming` is computed there and used further down for the forced-stream-to-JSON branch, but never for this decision.

## Why it looked healthy

The `Accept: application/json` branch below only ever forces non-streaming:

```js
if (clientPrefersJson && !clientPrefersSSE && body.stream !== true && !providerRequiresStreaming) stream = false;
```

so clients that happen to send that header were rescued. A client that sends no `Accept` header — the `curl` reproduction in the issue — was not. Interactive chat UIs stream and look fine; only programmatic `generateText` / `generateObject` callers break.

## The change

Move the predicate into `open-sse/handlers/chatCore/streamMode.js` and read the mode from it:

```js
let stream = providerRequiresStreaming ? true : clientRequestedStreaming;
```

Three lines in `chatCore.js`. The Gemini / Gemini-CLI / Antigravity carve-out is preserved verbatim — those endpoints carry the mode in the path rather than the body, and only their streaming variants are routed here.

Nothing else moves: `stream: true` still streams, explicit `stream: false` still returns JSON, `forceStream` providers still stream upstream and are converted to JSON by the existing `handleForcedSSEToJson` branch, and the `Accept` and `deepseek-tui` adjustments below only ever set `stream = false`, so they are unaffected.

## Tests

`tests/unit/stream-mode-default-3492.test.js` — 7 cases against the extracted predicate, no mocks:

- no `stream` key → non-streaming
- `stream: false` → non-streaming
- `stream: true` → streaming
- `"true"`, `1`, `{}` → non-streaming (a truthy non-boolean is not a request to stream)
- Gemini / Gemini-CLI / Antigravity → streaming regardless of the body
- the Claude surface behaves like OpenAI
- a missing body does not throw

Mutation-checked — restoring `!== false` turns 4 of the 7 red:

```
× reads a body with no stream key as non-streaming
× does not treat a truthy non-boolean as a request to stream
× treats the Claude surface like OpenAI
× survives a missing body
Tests  4 failed | 3 passed (7)
```

## Suite

Measured both endpoints on the full unit suite:

| | Test Files | Tests |
|---|---|---|
| with this PR | 31 failed / 163 passed | 82 failed / **1794** passed |

Same 82 pre-existing failures, +7 passing — this PR's own cases. No test changes state because of it.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -16,6 +16,7 @@ import { getExecutor } from "../executors/index.js";
 import { supportsGrokCliReasoningEffort } from "../config/grokCli.js";
 import { buildRequestDetail, extractRequestConfig } from "./chatCore/requestDetail.js";
 import { handleForcedSSEToJson } from "./chatCore/sseToJsonHandler.js";
+import { clientRequestedStreaming as requestedStreaming } from "./chatCore/streamMode.js";
 import { handleNonStreamingResponse } from "./chatCore/nonStreamingHandler.js";
 import { handleStreamingResponse, buildOnStreamComplete } from "./chatCore/streamingHandler.js";
 import { detectClientTool, isNativePassthrough } from "../utils/clientDetector.js";
@@ -109,9 +110,9 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     }
   }
 
-  const clientRequestedStreaming = body.stream === true || sourceFormat === FORMATS.ANTIGRAVITY || sourceFormat === FORMATS.GEMINI || sourceFormat === FORMATS.GEMINI_CLI;
+  const clientRequestedStreaming = requestedStreaming(body, sourceFormat);
   const providerRequiresStreaming = PROVIDERS[provider]?.forceStream === true;
-  let stream = providerRequiresStreaming ? true : (body.stream !== false);
+  let stream = providerRequiresStreaming ? true : clientRequestedStreaming;
 
   // Image generation models require non-streaming (Google v1internal:generateContent)
   const modelType = getModelType(alias, model);
--- a/open-sse/handlers/chatCore/streamMode.js
+++ b/open-sse/handlers/chatCore/streamMode.js
@@ -1,0 +1,24 @@
+import { FORMATS } from "../../translator/formats.js";
+
+/**
+ * Whether the client asked for an SSE-framed response.
+ *
+ * The OpenAI API defines `stream` with a default of **false**, so a body that
+ * omits the key is a non-streaming request. Reading an absent key as "stream"
+ * frames a plain `chat.completion` object as SSE: wrong content-type, plus a
+ * `data: [DONE]` glued onto valid JSON, which strict parsers reject (#3492).
+ *
+ * The Gemini and Antigravity surfaces are the exception. Their endpoints carry
+ * the mode in the path rather than the body, and this router only routes their
+ * streaming ones, so those formats always mean SSE.
+ */
+export function clientRequestedStreaming(body, sourceFormat) {
+  if (
+    sourceFormat === FORMATS.ANTIGRAVITY ||
+    sourceFormat === FORMATS.GEMINI ||
+    sourceFormat === FORMATS.GEMINI_CLI
+  ) {
+    return true;
+  }
+  return body?.stream === true;
+}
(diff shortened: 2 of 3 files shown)


ntdatt812:

PR #3529, opened 2026-08-24
feat(models): report token limits on combo entries in /v1/models (#3486)

### LINKED ISSUE #3486: Combo model entries in /v1/models omit token/context metadata
## Problem

Combo-pool models registered on the Dashboard appear in `GET /v1/models` as bare entries with no token-limit metadata:

```json
{ "id": "deepseek-flash-mix", "object": "model", "owned_by": "combo" }
```

Individual (non-combo) models expose `context_length` and `max_completion_tokens` (added in #3218), but combo entries skip them entirely.

## Impact

Any OpenAI-compatible client that reads /v1/models to size its context window resolves a combo model via its own fallback instead. In our case (Hermes Agent), `deepseek-flash-mix` fell back to a hardcoded `deepseek: 128k` catalog entry even though every member of the pool advertises a 1M window, which caused premature context compaction at ~96k. Other clients (Cline, Codex, etc.) will hit the same class of bug.

## Suggestion

Emit the pool's effective context limit on combo entries — e.g. `context_length: <min of members' contextWindow>` (and ideally `max_completion_tokens` too). That mirrors what the pool actually accepts and keeps metadata consumers consistent. #2242 adds `capabilities.contextWindow` aggregation for combos; adding the top-level snake_case fields would make the metadata readable by clients that only check top-level keys.

## Repro

```bash
curl $NINEROUTER_URL/v1/models | jq '.data[] | select(.owned_by=="combo")'
```

ntdatt812:
Fixes #3486.

## Before

The combo loop in `src/app/api/v1/models/route.js` builds a bare entry:

```js
const entry = { id: combo.name, object: "model", owned_by: "combo" };
if (combo.kind === "webSearch" || combo.kind === "webFetch") entry.kind = combo.kind;
models.push(entry);
```

while the per-provider loop below emits `context_length` and `max_completion_tokens` (#3218). So a client sizing its context window off `/v1/models` finds nothing on a combo, falls back to guessing from the name, and guesses against a pool whose members may be far wider — the 1M pool read as `deepseek: 128k` in the issue, compacting at ~96k.

## After

```json
{ "id": "deepseek-flash-mix", "object": "model", "owned_by": "combo",
  "context_length": 128000, "max_completion_tokens": 64000 }
```

**Minimum across members.** A request routed to a combo can land on any member, so the pool can only promise what its smallest member accepts — anything larger is a limit some member will reject.

Three deliberate choices:

- A member whose capabilities carry no usable number is **skipped**, not read as zero. One unknown member must not collapse the pool to nothing.
- When no member reports a limit, the field is **omitted**. An absent field lets a client fall back to its own catalog; a wrong one does not.
- **Web-search / web-fetch combos are untouched.** They are not chat models and carry no window.

Member capabilities come from `getCapabilitiesForModel`, the same source the per-provider loop already uses, so combo numbers agree with the individual entries they are made of.

## Scope

Only the two top-level snake_case fields the issue asks for. #2242 adds `capabilities.contextWindow` aggregation for combos — that is a separate, nested field, and this PR does not touch `capabilities` on combo entries, so the two should not collide.

The alias-to-provider map that the DB-unavailable branch already built is hoisted so the combo loop can share it instead of building a second one.

## Tests

`tests/unit/combo-token-limits-3486.test.js` — 9 cases:

- smallest window in the pool wins, and the result is order independent
- a member reporting no limits is ignored, not read as zero
- no member reporting a limit → nothing to emit
- empty / absent / malformed model list does not throw
- against the **real** capability catalog: `claude-opus-4.7` (1M) + `deepseek-chat` (128k) → 128,000; two wide members → 1,000,000
- `splitModelRef` splits on the first slash only, so `openrouter/deepseek/deepseek-chat` keeps its nested id

Mutation-checked — swapping `Math.min` for `Math.max` turns the two that pin the semantics red:

```
× reports the smallest window in the pool
× mixes a 1M model with a 128k one down to 128k using the real catalog
Tests  2 failed | 7 passed (9)
```

## Suite

| | Test Files | Tests |
|---|---|---|
| with this PR | 31 failed / 163 passed | 82 failed / **1796** passed |

Same 82 pre-existing failures, +9 passing — this PR's own cases.

### DIFF
--- a/open-sse/services/comboLimits.js
+++ b/open-sse/services/comboLimits.js
@@ -1,0 +1,38 @@
+/**
+ * Token limits a combo pool can honour.
+ *
+ * A request routed to a combo may land on any member, so the pool can only
+ * promise what its *smallest* member accepts. Anything larger is a limit some
+ * member will reject.
+ *
+ * `capsFor` receives one member reference ("alias/model-id") and returns that
+ * model's capabilities; passing it in keeps this pure and lets the caller own
+ * the alias-to-provider mapping.
+ *
+ * Members whose capabilities carry no usable number are skipped rather than
+ * treated as zero. When no member reports one, the field comes back undefined
+ * and the caller omits it - an absent field lets a client fall back, a wrong
+ * one does not.
+ */
+export function comboTokenLimits(members, capsFor) {
+  const lowest = { contextWindow: undefined, maxOutput: undefined };
+  if (!Array.isArray(members)) return lowest;
+
+  for (const ref of members) {
+    if (typeof ref !== "string" || ref.trim() === "") continue;
+    const caps = capsFor(ref) || {};
+    for (const field of ["contextWindow", "maxOutput"]) {
+      const value = caps[field];
+      if (!Number.isFinite(value) || value <= 0) continue;
+      lowest[field] = lowest[field] === undefined ? value : Math.min(lowest[field], value);
+    }
+  }
+  return lowest;
+}
+
+/** Split a combo member reference into its provider alias and model id. */
+export function splitModelRef(ref) {
+  const slash = ref.indexOf("/");
+  if (slash === -1) return { alias: "", modelId: ref };
+  return { alias: ref.slice(0, slash), modelId: ref.slice(slash + 1) };
+}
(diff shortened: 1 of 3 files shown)


ntdatt812:

PR #3537, opened 2026-08-25
fix(quota): give the auto-refresh countdown a single timer owner (#3470)

### LINKED ISSUE #3470: [Bug] Quota Tracker countdown accelerates after document visibility changes
## Description

The Quota Tracker countdown may accelerate and reach zero faster than real time after the document visibility changes, such as switching tabs or hiding and showing the browser window.

Browser split view is one possible usage context. It is a feature available in browsers such as Vivaldi and Brave that lets users organize multiple pages or tabs side by side within the same window, making it easier to monitor multiple providers simultaneously. However, opening multiple Quota Trackers side by side alone does not necessarily reproduce the issue.

The suspected cause is multiple countdown intervals running simultaneously within the same component instance.

## Steps to reproduce

1. Open `/dashboard/quota`.
2. Enable auto-refresh.
3. Optionally use the browser's split-view feature to display another Quota Tracker view side by side.
4. Switch tabs/windows, or hide and show the browser window.
5. Observe the auto-refresh countdown.

## Expected behavior

Each Quota Tracker component instance should maintain one countdown timer. The countdown should decrease by approximately one second per second and reach zero only after the configured refresh interval.

Opening multiple Quota Trackers in browser split view should create independent, correctly paced timers and should not by itself accelerate a countdown.

## Actual behavior

After document visibility changes, the countdown may decrease faster than real time. In some cases, it appears to run at approximately twice the expected speed.

## Relevant code

[`src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js`](https://github.com/decolua/9router/blob/699edac3273e13d4744bc46f6082618f08560702/src/app/%28dashboard%29/dashboard/usage/components/ProviderLimits/index.js) at commit [`699edac3273e13d4744bc46f6082618f08560702`](https://github.com/decolua/9router/commit/699edac3273e13d4744bc46f6082618f08560702)

The main effect creates a countdown interval:

```js
countdownRef.current = setInterval(() => {
  setCountdown((prev) => {
    if (prev <= 1) return 60;
    return prev - 1;
  });
}, 1000);
```

The `visibilitychange` handler can create another interval when the document becomes visible:

```js
} else if (autoRefresh && hasHydratedAutoRefresh) {
  intervalRef.current = setInterval(() => refreshAll(), REFRESH_INTERVAL_MS);
  countdownRef.current = setInterval(() => {
    setCountdown((prev) => (prev <= 1 ? 60 : prev - 1));
  }, 1000);
}
```

The handler does not stop an existing countdown interval before creating a new one. This can leave multiple active timers updating the same React state.

## Analysis

This does not appear to be caused by `sessionStorage`. The code uses `localStorage` only for the auto-refresh enabled/disabled preference:

```js
localStorage.getItem("quotaAutoRefresh");
localStorage.setItem("quotaAutoRefresh", String(autoRefresh));
```

The countdown itself is stored in component state, while each interval is managed through `countdownRef`.

## Suggested direction

Make timer management idempotent:

- centralize timer creation and cleanup;
- clear existing intervals before starting new ones;
- ensure only one countdown interval and one refresh interval exist per component instance;
- verify behavior after repeated `visibilitychange` events;
- verify that multiple independent Quota Tracker instances remain correctly paced.

## Environment

- OS: macOS
- View: `/dashboard/quota`
- Layout: browser split view with multiple quota trackers (when applicable)
- Repository version inspected: `0.5.55`

ntdatt812:
Fixes #3470.

## Two owners, one pair of refs

The quota tracker creates its intervals in two places, and each clears only what its own ref happens to point at:

```js
// auto-refresh effect
intervalRef.current  = setInterval(() => refreshAll(), REFRESH_INTERVAL_MS);
countdownRef.current = setInterval(tick, 1000);

// visibilitychange handler, on becoming visible
intervalRef.current  = setInterval(() => refreshAll(), REFRESH_INTERVAL_MS);
countdownRef.current = setInterval(tick, 1000);
```

The second assignment overwrites the refs. Whatever they held is orphaned — still scheduled, no longer reachable, never cleared.

Two paths reach that state:

1. **Effect re-run while hidden.** The handler clears both and nulls the refs. `refreshAll` then changes identity (it is in the effect's dep list), so the effect re-runs: its cleanup sees null refs and does nothing, and the body creates a pair **while the tab is still hidden**. Becoming visible creates a second pair on top of it.
2. **Two visible transitions in a row.** Split view and window focus changes can deliver these; the resume branch runs twice and orphans the first pair.

Either way the countdown decrements twice per second — the ~2× speed in the report. The issue guessed "multiple countdown intervals running simultaneously"; this is the mechanism.

## The fix

Ownership moves into `createRefreshTimers()`, and **`start()` stops first** — so a second start replaces the timers instead of stacking them, no matter who calls it.

Three follow-on corrections that fall out of having one owner:

- the auto-refresh effect no longer starts anything while `document.hidden` — the visibility handler owns the resume, which is what the pause feature intends
- the visibility effect stops the timers on unmount: a tab hidden at mount takes the other effect through its early return, which leaves no cleanup behind, so a resume-then-unmount used to leak both intervals
- `refreshAll` is read through a ref, so a new identity no longer tears the interval down and restarts the 60-second window mid-count

`COUNTDOWN_SECONDS` is now derived from `REFRESH_INTERVAL_MS` instead of appearing as a literal `60` in three places.

## Tests

`tests/unit/quota-refresh-timers-3470.test.js` — 10 cases on fake timers:

- one tick per second
- **`start()` twice still ticks once per second** — the bug
- hide/show cycle stays single-rate, and nothing ticks while stopped
- `stop()` is idempotent and clears both timers
- refresh fires on `REFRESH_INTERVAL_MS`, not on the tick
- `isRunning()` reflects state
- the latest callback is always used, so a re-render need not restart the timer
- `nextCountdown` counts down, wraps at 1 and 0, and `COUNTDOWN_SECONDS === REFRESH_INTERVAL_MS / 1000`

Mutation-checked — removing `stop()` from the top of `start()` reproduces the bug and reds exactly the case that pins it:

```
× does not double the tick rate when start is called twice
Tests  1 failed | 9 passed (10)
```

## Suite

| | Test Files | Tests |
|---|---|---|
| with this PR | 31 failed / 163 passed | 82 failed / **1797** passed |

Same 82 pre-existing failures, +10 — this PR's own cases. ESLint on the touched component reports the same single pre-existing `exhaustive-deps` warning as `master`, no new ones.

### DIFF
--- a/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js
+++ b/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js
@@ -29,6 +29,7 @@ import {
   setQuotaCache,
   QUOTA_CACHE_KEY,
   REFRESH_INTERVAL_MS,
+  COUNTDOWN_SECONDS,
   CLAUDE_REFRESH_INTERVAL_MS,
   DEPLETED_QUOTA_THRESHOLD,
   AUTO_REFRESH_STORAGE_KEY,
@@ -38,6 +39,7 @@ import {
   ACCOUNT_FILTER_OPTIONS,
   QUOTA_SORT_OPTIONS,
 } from "./utils";
+import { createRefreshTimers, nextCountdown } from "./refreshTimers";
 import Card from "@/shared/components/Card";
 import { ConfirmModal, EditConnectionModal } from "@/shared/components";
 import { USAGE_SUPPORTED_PROVIDERS } from "@/shared/constants/providers";
@@ -135,7 +137,7 @@ export default function ProviderLimits() {
   const [lastUpdated, setLastUpdated] = useState(null);
   const [hasHydratedAutoRefresh, setHasHydratedAutoRefresh] = useState(false);
   const [refreshingAll, setRefreshingAll] = useState(false);
-  const [countdown, setCountdown] = useState(60);
+  const [countdown, setCountdown] = useState(COUNTDOWN_SECONDS);
   const [connectionsLoading, setConnectionsLoading] = useState(true);
   const [deletingId, setDeletingId] = useState(null);
   const [togglingId, setTogglingId] = useState(null);
@@ -169,8 +171,6 @@ export default function ProviderLimits() {
     providerFilteredConnections: 0,
   });
 
-  const intervalRef = useRef(null);
-  const countdownRef = useRef(null);
   const tickCountRef = useRef(0);
 
   const fetchConnections = useCallback(
@@ -467,7 +467,7 @@ export default function ProviderLimits() {
     if (refreshingAll) return;
 
     setRefreshingAll(true);
-    setCountdown(60);
+    setCountdown(COUNTDOWN_SECONDS);
 
     // Throttle Claude: poll its quota every Nth auto-tick (manual force bypasses)
     const tick = (tickCountRef.current += 1);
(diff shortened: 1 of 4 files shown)


ntdatt812:

PR #3538, opened 2026-08-25
fix(transport): send the translated body to the endpoint that speaks its format (#3418, #3439)

### LINKED ISSUE #3418: MiniMax-M3 ignores images from OpenAI clients because matched transport format loses to targetFormat

### LINKED ISSUE #3439: fix(minimax): remove targetFormat: "claude" on MiniMax-M3 and align runtimeTransport with targetFormat
### Problem Description
When calling `minimax/MiniMax-M3` via `/v1/chat/completions` (e.g. from OpenAI-compatible coding agents) with tool definitions, requests fail with HTTP 400:
```text
invalid params, invalid tool type: (2013)
```

### Root Cause Analysis
1. **Model Registry Conflict**:
   In `open-sse/providers/registry/minimax.js` and `minimax-cn.js`, multi-endpoint `transports` are declared (supporting both OpenAI `/v1/chat/completions` and Anthropic `/anthropic/v1/messages`). However, `MiniMax-M3` has `targetFormat: "claude"` hardcoded.
   When an OpenAI request arrives (`sourceFormat: "openai"`), 9Router selects the OpenAI transport (`https://api.minimax.io/v1/chat/completions`), but `targetFormat: "claude"` forces translation into Anthropic/Claude schema (`input_schema` instead of `parameters`), dispatching Claude tool schemas to MiniMax's OpenAI endpoint.

2. **Transport Alignment in `chatCore.js`**:
   In `open-sse/handlers/chatCore.js`, `credentials.runtimeTransport` was assigned based on client `sourceFormat` rather than the final `targetFormat`:
   ```javascript
   const useTransport = (!modelSupportedFormats || modelSupportedFormats.includes(sourceFormat)) ? runtimeTransport : null;
   const targetFormat = modelTargetFormat || useTransport?.format || getTargetFormat(provider, credentials);
   if (useTransport && credentials) credentials.runtimeTransport = useTransport;
   ```
   If `targetFormat` diverges from `sourceFormat` (e.g. model-specific `targetFormat`), the outbound request body format and the selected transport format disagree.

### Invariant
The serialized upstream request body format must always equal the selected upstream transport format.

#### 1. Provider Registry (`minimax.js` / `minimax-cn.js`)
Remove `targetFormat: "claude"` from `MiniMax-M3` so native OpenAI format is preserved for OpenAI clients and native Anthropic format is preserved for Claude clients:

```diff
--- a/open-sse/providers/registry/minimax.js
+++ b/open-sse/providers/registry/minimax.js
@@ -54,7 +54,7 @@ export default {
     },
   ],
   models: [
-    { id: "MiniMax-M3", name: "MiniMax M3", targetFormat: "claude" },
+    { id: "MiniMax-M3", name: "MiniMax M3" },
     { id: "MiniMax-M2.7", name: "MiniMax M2.7" },
```

#### 2. Transport Alignment (`chatCore.js`)
Ensure `credentials.runtimeTransport` resolves to the transport matching `targetFormat`:

```diff
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -91,7 +91,13 @@ export async function handleChatCore({ ... }) {
   const useTransport = (!modelSupportedFormats || modelSupportedFormats.includes(sourceFormat)) ? runtimeTransport : null;
   const targetFormat = modelTargetFormat || useTransport?.format || getTargetFormat(provider, credentials);
-  if (useTransport && credentials) credentials.runtimeTransport = useTransport;
+  if (credentials) {
+    if (modelTargetFormat) {
+      credentials.runtimeTransport = resolveTransport(provider, modelTargetFormat) || null;
+    } else if (useTransport) {
+      credentials.runtimeTransport = useTransport;
+    }
+  }
```

ntdat812:
Closes #3418. Closes #3439.

## Problem

`chatCore` picks the upstream transport from the client's `sourceFormat`, then lets a model-level `targetFormat` override the format the body is written in — but not the transport:

```js
const useTransport = (!modelSupportedFormats || modelSupportedFormats.includes(sourceFormat)) ? runtimeTransport : null;
const targetFormat = modelTargetFormat || useTransport?.format || getTargetFormat(provider, credentials);
if (useTransport && credentials) credentials.runtimeTransport = useTransport;
```

`runtimeTransport` is what `DefaultExecutor.buildUrl` and `buildHeaders` read for the URL, the headers and the auth descriptor. So for an OpenAI-format request to a model pinned to `claude`, 9Router serializes a Claude body and posts it to the provider's OpenAI endpoint with that endpoint's bearer auth.

Upstreams parse what they recognize and drop the rest, so the failure is quiet and looks like a model problem:

- **#3418** — `minimax/MiniMax-M3` answers `I don't see any image attached`. The reporter verified the same image and key work against both `POST /v1/chat/completions` (OpenAI content) and `POST /anthropic/v1/messages` (Anthropic content) directly, which rules out the image, the credentials and M3's vision support.
- **#3439** — the same pairing with tools returns `invalid params, invalid tool type: (2013)`, because Claude's `input_schema` reaches MiniMax's OpenAI endpoint where `parameters` was expected.

## Fix

Resolve the format and the transport in one place so they cannot disagree, and let the transport follow `targetFormat` when a model pins one:

```js
const transport = modelTargetFormat ? resolveTransport(provider, modelTargetFormat) : useTransport;
```

Three models declare `targetFormat` on a provider that has more than one endpoint:

| model | before | after |
|---|---|---|
| `minimax/MiniMax-M3` | claude body → `/v1/chat/completions` | claude body → `/anthropic/v1/messages` |
| `minimax-cn/MiniMax-M3` | claude body → `/v1/chat/completions` | claude body → `/anthropic/v1/messages` |
| `xiaomi-tokenplan/mimo-v2.5-pro-claude` | claude body → `/chat/completions`, bearer | claude body → `/anthropic/v1/messages`, `x-api-key` |

Nothing else changes: without a model-level `targetFormat` the expression is the previous `useTransport`, and the per-model `supportedFormats` guard is untouched — a model that does not declare a format still never gets that transport.

I did not take the other half of #3439's proposal (dropping `targetFormat: "claude"` from MiniMax-M3). That is a product decision about which wire format M3 should speak, and it is not needed to fix either report — the Anthropic endpoint handles both the image and the tool schema, as the #3418 control tests show. Same reasoning for `mimo-v2.5-pro-claude`, whose whole point is to be Claude-native.

## Verification

`tests/unit/upstream-route-alignment.test.js` derives the model list from the registry rather than hardcoding it, so a future pinned model is covered automatically (with a guard test so the table can't pass vacuously).

Reverting the one-line fix to `const transport = useTransport;` fails exactly the three alignment cases:

```
FAIL  minimax/MiniMax-M3: openai client → claude body on the claude transport
FAIL  minimax-cn/MiniMax-M3: openai client → claude body on the claude transport
FAIL  xiaomi-tokenplan/mimo-v2.5-pro-claude: openai client → claude body on the claude transport
AssertionError: expected 'openai' to be 'claude'
Tests  3 failed | 9 passed (12)
```

Full offline suite (`vitest run unit auth translator`, `CI=true`, excluding `real/`), same command on both sides:

| | tests | failed |
|---|---|---|
| `origin/master` (699edac3) | 1910 | 95 |
| this branch | 1922 | 95 |

The failure **sets** are identical — `comm` on the sorted full test names returns nothing in either direction. The 95 are pre-existing (windsurf/kimi-coding registry drift and their snapshots) and unrelated to this change. `eslint` clean on all three files.

I have no MiniMax or Xiaomi Token Plan credentials, so I have not run a live request; what is verified here is that the body format and the endpoint now agree, which is the mechanism both issues identify.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -1,4 +1,4 @@
-import { detectFormat, getTargetFormat, resolveTransport } from "../services/provider.js";
+import { detectFormat } from "../services/provider.js";
 import { translateRequest } from "../translator/index.js";
 import { applyThinking, extractThinking, stripThinkingSuffix } from "../translator/concerns/thinkingUnified.js";
 import { FORMATS } from "../translator/formats.js";
@@ -6,7 +6,7 @@ import { normalizeClaudePassthrough, anchorClaudeCache } from "../translator/for
 import { createStreamController } from "../utils/streamHandler.js";
 import { refreshWithRetry } from "../services/tokenRefresh.js";
 import { createRequestLogger } from "../utils/requestLogger.js";
-import { getModelTargetFormat, getModelSupportedFormats, getModelStrip, getModelUpstreamId, getModelType, PROVIDER_ID_TO_ALIAS } from "../config/providerModels.js";
+import { getModelStrip, getModelUpstreamId, getModelType, PROVIDER_ID_TO_ALIAS } from "../config/providerModels.js";
 import { PROVIDERS } from "../config/providers.js";
 import { createErrorResult, parseUpstreamError, formatProviderError } from "../utils/error.js";
 import { HTTP_STATUS, TOKEN_SAVER_HEADER } from "../config/runtimeConfig.js";
@@ -15,6 +15,7 @@ import { trackPendingRequest, appendRequestLog, saveRequestDetail } from "@/lib/
 import { getExecutor } from "../executors/index.js";
 import { supportsGrokCliReasoningEffort } from "../config/grokCli.js";
 import { buildRequestDetail, extractRequestConfig } from "./chatCore/requestDetail.js";
+import { resolveUpstreamRoute } from "./chatCore/upstreamRoute.js";
 import { handleForcedSSEToJson } from "./chatCore/sseToJsonHandler.js";
 import { handleNonStreamingResponse } from "./chatCore/nonStreamingHandler.js";
 import { handleStreamingResponse, buildOnStreamComplete } from "./chatCore/streamingHandler.js";
@@ -77,20 +78,10 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
   if (bypassResponse) return bypassResponse;
 
   const alias = PROVIDER_ID_TO_ALIAS[provider] || provider;
-  const modelTargetFormat = getModelTargetFormat(alias, model);
   // Multi-endpoint providers: pick transport matching sourceFormat → zero translation.
-  // Per-model guard: only use the transport when the model declares support for that
-  // sourceFormat — opencode-go models differ in endpoint support (kimi/glm only do
-  // /chat/completions), so without this guard a claude-format request would wrongly
-  // route kimi to /messages.
-  const modelSupportedFormats = getModelSupportedFormats(alias, model);
-  const runtimeTransport = resolveTransport(provider, sourceFormat);
-  // Per-model guard: when a model declares supportedFormats, only use the
-  // sourceFormat-matched transport if that format is declared (opencode-go models
-  // differ — kimi/glm only do /chat/completions). Undeclared models keep the
-  // upstream default (use the transport), preserving behavior for glm/deepseek/...
-  const useTransport = (!modelSupportedFormats || modelSupportedFormats.includes(sourceFormat)) ? runtimeTransport : null;
-  const targetFormat = modelTargetFormat || useTransport?.format || getTargetFormat(provider, credentials);
+  // A model-level targetFormat overrides that choice, and the transport follows it so
+  // the body format and the endpoint never diverge.
+  const { targetFormat, transport: useTransport } = resolveUpstreamRoute({ provider, alias, model, sourceFormat, credentials });
   if (useTransport && credentials) credentials.runtimeTransport = useTransport;
   const stripList = getModelStrip(alias, model);
   const upstreamModel = getModelUpstreamId(alias, model);
--- a/open-sse/handlers/chatCore/upstreamRoute.js
+++ b/open-sse/handlers/chatCore/upstreamRoute.js
@@ -1,0 +1,26 @@
+import { getTargetFormat, resolveTransport } from "../../services/provider.js";
+import { getModelSupportedFormats, getModelTargetFormat } from "../../config/providerModels.js";
+
+// Resolve the wire format of the outbound body and the transport that carries it.
+//
+// Invariant: the transport must speak the format the body is serialized in. A
+// multi-endpoint provider picks its endpoint, headers and auth scheme from the
+// transport, so a Claude body sent over the OpenAI transport lands on
+// /v1/chat/completions with bearer auth — the upstream parses what it can and
+// silently drops the rest (images vanish, tool schemas 400).
+export function resolveUpstreamRoute({ provider, alias, model, sourceFormat, credentials }) {
+  const modelTargetFormat = getModelTargetFormat(alias, model);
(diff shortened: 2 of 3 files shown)


ntdat812:
Closing this: `master` has since decided the opposite of what it proposes, deliberately and with a documented reason.

This PR pinned the endpoint to a model's declared `targetFormat` for every client wire format, so that the body format and the endpoint could never diverge. `open-sse/handlers/chatCore.js` on current `master` now resolves it the other way round:

```js
// A source-format-matched endpoint keeps the request lossless. Prefer it
// over a model-level targetFormat, which is only the fallback for clients
// whose wire format has no supported transport (for example MiniMax-M3:
// OpenAI clients should stay on /chat/completions; other clients can fall
// back to its declared Claude target).
const targetFormat = useTransport?.format || modelTargetFormat || getTargetFormat(provider, credentials);
```

`tests/unit/upstream-route-alignment.test.js` in this branch asserts that an OpenAI client on `minimax/MiniMax-M3` lands on the Claude transport — which is exactly the case that comment singles out as wrong. Rebasing would therefore not resolve a text conflict; it would quietly revert a design decision, and the tests shipped here would enforce the revert. That is not a call for a PR to make on its own.

The per-model `supportedFormats` guard this PR also introduced is already on `master`. What is left unaddressed is the narrower invariant behind the original report — that when a model-level `targetFormat` *is* the one chosen, the transport should move with it rather than staying on the source-format endpoint. If that is still worth having, it is a much smaller change on top of the current precedence, and I am happy to open it as such.

PR #3539, opened 2026-08-25
fix(providers): let self-hosted TTS/STT/embedding set their base URL (#3467)

### LINKED ISSUE #3467: Self-hosted TTS/STT connections cannot configure providerSpecificData.baseUrl in the dashboard
## Summary

The Self-hosted TTS and Self-hosted STT providers read their upstream endpoint from:

```js
connection.providerSpecificData.baseUrl
```

However, the dashboard does not provide a Base URL field for either provider when adding or editing a connection.

As a result, connections created through the UI are saved without a `baseUrl` and silently use the built-in localhost defaults:

- Self-hosted TTS: `http://localhost:8880`
- Self-hosted STT: `http://localhost:8080/v1/audio/transcriptions`

In a Docker deployment, these localhost addresses refer to the 9router container itself, so separately hosted TTS/STT services cannot be reached.

## Environment

- 9router: `0.5.55`
- Deployment: Docker Compose
- Upstream: OpenAI-compatible gateway providing:
  - `POST /v1/audio/speech`
  - `POST /v1/audio/transcriptions`

## Steps to reproduce

1. Open **Providers**.
2. Select **Self-hosted TTS** or **Self-hosted STT**.
3. Click **Add API Key**.
4. Observe that the form contains Name, Priority, and API Key, but no Base URL field.
5. Save the connection.
6. Inspect the saved connection or make a request.

The saved connection has no usable endpoint override:

```json
{
  "providerSpecificData": {
    "connectionProxyEnabled": false,
    "connectionProxyUrl": "",
    "connectionNoProxy": ""
  }
}
```

The request then falls back to the provider's localhost default and fails in Docker.

The Edit Connection modal also has no Base URL field, so the endpoint cannot be corrected afterward.

## Expected behavior

The Add and Edit Connection forms should expose a Base URL field for these providers.

### Self-hosted TTS

Example input:

```text
http://ai-gateway:8100
```

Saved as:

```json
{
  "providerSpecificData": {
"baseUrl": "http://ai-gateway:8100"
  }
}
```

9router should append:

```text
/v1/audio/speech
```

### Self-hosted STT

Example input:

```text
http://ai-gateway:8100/v1/audio/transcriptions
```

Saved as:

```json
{
  "providerSpecificData": {
    "baseUrl": "http://ai-gateway:8100/v1/audio/transcriptions"
  }
}
```

The full transcription URL should be used as-is.

## Actual behavior

The backend and adapters support `providerSpecificData.baseUrl`, but the dashboard cannot set it.

Relevant code:

- `open-sse/providers/registry/selfhosted-tts.js`
- `open-sse/providers/registry/selfhosted-stt.js`
- `open-sse/handlers/ttsProviders/selfhostedTts.js`
- `src/app/(dashboard)/dashboard/providers/[id]/AddApiKeyModal.js`
- `src/shared/components/EditConnectionModal.js`

`AddApiKeyModal.buildProviderSpecificData()` currently creates `baseUrl` only for Ollama Local. Self-hosted TTS/STT fall through to `undefined`.

## Verified workaround

I manually added the values to the relevant `providerConnections.data` JSON in SQLite:

```json
{
  "providerSpecificData": {
    "baseUrl": "http://host.docker.internal:8100"
  }
}
```

and:

```json
{
  "providerSpecificData": {
    "baseUrl": "http://host.docker.internal:8100/v1/audio/transcriptions"
  }
}
```

After this change, both routes worked end-to-end:

- Self-hosted TTS returned valid MP3 audio through 9router.
- A non-default VOICEVOX speaker could be selected.
- The generated audio was sent through `selfhosted-stt/whisper-1`.
- STT returned HTTP 200 with a valid transcription.

This confirms that routing and adapters work; only the dashboard configuration path is missing.

## Suggested fix

Add provider-specific Base URL inputs to both:

- `AddApiKeyModal.js`
- `EditConnectionModal.js`

for:

```js
provider === "selfhosted-tts"
provider === "selfhosted-stt"
```

Then include the normalized value in:

```js
providerSpecificData.baseUrl

```

It would also be useful to:

1. Validate that Self-hosted TTS has a server root or full speech endpoint.
2. Validate that Self-hosted STT has the complete transcription endpoint.
3. Preserve the Base URL when editing other connection fields.
4. Add regression tests for add/edit payload construction.

ntdatt812:
Fixes #3467.

## The backend asks for a value the UI never collects

All three self-hosted providers read their endpoint from `providerSpecificData.baseUrl`, and their registry entries tell the user to set it:

```js
// selfhosted-tts
text: "Set providerSpecificData.baseUrl to the server root, e.g. http://host:8080 — /v1/audio/speech is appended."
```

The dashboard renders no field for it. So a connection created through the UI is saved without one and falls back to the built-in default:

| provider | default it falls back to |
|---|---|
| Self-hosted TTS | `http://localhost:8880` |
| Self-hosted STT | `http://localhost:8080/v1/audio/transcriptions` |
| Self-hosted Embedding | `http://localhost:8080/v1` |

Inside Docker `localhost` is the 9router container itself, so a separately hosted server is simply unreachable — and nothing in the UI says why.

## One mechanism, not two

The modal already had a base-URL field. It was hardcoded to a single provider:

```js
const isOllamaLocal = provider === "ollama-local";
...
{isOllamaLocal && <Input label="Ollama Host URL" placeholder="http://localhost:11434" ... />}
```

Rather than adding a second special case beside it, a provider whose endpoint is per-connection now declares it:

```js
baseUrlField: { label: "Base URL", placeholder: "http://localhost:8880", help: "Server root — /v1/audio/speech is appended." }
```

`buildProviderEntry` passes it through to the UI (next to `regions`, `authType`, `authHint`, which already work this way), and the modal renders whatever the registry declares. `ollama-local` moves onto the same mechanism with its label and placeholder preserved exactly, so there is one path instead of two.

Each placeholder is that provider's **real** fallback, so the box shows what the connection would otherwise silently use.

## Testability

`buildProviderSpecificData` moves out of the modal into `src/shared/utils/providerSpecificData.js`, so the saved shape can be tested without rendering — the saved value is what the executors read, and that is exactly what was wrong here.

Behaviour preserved: an empty box still saves nothing. Empty means "keep the provider's default", not "save an empty endpoint".

## Tests

`tests/unit/provider-base-url-field-3467.test.js` — 12 cases:

- all four providers declare a `baseUrlField` with a non-empty label and a placeholder
- each placeholder equals the default that provider actually falls back to (`ttsConfig.baseUrl`, `sttConfig.baseUrl`) — so the two cannot drift
- providers with a fixed endpoint declare none
- a typed value is trimmed and saved; empty / whitespace / `undefined` / `null` save nothing
- a typed value is ignored for a provider that does not declare the field
- the Azure, Cloudflare and region shapes are unchanged

Mutation-checked, twice:

```
# remove the baseUrl branch from the builder
× saves a trimmed baseUrl when the provider declares the field      1 failed | 11 passed

# remove baseUrlField from the selfhosted-tts registry entry
× selfhosted-tts declares a baseUrlField
× each placeholder matches the default that provider actually falls back to
                                                                    2 failed | 10 passed
```

## Suite

| | Test Files | Tests |
|---|---|---|
| with this PR | 31 failed / 163 passed | 82 failed / **1799** passed |

Same 82 pre-existing failures, +12 — this PR's own cases. ESLint clean on both touched files.

## Not in scope

There is no generic edit modal for an API-key connection, so this covers creation. Changing an existing connection's endpoint still means recreating it — worth a separate issue if you want it.

### DIFF
--- a/open-sse/providers/registry/ollama-local.js
+++ b/open-sse/providers/registry/ollama-local.js
@@ -11,6 +11,11 @@ export default {
     website: "https://ollama.com",
   },
   category: "apikey",
+  // The connection's own endpoint, stored as providerSpecificData.baseUrl.
+  baseUrlField: {
+    label: "Ollama Host URL",
+    placeholder: "http://localhost:11434",
+  },
   transport: {
     baseUrl: "http://localhost:11434/api/chat",
     format: "ollama",
--- a/open-sse/providers/registry/selfhosted-embedding.js
+++ b/open-sse/providers/registry/selfhosted-embedding.js
@@ -40,6 +40,12 @@ export default {
     website: "https://github.com/ggml-org/llama.cpp",
   },
   category: "apikey",
+  // Same gap as the TTS and STT entries.
+  baseUrlField: {
+    label: "Base URL",
+    placeholder: "http://localhost:8080/v1",
+    help: "OpenAI base URL — /embeddings is appended.",
+  },
   auth: {
     apiKey: {
       // Note the /v1: the adapter appends "/embeddings" to whatever it is given,
--- a/open-sse/providers/registry/selfhosted-stt.js
+++ b/open-sse/providers/registry/selfhosted-stt.js
@@ -28,6 +28,13 @@ export default {
     website: "https://github.com/ggml-org/whisper.cpp",
   },
   category: "apikey",
+  // See the TTS entry: without this the dashboard cannot point the connection
+  // at a separately hosted server.
+  baseUrlField: {
+    label: "Base URL",
+    placeholder: "http://localhost:8080/v1/audio/transcriptions",
+    help: "Full transcriptions endpoint.",
+  },
   auth: {
     apiKey: {
       text: "Set providerSpecificData.baseUrl to the full transcriptions URL, e.g. http://host:8080/v1/audio/transcriptions. The API key is not checked by local servers; any value works.",
--- a/open-sse/providers/registry/selfhosted-tts.js
+++ b/open-sse/providers/registry/selfhosted-tts.js
@@ -22,6 +22,14 @@ export default {
     website: "https://github.com/remsky/Kokoro-FastAPI",
   },
   category: "apikey",
+  // Without this the dashboard saves no baseUrl and the connection silently
+  // falls back to the localhost default below, which in Docker points at the
+  // 9router container itself.
+  baseUrlField: {
+    label: "Base URL",
+    placeholder: "http://localhost:8880",
+    help: "Server root — /v1/audio/speech is appended.",
+  },
   auth: {
     apiKey: {
       text: "Set providerSpecificData.baseUrl to the server root, e.g. http://host:8080 — /v1/audio/speech is appended. The API key is not checked by local servers; any value works.",
--- a/src/app/(dashboard)/dashboard/providers/[id]/AddApiKeyModal.js
+++ b/src/app/(dashboard)/dashboard/providers/[id]/AddApiKeyModal.js
@@ -5,6 +5,7 @@ import PropTypes from "prop-types";
 import { Button, Badge, Input, Modal, Select } from "@/shared/components";
 import { AI_PROVIDERS } from "@/shared/constants/providers";
 import { planBulkAdd } from "@/shared/utils/bulkAdd";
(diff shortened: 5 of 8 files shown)


ntdatt812:

Issue #3488, opened 2026-08-23
Usage rows silently lost when client aborts a streaming request (DISCONNECT: ResponseAborted) + sporadic rolled-back inserts skip usageHistory IDs


cryptyroot-ux:
# Title

Usage rows silently lost when client aborts a streaming request (`DISCONNECT: ResponseAborted`) + sporadic rolled-back inserts skip `usageHistory` IDs

# Environment

- 9Router **v0.5.55** (Docker image `decolua/9router:latest`, digest `sha256:f00fe389ef41a…`)
- DB driver: better-sqlite3 (WAL mode), single container
- Clients: OpenAI-compatible `/v1/chat/completions`, combo routing and direct provider models both affected

# Bug 1 — Aborted streaming requests are never recorded

## Steps to reproduce

```bash
# 1) note current max id
sqlite3 data.sqlite "SELECT MAX(id) FROM usageHistory;"          # -> 168

# 2) start a streaming request and kill the client mid-stream
curl -sN -m 3 http://localhost:20128/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" \
  -d '{"model":"oc/hy3-free","stream":true,"max_tokens":400,
       "messages":[{"role":"user","content":"Count slowly from 1 to 100"}]}' \
  | head -c 200   # client goes away after ~2-3 s

# 3) wait and re-check
sleep 6
sqlite3 data.sqlite "SELECT MAX(id) FROM usageHistory;"          # -> still 168
```

Server log shows the request was served and then disconnected:

```
[14:13:04] ▶ POST oc/hy3-free → opencode/hy3-free · FMT: openai→openai · STREAM · 1 MSG · ACC:Public
[14:13:06] ⚡ DISCONNECT: ResponseAborted · opencode/hy3-free · 2255ms
```

## Actual result

- No row in `usageHistory`, nothing in Recent Requests, no error logged.
- The upstream provider has already consumed tokens for the partial generation, so usage accounting silently under-counts.

## Expected result

A usage record should exist for aborted streams (e.g. `status:"aborted"`), or at minimum an entry should reach the in-memory recent-requests ring so the dashboard reflects the activity.

## Code pointer (v0.5.55)

In `open-sse/handlers/chatCore/streamingHandler.js` the recording calls are wired **only** into the `onStreamComplete` callback:

```js
onStreamComplete: (parsed, tokens, ttftRef) => {
  ...
[...]

PR #3542, opened 2026-08-25
fix(stream): record the turn when a client hangs up mid-stream (#3488)

ntdatt812:
Fixes bug 1 of #3488.

## flush() and cancel() are alternatives, and only flush() recorded

Every recording side effect — request detail, usage row, the `📊 done` line — hangs off `onStreamComplete`, and `onStreamComplete` was only called from the transform stream's `flush()`.

Per the Streams spec, `flush()` runs when the **writable** side ends and `cancel()` runs when the **readable** side is cancelled. One or the other, never both. Confirmed on the Node this ships with:

```js
const ts = new TransformStream({ transform, flush: () => log("flush"), cancel: r => log("cancel:" + r) });
reader.cancel("client_closed");
await writer.close();          // throws TypeError
// -> ["cancel:client_closed", "close threw: TypeError"]
```

So when a client goes away mid-stream the cancel path runs, `flush` never does, and nothing is recorded — no `usageHistory` row, no request detail, nothing in Recent Requests — even though the provider has already generated (and charged for) the partial answer. Agentic CLI clients cancel aggressively, so the dashboard can look idle while traffic flows. That is exactly the `DISCONNECT: ResponseAborted` trace in the report.

## The fix

Give the stream a `cancel()` that records what was accumulated, and route **both** endings through one `finishStream()` so the two paths cannot drift apart.

`onStreamComplete` gains a fourth argument, so an aborted turn is stored as `status: "aborted"` rather than passed off as a success — the issue asked for either that or at least an entry in the ring, and this gives both.

**Both stream modes are covered.** TRANSLATE keeps usage on `state`, PASSTHROUGH in a closure variable; the cancel path reads whichever applies. Reading only the first would have left half the providers reporting nothing on abort — I had that wrong in the first draft and the passthrough test caught it.

## Tests

`tests/unit/aborted-stream-usage-3488.test.js` — 8 cases driving a real transform stream and cancelling its reader, no network:

- a hang-up still reports the stream as finished
- it is reported as **aborted**, not as a normal completion
- the content generated before the hang-up is kept
- usage is reported, so the partial generation is accounted for
- a normal completion still reports exactly once, and not as aborted
- a late cancel after a normal completion does not report a second time
- a stream cancelled before anything arrived reports once, as aborted, without inventing content
- a passthrough stream records the partial turn too

Mutation-checked — removing the `cancel()` branch reproduces the bug:

```
× still reports the stream as finished
× reports it as aborted, not as a normal completion
× keeps the content generated before the hang-up
× reports usage so the partial generation is accounted for
× reports once, as aborted, without inventing content
Tests  5 failed | 2 passed
```

Two things I could **not** pin with a failing mutation, stated rather than glossed:

- the once-only `streamFinished` guard. I tried to make both endings fire and could not: after `cancel`, `writer.close()` throws and `flush` never runs. The guard is defensive against a future path that calls both, not load-bearing today.
- in the passthrough branch, reported-usage versus estimated-usage. The estimate fallback covers a null, so swapping the source still passes. The test asserts usage is present and non-zero, which is what it can prove.

## Suite

| | Test Files | Tests |
|---|---|---|
| with this PR | 31 failed / 163 passed | 82 failed / **1795** passed |

Same 82 pre-existing failures, +8 — this PR's own cases. ESLint clean on all three files.

## Not in scope

Bug 2 — a rolled-back insert skipping a `usageHistory` id, roughly 1 in 170 — is a different mechanism (`SQLITE_BUSY` contention, fixed by `PRAGMA busy_timeout` or a retry wrapper). I have no way to reproduce a 1-in-170 race here, so I have left it alone rather than ship an unverified change alongside a verified one.

## Runtime scope: Node only

**This fix does not fire under `start:bun`.** Every disconnect signal in Next descends from `res` `'close'`, and Bun's `node:http` `ServerResponse` never emits it on a client hangup — so nothing cancels the stream and `finishStream` only ever runs from `flush`, exactly as before. Measured on a bare `node:http` server with a real aborted `fetch`:

| Listener | Node 24.16.0 | Bun 1.3.14 |
| --- | --- | --- |
| `res` `'close'` | fires | **never** |
| `req` `'close'` | fires | fires |

Node is the runtime the Docker image ships (`node:22-alpine`), so this is a strict improvement where it runs, but the limitation is real and is now stated in the code next to `finishStream`.

The portable hook is `req` `'close'`, tracked in **#3559** — which also covers the larger consequence found while reviewing this PR: on Bun `handleDisconnect` never runs either, so the **upstream provider request is never aborted** and keeps being billed for a stream nobody is reading. That is a bigger and different bug than a missing usage row, and it belongs in its own PR with its own regression test.

Credit to @hiepau1231 for isolating the root cause and quantifying the waste.

### DIFF
--- a/open-sse/handlers/chatCore/streamingHandler.js
+++ b/open-sse/handlers/chatCore/streamingHandler.js
@@ -113,12 +113,13 @@ export async function handleStreamingResponse({ providerResponse, provider, mode
 export function buildOnStreamComplete({ provider, model, connectionId, apiKey, requestStartTime, body, stream, finalBody, translatedBody, clientRawRequest, pxpipe, reqTag, log }) {
   const streamDetailId = `${Date.now()}-${Math.random().toString(36).slice(2, 11)}`;
 
-  const onStreamComplete = (contentObj, usage, ttftAt) => {
+  const onStreamComplete = (contentObj, usage, ttftAt, { aborted = false } = {}) => {
     const latency = {
       ttft: ttftAt ? ttftAt - requestStartTime : Date.now() - requestStartTime,
       total: Date.now() - requestStartTime
     };
-    const safeContent = contentObj?.content || "[Empty streaming response]";
+    const safeContent =
+      contentObj?.content || (aborted ? "[Aborted streaming response]" : "[Empty streaming response]");
     const safeThinking = contentObj?.thinking || null;
 
     saveRequestDetail(buildRequestDetail({
@@ -130,13 +131,13 @@ export function buildOnStreamComplete({ provider, model, connectionId, apiKey, r
       providerResponse: safeContent,
       response: { content: safeContent, thinking: safeThinking, type: "streaming" },
       pxpipe,
-      status: "success"
+      status: aborted ? "aborted" : "success"
     }, { id: streamDetailId })).catch(err => {
       console.error("[RequestDetail] Failed to update streaming content:", err.message);
     });
 
     // Persist stream usage to DB (no console line; the "📊 done" line below is authoritative)
-    saveUsageStats({ provider, model, tokens: usage, connectionId, apiKey, endpoint: clientRawRequest?.endpoint, label: "STREAM USAGE", silent: true });
+    saveUsageStats({ provider, model, tokens: usage, connectionId, apiKey, endpoint: clientRawRequest?.endpoint, label: aborted ? "STREAM USAGE (aborted)" : "STREAM USAGE", silent: true });
     if (log?.line) log.line(reqTag, "📊", formatDoneLine({ usage, latency }));
   };
 
--- a/open-sse/utils/stream.js
+++ b/open-sse/utils/stream.js
@@ -79,7 +79,15 @@ export function createSSEStream(options = {}) {
 
   // Usage/logging tail, callable from transform() as well as flush(): a client that
   // closes right after the terminal event cancels the reader, and flush() never runs.
-  const finalizeStream = () => {
+  //
+  // RUNTIME SCOPE: the aborted path is Node-only. Every disconnect signal in Next
+  // — and therefore `cancel` in any of its shapes — descends from `res` `'close'`,
+  // which Bun's node:http ServerResponse never emits on a client hangup. Under
+  // `start:bun` nothing cancels this stream, so finalizeStream() only ever runs
+  // from flush(), exactly as before this fix. The portable hook is `req` `'close'`;
+  // see #3559, which also covers the larger consequence that the upstream provider
+  // request is never aborted on Bun either.
+  const finalizeStream = ({ aborted = false } = {}) => {
     if (finalized) return;
     finalized = true;
 
@@ -101,7 +109,7 @@ export function createSSEStream(options = {}) {
       onStreamComplete({
         content: accumulatedContent,
         thinking: accumulatedThinking
-      }, finalUsage, ttftAt);
+      }, finalUsage, ttftAt, { aborted });
     }
   };
 
@@ -482,6 +490,21 @@ export function createSSEStream(options = {}) {
       } catch (error) {
         console.log("Error in flush:", error);
         finalizeStream();
+      }
+    },
+
+    // The client hung up mid-stream. The Streams spec calls flush() or cancel(),
+    // never both, and finalizeStream() had no cancel() caller — so an aborted
+    // request left no usage row, no request detail and nothing in Recent
+    // Requests, even though the provider had already generated (and charged for)
+    // the partial answer (#3488). finalizeStream() estimates from
+    // totalContentLength when the provider reported no usage, which is the usual
+    // case here: the usage event arrives with the terminal chunk that never came.
+    cancel() {
+      try {
+        finalizeStream({ aborted: true });
+      } catch (error) {
+        console.log("Error in cancel:", error);
       }
     }
   });
--- a/tests/unit/aborted-stream-usage-3488.test.js
+++ b/tests/unit/aborted-stream-usage-3488.test.js
@@ -1,0 +1,214 @@
+import { describe, expect, it, vi } from "vitest";
+
+import { createSSETransformStreamWithLogger } from "../../open-sse/utils/stream.js";
+import { FORMATS } from "../../open-sse/translator/formats.js";
+
+/**
+ * Recording lived only in the transform stream's `flush`. The Streams spec
+ * calls `flush` when the writable side ends and `cancel` when the readable
+ * side is cancelled — one or the other, never both — so a client that hung up
+ * mid-stream produced no usage row, no request detail and nothing in Recent
+ * Requests, while the provider had already generated the partial answer
+ * (#3488).
+ */
+const encoder = new TextEncoder();
+
+function chunk(text) {
+  return encoder.encode(
+    `data: ${JSON.stringify({
+      choices: [{ index: 0, delta: { content: text }, finish_reason: null }],
+    })}\n\n`,
+  );
+}
+
+/** A frame carrying provider-reported usage, which the fallback estimate must not stand in for. */
+function usageChunk(usage) {
+  return encoder.encode(
+    `data: ${JSON.stringify({
+      choices: [{ index: 0, delta: {}, finish_reason: null }],
+      usage,
+    })}\n\n`,
+  );
+}
+
+/** Feed content, then a provider usage frame, then hang up. */
+async function abortAfterUsage(stream, usage) {
+  const reader = stream.readable.getReader();
+  const writer = stream.writable.getWriter();
+  // A pump rather than paired reads: the usage frame carries no delta content, so
+  // the stream need not surface anything for it, and a `write` with nobody pulling
+  // parks on backpressure forever.
+  const pump = (async () => {
+    try {
+      for (;;) {
+        const { done } = await reader.read();
+        if (done) return;
+      }
+    } catch {
+      // The cancel below is what ends this loop.
+    }
+  })();
+  await writer.write(chunk("Hello"));
+  await writer.write(usageChunk(usage));
+  await new Promise((resolve) => setTimeout(resolve, 50));
+  await reader.cancel("client_closed");
+  await pump;
+  await new Promise((resolve) => setTimeout(resolve, 20));
+}
+
+function build(onStreamComplete) {
+  return createSSETransformStreamWithLogger(
+    FORMATS.OPENAI,
+    FORMATS.OPENAI,
+    "openai",
+    null,
+    null,
+    "gpt-4.1",
+    "conn-1",
+    { messages: [{ role: "user", content: "hi" }] },
+    onStreamComplete,
+    "sk-test",
+  );
+}
+
+/** Feed some content, then hang up the way a disconnecting client does. */
+async function abortMidStream(stream) {
+  const reader = stream.readable.getReader();
+  const writer = stream.writable.getWriter();
+  writer.write(chunk("Hello"));
+  await reader.read();
+  writer.write(chunk(" world"));
+  await reader.read();
+  await reader.cancel("client_closed");
+  await new Promise((resolve) => setTimeout(resolve, 20));
+}
+
+/** Feed some content and let the upstream finish normally. */
+async function completeStream(stream) {
+  const reader = stream.readable.getReader();
+  const writer = stream.writable.getWriter();
+  const drain = (async () => {
+    for (;;) {
+      const { done } = await reader.read();
+      if (done) return;
+    }
+  })();
+  await writer.write(chunk("Hello"));
+  await writer.close();
+  await drain;
+  // Released so a caller can still cancel the readable afterwards. A locked
+  // readable rejects `cancel()` with `TypeError: Invalid state: ReadableStream
+  // is locked`, which a swallowing `.catch()` turns into a test that asserts
+  // nothing about cancellation at all.
+  reader.releaseLock();
+}
+
+describe("a client that hangs up mid-stream", () => {
+  it("still reports the stream as finished", async () => {
+    const onStreamComplete = vi.fn();
+    await abortMidStream(build(onStreamComplete));
+    expect(onStreamComplete).toHaveBeenCalledTimes(1);
+  });
+
+  it("reports it as aborted, not as a normal completion", async () => {
+    const onStreamComplete = vi.fn();
+    await abortMidStream(build(onStreamComplete));
+    const [, , , meta] = onStreamComplete.mock.calls[0];
+    expect(meta?.aborted).toBe(true);
+  });
+
+  it("keeps the content generated before the hang-up", async () => {
+    const onStreamComplete = vi.fn();
+    await abortMidStream(build(onStreamComplete));
+    const [contentObj] = onStreamComplete.mock.calls[0];
+    expect(contentObj.content).toContain("Hello");
+  });
+
+  it("reports usage so the partial generation is accounted for", async () => {
+    const onStreamComplete = vi.fn();
+    await abortMidStream(build(onStreamComplete));
+    const [, usage] = onStreamComplete.mock.calls[0];
+    expect(usage).toBeTruthy();
+    expect(usage.completion_tokens).toBeGreaterThan(0);
+  });
+});
+
+describe("a stream that ends normally", () => {
+  it("still reports exactly once, and not as aborted", async () => {
+    const onStreamComplete = vi.fn();
+    await completeStream(build(onStreamComplete));
+    expect(onStreamComplete).toHaveBeenCalledTimes(1);
+    const [, , , meta] = onStreamComplete.mock.calls[0];
+    expect(meta?.aborted ?? false).toBe(false);
+  });
+
+  // Named for what it is. `cancel()` on an already-closed readable resolves without
+  // running the underlying cancel algorithm, so this canNOT prove the exactly-once
+  // guard -- removing that guard leaves this green. It is an API smoke test.
+  it("accepts a late cancel on an already-closed readable without reporting again", async () => {
+    const onStreamComplete = vi.fn();
+    const stream = build(onStreamComplete);
+    await completeStream(stream);
+    expect(onStreamComplete).toHaveBeenCalledTimes(1);
+
+    // Asserted, not swallowed: if this rejects the cancel never happened and the
+    // count below would prove nothing.
+    await expect(stream.readable.cancel("late")).resolves.toBeUndefined();
+    await new Promise((resolve) => setTimeout(resolve, 20));
+    expect(onStreamComplete).toHaveBeenCalledTimes(1);
+  });
+});
+
+describe("a stream cancelled before anything arrived", () => {
+  it("reports once, as aborted, without inventing content", async () => {
+    const onStreamComplete = vi.fn();
+    const stream = build(onStreamComplete);
+    await stream.readable.cancel("client_closed");
+    await new Promise((resolve) => setTimeout(resolve, 20));
+    expect(onStreamComplete).toHaveBeenCalledTimes(1);
+    const [contentObj, , , meta] = onStreamComplete.mock.calls[0];
+    expect(meta?.aborted).toBe(true);
+    expect(contentObj.content).toBe("");
+  });
+});
+
+describe("passthrough streams", () => {
+  /**
+   * TRANSLATE keeps usage on `state`; PASSTHROUGH keeps it in a closure
+   * variable. The cancel path reads whichever applies, so both modes record
+   * something rather than one of them reporting nothing.
+   */
+  it("also record the partial turn when the client hangs up", async () => {
+    const { createPassthroughStreamWithLogger } = await import("../../open-sse/utils/stream.js");
+    const onStreamComplete = vi.fn();
+    const stream = createPassthroughStreamWithLogger(
+      "openai",
+      null,
+      "gpt-4.1",
+      "conn-1",
+      { messages: [{ role: "user", content: "hi" }] },
+      onStreamComplete,
+      "sk-test",
+    );
+
+    await abortAfterUsage(stream, {
+      prompt_tokens: 11,
+      completion_tokens: 7,
+      total_tokens: 18,
+    });
+
+    expect(onStreamComplete).toHaveBeenCalledTimes(1);
+    const [contentObj, usage, , meta] = onStreamComplete.mock.calls[0];
+    expect(meta?.aborted).toBe(true);
+    expect(contentObj.content).toContain("Hello");
+    expect(usage?.completion_tokens).toBeGreaterThan(0);
+
+    // Characterization, not an endorsement. A usage-only frame is dropped by the
+    // `hasValuableContent` guard BEFORE `extractUsage` runs, so the closure never
+    // sees the provider's numbers and the abort path falls back to `estimateUsage`.
+    // Asserting the provider values here fails today (11 -> 2012). Pinned so the
+    // gap is visible and this test starts failing the moment it is closed, rather
+    // than quietly passing on an estimate as it did before.
+    expect(usage.prompt_tokens).not.toBe(11);
+  });
+});


hiepau1231 (changes requested):
## Request changes

I found one production blocker and two test gaps.

### [P1] `open-sse/utils/stream.js` around `cancel()`

`cancel()` is added as a member of the object passed to `new TransformStream(...)`. In the standard Web Streams API, transformer callbacks are `start`, `transform`, and `flush`; readable-side cancellation is handled by the underlying readable source, not by a transformer `cancel()` method. Therefore `stream.readable.cancel()` will not invoke this code, so `finishStream(reported, { aborted: true })` remains unreachable for a client disconnect and the #3488 usage/request-detail regression remains.

Please move the accounting hook to the actual readable cancellation boundary or the HTTP request abort/close signal.

### [P2] `tests/unit/aborted-stream-usage-3488.test.js` around the late-cancel test

`completeStream()` keeps the reader lock and never calls `releaseLock()`. The subsequent `stream.readable.cancel("late")` therefore rejects with `TypeError`, and the test swallows it with `.catch(() => {})`. The assertion can pass without exercising cancellation or exactly-once behavior. The stream is also already closed, so even an unlocked cancel would not test the intended race.

Please cancel while the readable is still open after the first completion callback, assert that cancellation succeeds, and then assert the callback count.

### [P2] `tests/unit/aborted-stream-usage-3488.test.js` around the passthrough test

The `chunk()` helper never emits provider-reported usage. The `completion_tokens > 0` assertion can therefore pass using the `estimateUsage()` fallback, without testing the passthrough closure usage path.

Please add a frame with explicit provider usage and assert the exact usage values received by `onStreamComplete`.

ntdatt812:
Thank you — the two test findings were both right, and following the third one turned up a real gap I had papered over. The P1 I have to push back on, with a measurement.

## P1 — `transformer.cancel()` does fire

`cancel` was added to the transformer interface in the Streams Standard in 2023 (whatwg/streams#1290), alongside `start` / `transform` / `flush`, and Node implements it. Measured on this machine:

```js
const ts = new TransformStream({
  start(){seen.push('start')}, transform(c,ctl){ctl.enqueue(c)},
  flush(){seen.push('flush')}, cancel(reason){seen.push('cancel:'+reason)},
});
ts.writable.getWriter().write('x');
await ts.readable.cancel('client-hangup');
```

```
node v24.16.0
after readable.cancel -> ["start","cancel:client-hangup"]
```

`flush` is absent and `cancel` ran, which is the contract the PR relies on. The repo declares no `engines` floor and the only Node pin in CI is `node-version: 24`, so this is the runtime it targets. The four tests in this file that already drive `readable.cancel()` also pass — they could not if the hook were unreachable.

If your concern is the step before that — whether a real client disconnect makes the runtime cancel the response body's readable, rather than whether the hook fires once it does — say so and I will chase that end specifically. That is a different claim from the one about the transformer interface, and it is the one worth nailing down.

## P2 — reader lock: correct, and worse than described

Confirmed the exact failure mode:

```
node -e "const ts=new TransformStream(); ts.readable.getReader(); ts.readable.cancel('late')"
REJECTED: TypeError: Invalid state: ReadableStream is locked
```

`completeStream()` held the lock, the cancel rejected, and `.catch(() => {})` ate it — so the exactly-once assertion passed without any cancellation happening. `completeStream()` now calls `releaseLock()`, and the cancel is asserted as resolving rather than swallowed:

```js
await expect(stream.readable.cancel("late")).resolves.toBeUndefined();
```

## P2 — provider usage: correct, and it exposed a product gap

You were right that `completion_tokens > 0` could be met by `estimateUsage()`. I added a frame carrying explicit provider usage — and the assertion **fails**:

```
AssertionError: expected { prompt_tokens: 2012, … } to match object { prompt_tokens: 11, … }
```

Not a timing artefact — raising the settle from 10 ms to 200 ms changes nothing. The cause is in the passthrough transform:

```js
if (!hasValuableContent(parsed, FORMATS.OPENAI)) {
  continue;                 // ← usage-only frame dropped here
}
…
const extracted = extractUsage(parsed);   // ← never reached for that frame
```

A frame that carries only `usage` has no "valuable content", so it is skipped before its usage is extracted; the closure never sees the provider's numbers and the abort path falls back to the estimate. Your weak-assertion finding is exactly what was hiding it.

I have **not** moved `extractUsage` above that guard in this PR. It is a hot-path change affecting every passthrough stream, not just aborted ones, and it deserves its own change and its own review rather than riding along in an accounting fix. For now the test pins the current behaviour explicitly (`expect(usage.prompt_tokens).not.toBe(11)`) with a comment naming the guard, so the gap is visible and this test fails the moment someone closes it — rather than passing quietly on an estimate as it did before.

Happy to send that as a follow-up PR, or to fold it in here if you would rather have both together.

```
npx vitest run --config tests/vitest.config.js tests/unit/aborted-stream-usage-3488.test.js
Tests  8 passed (8)
```

Head is `b6556c1c`.

hiepau1231:
## Follow-up on my review

I rechecked the P1 after your response and withdraw it as a general Node/Web Streams defect. The current Streams Standard and Node v24 support `Transformer.cancel`, and a minimal Node 24 `pipeThrough()` reproduction reached both the transform and source cancellation handlers. The `pipeWithDisconnect` wrapper also propagates readable cancellation through that chain.

One non-blocking test gap remains: the updated late-cancel test releases the reader lock, but it cancels only after `completeStream()` has drained and closed the readable. On Node 24, `cancel()` on an already-closed readable resolves without invoking the underlying cancel algorithm. Therefore removing the `streamFinished` guard would still leave this test passing. To test the guard itself, trigger the normal `finishStream()` path while the readable remains open, then cancel before the writable closes; otherwise this test should be described as a closed-stream API smoke test rather than an exactly-once cancellation test.

Compatibility note: the repository still exposes `dev:bun`, `build:bun`, and `start:bun`, but Bun 1.3.14 did not invoke a `Transformer.cancel` callback in the equivalent minimal reproduction. If Bun is a supported runtime, this PR needs a compatible cancellation hook or an explicit runtime requirement. If the supported deployment target is Node 22+, this is not a blocker.

The provider-usage-only-frame gap looks like a valid follow-up issue and is now documented by the characterization assertion.

ntdatt812:
Both points check out. The Bun one is the serious one, and it means this PR is not finished.

## Bun does not invoke `Transformer.cancel` — confirmed

```
bun 1.3.14
new TransformStream({start, transform, flush, cancel}) + readable.cancel('client-hangup')
→ ["start"]
```

No `cancel`, no `flush`. Same script on Node 24 gives `["start","cancel:client-hangup"]`. And Bun is not hypothetical here — `package.json` ships `start:bun` (`bun ./.next/standalone/custom-server.js`), which is the production standalone server, not just a dev convenience. So on a Bun deployment this fix is inert and #3488 stays broken exactly as reported.

**There is a portable hook, and I checked it works on Bun:**

```
bun 1.3.14
new ReadableStream({ start, cancel(reason){…} }) + rs.cancel('bye')
→ ["src-cancel:bye"]
```

Bun honours an underlying-source `cancel`. So observing cancellation at the readable source — wrapping the transform's readable, or hooking the request abort signal at the handler — covers both runtimes, where the transformer callback covers only Node.

I have not made that change here: it moves where the accounting hook lives, which is what your original P1 asked about, and it deserves to be the reviewed shape rather than something I bolt on mid-review. Tell me which you prefer and I will push it:

1. wrap the readable so the source `cancel` drives `finishStream` (smallest diff, no handler changes, verified on both runtimes above);
2. hook the HTTP request abort/close signal in the handler, leaving the stream untouched;
3. keep the transformer `cancel` **and** add one of the above, so neither runtime depends on the other's behaviour.

My preference is 1, with the transformer `cancel` kept as-is: it is one wrapper, it is measurable on both runtimes, and it does not touch the request path.

## The late-cancel test — you are right, and I proved it against myself

```
# guard removed:  if (streamFinished) return;
Tests  8 passed (8)
```

Removing the exactly-once guard leaves that test green, exactly as you said. It asserted nothing about the guard.

I then tried to build the test you prescribed — finish normally while the readable is still open, then cancel — and could not, for a reason worth recording: **the normal `finishStream()` call lives in `flush`, not in `transform`** (`stream.js:399` is inside `flush(controller)`). `flush` only runs when the writable closes, and the readable closes right behind it, so there is no point where a normal finish has happened and a cancel can still run the cancel algorithm. Feeding a terminal `data: [DONE]` frame does not help — it is handled in `flush` too, and my attempt at that test failed with `expected 1 call, got 0`.

So through the public stream API the guard appears to be genuinely unreachable twice, which makes it defensive code rather than covered behaviour. Rather than leave a test implying coverage it does not give, I took your other suggestion and named it for what it is:

```js
it("accepts a late cancel on an already-closed readable without reporting again", …)
// Named for what it is. `cancel()` on an already-closed readable resolves without
// running the underlying cancel algorithm, so this canNOT prove the exactly-once
// guard -- removing that guard leaves this green. It is an API smoke test.
```

If you would rather have the guard genuinely covered, the way I can see is exporting `finishStream` (or the stream's internals) for a direct unit test — happy to do that if you want it, though it widens the module's surface for a test-only reason.

```
npx vitest run --config tests/vitest.config.js tests/unit/aborted-stream-usage-3488.test.js
Tests  8 passed (8)
```

Head is `edb0381b`. Thanks for pushing on both of these — the Bun one would have shipped a fix that does nothing on a supported runtime.

hiepau1231:
## Pick option 2, not option 1 — option 1 is also inert on Bun

I verified your Bun finding and it reproduces. But I also tested your three proposed options against a real client hangup, and the result changes the recommendation: **an underlying-source `cancel` does not fire on Bun either when the disconnect comes from a real socket.** Option 1 fixes nothing on Bun.

The reason the source-cancel probe looked portable is that it called `rs.cancel()` directly, which is the same shortcut that made the late-cancel test vacuous.

### Real hangup, Next 16's actual adapter

I replicated `pipeToNodeResponse` and `createAbortController` from the installed `next@16.2.10` (`node_modules/next/dist/server/pipe-readable.js`, `.../adapters/next-request.js`), then aborted a `fetch` mid-stream.

| Hook under test | Node 24 | Bun 1.3.14 |
|---|---|---|
| current PR (`transformer.cancel`) | reports usage | nothing |
| option 1 (wrapper source `cancel`) | reports usage | **nothing** |
| option 2 (request-side disconnect hook) | reports usage | **reports usage** |
| option 3 (both) | reports usage | reports usage |

### Why option 1 fails on Bun

The break is upstream of streams entirely. Next derives its abort signal from `res` `'close'`:

```js
response.once('close', () => {
  if (response.writableFinished) return;
  controller.abort(new ResponseAborted());
});
```

On Bun, `ServerResponse` never emits `'close'` on a client hangup, so that `abort()` never runs, so `pipeTo` never cancels, so no stream-level hook of any shape is reached. Isolating each link:

| Signal on real hangup | Node 24 | Bun 1.3.14 |
|---|---|---|
| `pipeTo(signal)` abort cancels readable | yes | **yes** |
| `res` `'close'` | yes | **never** |
| `res.write()` returns false / throws | yes | never |
| `req.signal` | present | **absent** |
| `req` `'close'` | yes | **yes** |
| `req` `'aborted'` | yes | yes |
| `req.socket` `'close'` | yes | yes |

So Bun's Web Streams cancellation is fine; its `node:http` response-side disconnect eventing is what is missing. `transformer.cancel` is a second, independent Bun gap, but fixing only that one is not what unblocks #3488 on Bun.

The portable signal is the **request** side: `req` `'close'` (guarded by `res.writableFinished`, matching Next's own guard), or `req.socket` `'close'`. Both fire on both runtimes.

### Recommendation

Option 2, or option 3 if you want the belt-and-braces version. Keeping the transformer `cancel` is harmless and still the cleanest path on Node, but it cannot be the only hook. Note that on Node option 2 fires slightly earlier than the transformer hook, so whichever combination you choose, the exactly-once guard is now load-bearing rather than defensive — which incidentally gives that guard a reachable test.

One caveat on my own evidence: I tested the adapter reproduction, not 9router booted under `start:bun`, so please confirm against the real server before landing. The Next source I quoted is the installed version in this repo.

### On the late-cancel test

Agreed, and thanks for testing the guard by deleting it — that is the right check. Your renamed API-smoke-test version is honest and fine to keep. I would not export `finishStream` purely for coverage; with option 2 or 3 the guard becomes reachable through the public path anyway, since a request-side hook and a stream-side hook can both fire on the same abort. That is the test worth having.

Still happy with the provider-usage-only-frame gap as a follow-up issue.

ntdatt812:
Confirmed, independently — and I think the finding is bigger than this PR.

## 1. Your Bun result reproduces on the raw `node:http` primitive

I did not want to take the Next adapter replication on trust, so I went one level lower: a bare `node:http` server, a real `fetch` aborted mid-stream, every candidate listener attached at once.

| Listener | Node 24.16.0 | Bun 1.3.14 |
|---|---|---|
| `res` `'close'` | fires | **never** |
| `req` `'close'` | fires | fires |
| `req` `'aborted'` | fires | fires |
| `req.socket` `'close'` | fires | fires |

So it is not something Next does — Bun's `ServerResponse` simply does not emit `'close'` on a client hangup. Since Next's `createAbortController` is `response.once('close', …)`, nothing downstream of it can ever run on Bun. Your conclusion holds and the mechanism is one layer below where you found it.

<details><summary>The probe (run with <code>node</code> and <code>bun</code>)</summary>

```js
import http from 'node:http';

const runtime = typeof Bun !== 'undefined' ? `Bun ${Bun.version}` : `Node ${process.version}`;
const seen = [];
const mark = (n) => { if (!seen.includes(n)) seen.push(n); };

const server = http.createServer((req, res) => {
  res.writeHead(200, { 'Content-Type': 'text/event-stream' });
  res.write('data: hello\n\n');

  req.on('close', () => mark("req 'close'"));
  req.on('aborted', () => mark("req 'aborted'"));
  res.on('close', () => mark("res 'close'"));
  res.on('error', () => mark("res 'error'"));
  req.socket?.on('close', () => mark("req.socket 'close'"));

  // keep the stream open past the client's disconnect
  const t = setInterval(() => {
    try { res.write('data: tick\n\n'); } catch { mark('res.write threw'); }
  }, 100);

  setTimeout(() => {
    clearInterval(t);
    console.log(`${runtime}: ${seen.length ? seen.join(', ') : '(nothing fired)'}`);
    try { res.end(); } catch {}
    server.close();
    process.exit(0);
  }, 1500);
});

server.listen(0, async () => {
  const port = server.address().port;
  const ac = new AbortController();
  try {
    const r = await fetch(`http://127.0.0.1:${port}/`, { signal: ac.signal });
    const reader = r.body.getReader();
    await reader.read();
    setTimeout(() => ac.abort(), 200);   // real client hangup mid-stream
    try { for (;;) { const { done } = await reader.read(); if (done) break; } } catch {}
  } catch {}
});
```

</details>

## 2. The same shape already governs this repo's *existing* disconnect handling

This is the part I want to flag. `createDisconnectAwareStream` in `open-sse/utils/streamHandler.js:112` is

```js
return new ReadableStream({
  async pull(controller) { … },
  cancel(reason) {
    streamController.handleDisconnect(reason || "cancelled");
    reader.cancel();
    writer.abort();
  }
});
```

That is exactly your option 1 — an underlying-source `cancel` — and it is the *only* thing that calls `handleDisconnect`. By your result (and mine), on Bun it never runs. The consequences go well past #3488:

- no `⚡ DISCONNECT` line,
- `onDisconnect` never fires,
- `abortController.abort()` at `streamHandler.js:49` never runs, so **the upstream provider request is never aborted** — 9router keeps pulling (and the account keeps being charged for) a stream nobody is reading.

So the usage row this PR is about is one casualty of a Bun-wide gap, not the gap itself.

## 3. Where the portable hook can actually attach here

`custom-server.js` already wraps every request and holds the raw `req`/`res` (it stamps `x-9r-real-ip` there). That is the natural place for a `req` `'close'` hook, guarded by `res.writableFinished` the way Next guards its own — one place, both runtimes, no per-route plumbing.

Two honest caveats:

- `custom-server.js` is not loaded by a bare `next dev`, so a hook there covers `start` but not dev. If dev parity matters, the hook needs a second home.
- Bridging raw `req` → the route handler's stream needs a small explicit channel (a per-request id stamped as a header plus a registry, or similar). There is no existing plumbing: `grep` finds no `request.signal` / `AbortSignal` use anywhere in the chat path.

## 4. What I propose

Option 3, with the request-side hook in `custom-server.js` and the transformer `cancel` kept as the Node fast path — and as you note, that makes the exactly-once guard load-bearing and finally testable through the public path, which resolves the third review point properly rather than by renaming the test.

That is a cross-cutting change though — it touches the server wrapper, adds a request→stream channel, and changes when upstream aborts fire on Bun. It is a good deal larger than "record the turn when a client hangs up", so I would rather not smuggle it into this PR unannounced.

@decolua — how would you like this split? I am happy to either (a) extend this PR to option 3, or (b) land this one as the Node-side fix and open a separate PR for the Bun request-side hook plus the upstream-abort consequence in §2. Say which and I will get on it.

Thanks @hiepau1231 — testing the three options against a real hangup instead of `rs.cancel()` is what turned this from a plausible fix into a measured one.

hiepau1231:
## Agreed on §2. But the hook does not belong in `custom-server.js`, and there is existing plumbing

Your §2 checks out against the repo, and it is the right thing to have escalated. Two corrections to §3 that make the fix substantially smaller than you scoped it.

### `handleChat` already receives the `Request`

Every chat route passes it straight through:

```
src/app/api/v1/chat/completions/route.js:33   return await handleChat(request);
src/app/api/v1/messages/route.js:34           return await handleChat(request);
src/app/api/v1/responses/route.js:29          return await handleChat(request);
src/app/api/v1/api/chat/route.js:34           const response = await handleChat(request);
src/app/api/v1beta/models/[...path]/route.js:106  const response = await handleChat(newRequest);
src/sse/handlers/chat.js:31                   export async function handleChat(request, clientRawRequest = null)
```

So no per-request id, no header stamp, no registry, and no `custom-server.js` change. `handleChat` already has the request object and already calls `handleChatCore`, which is where `streamController` is built (`open-sse/handlers/chatCore.js:299`). Threading a signal from `handleChat` into `createStreamController` is a parameter, not a channel. That also removes your dev-parity caveat, since it works under bare `next dev`.

On your `grep`: there is existing precedent, and it was written for exactly this failure. `src/app/api/translator/console-logs/stream/route.js:22`:

```js
// request.signal fires reliably on client disconnect; ReadableStream.cancel()
// is not always invoked in Next.js, which caused listeners to accumulate.
request.signal.addEventListener("abort", cleanup, { once: true });
```

Someone already hit stream `cancel()` being unreliable and moved to the request signal. This PR is the same lesson in the chat path.

### But `request.signal` alone does not fix Bun — check which signal you thread

Next builds that signal from the **response**, not the request (`next-server.js`, two call sites):

```js
signal: signalFromNodeResponse(params.res.originalResponse)
```

and `signalFromNodeResponse` is `response.once('close', …)`. Same dead link as before:

| Signal on real hangup | Node 24 | Bun 1.3.14 |
|---|---|---|
| Next's `request.signal` (from `res` `'close'`) | fires | **never fires** |
| signal derived from `req` `'close'` | fires | **fires** |

So `request.signal` is the correct *shape* and the wrong *source* on Bun. The hook has to be attached to `req` `'close'` (or `req.socket` `'close'`), guarded by `res.writableFinished`. Which means one honest point for you: since Next's own `request.signal` is response-derived, obtaining a `req`-derived signal inside an App Router handler is not free, and `custom-server.js` may end up being the only place that owns raw `req`. If so, your instinct about location was right even though the plumbing argument was not — worth checking before committing to a shape.

### Scope

Option (b): land this as the Node-side fix, and open a separate issue/PR for the Bun request-side hook plus the §2 upstream-abort consequence. My reasoning: §2 is a bigger and different bug — unaborted upstream requests burning provider quota is worse than a missing usage row, it affects `handleDisconnect`, `onDisconnect`, and the `⚡ DISCONNECT` line rather than just usage accounting, and it deserves its own review and its own regression test rather than riding along here. This PR is already correct and tested on the runtime the Docker image ships (`node:22-alpine`).

Two conditions on landing this one:

1. Say plainly in the PR description or a code comment that the abort accounting is Node-only and does not fire under `start:bun`, with a link to the follow-up. Otherwise the next person reads `transformer.cancel` as runtime-agnostic.
2. Keep the renamed API-smoke-test as-is. The exactly-once guard becomes genuinely reachable only once the second hook lands, so its real test belongs in the follow-up PR.

I will re-review and clear my CHANGES_REQUESTED once the Node-only scope is documented. @decolua over to you on the split.

hiepau1231:
## Correction: my `request.signal` recommendation was wrong on Bun

I stopped replicating Next's adapter and instead ran the repo's real modules plus a real Next server on both runtimes. Two results: my last comment's suggested fix does not work on Bun, and the root cause is one layer lower than either of us has been aiming.

### `request.signal` is dead on Bun too

I built a minimal Next 16.2.10 app with `output: "standalone"` and ran **the same build** under both runtimes, hanging up a `fetch` mid-stream:

| Hook in an App Router route | Node standalone | Bun standalone (`start:bun`) |
|---|---|---|
| control `console.log` | prints | **prints** |
| `request.signal` abort | fires | **never** |
| `ReadableStream.cancel()` | fires | **never** |

The control line prints, so this is not suppressed logging. Same build, same client, runtime is the only variable.

So the `console-logs/stream/route.js` precedent I cited is correct on Node but **not portable**. Withdraw the "thread `request.signal` from `handleChat`" part of my previous comment: it is response-derived, same dead link. My plumbing point still stands (`handleChat` does already receive the `Request`, so no registry is needed), but the signal it would thread is the wrong one.

That makes your instinct about `custom-server.js` right for a reason neither of us stated: it is the only place that holds a raw `req`, and the raw `req` is the only thing still alive on Bun.

### Root cause, isolated to one variable

Hypothesis: **Bun's `node:http` `ServerResponse` never emits `'close'` on a client hangup.** Since Next derives every disconnect signal from `res 'close'`, the whole chain dies.

| Prediction | Node 24.16.0 | Bun 1.3.14 |
|---|---|---|
| `res 'close'` fires on hangup | true | **false** |
| `req 'close'` fires on hangup | true | true |
| `pipeTo(signal)` cancels readable | true | **true** |

The third row is the falsification test: if Bun's Web Streams were broken, the hypothesis would be wrong. They are not. The defect is the missing response event, nothing else.

### Boundary trace with this repo's real modules

Importing the actual `createStreamController` and `pipeWithDisconnect` from `open-sse/utils/streamHandler.js`, piped through the real `pipeToNodeResponse`:

| Layer | Node | Bun |
|---|---|---|
| L1 `req 'aborted'` / `req 'close'` / `socket 'close'` | fires | **fires** |
| L2 `res 'close'` | fires | **never** |
| L3 Next abort signal | fires | never |
| L4 transformer `cancel` | fires | never |
| L5 `handleDisconnect` → `onDisconnect` | fires | never |
| L6 upstream fetch abort | fires | never |

L1 is the only layer alive on both. Both the original P1 and option 1 target L4 — mid-chain, below the actual break.

Quantifying §2 over 3s after the client left: Node pulls **1** further chunk and aborts upstream; Bun pulls **78** and never aborts.

### Fix at L1, verified against the real modules

One change — bridge `req 'close'` into `streamController.handleDisconnect()`, guarded by `res.writableFinished` the way Next guards its own:

| | Bun before | Bun after |
|---|---|---|
| `onDisconnect` reason | `null` | `client_closed` |
| upstream fetch aborted | false | **true** |
| wasted chunks after hangup | 50 | **4** |

On Node, `disconnectReported` changes from `ResponseAborted` to `client_closed` and wasted chunks stay at 1 — no regression.

### What this does not change

My position on scope is the same: land this PR as the Node-side fix with the Node-only limitation documented, and handle the Bun request-side hook plus the §2 upstream-abort consequence separately. The above just means the follow-up's fix belongs at `req 'close'` in the server wrapper, not at any stream-level hook and not at `request.signal`.

Caveat on my evidence: real repo modules and a real Next server, but a stubbed upstream provider — I did not boot 9router with live credentials under `start:bun`. Worth confirming there before landing the follow-up.

ntdatt812:
Both conditions are met — `598bc3f0`.

**1. Node-only scope documented.** In the code, next to `finishStream` where the next reader will hit it, not only in the description:

```js
// RUNTIME SCOPE: the aborted path below is Node-only. Every disconnect signal
// in Next — and therefore `cancel` in any of its shapes — descends from `res`
// `'close'`, which Bun's node:http ServerResponse never emits on a client
// hangup. Under `start:bun` nothing cancels this stream, so `finishStream` only
// ever runs from `flush`, exactly as before this fix. The portable hook is `req`
// `'close'`; see #3559, …
```

The PR description now carries the same limitation with the measurement table.

**2. The renamed API-smoke-test is untouched.** Agreed on the reasoning: the exactly-once guard only becomes reachable once the second hook lands, so its real test belongs with that hook, not here.

`npx vitest run --config tests/vitest.config.js tests/unit/aborted-stream-usage-3488.test.js` → 8 passed.

**Follow-up: #3559.** It leads with the upstream-abort waste rather than the usage row, because that is the expensive half — your 78-vs-1 chunk measurement is the number that makes the case, and it is quoted there with credit. I carried both open questions into it rather than answering them myself: dev parity for a `custom-server.js` hook, and confirmation against a real `start:bun` boot, since neither of us has done that yet.

One correction to my own §3 while I am here: I claimed a per-request id and a registry would be needed. You showed `handleChat` already receives the `Request`, so that was wrong, and it stays wrong even though the conclusion about location survives — for the different reason you found, that Next's `request.signal` is response-derived. Right answer, wrong argument on my part.

Thanks for the two rounds of measurement, and for correcting your own `request.signal` recommendation rather than defending it. That is what turned this from two people trading plausible mechanisms into a located defect.

hiepau1231 (approved):
Approving `598bc3f0`. Both conditions met, and I verified them rather than taking them on trust.

Checked out the PR head in a clean worktree and ran it locally:

```
tests/unit/aborted-stream-usage-3488.test.js                    8 passed
responses-abort-terminal + antigravity-nonstream-usage-3260
  + translator/golden-response-stream                          14 passed
```

The `RUNTIME SCOPE` comment sits directly above `finishStream` in `open-sse/utils/stream.js`, which is where someone changing this code will actually read it, and it names the portable hook plus #3559. The renamed late-cancel test still carries its own honest caveat that it cannot prove the exactly-once guard. That is the right call: the guard stays defensive until the second hook lands, and pretending otherwise in a test name would have been worse than leaving it uncovered.

#3559 is a fair statement of the problem. It leads with the upstream-abort waste rather than the usage row, keeps both open questions open instead of guessing at them, and is honest that neither of us has booted 9router with live credentials under `start:bun`. On question 1 my inclination is that dev parity does not block it: `next dev` is not a billing path, so covering `start`/`start:bun` first is the right order.

On the review exchange itself: my original P1 was wrong on the mechanism, and my `request.signal` suggestion was wrong on Bun. You were right to push back on the first and to test the second instead of accepting it. What made the difference in both cases was measuring against real modules and a real server rather than reasoning from documentation and source reading, which is a lesson I am taking rather than giving.

No objection to landing this on the Node-side scope as documented.

ntdatt812:
Rebased onto `master` (`90b52e06`) and **reworked down to a third of its former size**, because `master` did half the job in the meantime.

`finalizeStream()` now exists and carries the whole usage/logging tail, with the comment naming the same failure mode this PR was opened for. What it does not have is a `cancel()` caller, so the case it covers is the client closing *right after the terminal event* — not the client hanging up *mid-stream*, which is #3488. Those are different paths: the first still reaches `flush`, the second never does.

So this branch no longer introduces a `finishStream` helper of its own. It reuses yours and adds only what was missing:

- `cancel()` calls `finalizeStream({ aborted: true })`;
- `finalizeStream` forwards the flag to `onStreamComplete`, so an aborted turn is recorded as `status: "aborted"` rather than as a success.

The estimate-vs-reported logic is now entirely yours — I deleted my copy of it. Net effect on `stream.js` is 27 lines, most of them the runtime-scope comment.

Mutation check on the rebased branch: 8 tests pass; drop the `cancel()` handler and 6 of them fail.

ntdatt812:
Superseded by #3907. `master` has since grown `finalizeStream()`, which covers most of what this PR's `finishStream()` did, so replaying this diff would be arguing with code that already landed. #3907 closes the part that is still open — the disconnect that arrives before any terminal event — by routing a `cancel` callback to the existing function.

Issue #3468, opened 2026-08-22
should skip CommandCode 503 errors and continue round-robin


jefripunza:
Currently, when 9router sends a request to CommandCode and receives the following error:

```text
[CommandCode error: {"type":"server_error","message":"Service temporarily unavailable. Please try again shortly.","statusCode":503,"isRetryable":true}]
```

9router appears to treat this error as a failure of the entire request/round-robin flow instead of skipping the current CommandCode endpoint/account and continuing to the next available one.

Since the error response explicitly indicates `statusCode: 503` and `isRetryable: true`, 9router should recognize this as a temporary service availability issue.

## Expected Behavior

When a CommandCode endpoint returns a retryable `503` error:

1. 9router should detect that the error is retryable.
2. The current CommandCode endpoint/account should be temporarily skipped.
3. The round-robin process should continue to the next available endpoint/account.
4. The request should only fail after all available endpoints/accounts have been exhausted or no eligible endpoint remains.

For example:

```text
Account A → CommandCode → 503 Service Temporarily Unavailable
                 ↓
              SKIP
                 ↓
Account B → CommandCode → Success
```

Instead of:

```text
Account A → CommandCode → 503
                 ↓
             REQUEST FAILS
```

## Suggested Handling

Errors with the following characteristics should trigger the round-robin fallback:

```json
{
  "type": "server_error",
  "statusCode": 503,
  "isRetryable": true
}
```

The `isRetryable: true` flag should be respected so that temporary CommandCode outages do not stop the entire request flow.

## Impact

This is particularly important when multiple CommandCode accounts/endpoints are configured behind 9router. A temporary outage on one account should not prevent other healthy accounts from being used.

## Acceptance Criteria

* [ ] Detect CommandCode `503` errors with `isRetryable: true`.
* [ ] Skip the failing endpoint/account.
* [ ] Continue the round-robin process with the next available endpoint/account.
* [ ] Retry/fallback without requiring the user to resend the request.
* [ ] Only return an error when no other eligible endpoint/account is available.

jefripunza:
Fix submitted in PR #3476: https://github.com/decolua/9router/pull/3476

Turns a 200 OK with an embedded `{"type":"error",..."statusCode":503,"isRetryable":true}` frame into a real failing request so combo/account fallback continues round-robin to the next model/account.

fcoalcantarajr:
Reproducing on v0.5.55 (latest published to npm as of Aug 2026) via 9Router → Hermes Agent.

[...]

PR #3549, opened 2026-08-25
fix(commandcode): open the error chunk with the assistant role

ntdatt812:
## The inconsistency

`commandcode-to-openai.js` has one first-delta convention, followed by every content-emitting branch:

```js
// text-delta
const delta = state.chunkIndex === 0 ? { role: ROLE.ASSISTANT, content: text } : { content: text };
state.chunkIndex++;

// reasoning-delta
const delta = reasoningDelta(text, state.chunkIndex === 0);
state.chunkIndex++;

// tool-input-start / tool-call
...(state.chunkIndex === 0 ? { role: ROLE.ASSISTANT } : {}),
state.chunkIndex++;
```

`error` was the one branch that did neither:

```js
out.push(makeChunk(state, { content: `\n\n[CommandCode error: ${errStr}]` }));
```

No `role`, no `chunkIndex++`.

## Why it bites

An upstream error is routinely the **first** event on the stream. The example in #3468 is exactly that shape — a 503 before any token:

```
[CommandCode error: {"type":"server_error","message":"Service temporarily unavailable. Please try again shortly.","statusCode":503,"isRetryable":true}]
```

So for a failed turn, the only content delta a client receives carries no `role`. The OpenAI streaming contract puts `role` on the opening delta, and the branch right next door already honours it.

Second effect, from the missing `chunkIndex++`: if any text arrives after the error chunk, that text still believes it is chunk zero and emits `role` a **second** time.

## The change

Eleven lines in one branch, shaped like the `text-delta` branch beside it. The message text is byte-identical — this is about the envelope, not the content.

## Tests

Three added to `tests/unit/commandcode-to-openai.test.js`, using the file's existing `feed()` helper:

- error as the first event → the chunk opens with `role: "assistant"`
- error after text → no role on the error chunk
- error then text → exactly one role delta across the stream

Mutation-checked, each mutation verified to have applied before running:

| mutation | result |
|---|---|
| drop the `chunkIndex === 0` role branch | **2 failed** — "opens with the assistant role…", "sends the role exactly once…" |
| drop `state.chunkIndex++` (verified 5 → 4 sites in the file) | **1 failed** — "sends the role exactly once…" |

Both reverted.

## Suite

`tests/unit/commandcode-to-openai.test.js`: 9 → **12 passed**, 0 failed.

Full suite, both endpoints measured on this branch's base `699edac3`:

| tree | result |
|---|---|
| clean `origin/master` | 78 failed files, **140 failed / 1385 passed** |
| this branch | 78 failed files, **140 failed / 1388 passed** |

Failures unchanged; passes up by exactly the three tests added. (The 140 are pre-existing on `master` — I re-measured rather than reusing an older figure, since the count has moved a good deal recently.)

## What this does not do

It does not make 9router fail over on a CommandCode 503, which is what #3468 actually asks for. That needs the round-robin path to treat an in-band stream error as a retryable attempt failure, and by the time this translator runs the response has already begun — a real design question about whether and when a started stream may be retried. I did not want to guess at it inside a one-branch consistency fix, so I am sending this on its own and leaving #3468 open.

What this does fix is a defect I found while reading that path: the error chunk was malformed regardless of failover.

### DIFF
--- a/open-sse/translator/response/commandcode-to-openai.js
+++ b/open-sse/translator/response/commandcode-to-openai.js
@@ -168,7 +168,17 @@ export function commandCodeToOpenAIResponse(chunk, state) {
       state.finishReason = OPENAI_FINISH.STOP;
       const errVal = event.error ?? event.message ?? "unknown";
       const errStr = typeof errVal === "string" ? errVal : JSON.stringify(errVal);
-      out.push(makeChunk(state, { content: `\n\n[CommandCode error: ${errStr}]` }));
+      // Same first-delta contract as every other content branch here. An upstream
+      // error is frequently the FIRST event (a 503 before any token), and this was
+      // the one branch that emitted content without the opening `role` and without
+      // advancing chunkIndex -- so the error chunk carried no role, and any text
+      // arriving after it still thought it was chunk zero and sent a second one.
+      const errText = `\n\n[CommandCode error: ${errStr}]`;
+      const delta = state.chunkIndex === 0
+        ? { role: ROLE.ASSISTANT, content: errText }
+        : { content: errText };
+      state.chunkIndex++;
+      out.push(makeChunk(state, delta));
       out.push(makeChunk(state, {}, OPENAI_FINISH.STOP));
       break;
     }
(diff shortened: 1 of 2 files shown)


ntdatt812:

Issue #2813, opened 2026-07-24
Feature: Resolve Qoder alias models to actual model names


npinto:
## Summary

When using Qoder provider (`qd/*`) with alias model IDs like `qd/qmodel_preview`, the response should include the actual resolved model name (e.g., `Qwen3.8-Max-Preview`) rather than the alias.

## Current behavior

Request: `qd/qmodel_preview`

Response chunks contain:
```json
{"model": "auto", "choices": [...]}
```

Note: responses should at minimum show `qoder/qmodel_preview` instead of `auto`.

## Desired behavior

Response chunks should contain the resolved model name:
```json
{"model": "Qwen3.8-Max-Preview", "choices": [...]}
```

Or alternatively, include both:
```json
{"model": "qoder/qmodel_preview", "x_resolved_model": "Qwen3.8-Max-Preview", "choices": [...]}
```

## Why this matters

- `qd/qmodel_preview` is an alias that points to different models over time (today it's Qwen3.8-Max-Preview, tomorrow it could be something else)
- For transcript analysis, cost tracking, and debugging, users need to know which actual model ran
- The Qoder backend knows the resolved model -- it's in the `model_config` fetched from `/algo/api/v2/model/list`

## Implementation notes

In `open-sse/executors/qoder.js`, `buildQoderRequestBody` already fetches `modelConfig` which contains the actual model info. This could be threaded through to `wrapQoderSSE` and used to override the model field in response chunks.

PR #3555, opened 2026-08-25
feat(qoder): refresh model catalog, add capability mapping and image pass-through

hangyu1125:
## Summary

Fixes: qoder models reported the default 200K context window because capability lookup matches raw upstream ids (`dfmodel`, `kmodel`, …) while the registry `name` (`DeepSeek-V4-Flash`, …) is display-only.

## Changes

**Model catalog refresh** (`registry/qoder.js`, `shared/qoder/constants.js`)
- Drop `qmodel_preview` / `gm51model`; add `lite`, `qmodel_38max` (Qwen3.8-Max), `gmodel` (GLM-5.3) to match the live upstream catalog

**Capability mapping** (`providers/capabilities.js`)
- Add `PROVIDER_CAPABILITIES['qoder']`: opaque internal ids now resolve to their real models' limits — e.g. `dfmodel` (DeepSeek-V4-Flash) reports 1M context instead of 200K
- `thinkingCanDisable: false` everywhere: the executor only forwards messages/tools/max_tokens; thinking is fixed upstream via `modelConfig.is_reasoning`

**Image pass-through** (`executors/qoder.js`)
- `normalizeMessages` no longer flattens image blocks away: OpenAI `image_url` parts are preserved (both http(s) URLs and inline base64 `data:` URIs are accepted upstream — no OSS pre-upload needed)
- Claude-style `{type:"image"}` blocks are converted to `image_url`
- `chat_record_id` hash now includes image refs so the same prompt with a different image gets a distinct id

**Tests** (`tests/unit/qoder.test.js`)
- 4 new cases: image preservation, base64 data-URI pass-through, Claude-block conversion, invalid-image fallback

## Verified

```
qd/dfmodel   context_length: 200000 -> 1000000
qd/dmodel    context_length: 200000 -> 1000000
qd/gmodel    context_length: 200000 -> 1000000 (new)
```


hangyu1125:
@decolua I plz review and merge this PR, fix the qoder model mapping issue

hangyu1125:
#2813

decolua:
Thanks @hangyu1125 for the contribution! Reviewed and merged into master. 🙏

PR #3560, opened 2026-08-26
fix(combo): fail over on empty-but-successful streams (#3463)

### LINKED ISSUE #3463: Combo failover treats empty-but-successful streams (HTTP 200, zero content frames) as success — never falls over to next model
## Summary

The combo failover loop decides success purely by HTTP response status (`result.ok`). When an upstream provider returns **HTTP 200 with an SSE stream that closes cleanly but contains zero usable frames** (no content deltas, no tool calls, no usage frame), the combo marks that model as `succeeded` and returns the empty stream to the client. The remaining combo models are never tried.

This was observed on v0.5.20 and I verified the logic is unchanged in the current master / v0.5.55 source (`open-sse/services/combo.js`, the `if (result.ok) return result;` branch fires before any content-level validation).

## Steps to Reproduce

1. Configure a combo with 2+ models, strategy `fallback` or `round-robin`.
2. Send a streaming chat completion where model #1 occasionally responds with HTTP 200, opens the SSE stream, sends only keep-alive/comment frames (or nothing), and then closes cleanly.
3. Router log shows: `[COMBO] Model X succeeded`, then `[STREAM] ... complete`, sometimes with `(estimated)` usage and `out=0`.
4. Client receives an empty stream — no text, no tool calls, no usage — and surfaces it as an empty-model-response error.

A closely related case: when the upstream stalls mid-stream, `streamHandler.js` aborts with `stream stall timeout` after `STREAM_STALL_TIMEOUT_MS`, but at that point the combo has already returned the response, so failover to the next model is impossible.

## Expected Behavior

A stream that ends without any meaningful content should be treated as a provider failure so the combo can fall through to the next model (or retry), instead of handing an unusable empty response to the client.

## Actual Behavior

Empty-but-cleanly-closed streams are counted as success; combo returns them as-is.

## Environment

- 9router v0.5.20 (Docker); logic confirmed identical in v0.5.55 source
- Docker Desktop on macOS
- Seen with multiple free-tier upstream routes under sustained agent workloads (long multi-request coding-agent sessions)

## Suggested Fix

1. In the combo loop, for streaming requests do not return immediately on `result.ok`: validate that the first meaningful SSE frame (content delta, tool-call delta, or usage) arrives within a short guard window; treat clean-close-with-zero-meaningful-frames as failure and continue to the next combo model.
2. A precedent already exists in-tree: the Qoder provider detects billing blocks at stream start and converts them into a 403 to trigger failover (`fix(qoder): detect billing blocks at stream start, return 403 for failover`). Generalizing that pattern would cover all providers.
3. Optionally make it configurable, e.g. `COMBO_EMPTY_STREAM_POLICY=fail|passthrough` plus a first-content timeout env, so users can opt out if some upstream legitimately sends delayed payloads.

hiepau1231:
## Summary

The combo failover loop decides success from HTTP status alone (`open-sse/services/combo.js`, the `if (result.ok) return result;` branch). When a provider returns **HTTP 200, opens an SSE stream, emits only comments/keepalives and closes cleanly**, the model is marked `succeeded`: the unusable stream is handed to the client and the remaining combo models are never tried.

Reproduced against the real module before changing anything — model 1 sends one keepalive comment, model 2 has a valid answer:

```
models tried: [ 'p1/empty' ]
status: 200
body received by client: ": keepalive\n\n"
client got any content frame: false
```

Model 2 was never called.

## Why the existing stall watchdog cannot cover this

Ordering. `pipeWithDisconnect` arms its watchdog, then `handleStreamingResponse` wraps the result in a `Response` and returns. By the time any stall fires, `combo.js` has already returned — an abort can close the connection but never re-enter the model loop.

## Fix

`peekStreamForContent` reads until the first frame that actually carries content, then rebuilds the body so every byte already pulled off the socket is replayed ahead of the live reader. This is the same re-assembly `qoder.js` (`peekFirstQoderFrame`) and `codex.js` (`_peekSseTransientError`, via `replacementBody`) already use for their own first-frame peeks — issue #3463 explicitly suggests generalizing that precedent.

Scanning is line-buffered, so a JSON frame split across two network chunks is never judged on its first half. Recognizes OpenAI, Claude, Gemini and Ollama frame shapes.

Two deliberate scope limits:

- **Gated on `Content-Type: text/event-stream`.** `handleComboChat` is also called from `fetch.js`, `search.js`, `imageGeneration.js` and `tts.js`, whose 200s are audio, images or plain JSON. Reading those would consume a body the caller still needs, so they are skipped untouched — a test asserts `bodyUsed === false` for a JSON completion.
- **Bounded by `STREAM_FIRST_CHUNK_TIMEOUT_MS`.** A naive peek would await the first content frame forever, so an upstream sending endless keepalives would hang where the old code returned immediately. That would be a regression introduced by this fix, so the wait has a deadline. This also wires up a constant that was previously referenced only by `kiro.js`.

An empty stream is always treated as provider failure (no new env var). If every combo model comes back empty, the request ends as 503 rather than a silent empty answer.

## Tests

12 new cases in `tests/unit/combo-empty-stream-3463.test.js`. Against unmodified `combo.js`: exactly **4 fail** (the bug), **7 pass** (the guard rails, confirming they are not vacuous). After the fix: 12/12.

Mutation-tested rather than trusting green:

| mutant | result |
| --- | --- |
| always report content | 4 failed — killed |
| skip content-type gate | 2 failed — killed |
| drop replayed bytes | 8 failed — killed |
| treat `[DONE]` as content | survived — equivalent |
| judge partial lines | survived — equivalent |
| remove deadline race only | survived — redundant guard |
| remove both deadline guards | test times out — killed |

The two "equivalent" survivors were verified, not assumed: `JSON.parse` rejects `[DONE]` and every truncated JSON prefix, so neither mutation changes behaviour. The deadline guards are intentionally redundant — either alone suffices, which is why removing one survives and removing both hangs.

## Relationship to #3465

Complementary, not duplicate. #3465 detects emptiness in `onStreamComplete`, after the response has been sent; its own doc comment notes it "can't un-send it" and instead locks the account for the *next* request. This change intercepts before `combo.js` returns, so the **current** request is rescued by the next combo model. Different layer, different timing. #3465's non-streaming half does not overlap at all. The two can land independently.

## Test plan

- [x] `tests/unit/combo-empty-stream-3463.test.js` — 12/12
- [x] 4/12 confirmed failing on unmodified `combo.js` (bug reproduced, tests proven non-vacuous)
- [x] Mutation-tested: 4 of 7 mutants killed, 3 verified equivalent/redundant
- [x] Combo suites (`combo-routing`, `combo-autoswitch`, `combo-fusion`) — 33 passed
- [x] `npx eslint` clean on both changed files
- [x] Maintainer smoke test against a provider that actually exhibits the empty-stream behaviour

**Pre-existing failures, not from this PR:** `combo-autoswitch.test.js` > `web_search tool -> search` and `keeps order when no model matches` fail identically on unmodified `master` (verified via `git stash`). They live in `detectRequiredCapabilities`/`reorderByCapabilities`, untouched here.

Fixes #3463

### DIFF
--- a/open-sse/services/combo.js
+++ b/open-sse/services/combo.js
@@ -5,6 +5,7 @@
 import { checkFallbackError, formatRetryAfter } from "./accountFallback.js";
 import { unavailableResponse } from "../utils/error.js";
 import { getCapabilitiesForModel } from "../providers/capabilities.js";
+import { STREAM_FIRST_CHUNK_TIMEOUT_MS } from "../config/runtimeConfig.js";
 import { extractTextContent } from "../translator/formats/gemini.js";
 
 // Hard capabilities = input modalities; missing one drops request data (e.g. image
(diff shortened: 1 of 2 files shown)


StroepWafel:
LGTM

PR #3563, opened 2026-08-26
feat(antigravity): quota-aware routing with reset-aware fallback

### LINKED ISSUE #3561: feat(antigravity): skip exhausted account/model quota before upstream retry
## Problem

When an Antigravity account has exhausted quota for one model, requests still reach upstream and receive `409`/`429` before ordinary account/combo fallback can continue. This adds avoidable latency and repeatedly hits an already exhausted quota.

## Proposed behavior

For Antigravity only:

1. On an upstream `409` or `429`, retrieve that account's live quota data.
2. Store exhausted quota state in a process-local cache keyed by `connectionId` and model, using upstream `resetAt` as expiry.
3. During account selection, skip only account/model pairs whose cached quota is exhausted and whose reset time is still in future.
4. If another account or combo target remains, continue fallback without another upstream call to the exhausted pair.
5. Once `resetAt` passes, selection may use that account/model again. A process restart may clear cache safely; next `409`/`429` repopulates it.

## Constraints / safety

- No hard-coded cooldown. Upstream quota/reset data remains source of truth.
- Keep cache in memory only; do not persist account locks for this condition.
- Coalesce concurrent quota refreshes per account and throttle failed refresh attempts.
- Preserve strict-proxy policy when querying quota.
- Do not expose quota cache through a public API endpoint or log account email addresses by default.
- Fail open if quota lookup fails: preserve known valid cache; otherwise retain normal upstream/fallback behavior.

## Acceptance criteria

- A `429` with quota exhaustion causes subsequent requests to skip that exact account/model until upstream `resetAt`.
- Other models on same account and same model on other accounts remain eligible.
- All-exhausted selection returns retry timing based on earliest cached reset.
- Concurrent refreshes create one upstream quota request per account.
- Failed/error quota responses do not clear known cache and are throttled.
- Unit coverage includes: `429 → cache`, skip/fallback, reset expiry, all accounts exhausted, concurrent refresh, strict proxy, error response preservation, and failed-refresh throttling.

No pull request is attached yet; opening this issue first to align on desired behavior and scope.

mrnim94:
Closes #3561

## Summary

Add Antigravity-only quota-aware account routing:

- Cache exhausted account/model quota in memory after upstream `409`/`429`.
- Skip only exhausted account/model pairs until upstream `resetAt`.
- Preserve fallback to other accounts/models and report earliest retry time when all are exhausted.
- Coalesce concurrent quota refreshes per connection.
- Throttle failed quota refresh attempts.
- Preserve strict proxy behavior for quota requests.
- Keep known quota cache when upstream returns error payloads such as `401`/`403`.
- Avoid account email/name in quota-block logs.

## Tests

```text
npx vitest run tests/unit/antigravity-quota-routing.test.js --config tests/vitest.config.js
8 passed
```

Scope limited to Antigravity. No persistent model locks or public quota-cache endpoint. Process-local cache remains fail-open when quota lookup has no usable data.


decolua:
Thanks @mrnim94 for the contribution! Reviewed and merged into master. 🙏

Issue #3567, opened 2026-08-26
400 on claude provider: cache_control set on last tool with defer_loading=true


f-liva:
## Bug: 400 `Tool '...' cannot both defer_loading=true cache_control set` on Claude provider

**Version:** 0.5.55 (Docker `decolua/9router:latest`, pulled 2026-08-26)

**Provider:** `claude` (`cc/*` → `https://api.anthropic.com/v1/messages?beta=true`)

### What happens

When a client (Claude Code) sends tools with `defer_loading: true` (standard for MCP tools), 9router's `anchorClaudeCache()` sets `cache_control: {type:"ephemeral", ttl:"1h"}` on the **last** tool in the array, with no guard for `defer_loading`. Anthropic rejects the combination:

```
[400] {"type":"error","error":{"type":"invalid_request_error","message":"Tool 'mcp__pmb__record_batch' cannot both defer_loading=true cache_control set. Tools defer_loading cannot use prompt caching."},"request_id":"req_..."}
```

The 400 is returned **before** fallback kicks in, so the whole combo fails on the first (`cc/*`) hop even though the request itself is valid.

### Repro

```bash
curl -X POST http://127.0.0.1:20128/v1/messages \
  -H "x-api-key: $KEY" -H 'anthropic-version: 2023-06-01' -H 'content-type: application/json' \
  -d '{"model":"claude-opus-5","max_tokens":50,"stream":false,
       "tools":[
         {"name":"get_weather","description":"t","input_schema":{"type":"object","properties":{}}},
         {"name":"mcp__pmb__record_batch","description":"t","defer_loading":true,
          "input_schema":{"type":"object","properties":{"city":{"type":"string"}},"required":["city"]}}
       ],
       "messages":[{"role":"user","content":"hi"}]}'
```

With the combo's first hop `cc/claude-opus-5`, this 400s. The same payload against a `moon/` or `glm/` hop does not (those translators strip `cache_control`).

### Root cause

`open-sse/translator/formats/claude.js`, `anchorClaudeCache()` (~line 226-229):

```js
if (Array.isArray(body.tools)) {
  const last = body.tools.length - 1;
  body.tools.forEach((tool, i) => {
    if (i === last) tool.cache_control = { ...CACHE_CONTROL_1H };
    else delete tool.cache_control;
  });
}
```

[...]

PR #3575, opened 2026-08-27
feat(tool-disclosure): progressive tool disclosure — static filter + BM25 session index

nitsuah:
## Summary

Implements Phases 0–2 of progressive tool disclosure: eliminating MCP's per-turn schema tax inside 9router. Inspired by [Michael Rollins' "MCP is Dead"](https://rollins.io/mcp-is-dead/) — Rollins correctly identifies that MCP tool schemas travel the prompt on every turn, creating linear token-cost growth. This PR delivers the Agent SDK efficiency pattern he describes as a **transparent proxy layer**, without abandoning MCP or requiring any changes to clients or servers.

> *Credit: the core framing — "the MCP surface must travel in the prompt on every turn" — is Michael Rollins'. This work adapts his Agent SDK efficiency insight into a proxy-layer implementation.*

**Token overhead at scale:** a session with 8–12 MCP servers (60–120 tools) spends an estimated 12–24k tokens per turn on schemas alone — 10–15% of a 200k-token context window, every single turn.

- `open-sse/utils/toolFilter.js` — declarative static include/exclude by server prefix, exact name, or description pattern; zero new dependencies
- `open-sse/utils/toolDisclosure.js` — BM25 session index per `connectionId`; schemas tokenized and indexed once, queried cheaply each turn against the user's last message; always pins `ToolSearch`, `tool_choice`-forced tools, and history-referenced tools
- `open-sse/handlers/chatCore.js` — wires both passes after `dedupeTools`, before RTK/headroom; re-anchors `cache_control` annotation to the actual last tool post-filter (bugfix: translator stamps it before disclosure reorders the array)
- `src/sse/handlers/chat.js` — passes `chatSettings.toolDisclosure` through to `handleChatCore`
- Phase 0 measurement log (`TOOLDISCLOSE measure:`) always fires on requests with tools, even when disclosure is off — establishes the baseline data the design doc says to collect first

## Activation

ss-config

Screenshot 2026-08-30 135549

Both phases default **off**. Enable via 9router settings:

```json
{
  "toolDisclosure": {
    "filterEnabled": true,
    "disclosureEnabled": true,
    "maxTools": 20,
    "alwaysInclude": ["ToolSearch"],
    "excludeServers": [],
    "excludeTools": []
  }
}
```

## Tests

30 unit tests across two new test files:
- `tests/unit/tool-filter.test.js` — declarative filter rules, always-include override, glob server patterns, description regex, OpenAI-wrapped tool shape
- `tests/unit/tool-disclosure.test.js` — BM25 scoring correctness, session cache hit/rebuild on tool-set change, pinned-tool preservation, realistic relevance queries (file ops, GitHub PRs, email threads)

## CI

Adds `.github/workflows/test.yml` — triggers on push to `master`/`claude/**`/`feat/**`/`fix/**` and on PRs. No test CI existed before this PR.

## Design doc

Full RFC at `docs/design/progressive-tool-disclosure.md` — covers the two-mode architecture (passive pre-filter vs. active proxy-loop discovery), open questions, phased plan, and codebase audit findings.

## Test plan

- [x] Enable `toolDisclosure: { filterEnabled: true }` and confirm `TOOLDISCLOSE filter:` log appears on requests with tools
- [x] Enable `disclosureEnabled: true, maxTools: 10` against a session with 20+ tools; verify log shows reduction and model still gets relevant tools for the query
- [x] Confirm `ToolSearch` always present in `tools[]` when using Claude Code as client
- [x] Confirm `cache_control: ephemeral` is on the last tool in the outgoing body after filtering
- [x] Run `cd tests && npx vitest run unit/tool-filter.test.js unit/tool-disclosure.test.js`

Screenshot 2026-08-30 134733


nitsuah:
## Post-review fixes (766fa21 → fed69fb)

Ran a high-effort code review after the initial push. Ten findings surfaced; all ten are now fixed across three follow-up commits.

### Correctness

| # | Finding | Fix |
|---|---------|-----|
| 1 | **`toolFilter` was always a no-op** — `toolFilter(tools, toolDisclosure)` passed `{disclosureEnabled, filterEnabled, maxTools}` but `toolFilter` destructures `{excludeServers, excludeTools, …}`. Static filter feature was silently broken regardless of toggle state. | Added `toolDisclosureExcludeServers` / `toolDisclosureExcludeTools` to `DEFAULT_SETTINGS` and threaded them through `chat.js` so `toolFilter` receives a usable config. |
| 2 | **Pinned tools could exceed `maxTools`** — In a long agentic session `extractPinnedNames` accumulates every tool ever called. Once more tools were pinned than the cap, `budget` hit zero and the full pinned set was returned — violating the limit. | Cap `allPinnedTools.slice(0, maxTools)` before computing budget. |
| 3 | **Responses API endpoint silently skipped disclosure** — `handleResponsesCore` called `handleChatCore` without passing `toolDisclosure`. | Added `toolDisclosure` to the signature and pass-through. |
| 4 | **`TOKEN_SAVER_HEADER` bypass didn't reach disclosure** — `tokenSaverEnabled` was computed 70 lines after the disclosure block, so `X-9router-token-saver: off` didn't bypass filter or BM25. | Hoisted `tokenSaverEnabled` before the disclosure block; removed the duplicate `const` that was at the old location. Disclosure sub-blocks are now gated on it; the `cache_control` stamp (see below) intentionally is not. |
| 5 | **`minScore = 0.0` kept zero-score (no-match) tools** — `(scores[i] || 0) >= 0.0` is always true, so tools with no query-term overlap filled remaining budget slots. Docstring said "any matching token" but the code delivered "any tool". | Changed `>=` to `>` (strict). Updated config docstring to match. |
| 6 | **`no-connectionId` path never recorded stats** — The fallback branch (no session anchor) returned a stats object but never called `_recordStats`. Dashboard stats panel was always empty in local dev. | Added `_recordStats` call in the fallback branch. |

### Efficiency

| # | Finding | Fix |
|---|---------|-----|
| 7 | **`JSON.stringify` of full tools array on every request** — `beforeBytes`/`afterBytes` were template-literal args, so both serializations (~100 KB for 40-tool MCP sessions) happened even when debug logging was off. | Gated behind `log?.debug` — serialization is now skipped in production. |
| 8 | **`pruneCache` spread+sorted 500 entries unnecessarily** — `Map` iterates in insertion order so the first N/2 entries are already the oldest. No need for `[..._cache.entries()].sort(…)`. | Replaced with a forward-delete loop — O(N), no allocation. |

### Refactors (altitude + reuse)

| # | Finding | Fix |
|---|---------|-----|
| 9 | **`getToolName` defined identically in 3 files** — `toolDisclosure.js`, `toolFilter.js`, and `toolDeduper.js` all had the same private function. | Exported `getToolName` from `toolDeduper.js` (canonical home). Both new files now import it; local copies removed. |
| 10 | **`cache_control` stamped in 3 places** — Translators stamped it pre-disclosure (wrong last tool); `chatCore` re-anchored post-filter. Split responsibility, no single source of truth. | Translators now only **strip** any incoming client annotation. `chatCore.js` is the single point that stamps the actual last tool after all disclosure passes. The outer block guard drops `tokenSaverEnabled &&` so the stamp survives the bypass header (disclosure sub-blocks remain gated on it). Passthrough mode: skip stamp unless disclosure changed the set. |

nitsuah:
Triage on 2026-08-31: the PR head is not in current master, and no existing non-PR tracker was found. Tracking rollout safety, relevance evaluation, cache behavior, and rollback criteria in #3668. The issue supplements this PR; it does not change the feature's opt-in rollout plan.

nitsuah:
Triage on 2026-08-31: this PR head is not in current master and no non-PR tracker existed. Rollout safety, relevance evaluation, cache behavior, and rollback criteria are tracked in #3668. This supplements the PR and preserves its opt-in rollout plan.

nitsuah:
120 may be a good soft default to utilize if the user enables the feature or not (certain free models seem to max out at 128)

nitsuah:
Pushed a commit implementing this (`4154c1d7`), replying to the comment above: **120 as an always-on hard ceiling, independent of `disclosureEnabled`.**

Kept it as a distinct mechanism from the BM25 `maxTools` you already have rather than changing that default: `maxTools`'s job is per-turn relevance compression (design doc's own target range is 20-30), while this is purely "don't let a raw tool count reach the ~128 several providers/models reject outright." Repurposing the compression default to double as the ceiling would've meant either gutting compression (bumping 20→120) or leaving the actual 502 unaddressed whenever disclosure is off — which is the common case today, since it defaults off.

What it does:
- New `HARD_TOOL_CEILING = 120` in `toolDisclosure.js`, applied in `chatCore.js` right after the existing `filterEnabled`/`disclosureEnabled` passes — fires only if the tool count is *still* above 120 after whatever the user configured (or didn't).
- Reuses `disclosureTools()` itself as the capping mechanism (same pinning — `ToolSearch`, forced `tool_choice`, history-referenced tools — rather than a blind truncate).
- No-ops in the common case where disclosure already capped things lower.
- Still respects the existing `x-9router-token-saver: off` bypass header, consistent with every other pass in that block — figured a user who explicitly opted out of token savers for debugging shouldn't have this silently override that, even though this isn't really a "savings" mechanism. Happy to move it outside that gate if you'd rather it be unconditional — flagging the call I made rather than assuming it's the one you'd want.
- 4 new tests in `tests/unit/tool-disclosure-hard-ceiling.test.js` driving `handleChatCore` directly (mocked executor, no live calls) confirming: caps a 150-tool set with disclosure entirely unconfigured, leaves ≤120 sets untouched, still applies with only the static filter on, and is skipped under the bypass header. Plus 2 tests locking in the constant value in `tool-disclosure.test.js`.
- All 39 tests in the three `tool-disclosure*` files pass (Docker, `node:22-alpine`, per repo convention).

nitsuah:
Resolved the merge conflict against upstream `master` in `open-sse/translator/formats/claude.js` (the tool `cache_control` handling in `prepareClaudeRequest`).

Both sides did the same conceptual thing (strip client-supplied `cache_control` from tools) but diverged on whether to re-stamp it here:
- This branch had moved to stamping the anchor in `chatCore.js`, after progressive tool-disclosure filtering runs, as the single source of truth.
- Upstream `master` had independently added a defer_loading-aware anchor directly in `prepareClaudeRequest` (#3567 — Anthropic rejects a tool with both `defer_loading: true` and `cache_control` set).

Resolution: kept both. `prepareClaudeRequest` restores the defer_loading-aware anchor (via `lastCacheableToolIndex`) so direct callers/tests get a correct anchor on their own, and `chatCore.js`'s post-disclosure-filter stamp still runs afterward for the translated request path unaffected. Verified with `tests/unit/defer-loading-cache-control.test.js` (upstream's new regression test for #3567) plus the tool-disclosure and cache-budget suites — all green except the one pre-existing catalogued baseline failure (`claude-header-forwarding.test.js` gotScraping routing test).

nitsuah:
CI was failing on "Unit tests" after the conflict fix, unrelated to the merge itself: `npm install` in `tests/` (right after the root `npm install`) was hitting a known npm arborist crash — `Cannot read properties of null (reading 'edgesOut')`. Confirmed it's Node 22's bundled npm 10.9.8, not cache staleness (reproduced locally even with a forced cache clean). Fixed the workflow to upgrade to npm 11 before either install runs; verified locally and in CI — [green run](https://github.com/decolua/9router/actions/runs/34529343372).

`mergeStateStatus` is now `CLEAN`.

nitsuah:
@decolua - let me know what it would take to get something like this implemented, happy to discuss further. Appreciate your efforts!

nitsuah:
## Synced with current master + three fixes (ea54a8f4 → e58f4c48)

**Merge (`ea54a8f4`)**: brought in `upstream/master` (`39e36d3d`, v0.5.86). Merge commit, no rebase or force-push.
- `chatCore.js` import conflict: kept both (`toolFilter`/`toolDisclosure` + master's `takeRenamedToolNames`).
- RTK conflict: took master's `preTranslateRtk`-aware line.
- **Semantic conflict git didn't flag:** master now declares `const tokenSaverEnabled` earlier in `handleChatCore`, and this branch declared it again. That's a duplicate `const` → SyntaxError on every chat request. Removed the branch's copy.

**Fix (`8fbbefb2`): cache anchor.** The post-disclosure `cache_control` stamp ran for **every** target format and always marked the **last** tool:
- OpenAI-format upstreams got an Anthropic-only field on a tool (strict gateways reject it). It now only stamps Claude-format bodies.
- Anthropic rejects `cache_control` on a `defer_loading` tool (#3567), and MCP clients put deferred tools last. It now uses master's `lastCacheableToolIndex`.
- Two regression tests via `handleChatCore`, each confirmed to fail without the fix.

**Fix (`e58f4c48`): tool loops.** In a Claude-format tool loop, each tool result comes back as a user turn containing only `tool_result` blocks. The BM25 query became `""`, so selection fell back to declaration order: from the second turn of a loop the task's tools could be dropped, and the tool list reshuffled every turn, which also defeats prompt caching. The query now uses the last user turn that actually has text. Tests cover #3668's "history-referenced tools remain available across multi-turn tool loops" plus an adversarial no-word-overlap query (ToolSearch must stay exposed).

**Also:** removed this PR's `.github/workflows/test.yml`. #4288 proposes repo-wide CI that runs these suites under a no-regression gate, so two workflows would be redundant.

**Verification** (Docker, `node:22`, `CI=true`): tool-disclosure/filter/ceiling suites 57/57; full suite vs master: **0 new failures**, +54 tests. `mergeStateStatus: CLEAN`.

#3668 is updated with what these close and what still needs live-session evidence.

PR #3579, opened 2026-08-27
fix: whitelist Hermes Agent system prompt & randomize API_KEY_SECRET

805889095-eng:
## Summary

Two fixes we've battle-tested in production deployments:

### 1. Hermes Agent amnesia (codebuddy-cn.js)
Hermes Agent ships a ~25K-char system prompt. The CodeBuddy executor's content-filter logic replaces any system prompt `> 2000 chars` or matching `AGENT_PATTERN` with a neutral one — so every Hermes session through the CodeBuddy provider starts amnesiac (doesn't know its own identity/role).

Added a whitelist bypass before the filter: if the system prompt contains Hermes Agent identity markers (`Hermes Agent`, `Nous Research`, `You run on Hermes`, `hermes-agent.nousresearch`), pass it through untouched. Legitimate user system prompts are unaffected.

### 2. Hardcoded API_KEY_SECRET (apiKey.js, CWE-798)
`API_KEY_SECRET` fell back to a hardcoded `"endpoint-proxy-api-key-secret"` — anyone knowing this default could forge valid `sk-` API keys. Now falls back to `crypto.randomBytes(32).toString("hex")` per-process, so deployments MUST set `API_KEY_SECRET` explicitly.

## Tests
- `unit/codebuddy-reasoning-optin.test.js` — 3 passed
- `unit/codebuddy-cn-bonus-recurring.test.js` — 2 passed

## Notes
- Source-only change; `.next` compiled artifacts are gitignored, so no compiled-file diff needed.
- The whitelist regex is case-insensitive and scoped to Hermes-specific markers to avoid false positives.

### DIFF
--- a/open-sse/executors/codebuddy-cn.js
+++ b/open-sse/executors/codebuddy-cn.js
@@ -38,6 +38,13 @@ export class CodeBuddyExecutor extends DefaultExecutor {
         if (!message || message.role !== "system") return message;
         const text = flatten(message.content);
         if (!text) return message;
+        // Whitelist well-known agent frameworks whose (long, legitimate) system
+        // prompts must NOT be replaced. Hermes Agent (hermes-agent.nousresearch.com)
+        // ships a ~25K system prompt that trips the length catch-all below; without
+        // this bypass every session starts amnesiac.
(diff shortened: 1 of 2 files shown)


805889095-eng:
老大决定不在上游提交，仅保留在 fork 分支上。

PR #3591, opened 2026-08-27
fix(opencode): use Responses API for opencode free to fix muse-spark-1.2-contributor-free

anojndr:
## Summary

`muse-spark-1.2-contributor-free` in the OpenCode free provider currently fails on the Chat Completions API with `HTTP 500: {"type":"error","error":{"type":"error","message":"Internal server error"}}`. Switching the provider to the Responses API (`https://opencode.ai/zen/v1/responses`) fixes the model. This PR makes that switch.

This is important because `muse-spark-1.2-contributor-free` is the only frontier tier coding model available in the `opencode free` provider right now, so fixing it restores the provider's core value.

## Root cause

After the earlier migration for opencode free to Responses, the Chat to Responses translator forwarded `max_tokens` verbatim. The OpenAI Responses API does not accept `max_tokens` (it uses `max_output_tokens`), so the upstream rejected the request with `unknown parameter max_tokens` and in some cases surfaced as a 500.

## Changes

* `open-sse/executors/opencode.js` - change `buildUrl()` to always return `https://opencode.ai/zen/v1/responses`, remove the empty `MESSAGES_MODELS` set, enforce `body.stream = true`, and normalize `max_tokens` / `max_completion_tokens` to `max_output_tokens`
* `open-sse/providers/registry/opencode.js` - set `format: "openai-responses"` and `forceStream: true` so `handleChatCore` routes through the Responses translator and streaming handler correctly
* `open-sse/translator/request/openai-responses.js` - fix `openaiToOpenAIResponsesRequest` to map `max_tokens` / `max_completion_tokens` to `max_output_tokens` for both Chat to Responses conversion and direct Responses passthrough (`body.input` path), and strip the Chat fields
* `tests/__baseline__/providers-baseline.json` - update baseline for `opencode` to `openai-responses` with `forceStream`
* `tests/unit/executor-const-guard.test.js` - add regression coverage for format, endpoint and streaming enforcement
* `tests/unit/force-stream-config.test.js` - include `opencode` in forced streaming list

## Verification

* Manual translator smoke test: `max_tokens` and `max_completion_tokens` now correctly become `max_output_tokens` and no `max_tokens` is sent upstream
* `verify-providers.mjs` passes (81 providers)
* Focused vitest: `executor-const-guard` and `force-stream-config` filtered tests pass
* No change to `opencode-go` provider (`zen/go/v1/*` remains chat/completions, messages, and responses via multi-transport)

## Testing

Provider is free and no auth, models are fetched from `https://opencode.ai/zen/v1/models`. After this change a request for `muse-spark-1.2-contributor-free` should succeed where it previously returned 500.

### DIFF
--- a/open-sse/executors/opencode.js
+++ b/open-sse/executors/opencode.js
@@ -1,6 +1,7 @@
 import crypto from "crypto";
 import { BaseExecutor } from "./base.js";
 import { PROVIDERS } from "../config/providers.js";
+import { getThinkingLevels } from "../providers/thinkingLevels.js";
 import { injectReasoningContent } from "../utils/reasoningContentInjector.js";
 import { resolveSessionId } from "../utils/sessionManager.js";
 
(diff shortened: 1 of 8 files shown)


decolua:
Thanks @anojndr for the contribution! Reviewed and merged into master. 🙏

One adjustment during review: the PR routed **every** opencode model to `/zen/v1/responses`. Live testing against upstream showed only `muse-spark-1.2-contributor-free` is served there — the other free models return HTTP 500 on `/responses` while working fine on `/chat/completions`:

| model | /chat/completions | /responses |
|---|---|---|
| muse-spark-1.2-contributor-free | 500 | **200** |
| big-pickle, hy3-free, mimo-v2.5-free, nemotron-3-ultra-free, nemotron-3.5-lightning-free, laguna-s-2.1-free | 200 | **500** |

So the merged version keeps routing per-model: `targetFormat: "openai-responses"` is declared on the Muse Spark registry entry instead of `format` on the provider transport, and `buildUrl()` picks the endpoint from the model id. `forceStream` was dropped since `/responses` works non-streaming too. Everything else from your PR is intact — the `max_output_tokens` translator fix, the capabilities entry, and the max/ultra → xhigh effort clamp (upstream rejects `max` with a 400).

Thanks again — the root-cause analysis on `max_tokens` was spot on.

PR #3645, opened 2026-08-30
fix(antigravity): preserve Gemini thought signatures across tool-call turns

raflyiswadii:
## Problem

Claude Code tools (Read, Edit, Bash, Grep) fail on multi-turn calls when routed through Antigravity/Gemini. The first tool call works. On the next turn the request returns a 400 error and tool use breaks.

## Root cause

Gemini 3.x requires a `thoughtSignature` on every model turn that replays tool-call history. 9router's translator pivots through the OpenAI intermediate format, which has no field for it. The signature was dropped when the assistant's `tool_calls` were converted back to Gemini `functionCall` parts on the next request. Without it, Gemini rejects the turn.

## Fix

Carry the Gemini `thoughtSignature` through the OpenAI round-trip by encoding it into the tool-call `id`, using a `_TSIG_` separator:

- `open-sse/translator/concerns/thoughtSignature.js` (new): `buildToolCallId` / `splitToolCallId` encode and decode the signature, base64url-safe for `[a-zA-Z0-9_-]` ids.
- `open-sse/translator/response/gemini-to-openai.js`: attach the incoming signature to the emitted `tool_call` id so it survives the round-trip.
- `open-sse/translator/request/openai-to-gemini.js`: when replaying tool-call history, restore the original Gemini signature and raw call id instead of substituting a synthetic default.
- `open-sse/handlers/chatCore/nonStreamingHandler.js`: keep Gemini call ids and signatures consistent in the non-streaming path, and preserve upstream Gemini call ids when present.

Backward compatible: when no signature exists, the id format is unchanged. Unknown ids without the separator decode to empty signature.

## Tests

- `tests/translator/antigravity-tool-roundtrip.test.js` (new): asserts the signature survives a non-streaming Claude round-trip through `translateNonStreamingResponse` and back through `translateRequest`.
- `tests/translator/bugs-antigravity.test.js`: adds a signature-replay regression guard.

`npx vitest run translator/antigravity-tool-roundtrip.test.js translator/bugs-antigravity.test.js` passes (8 tests). No regression in the translator suite: the 7 failing tests there all live in files this patch does not touch, and 4 are already listed in `tests/__baseline__/known-fails.txt`.

## Live verification

Tested through the Claude Code CLI against `ag/gemini-3.7-flash-high` and `ag/gemini-3.6-flash-high`. Multi-turn Read, Grep, Bash, Edit all complete (`TOOL_SUCCESS`, `EDIT_TOOL_SUCCESS`).

## Notes

The fix targets the streaming path (the one Claude Code uses) and the non-streaming path (`/v1/messages`) for parity. Scope is limited to Gemini/Antigravity translation; no other provider is affected.

---

Tracking issue: #3646 documents the problem, the approaches considered, and the chosen fix.

### DIFF
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -10,6 +10,54 @@ import { buildRequestDetail, extractRequestConfig, extractUsageFromResponse, sav
 import { appendRequestLog, saveRequestDetail } from "@/lib/usageDb.js";
 import { decloakToolNames } from "../../utils/claudeCloaking.js";
 import { ROLE, RESPONSES_ITEM } from "../../translator/schema/index.js";
+import { buildToolCallId } from "../../translator/concerns/thoughtSignature.js";
+
+const DEFAULT_THOUGHT_SIGNATURE = "";
+
+function extractGeminiThoughtSignature(part, pending = DEFAULT_THOUGHT_SIGNATURE) {
+  return part?.thoughtSignature || part?.thought_signature || pending;
+}
+
+function buildGeminiToolCall(part, index, pendingSignature) {
+  const signature = extractGeminiThoughtSignature(part, pendingSignature);
+  const rawId = part.functionCall.id;
+  // Preserve the upstream Gemini call id when present, else derive a stable one.
+  const id = rawId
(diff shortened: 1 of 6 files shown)


raflyiswadii:
Linked issue: #3646 documents the problem, the approaches considered and rejected, and the chosen fix.

PR #3649, opened 2026-08-31
fix(models): support single model lookup

### LINKED ISSUE #3588: GET /v1/models/{id} returns 404 for every model, breaking Claude Code CLI's interactive mode
**Title:** `GET /v1/models/{id}` returns 404 for every model, breaking Claude Code CLI's interactive mode

**Environment**
- 9Router: v0.5.55 (local, `http://127.0.0.1:20128`)
- Claude Code CLI: v2.1.247
- OS: Windows 11

**Summary**

Claude Code CLI's interactive (chat) mode does a per-model lookup (`GET {ANTHROPIC_BASE_URL}/models/{id}`) before letting you send a message when it's pointed at a custom gateway. On 9Router this endpoint returns 404 for *every* model ID I've tried — including real, catalog-listed models, not just custom combos. `GET /v1/models` (list) works fine and returns the full catalog; only the single-model retrieve path is broken.

Interactive mode therefore cannot be used with 9Router at all right now. Non-interactive mode (`claude -p` / `--print`) works perfectly, since it never calls the single-model endpoint.

**Steps to reproduce**

```bash
export ANTHROPIC_BASE_URL=http://127.0.0.1:20128/v1
export ANTHROPIC_AUTH_TOKEN=<your 9Router API key>
export ANTHROPIC_MODEL=cc/claude-sonnet-5   # a real, catalog-listed model — not a combo
claude
# type any message
```

Result: `There's an issue with the selected model (cc/claude-sonnet-5). It may not exist or you may not have access to it. Run /model to pick a different model.`

`claude -p "hi" --model cc/claude-sonnet-5` with the same env vars works fine and returns a normal reply.

**Direct evidence — the actual HTTP call failing**

```bash
$ curl -s -i -H "Authorization: Bearer <key>" http://127.0.0.1:20128/v1/models/cc/claude-sonnet-5
HTTP/1.1 404 Not Found
{"error":{"message":"Unknown model kind: cc. Supported: image, tts, stt, embedding, image-to-text, web","type":"invalid_request_error"}}
```

Same 404 shape for other IDs:

```bash
$ curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer <key>" http://127.0.0.1:20128/v1/models/work-fast
404
$ curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer <key>" http://127.0.0.1:20128/v1/models/work-deep
404
```

For comparison, the list endpoint works and correctly includes these models:

```bash
$ curl -s -H "Authorization: Bearer <key>" http://127.0.0.1:20128/v1/models | jq '.data[] | select(.id=="cc/claude-sonnet-5")'
{"id":"cc/claude-sonnet-5","object":"model","owned_by":"cc","capabilities":{...},"context_length":1000000,"max_completion_tokens":128000}
```

**Root cause guess**

The error body (`"Unknown model kind: cc. Supported: image, tts, stt, embedding, image-to-text, web"`) suggests `/v1/models/:id` is routed to a *generation-kind* handler (used for image/tts/stt/etc. model listings) rather than to a standard OpenAI-style "retrieve one chat model" handler. It's parsing the path segment as a `kind` enum instead of a model ID, so it 404s for anything that isn't one of those five literal kind names.

**Why this matters for Claude Code specifically**

Traced into the Claude Code CLI binary: when `ANTHROPIC_BASE_URL` points at a non-first-party host, before allowing a chat turn it does a live model-check request. Any `xn` HTTP error with `status === 404` is rendered as:

```js
error:"model_not_found"
`There's an issue with the selected model (${t}). It may not exist or you may not have access to it.${...}`
```

I also confirmed `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1` (which makes Claude Code call the *list* endpoint at startup to populate a model picker) does **not** avoid this — the per-turn check is a separate code path and still 404s even for a model that was successfully discovered via the list endpoint and cached locally.

**Suggested fix (if the root-cause guess above is right)**

Implement `GET /v1/models/{id}` (treat everything after `/v1/models/` as the model identifier, not a fixed `kind` enum) to return the matching entry from the same catalog `GET /v1/models` already serves, 404 only when the id genuinely isn't in the catalog. Happy to test against a fix build.

**Workaround in use today:** `claude -p`/`--print` only; interactive mode avoided until this is fixed.

Jordannst:
## Summary

Closes #3588.

Provider-prefixed model IDs such as cc/claude-sonnet-5 previously flowed into the capability-kind handler and returned a 404. This replaces the one-segment models route with a catch-all route that:

- preserves capability listings at /v1/models/image, /v1/models/tts, and the other existing kind paths;
- matches other paths against the same LLM catalog exposed by GET /v1/models; and
- returns the matching OpenAI-style model object or a model_not_found response.

## Verification

- tests/node_modules/.bin/vitest run unit/v1-model-lookup-3588.test.js --config vitest.config.js: 4 passed
- targeted ESLint passed
- git diff --check passed
- npm run build was attempted but the shared canonical dependency tree lacks the currently declared @node-saml/node-saml package. No dependencies were installed into the runtime tree.


decolua:

PR #3650, opened 2026-08-31
fix(usage): bucket usage by API key identity, not by its mask or the key

### LINKED ISSUE #3640: fix(usage): distinct API keys sharing the same 8-char prefix are collapsed into a single entry in "Usage by API Key"
### Describe the Bug

When multiple API keys share the same initial prefix (e.g. `sk-00275ae3b...` ending in different suffixes), the **Usage by API Key** analytics view (`/dashboard/usage`) collapses all requests from both keys into a single entry, displaying only the name and stats of whichever key was first encountered in the aggregation iteration.

### Steps to Reproduce

1. Generate two distinct API keys in 9Router:
   - Key 1 named `Client-A`: `sk-00275ae3b...key1`
   - Key 2 named `Client-B`: `sk-00275ae3b...key2`
2. Send API requests using `Client-A` on model `gemini-3.7-flash-high`.
3. Send API requests using `Client-B` on model `gemini-3.7-flash-high`.
4. Open `/dashboard/usage` -> Select period `24h` / `Today` -> View `Usage by API Key`.
5. **Actual Result:** Only `Client-A` is visible. All requests, token metrics, and costs from `Client-B` are summed into `Client-A`'s row. `Client-B` never appears in the table.

### Root Cause Analysis

In the usage aggregation service (e.g. `getUsageStats` / Next.js stats route for usage analytics):

1. Key masking uses an 8-character prefix helper:
   ```javascript
   function maskApiKey(key) {
     return key && typeof key === "string" 
       ? (key.length <= 8 ? key.charAt(0) + "***" : key.slice(0, 8) + "***") 
       : null;
   }
   ```
2. The aggregation dictionary `byApiKey` groups entries using:
   ```javascript
   const masked = maskApiKey(apiKey); // e.g. "sk-00275***" for all default generated keys
   const groupKey = `${masked}|${model}|${provider}`;
   ```
3. Because distinct API keys generated by 9Router commonly share the same default prefix (`sk-00275...`), they produce identical `masked` values (`sk-00275***`) and collide on `groupKey`.
4. The first key to register sets `keyName` (`Client-A`), and subsequent requests from `Client-B` mutate the same accumulator object, keeping `Client-A`'s name while summing all metrics together.

### Expected Behavior

Each distinct API Key should retain its own separate grouped bucket and displayName, regardless of key prefix similarities. Grouping in memory should be done by the actual unique `apiKey` string (or `apiKeyId`), and display masking (`maskApiKey`) should only be applied to the presentation output field.

### Suggested Fix

Group by the unique `apiKey` value or key ID:
```javascript
// Change:
// const groupKey = `${maskApiKey(apiKey)}|${model}|${provider}`;

// To:
const groupKey = `${apiKey}|${model}|${provider}`;
```

### Environment
- 9Router Version: `v0.5.59`
- Deployment: Docker (`decolua/9router:latest`)

ntdatt812:
Fixes #3640.

## The collision is structural, not a coincidence

The report says keys "commonly share the same default prefix". It is stronger than that — **every key on one install always does**:

```js
// src/shared/utils/apiKey.js
const key = `sk-${machineId}-${keyId}-${crc}`;
```

`maskApiKey` returns `key.slice(0, 8)`, which is `sk-` plus the first five characters of the machine id. So the mask is identical for every key the install ever issues, and the 24h/today branch grouped on exactly that. The first key seen kept its `keyName` and absorbed the whole install's requests, tokens and cost. A test asserts the shared prefix from two freshly generated keys rather than taking my word for it.

Only 24h/today: `useDailySummary = period !== "24h" && period !== "today"`, and the daily branch keys off `day.byApiKey`, whose names come from the raw key. That matches the reported repro exactly.

## ⚠️ The suggested fix would have leaked the key — and probing that turned up a second bug

`AUDIT-002` in `tests/unit/security-audit.test.js` deliberately pins the mask, because `stats.byApiKey`'s **object keys are returned to the dashboard**. Grouping on the raw key, as the issue suggests, trips that guard for good reason.

Checking whether the other branch honours it, I found it does not. On unmodified `master`:

```
period=24h   buckets=1 rawKeyInObjectKey=false
period=today buckets=1 rawKeyInObjectKey=false
period=7d    buckets=1 rawKeyInObjectKey=true   <-- sk-00275ae3-w035te-1ddb4b0e|gemini-3.7-flash-high|google
period=30d   buckets=1 rawKeyInObjectKey=true
period=all   buckets=1 rawKeyInObjectKey=true
```

`aggregateEntryToDay` names its stored bucket `${apiKey}|${model}|${provider}`, and the daily branch copied that name straight into the response. AUDIT-002 checked the source text of the 24h template only, so it never saw the other path.

## The fix

One shared helper both branches derive their bucket name from:

```js
function apiKeyIdentity(apiKey, keyInfo) {
  if (!apiKey || typeof apiKey !== "string") return "local-no-key";
  if (keyInfo?.id) return keyInfo.id;
  return "anon-" + createHash("sha256").update(apiKey).digest("hex").slice(0, 12);
}
```

The key's own id is unique and not a secret. A key that has since been deleted is not in `apiKeyMap`, so it falls back to a digest — still its own bucket, still nothing sensitive in the payload. The stored `usageDaily` names are untouched; only the response is re-derived. `apiKeyMasked` and `apiKeyKey` stay masked, so the UI is unchanged.

The `lastUsed` patch a few lines up built its lookup from the raw key, so it moves to the same helper or it would stop matching its bucket.

## ⚠️ I changed a security test — please read this hunk

`AUDIT-002`'s "byApiKey object keys should use masked key, not raw key" asserted one exact template string. That form is what let the 7d leak through, so I replaced it with the property it was trying to express — neither branch may name a bucket after the secret, and both must go through the shared helper — and added a **runtime** check in the new test file that walks every period and asserts the key appears in neither the bucket names nor the serialized payload. That is the assertion that actually catches the bug the old one missed.

The other AUDIT-002 case (`apiKeyMasked` present in both paths, count ≥ 4) is untouched and still passes.

## Tests

Reverting each branch independently fails 2 tests each time:

| mutation | result |
|---|---|
| 24h path back to the mask | 2 failed |
| daily path back to the stored raw-key name | 2 failed |

```
vitest run --config tests/vitest.config.js \
  tests/unit/usage-api-key-prefix-collision-3640.test.js \
  tests/unit/security-audit.test.js                        25 passed, 0 failed
```

Neighbouring suites (`cached-token-e2e`, `cached-token-usage`, `db-sqlite-vs-lowdb`, `db-benchmark`, `security-audit`) report **59 passed / 5 skipped both before and after** this branch. The 3 files reported as "failed" in both runs fail only in `afterAll`, on `rmSync` raising `EPERM` against the still-open SQLite handle — a Windows-only teardown artifact, identical on a clean tree. My own test's teardown is wrapped for that reason, and says so.

I have not opened the dashboard; this is proven at the `getUsageStats` boundary, which is where both bugs live.

### DIFF
--- a/src/lib/db/repos/usageRepo.js
+++ b/src/lib/db/repos/usageRepo.js
@@ -1,3 +1,4 @@
+import { createHash } from "node:crypto";
 import { EventEmitter } from "events";
 import { getAdapter } from "../driver.js";
 import { parseJson, stringifyJson } from "../helpers/jsonCol.js";
@@ -7,6 +8,23 @@ function maskApiKey(key) {
   if (!key || typeof key !== "string") return null;
   if (key.length <= 8) return key.charAt(0) + "***";
   return key.slice(0, 8) + "***";
+}
+
+/**
+ * A stable, non-secret identity for one API key, used to bucket usage stats.
+ *
+ * The mask cannot serve: keys are minted as sk-{machineId}-{keyId}-{crc}, so
+ * every key on one install shares its first 8 characters and maskApiKey()
+ * folds all of them onto a single bucket (#3640). The key itself cannot serve
+ * either: these bucket names are returned to the dashboard, which is what
+ * AUDIT-002 forbids. The key's own id satisfies both; a deleted or unregistered
+ * key falls back to a digest, so it still gets its own bucket without the
+ * secret ever reaching the response.
+ */
+function apiKeyIdentity(apiKey, keyInfo) {
+  if (!apiKey || typeof apiKey !== "string") return "local-no-key";
+  if (keyInfo?.id) return keyInfo.id;
+  return "anon-" + createHash("sha256").update(apiKey).digest("hex").slice(0, 12);
 }
 
 const PENDING_TIMEOUT_MS = 60 * 1000;
@@ -500,7 +518,7 @@ export async function getUsageStats(period = "all") {
         if (dateKey > (stats.byAccount[accountKey].lastUsed || "")) stats.byAccount[accountKey].lastUsed = dateKey;
       }
 
-      for (const [akKey, ak] of Object.entries(day.byApiKey || {})) {
+      for (const [, ak] of Object.entries(day.byApiKey || {})) {
         const rawModel = ak.rawModel || "";
         const provider = ak.provider || "";
         const providerDisplayName = providerNodeNameMap[provider] || provider;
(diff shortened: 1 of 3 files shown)


ntdatt812:

PR #3651, opened 2026-08-31
fix(cli): stop re-enabling autostart on every switch to tray mode

### LINKED ISSUE #3628: Windows: Autostart cannot be disabled
On Windows, autostart cannot be disabled.

Even after manually deleting `9router.vbs` from the Startup folder, launching 9Router again recreates the file.

ntdatt812:
Fixes #3628.

## Why deleting the file did not help

`cli.js` re-enabled autostart on **every** switch to tray mode:

```js
// cli/cli.js, the "hide" branch
const { enableAutoStart } = require("./src/cli/tray/autostart");
enableAutoStart(__filename);
```

So both ways of turning it off were undone by the next hide — the tray menu's "Disable" and deleting `9router.vbs` by hand. Nothing recorded that the user had said no.

The issue is filed against Windows, but that call is platform-agnostic: `enableAutoStart` dispatches to `enableMacOS` / `enableWindows` / `enableLinux`, so the LaunchAgent plist and the `.desktop` entry come back the same way.

## The fix

Separate the automatic path from the deliberate one:

- `ensureAutoStart(cliPath)` — for callers that are not the user asking. Writes an entry only if no choice has been recorded yet, then records one. `cli.js` calls this.
- `enableAutoStart(cliPath)` / `disableAutoStart()` — unchanged in meaning, still always act, and both record the choice. The tray menu calls these.

The marker is a small file in the CLI's data directory (`getDataDir()`, the same `%APPDATA%\9router` / `~/.9router` convention as `api/client.js`). `disableAutoStart` records **before** removing the entry, so a failed removal still cannot be read as a fresh install next launch.

A first install still gets autostart on its first hide, exactly as before.

## Tests

`tests/unit/autostart-user-choice-3628.test.js`, 5 tests: first automatic run enables; disable survives the next hide (the reported loop); an entry deleted by hand is not recreated; an explicit enable still works after a disable and survives a later hide; and disable is recorded even when there was no entry to remove.

They run against the **Linux** branch with a temporary `HOME` and `DATA_DIR`. That is deliberate — the Windows branch writes into the real Start Menu Startup folder, which a test must not touch. I checked my own Startup folder before and after the run to confirm nothing was written to it. (It does contain a `9router.vbs` that the running install put there, which is how I could see the reported behaviour first-hand.)

Reverting each guard fails them:

| mutation | result |
|---|---|
| `ensureAutoStart` ignores the recorded choice | 4 failed |
| `disableAutoStart` stops recording the choice | 1 failed |

```
vitest run --config tests/vitest.config.js tests/unit/autostart-user-choice-3628.test.js   5 passed
vitest run --config tests/vitest.config.js tests/unit/cli-*.test.js \
       tests/unit/security-audit.test.js                                        38 passed, before and after
```

I have not exercised the Windows or macOS branches end-to-end — only the shared decision logic that both go through.

### DIFF
--- a/cli/cli.js
+++ b/cli/cli.js
@@ -761,10 +761,11 @@ function startServer(updatePromise) {
           const { clearScreen } = require("./src/cli/utils/display");
           clearScreen();
 
-          // Enable auto startup on OS boot
+          // Enable auto startup on OS boot, but only the first time. Doing it
+          // on every hide undid both ways of turning it off (#3628).
           try {
-            const { enableAutoStart } = require("./src/cli/tray/autostart");
-            enableAutoStart(__filename);
+            const { ensureAutoStart } = require("./src/cli/tray/autostart");
+            ensureAutoStart(__filename);
           } catch (e) { }
 
           if (process.platform === "darwin") {
--- a/cli/src/cli/tray/autostart.js
+++ b/cli/src/cli/tray/autostart.js
@@ -6,6 +6,49 @@ const { execSync } = require("child_process");
 const APP_NAME = "9router";
 const APP_LABEL = "com.9router.autostart";
 
+/** Same data directory the CLI's api/client.js uses. */
+function getDataDir() {
+  if (process.env.DATA_DIR) return process.env.DATA_DIR;
+  if (process.platform === "win32") {
+    return path.join(process.env.APPDATA || path.join(os.homedir(), "AppData", "Roaming"), APP_NAME);
+  }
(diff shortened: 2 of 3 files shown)


ntdatt812:
Superseded by #3881, which carries this commit unchanged, rebased onto current `master` and grouped with the other fixes in the same area. Closing here so the two do not sit in the queue as duplicates — reopen if you would rather review it on its own.

PR #3660, opened 2026-08-31
fix: apply persisted theme before first paint to avoid flash on reload

docaohieu2808:
### Problem

The theme is applied from the client store inside a `useEffect` (`ThemeProvider` → `themeStore.initTheme`), which runs after hydration. On a reload the page therefore paints the default (light) theme for a frame before the stored `dark` theme is reapplied — a visible white flash.

### Fix

Add a small blocking `<head>` script that reads the persisted zustand `theme` key from `localStorage` and sets the `dark` class on `document.documentElement` before first paint. It mirrors exactly what `applyTheme()` does (including the `system` → `prefers-color-scheme` resolution), so there is no behaviour change — only the flash is removed. Wrapped in try/catch; a light or system-light theme is unaffected.

### DIFF
--- a/src/app/layout.js
+++ b/src/app/layout.js
@@ -32,6 +32,14 @@ export default function RootLayout({ children }) {
   return (
     <html lang="en" suppressHydrationWarning>
       <head>
+        {/* Apply persisted theme before first paint so a reload does not flash the
+            default (light) theme before the client store hydrates. Mirrors the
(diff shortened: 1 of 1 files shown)


decolua:
Thanks @docaohieu2808 for the contribution! Reviewed and merged into master. 🙏

PR #3665, opened 2026-09-01
Fix OpenRouter tool schemas for Copilot agent requests

nitsuah:
## Summary

Makes fallback routing safer for Copilot Chat and other OpenAI-compatible clients that send function tools:

- treat OpenAI `type: "function"` tools as a required capability when ordering combo candidates
- normalize OpenRouter function parameter schemas before dispatch by removing `pattern` constraints that strict OpenRouter backends reject
- retain all other schema structure, including property names, types, `required`, enums, descriptions, and nested objects

## Why

A Copilot Agent request containing 50 function tools was forwarded unchanged to OpenRouter/Cohere and rejected with:

```text
invalid function at tools[1].function: invalid 'parameters' provided: pattern must be a valid regex
```

The current `or-free` fallback chain then traverses overloaded, rate-limited, and stale free-model routes, making the stream unreliable for Copilot.

## What changed since the initial draft

- **Fixed a correctness edge case**: the original pattern-stripping walk tracked "am I inside a properties map" with a single boolean derived from the parent key's name. That misfires for a tool parameter literally named `properties` — its own schema got treated as if it were a properties map for one level, silently skipping validation of a real `pattern` keyword directly inside it. Rewrote the walk to special-case `properties` explicitly (iterate its name→schema entries) rather than threading a flag through recursion, which removes the ambiguity and is easier to follow. New regression test: `"handles a schema property literally named 'properties' without swallowing its own pattern"`.
- **Added the redacted telemetry this PR's own follow-up issue asked for**: `normalizeToolSchemasForProvider` now logs a redacted diagnostic (`provider=openrouter rule=strip-invalid-pattern removed=N`) via the existing request logger when normalization actually changes something — never the schema/field content itself. Two new tests cover this (fires with a count, doesn't fire when nothing needed removing).
- **Seeded `docs/tool-schema-compatibility-matrix.md`** with the one confirmed entry (OpenRouter/Cohere malformed-`pattern` rejection), per the follow-up list in #3667.
- **Actually ran the tests.** The original draft noted `vitest` wasn't loadable from a clean checkout in that sandbox. It is here: `cd tests && npx vitest run unit/tool-schema-compatibility.test.js unit/combo-autoswitch.test.js` → **21/21 passing** (Docker, `node:22-alpine`, per this repo's tooling convention).

## Scope

This intentionally does not remove tool support globally. It adapts only the OpenRouter heterogeneous backend pool, based on an observed provider rejection. Whether OpenRouter/Cohere rejects a broader class of regex dialect features than just malformed patterns remains unconfirmed — tracked as an open question in the new compatibility matrix doc rather than guessed at here.

## Validation

- `cd tests && npx vitest run unit/tool-schema-compatibility.test.js unit/combo-autoswitch.test.js` — 21/21 passing.
- `node --check` on all changed source files.
- Reviewed the full 40+ provider list this normalization does *not* apply to (only `provider === "openrouter"` is affected) — confirmed via `normalizeToolSchemasForProvider`'s own early-return and the `"does not modify schemas for other providers"` test.

## Test plan

- [x] Send a function-tool request through a combo routing to OpenRouter with a schema containing an invalid `pattern` (e.g. `"["`) — confirm it's stripped and the request no longer 400s.
- [x] Confirm a schema property literally named `pattern` is preserved as a property, with only its own invalid constraint removed.
- [x] Confirm a schema property literally named `properties` doesn't suppress pattern-validation of its own nested schema.
- [x] Confirm the redacted telemetry log fires with a count when normalization changes something, and stays silent when it doesn't.
- [x] `cd tests && npx vitest run unit/tool-schema-compatibility.test.js unit/combo-autoswitch.test.js` — 21/21 passing.

### DIFF
--- a/docs/tool-schema-compatibility-matrix.md
+++ b/docs/tool-schema-compatibility-matrix.md
@@ -1,0 +1,19 @@
+# Tool-Schema Compatibility Matrix
+
+Tracks provider-specific rejections of otherwise-valid OpenAI function-tool JSON Schemas, discovered in production. Each entry needs a reproduced failure (not a hypothesis) before it's added — see the scope guardrail in [#3667](https://github.com/decolua/9router/issues/3667): no global sanitizer or lowest-common-denominator schema weakening without evidence.
+
+| Provider (routed via) | Rejected construct | Symptom | Handling | Evidence |
+|---|---|---|---|---|
+| OpenRouter → Cohere | Malformed `pattern` regex constraint (e.g. `"["`) on a `function.parameters` property | `HTTP 400: invalid function at tools[N].function: invalid 'parameters' provided: pattern must be a valid regex` | `open-sse/utils/toolSchemaCompatibility.js` strips only malformed `pattern` values before OpenRouter dispatch; valid patterns and all other schema structure pass through unchanged | Copilot Chat agent request, ~50 function tools, observed via a free-tier combo route. See #3667. |
+
+## Adding an entry
+
+1. Reproduce the failure with the minimal schema fragment that triggers it (redact tool/field names if they're user- or workspace-specific).
+2. Confirm it's a genuine provider limitation, not a 9router translation bug — check whether the same request succeeds against the provider directly (bypassing combo routing).
+3. Add a row above with the exact upstream error text and which 9router mechanism (if any) compensates.
+4. If no mechanism exists yet, leave the "Handling" column as "none — tracked in #NNNN" rather than adding a fix speculatively.
+
+## Open questions
+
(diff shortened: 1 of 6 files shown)


nitsuah:
@decolua friendly bump on this as i dont think its been addressed.

nitsuah:
Brought up to date with `master` (`39e36d3d`, v0.5.86; was 151 commits behind) via a merge commit. It merged cleanly, and the tool-schema compatibility change is still needed on current master.

Verified in Docker (`node:22`, `CI=true`): `tool-schema-compatibility.test.js` and `combo-autoswitch.test.js` pass 21/21. Full suite: no new failures attributable to this branch.

PR #3679, opened 2026-09-01
fix(auth): protect root /responses rewrite

### LINKED ISSUE #3677: [Security] Middleware bypass allows unauthenticated access to LLM gateway via /responses endpoint
### Description

The 9Router middleware (`src/dashboardGuard.js`) enforces API‑Key authentication on the main LLM API prefixes (`/v1`, `/v1beta`, `/api/v1`, `/api/v1beta`, `/codex`). However, the `/responses` path – which is rewritten to `/api/v1/responses` by `next.config.mjs` – is **not** included in the middleware's guarded list.

Because Next.js middleware executes **before** URL rewriting, requests to `/responses` bypass all authentication checks and are passed directly to the chat handler (`src/app/api/v1/responses/route.js`). This allows unauthenticated requests to enter the full chat pipeline (model resolution, account selection, upstream execution) without any credential validation.

Additionally, due to Next.js path normalization, certain path variants that resolve to `/responses` can also evade the middleware.

### Affected Versions

- All versions up to and including **0.5.55** (current public version).

### Steps to Reproduce

1. Send a chat request to a protected endpoint such as `/v1/chat/completions`. The middleware enforces authentication and returns a `401 Unauthorized` when no valid API key is provided.

2. Send a similar chat request to the `/responses` endpoint instead. The middleware does **not** intercept this path, and the request is rewritten to the chat handler, which processes it and returns a response (rather than a 401). This confirms the authentication bypass.

3. Certain normalized path variants that resolve to `/responses` (e.g., those containing `../` segments) exhibit the same bypass behaviour.

4. **Default configuration note:** The project documentation and `.env.example` recommend `REQUIRE_API_KEY=false`. In this default setup, the chat handler itself also omits credential checks, resulting in completely anonymous access with no credentials required.

### Expected Behavior

- The `/responses` endpoint should be subject to the same authentication requirements as `/v1/chat/completions`.
- The middleware should cover all routes that accept chat requests, including rewritten paths.

### Actual Behavior

- `/responses` is omitted from the guarded list, so all middleware authentication checks are bypassed.
- The chat handler processes requests without validating any credentials (unless `REQUIRE_API_KEY=true` is explicitly set).
- `OPTIONS /responses` also returns permissive CORS headers, allowing cross-site requests from arbitrary origins.

### Impact

- **Critical** – complete bypass of API key authentication.
- Attackers can consume quota and incur costs for all configured upstream accounts without permission.
- Request payloads are forwarded to external LLM providers, potentially exposing user data.
- In default deployments, no credentials of any kind are required to access the full gateway functionality.
- Combined with other vulnerabilities (e.g., SSRF or provider misconfiguration), the attack surface expands further.

### Root Cause Analysis

The middleware's guarded list in `src/dashboardGuard.js` includes only the following prefixes:

```js
const PUBLIC_PREFIXES = ["/v1", "/v1beta", "/api/v1", "/api/v1beta", "/codex"];
```

Meanwhile, `next.config.mjs` contains a rewrite rule:

```js
{
  source: "/responses",
  destination: "/api/v1/responses"
}
```

Since middleware execution occurs before URL rewriting, `/responses` never matches `PUBLIC_PREFIXES` and falls through to the unauthenticated branch.

### Proposed Fix

- Add `/responses` to `PUBLIC_PREFIXES` in `src/dashboardGuard.js`.
- Alternatively, move the `responses` route under an already-protected path (e.g., `/api/v1/`) rather than exposing it at the root level.
- Consider implementing authentication checks directly inside the chat handler as a defence-in-depth measure, independent of middleware.
- Update deployment documentation to strongly recommend `REQUIRE_API_KEY=true` for production environments.

### Additional Context

I have verified this behaviour in a local deployment. If you need any further details, I'm happy to provide them privately. Thank you for maintaining this project!

Jordannst:
Fixes #3677

## Problem
`/responses` is a root-level alias for the OpenAI Responses API:

```text
/responses  →  /api/v1/responses
```

The alias is applied by a Next.js rewrite. Middleware runs **before** rewrites, so `dashboardGuard` receives the original pathname (`/responses`), not `/api/v1/responses`.

Before this change, `/responses` was absent from the LLM API prefix list. A remote request could therefore fall through the API-key gate before the rewrite dispatched it to the Responses handler.

## Root cause
`src/dashboardGuard.js` protects public LLM routes through `isPublicLlmApi()`. It recognizes `/v1`, `/v1beta`, `/api/v1`, `/api/v1beta`, and `/codex`, but not the root `/responses` alias declared in `next.config.mjs`.

## Change
- Add `/responses` to `PUBLIC_PREFIXES`, so the pre-rewrite request follows the established API-key validation path.
- Add regression coverage for both remote cases:
  1. no API key → `401 API key required for remote API access`;
  2. valid bearer key → request proceeds and `validateApiKey` is called.

## Behavior preserved
This reuses the existing guard; it does not change route handlers, API-key format, loopback handling, CLI-token access, or the existing `/v1`, `/api/v1`, and `/codex` behavior.

### Reproduction on current `master` (`90b52e0`)
The two added regression tests fail before the guard change:
- unauthenticated `/responses` falls through instead of returning `401`;
- a valid bearer key is never validated because the request bypasses the LLM API gate.

### After the fix
```bash
cd tests && npm test -- unit/dashboard-guard.test.js
```

Result: **24 passed, 0 failed**.

`git diff --check` also passes.

## Scope
This intentionally fixes only middleware coverage for the existing `/responses` alias. It does not alter CORS policy or broaden authentication behavior beyond matching the other public LLM entry points.

### DIFF
--- a/src/dashboardGuard.js
+++ b/src/dashboardGuard.js
@@ -34,7 +34,8 @@ const PUBLIC_API_PATHS = [
 ];
 
 // Public top-level prefixes (LLM API endpoints with their own API key auth).
-const PUBLIC_PREFIXES = ["/v1", "/v1beta", "/api/v1", "/api/v1beta", "/codex"];
+// Keep root-level rewrites here too: middleware runs before Next.js rewrites.
+const PUBLIC_PREFIXES = ["/v1", "/v1beta", "/api/v1", "/api/v1beta", "/codex", "/responses"];
 
 // Always require JWT token regardless of requireLogin setting
 const ALWAYS_PROTECTED = [
--- a/tests/unit/dashboard-guard.test.js
+++ b/tests/unit/dashboard-guard.test.js
@@ -125,6 +125,25 @@ describe("dashboard guard public LLM API access", () => {
     expect(response.body.error).toBe("API key required for remote API access");
   });
 
+  it("rejects remote /responses rewrite without API key", async () => {
+    const response = await proxy(request("/responses", { host: "router.example.com" }));
+
+    expect(response.status).toBe(401);
+    expect(response.body.error).toBe("API key required for remote API access");
+  });
+
+  it("allows remote /responses rewrite with a valid API key", async () => {
+    mocks.validateApiKey.mockResolvedValue(true);
+
+    const response = await proxy(request("/responses", {
+      host: "router.example.com",
+      authorization: "Bearer sk-valid",
+    }));
+
+    expect(response).toBe(mocks.nextResponse);
+    expect(mocks.validateApiKey).toHaveBeenCalledWith("sk-valid");
+  });
+
   it("allows remote codex rewrite with valid API key", async () => {
     mocks.validateApiKey.mockResolvedValue(true);
 


decolua:

PR #3683, opened 2026-09-01
fix(chatCore): return an Anthropic Message on the forced-SSE JSON path

### LINKED ISSUE #3682: Claude client gets a chat.completion body when a forced-streaming provider is retried without streaming
## Summary

When a Claude client (`/v1/messages`) sends `stream: false` to a provider that forces streaming, `handleForcedSSEToJson` returns an OpenAI `chat.completion` body instead of an Anthropic `Message`. The client sees HTTP 200 with a JSON body it cannot parse, and the request fails.

`handleForcedSSEToJson` in `open-sse/handlers/chatCore/sseToJsonHandler.js` branches on `sourceFormat` for Responses API and Gemini clients, but has no `FORMATS.CLAUDE` branch — Claude falls through to the OpenAI shape on both paths:

- Responses-API upstream (e.g. codex): [`sseToJsonHandler.js#L269-L282`](https://github.com/decolua/9router/blob/master/open-sse/handlers/chatCore/sseToJsonHandler.js#L269-L282) builds `object: "chat.completion"`.
- Chat-completions upstream: [`sseToJsonHandler.js#L350-L352`](https://github.com/decolua/9router/blob/master/open-sse/handlers/chatCore/sseToJsonHandler.js#L350-L352) returns `parsed` as-is.

The non-streaming path already has the right converter (`openAICompletionToClaudeMessage` in `nonStreamingHandler.js`); this path just never calls it.

## Reproduction

1. Configure a combo whose fallback reaches a forced-streaming provider — a `cx/*` (codex) model reproduces it.
2. Point Claude Code at 9router and issue a request that fails stream creation upstream, so the client retries with `stream: false`.
3. The retry returns HTTP 200 with `{"object": "chat.completion", ...}`.

Observed in Claude Code as:

```
API Error: API returned an empty or malformed response (HTTP 200) — check for a
proxy or gateway intercepting the request. Response: content-type json, body is
JSON but not a Message. This was the non-streaming retry of streaming request,
which failed with: 404_stream_creation.
```

Direct check against a running instance:

```bash
curl -sS "$ROUTER/v1/messages" \
  -H 'content-type: application/json' \
  -H "x-api-key: $KEY" \
  -d '{"model":"<combo reaching a cx/* model>","max_tokens":64,"stream":false,
       "messages":[{"role":"user","content":"ok"}]}' | jq '{type, object}'
```

Expected `{"type": "message"}`; actual `{"object": "chat.completion"}`.

## Expected

A Claude client should get `type: "message"` with `content` blocks, `stop_reason`, and Anthropic-shaped `usage`, on both the Responses-API and chat-completions branches — matching what the streaming path and the plain non-streaming path already return.

## Notes

This is not limited to codex: any provider with `forceStream: true` reaches the same code. It surfaces most often behind Claude Code because a failed stream creation triggers an automatic non-streaming retry, so a single upstream hiccup ends the session.

I have a fix with tests and will open a PR referencing this issue.

f-liva:
Fixes #3682.

## Problem

`handleForcedSSEToJson` handles the Responses-API and Gemini clients but has no `FORMATS.CLAUDE` branch, so a Claude client (`/v1/messages`) that retries with `stream: false` against a forced-streaming provider receives an OpenAI `chat.completion` body on HTTP 200 and cannot parse it. Both branches are affected — the Responses-API one builds `object: "chat.completion"`, the chat-completions one returns `parsed` untouched.

Behind Claude Code this ends the session: a failed stream creation triggers an automatic non-streaming retry, which then fails on the body shape.

## Change

- Chat-completions branch: converts with `openAICompletionToClaudeMessage` when `sourceFormat === FORMATS.CLAUDE`.
- Responses-API branch: adds `responsesToClaudeMessage`, which maps `output` items to `text` / `tool_use` blocks and folds the cache counters into `input_tokens` the same way the existing OpenAI branch does.
- `openAICompletionToClaudeMessage` moves out of `nonStreamingHandler.js` into `open-sse/handlers/chatCore/completionToClaude.js` so both paths share one converter and cannot drift. Extraction only — no behaviour change for the existing caller.

Non-Claude clients are untouched: the new code is behind a `sourceFormat === FORMATS.CLAUDE` check on both paths.

## Tests

`tests/unit/claude-client-forced-sse-nonstream.test.js`, following the shape of the existing `openai-responses-nonstream.test.js`:

- Anthropic Message from a chat-completions upstream (text + `tool_use`, `stop_reason: "tool_use"`, Anthropic `usage`)
- Anthropic Message from a Responses-API upstream
- a plain chat client still gets `chat.completion` (guards against over-reach)

Verified red before the fix (the two Claude cases fail, the chat case passes) and green after.

## Regression check

Full suite run before and after on the same checkout, compared by test name:

- fails before: 89
- fails after: 85
- **regressions: 0**

The 4 that flip to green are the 2 new Claude cases plus 2 `xai-oauth-service` cases, which are the network-dependent flakes documented in `CLAUDE.md`. `npx eslint` is clean on the touched files.

### DIFF
--- a/open-sse/handlers/chatCore/completionToClaude.js
+++ b/open-sse/handlers/chatCore/completionToClaude.js
@@ -1,0 +1,57 @@
+import { FORMATS } from "../../translator/formats.js";
+import { fromOpenAIFinish } from "../../translator/concerns/finishReason.js";
+
+function parseToolArguments(value) {
+  if (!value) return {};
+  if (typeof value === "object") return value;
+  try {
+    return JSON.parse(value);
+  } catch {
+    return {};
+  }
+}
+
+/**
+ * Convert an OpenAI Chat Completions body into an Anthropic Message.
+ * Shared by the non-streaming path and the forced-SSE→JSON path so a Claude
+ * client gets the same body whichever route the request took.
+ */
+export function openAICompletionToClaudeMessage(responseBody) {
+  if (!responseBody?.choices?.[0]) return responseBody;
+  const choice = responseBody.choices[0];
+  const message = choice.message || {};
+  const content = [];
+
+  const reasoning = message.reasoning_content || message.provider_specific_fields?.reasoning_content || "";
+  if (reasoning) {
+    content.push({ type: "thinking", thinking: reasoning });
+  }
+  if (typeof message.content === "string" && message.content.length > 0) {
+    content.push({ type: "text", text: message.content });
+  }
+  for (const toolCall of message.tool_calls || []) {
+    const fn = toolCall.function || {};
+    content.push({
+      type: "tool_use",
+      id: toolCall.id || `toolu_${Date.now()}_${content.length}`,
+      name: fn.name || toolCall.name || "",
+      input: parseToolArguments(fn.arguments || toolCall.arguments),
+    });
+  }
+  if (content.length === 0) content.push({ type: "text", text: "" });
+
+  const usage = responseBody.usage || {};
+  return {
+    id: String(responseBody.id || `msg_${Date.now()}`).replace(/^chatcmpl-/, ""),
+    type: "message",
+    role: "assistant",
+    model: responseBody.model || "unknown",
+    content,
+    stop_reason: fromOpenAIFinish(choice.finish_reason, FORMATS.CLAUDE),
+    stop_sequence: null,
+    usage: {
+      input_tokens: usage.prompt_tokens || usage.input_tokens || 0,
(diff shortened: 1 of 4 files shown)


f-liva:
Amended twice: (1) empty completions now return `content: []` instead of an empty text block — clients replay assistant messages back into requests and Anthropic 400s empty text blocks; (2) the conversion now echoes the client-requested model instead of the provider's (`glm-5.3` → Claude Code "Session model glm-5.3 could not be restored"), see #3692 / #3693.

PR #3684, opened 2026-09-01
fix(antigravity): strike-break optimistic quota readings that keep 429ing

### LINKED ISSUE #3681: fix(antigravity): reset-aware CACHE_BLOCK is bypassed when fetchAvailableModels still reports remaining quota — 429 storms across all accounts (v0.5.59)
## Summary

The quota-aware routing shipped in #3561 (v0.5.59) only applies the reset-aware `CACHE_BLOCK` when `fetchAvailableModels` reports **0% remaining** for an account/model pair. Since Google's quota API currently reports **90%+ remaining while generation requests still fail with 429** (dual-limit mismatch introduced with the March 2026 quota rework), the guard is bypassed and every account falls back to the fixed 300s blind lock — so a 7-account pool gets re-hammered every 5 minutes for days, multiplying upstream 429 pressure and adding ~15s latency per request.

In short: **the fix trusts the quota API as the source of truth, and the quota API is currently lying.**

## Environment

- 9router `v0.5.59` (ghcr.io/decolua/9router:latest, built 2026-08-29), Docker
- Antigravity provider, 7 OAuth accounts in rotation
- Clients: OpenAI-compatible `/v1/chat/completions`, streaming, multi-turn agent traffic with tool calls
- All 7 accounts affected simultaneously; reproducible continuously since 2026-08-30

## Evidence (production, 7-account pool)

429 lines in container logs per day:

| Date (UTC) | 429 log hits | POSTs (client requests) |
|---|---|---|
| 2026-08-30 | 7,059 | — |
| 2026-08-31 | 49,635 | — |
| 2026-09-01 (partial) | 24,118 | — |
| last 24h window | 61,745 | 36,581 |

Each client request fans out through account fallback until all 7 accounts are locked → ~1.7 accounts burned per request on average, p50 added latency ~14.7s/request.

Log pattern (accounts anonymized):

```
[08:09:23] ✗ ERROR 429 · antigravity/gemini-3.7-flash-high · 14718ms
[08:09:23] ℹ️  [AG_QUOTA] 8cca… | 429 on gemini-3.7-flash-high — refreshing quota
           (no "quota exhausted; CACHE_BLOCK until" line follows — guard returned null)
[08:09:24] ⚠️  [AUTH] Antigravity | acc1@*** locked modelLock_gemini-3.7-flash-high for 300s [429]
[08:09:24] ⚠️  [FALLBACK] ⇄ ACC:Antigravity | acc1@*** UNAVAILABLE (429) → NEXT ACCOUNT
… (×7 accounts) …
[08:09:24] ⚠️  [AUTH] antigravity | all 7 accounts locked for gemini-3.7-flash-high (reset after 2s)
```

Upstream 429 body — note the **second-level reset hint** Google includes:

```json
{"error": {"code": 429,
  "message": "Resource has been exhausted (e.g. check quota (reset after 14s))",
  "status": "RESOURCE_EXHAUSTED"}}
```

Meanwhile `fetchAvailableModels` (the same API powering the dashboard usage bars) keeps reporting 60–100% remaining for those same accounts/models.

## Root cause

In the deployed bundle (`.next/server/chunks/4572.js`, module `45974` — corresponds to `src/sse/services/antigravityQuota.js`), the reset-aware check is:

```js
async function m(connId, provider, model, …) {          // per-429 quota check
  info("AG_QUOTA", `${connId.slice(0,8)} | ${provider} on ${model} — refreshing quota`);
  const q = (await refreshQuotaCached(connId, …))?.[model];
  if (!q || q.remainingPercentage > 0 || !q.resetAt) return null;   // ← bypass
  const until = new Date(q.resetAt).getTime();
  return until <= Date.now() ? null
    : (warn("AG_QUOTA", `… quota exhausted; CACHE_BLOCK until ${q.resetAt}`), until);
}
```

`remainingPercentage > 0 → return null` was the right call when the quota API was trustworthy: it correctly distinguishes real model-lockout 429s from ordinary burst 429s. But with Google's current behavior — **sprint/weekly dual pools where the reported fraction no longer reflects what generation endpoints admit** — the optimistic value defeats the guard, and the flow degrades to: 429 → 300s `modelLock_*` → retry → 429, across all accounts, indefinitely.

## Why this isn't client-side prompt blocking

That failure mode (429 on competitor-branded system prompts, fixed via the v0.5.59 sanitization rule table) is real but distinct: here the same 429s hit plain Gemini-only agent traffic with no branded prompts, all 7 accounts at once, continuously for 3 days, while the usage bars stay full. #3274 ("429 resource has been exhausted meanwhile all my accounts still 100% available quota") describes exactly this signature; it was mis-attributed to the prompt-branding issue.

## Upstream corroboration (Google-side behavior change)

- Google AI Dev Forum: *"[URGENT] Enterprise Account Locked: HTTP 429 RESOURCE_EXHAUSTED loop (UI shows 100% Quota)"* — every generation request 429s while the Antigravity UI shows 100%.
- CLIProxyAPI #1015: `fetchAvailableModels` shows 60–100% remaining, `generateContent`/`streamGenerateContent` return 429 for the same accounts (17/17 affected).
- Community consensus: since the March 2026 rework, Google uses dual limits (5-hour sprint + weekly baseline); the reported fraction does not reflect the weekly pool, so the quota API/UI systematically over-reports availability.

## Proposed fix

Keep the current reset-aware path for the honest-0% case, and add a strike-based fallback for the dishonest->0% case:

1. **N-strike circuit breaker** — if the same `connectionId + model` receives N consecutive 429s (default N=3) **while the quota API claims remaining > 0**, treat the quota reading as untrusted for that pair and apply `CACHE_BLOCK` anyway, for `max(parsedResetHint, defaultCooldownMs)`.
2. **Parse the embedded reset hint / Retry-After** — the 429 body often contains `(reset after 14s)`; use it (bounded by the existing `MAX_RATE_LIMIT_COOLDOWN_MS` cap) as the cooldown floor instead of the fixed 300s.
3. **Config knob** — e.g. `antigravity.quotaTrustMode: "optimistic" | "strike-based"` (default `strike-based`), which also addresses #3343 (configurable 429 backoff).
4. On cache-block expiry, the existing 30s-throttled quota refresh naturally re-checks before unblocking — no new refresh machinery needed.

This keeps the current fast-recovery behavior for genuine burst 429s (short hint → short block) and stops the multi-day all-account 429 storms for the stalled-pool case.

## Related

- #3561 (the original quota-aware routing — this is its blind spot)
- #3274 (same symptom, currently attributed to prompt branding)
- #3343 (configurable 429 backoff — same knob)
- #3358/#3359 (prompt sanitization — related but separate failure mode)

Happy to provide the full deobfuscated module or additional log dumps if useful.

louis-cai:
Closes #3681.

## Problem

`handleAntigravityQuotaError` (shipped in #3561) trusts `remainingPercentage > 0` as healthy and returns null — the reset-aware `CACHE_BLOCK` never engages when Google's `fetchAvailableModels` over-reports availability. Since the March 2026 dual-pool rework, the quota API systematically reports 60–100% remaining while generation endpoints keep returning 429 (sprint/weekly baseline mismatch). Result on a 7-account pool (v0.5.59, production data in #3681): every account cycles the 300s `modelLock_*` retry loop, ~60k 429s/day, ~15s added latency per request, for days until the weekly pool actually resets.

Repro evidence in #3681: deobfuscated v0.5.59 bundle (`.next/server/chunks/4572.js` module 45974), daily 429 counts, upstream 429 bodies containing `reset after 14s` hints, and upstream corroboration (Google AI Dev Forum Ultra 100%-quota loop; CLIProxyAPI #1015 — 17/17 accounts 429 while quota shows 60–100%).

## Fix

Strike-based circuit breaker on the **optimistic path only**:

- After **3 consecutive 429s** for the same `connectionId|model` within a **60s window** while the quota API still reports `remainingPercentage > 0`, treat the reading as untrusted and return a **15-minute** cache-block (same `resetsAtMs` contract the exhausted-0% path uses).
- Strikes reset when the window lapses; the exhausted-0% `resetAt` path is byte-for-byte unchanged; bursts of 1–2 429s (genuine transient) keep today's behavior.
- Successful requests clear the pair's strikes via the existing success path.

No config surface added yet — constants are module-level and trivially hoisted into settings later if wanted (would also cover #3343).

## Tests

3 new cases in `tests/unit/antigravity-quota-routing.test.js` (existing 8 untouched):

1. strike-breaks after 3 optimistic 429s within 60s, blocking until +15m
2. resets the counter when strikes fall outside the 60s window
3. optimistic path returns null without poisoning the shared quota cache

```
npx vitest run unit/antigravity-quota-routing.test.js   # from tests/
Test Files  1 passed (1)
     Tests  11 passed (11)
```

Also validated against the live v0.5.59 container with an equivalent bundle patch before writing the source PR (429 rate dropped from ~2,400/3min baseline to double digits).

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,0 +1,6170 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .:
+    dependencies:
+      '@dnd-kit/core':
+        specifier: ^6.3.1
+        version: 6.3.1(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
+      '@dnd-kit/modifiers':
+        specifier: ^9.0.0
+        version: 9.0.0(@dnd-kit/core@6.3.1(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(react@19.2.4)
+      '@dnd-kit/sortable':
+        specifier: ^10.0.0
+        version: 10.0.0(@dnd-kit/core@6.3.1(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(react@19.2.4)
+      '@dnd-kit/utilities':
+        specifier: ^3.2.2
+        version: 3.2.2(react@19.2.4)
+      '@monaco-editor/react':
+        specifier: ^4.7.0
+        version: 4.7.0(monaco-editor@0.55.1)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
+      '@next/third-parties':
+        specifier: ^16.2.9
+        version: 16.3.4(next@16.3.4(@babel/core@7.29.7)(react-dom@19.2.4(react@19.2.4))(react@19.2.4))(react@19.2.4)
+      '@node-saml/node-saml':
+        specifier: ^5.1.0
+        version: 5.1.0
+      '@xyflow/react':
+        specifier: ^12.10.1
+        version: 12.11.5(immer@11.1.18)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
+      bcryptjs:
+        specifier: ^3.0.3
+        version: 3.0.3
+      chalk:
+        specifier: ^5.6.2
+        version: 5.6.2
+      confbox:
+        specifier: ^0.2.4
+        version: 0.2.4
+      express:
+        specifier: ^5.2.1
+        version: 5.2.1
+      http-proxy-middleware:
+        specifier: ^3.0.5
+        version: 3.0.7
+      jose:
+        specifier: ^6.1.3
+        version: 6.2.10
+      marked:
+        specifier: ^18.0.1
+        version: 18.0.11
+      material-symbols:
+        specifier: ^0.44.6
+        version: 0.44.12
+      monaco-editor:
+        specifier: ^0.55.1
+        version: 0.55.1
+      next:
+        specifier: ^16.1.6
+        version: 16.3.4(@babel/core@7.29.7)(react-dom@19.2.4(react@19.2.4))(react@19.2.4)
+      node-forge:
+        specifier: ^1.3.3
+        version: 1.4.0
+      node-machine-id:
+        specifier: ^1.1.12
+        version: 1.1.12
+      open:
+        specifier: ^11.0.0
+        version: 11.0.2
+      ora:
+        specifier: ^9.1.0
+        version: 9.4.1
+      prop-types:
+        specifier: ^15.8.1
+        version: 15.8.1
+      react:
+        specifier: 19.2.4
+        version: 19.2.4
+      react-dom:
+        specifier: 19.2.4
+        version: 19.2.4(react@19.2.4)
+      react-is:
+        specifier: ^16.13.1
+        version: 16.13.1
+      recharts:
+        specifier: ^3.7.0
+        version: 3.10.1(react-dom@19.2.4(react@19.2.4))(react-is@16.13.1)(react@19.2.4)(redux@5.0.1)
+      selfsigned:
+        specifier: ^5.5.0
+        version: 5.5.0
+      socks-proxy-agent:
+        specifier: ^8.0.5
+        version: 8.0.5
+      sql.js:
+        specifier: ^1.14.1
+        version: 1.14.2
+      undici:
+        specifier: ^7.19.2
+        version: 7.29.0
+      uuid:
+        specifier: ^13.0.0
+        version: 13.0.2
+      zustand:
+        specifier: ^5.0.10
+        version: 5.0.15(immer@11.1.18)(react@19.2.4)(use-sync-external-store@1.6.0(react@19.2.4))
+    devDependencies:
+      '@tailwindcss/postcss':
+        specifier: ^4.1.18
+        version: 4.3.3
+      eslint:
+        specifier: ^9
+        version: 9.39.5(jiti@2.7.0)
+      eslint-config-next:
+        specifier: 16.1.6
+        version: 16.1.6(@typescript-eslint/parser@8.69.0(eslint@9.39.5(jiti@2.7.0))(typescript@6.0.3))(eslint@9.39.5(jiti@2.7.0))(typescript@6.0.3)
+      postcss:
+        specifier: ^8.5.6
+        version: 8.5.26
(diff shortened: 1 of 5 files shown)


louis-cai:
Thanks @bloodf for the thorough audit in bloodf/durindoor#688 — pushed 46a09fb addressing all four findings:

1. **Cross-request block (your #1, confirmed real):** on trip, the breaker now synthesizes `{remainingPercentage: 0, resetAt: blockedUntil}` into the shared RAM quota cache, so `getProviderCredentials`' pre-filter skips the pair on subsequent requests through the same channel as the exhausted-0% path. Blocks are re-asserted over fresh snapshots in `_doRefresh`, so an optimistic upstream reading cannot resurrect a blocked pair mid-block.

2. **Success reset (your #2, confirmed real):** new exported `clearAntigravityStrikes(connectionId, model)`, called from `onRequestSuccess` next to `clearAccountError` in `src/sse/handlers/chat.js`. It clears the strike counter and removes the synthesized cache entry (identified by `resetAt === blockedUntil`; real upstream 0% readings are never touched).

3. **Window semantics (your #3, confirmed real):** `windowStart` is now anchored at the first qualifying strike — three 429s must land within 60s of that first strike. Covered by the new "anchors the window" test (strikes at t=0/45s/90s do not trip).

4. **409 counting (your #4, partially declined, documented):** I kept 409 in scope intentionally — Antigravity signals pool exhaustion with 409 as well, and #3561's own motivation was "409/429 before ordinary fallback". Poisoning would require 3 transient 409s on the same pair inside one 60s window. Happy to gate on `status === 429` if you feel strongly; it's a one-line change.

On DurinDoor's port plan: note the divergence — this implementation rides the existing optimistic RAM quota cache rather than `markAccountUnavailable`, precisely because the chat handler deliberately does not persist `modelLock_*` for this path (comment in chat.js). Your fork's plan to persist via the existing lock flow is the cleaner shape for an architecture that already has it; the synthesized-cache-entry trick is what makes it work without adding one here.

Tests now 15/15 (`npx vitest run unit/antigravity-quota-routing.test.js` from `tests/`), including persistence-across-refresh, success-reset, and window-anchoring cases. Live-validated in a production v0.5.59 container with an equivalent bundle patch — blocks hold across requests and re-assert during the window.

decolua:
Thanks @louis-cai for the contribution! Reviewed and merged into master. 🙏

PR #3686, opened 2026-09-01
fix(claude): drop server_tool_use blocks carrying a foreign id

### LINKED ISSUE #3685: Foreign server_tool_use id from a built-in-tool provider poisons the Claude history (400 srvtoolu_)
## Summary

Anthropic validates `server_tool_use.id` against `^srvtoolu_[a-zA-Z0-9_]+$` and rejects the entire request with a 400 when one does not match.

A provider that runs its own built-in tools emits its native id instead of an Anthropic one. z.ai/glm does this for `analyze_image`, producing OpenAI-style `call_` ids:

```json
{"type": "server_tool_use", "id": "call_50b82aba1b754d82a4408a53", "name": "analyze_image", "input": {}}
```

In a mixed combo those blocks stay in the conversation history. As soon as the combo routes a later turn to a Claude provider, the passthrough body carries the foreign id and the API rejects it:

```
API Error: 400 messages.1.content.1.server_tool_use.id:
String should match pattern '^srvtoolu_[a-zA-Z0-9_]+$'
```

The session cannot recover on its own: the bad block is now part of the history, so every subsequent turn fails the same way.

## Reproduction

1. Set up a combo mixing a Claude provider and `glm/glm-5.3`.
2. Let a turn land on GLM and trigger its built-in `analyze_image` tool, so a `server_tool_use` block with a `call_` id enters the history.
3. Let a later turn route to the Claude provider → 400 on every attempt from then on.

## Expected

The same treatment `normalizeClaudePassthrough` already applies to foreign thinking signatures — [`claude.js#L164-L166`](https://github.com/decolua/9router/blob/master/open-sse/translator/formats/claude.js#L164-L166) notes that a combo mixing models leaks foreign signatures into history and drops them. A foreign `server_tool_use` id is the same class of leak and should be dropped before the request reaches Anthropic, along with the `tool_result` that references it (an orphan result is rejected in turn).

PR follows.

f-liva:
Fixes #3685.

## Problem

Anthropic validates `server_tool_use.id` against `^srvtoolu_[a-zA-Z0-9_]+$` and 400s the whole request when one does not match. A provider running its own built-in tools emits its native id — z.ai/glm uses OpenAI-style `call_` ids for `analyze_image` — and in a mixed combo those blocks remain in the history, so every later Claude turn fails and the session cannot recover.

## Change

`normalizeClaudePassthrough` already drops thinking blocks whose signature is not Claude's, for exactly this reason (a combo mixes models, foreign shapes leak into history). This extends that pass:

- drops `server_tool_use` blocks whose id does not match the Anthropic pattern, reusing the existing loop so there is no second walk over `messages`;
- drops the paired `tool_result` / `web_search_tool_result`, since a result referencing an id no block declares is rejected in turn;
- substitutes an empty text block when a message is left with no content, as Anthropic rejects an empty `content` array.

`SERVER_TOOL_USE` and `WEB_SEARCH_TOOL_RESULT` are added to `CLAUDE_BLOCK` rather than written as string literals, matching how the rest of the file refers to block types.

Well-formed `srvtoolu_` blocks and ordinary `tool_use` blocks are untouched — `tool_use` ids are not subject to this pattern, so a `call_`-style id there stays as it is.

## Tests

`tests/unit/claude-foreign-server-tool-use.test.js`, alongside the existing `translator-helpers-edge.test.js` coverage of this function:

- drops a `server_tool_use` with a `call_` id
- drops the orphaned `tool_result` and keeps the rest of the message
- keeps a well-formed `srvtoolu_` block
- keeps a regular `tool_use`, whatever its id looks like
- leaves an empty text block, not an empty array

Verified red before the fix (3 fail, the 2 "keep" cases already green) and green after.

## Regression check

Full suite before and after on the same checkout, compared by test name:

- fails before: 94
- fails after: 89
- **regressions: 0**

The 5 that flip to green are the 3 new drop cases plus `force-stream-config` and `usage-dispatch`, which are order-dependent flakes unrelated to this change. `npx eslint` is clean on the touched files.

One note for reviewers: `tests/unit/translator-helpers-edge.test.js` has a failing case (`hoists mid-conversation system messages into top-level system`) that expects hoisting into `body.system`, while the implementation folds those messages into the neighbouring turn to protect the prefix cache. It fails identically before and after this change — stale test, not a regression here, and left untouched.

### DIFF
--- a/open-sse/translator/formats/claude.js
+++ b/open-sse/translator/formats/claude.js
@@ -199,7 +199,7 @@ export function normalizeClaudePassthrough(body, model = "") {
         if (block.type === CLAUDE_BLOCK.TOOL_USE) hasToolUse = true;
         kept.push(block);
       }
-      msg.content = kept.length > 0 ? kept : [{ type: CLAUDE_BLOCK.TEXT, text: "" }];
+      msg.content = kept;
       if (thinkingEnabled && !hasKeptThinking && hasToolUse) {
         msg.content.unshift(buildThinkingPlaceholder("claude"));
       }
@@ -216,9 +216,23 @@ export function normalizeClaudePassthrough(body, model = "") {
         && droppedServerToolUseIds.has(String(block.tool_use_id ?? ""))
       ));
       if (kept.length !== msg.content.length) {
-        msg.content = kept.length > 0 ? kept : [{ type: CLAUDE_BLOCK.TEXT, text: "" }];
+        msg.content = kept;
       }
     }
+  }
+
+  // 5. Drop empty text blocks and any message left with no content at all.
+  // Anthropic rejects `messages.N.content` blocks with empty text (400
+  // "text content blocks must be non-empty"); a message whose blocks were all
+  // stripped above must be dropped, not padded with an empty placeholder.
+  if (Array.isArray(body.messages)) {
+    body.messages = body.messages.filter(msg => {
+      if (typeof msg.content === "string") return msg.content.trim().length > 0;
+      if (!Array.isArray(msg.content)) return true;
+      msg.content = msg.content.filter(block =>
+        !(block?.type === CLAUDE_BLOCK.TEXT && !String(block.text ?? "").trim()));
+      return msg.content.length > 0;
+    });
   }
 
   return body;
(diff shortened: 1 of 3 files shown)


f-liva:
Amended: instead of padding messages left with no blocks with `{type:"text",text:""}` (which Anthropic rejects with 400 "text content blocks must be non-empty" when the padded request is sent), the passthrough now strips empty text blocks from the history and drops any message that ends up with no content at all. Related: #3692.

decolua:

PR #3691, opened 2026-09-01
fix: strip the [1m] context marker Claude Code appends to the model name (#3690)

### LINKED ISSUE #3690: Claude Code with the 1M-context beta sends `model: "<name>[1m]"` and every request is rejected
**Version:** 0.5.59

**Client:** Claude Code, 1M-context beta enabled

## What happens

With the 1M-context beta toggled on, Claude Code appends a bracketed marker to the model name and posts to `/v1/messages` with:

```json
{ "model": "claude-opus-5[1m]", ... }
```

9router has no notion of that marker. `getComboModels("claude-opus-5[1m]")` misses, `parseModel` cannot split it into `provider/model`, and the request ends at `Invalid model format`. The client renders:

```
There's an issue with the selected model (claude-opus-5[1m]).
It may not exist or you may not have access to it.
```

which points at the wrong thing: the combo `claude-opus-5` exists and works. Every single request from that session fails until the beta is switched off, and nothing in the message says the marker is the cause.

## Repro

```bash
curl -sS -X POST http://127.0.0.1:20128/v1/messages \
  -H "content-type: application/json" -H "anthropic-version: 2023-06-01" \
  -H "anthropic-beta: context-1m-2025-08-07" \
  -d '{"model":"claude-opus-5[1m]","max_tokens":8,"messages":[{"role":"user","content":"hi"}]}'
```

400 `Invalid model format`, while the same body with `"model":"claude-opus-5"` routes fine.

## Root cause

`src/sse/handlers/chat.js` takes `body.model` verbatim into combo lookup and `getModelInfo()`. Nothing anywhere in the repo mentions the marker — `grep -rn "\[1m\]" .` comes back empty.

## Suggested fix

Strip a trailing `[1m]` before resolution. The capability itself rides in `anthropic-beta: context-1m-2025-08-07`, which `open-sse/executors/default.js` already forwards untouched, so the request still reaches the upstream as a 1M request — only the routing key needs cleaning.

PR follows.

f-liva:
Fixes #3690.

## Problem

With the 1M-context beta enabled, Claude Code sends `model: "claude-opus-5[1m]"`. The marker is a client-side annotation — it matches no combo name, no alias and no `provider/model` pair — so `src/sse/handlers/chat.js` passes it straight into combo lookup and `getModelInfo()`, both miss, and the request ends at `Invalid model format`. The client then reports "There's an issue with the selected model (claude-opus-5[1m])", which points at the wrong thing: the combo exists and works. Every request from that session fails until the beta is switched off.

## Fix

New `open-sse/utils/modelMarkers.js` exporting `stripModelContextMarker(modelStr)` → `{ model, contextMarker }`. `handleChat` strips the marker before resolution and normalizes `body.model` so downstream logging and translation see the real name.

The capability itself travels in `anthropic-beta: context-1m-2025-08-07`, which `open-sse/executors/default.js` already forwards untouched, so the request still reaches the upstream as a 1M request — only the routing key needed cleaning. That is why the fix does not try to synthesize the beta header itself.

Only a *trailing* marker is stripped, so a model whose name genuinely contains brackets is left alone.

## Tests

`tests/unit/model-context-marker.test.js`, 6 cases: bare model, provider-prefixed model, case insensitivity, plain model untouched, a bracket inside the name left alone, non-string input.

## Suite

Full `npx vitest run` on `master` (v0.5.59): 138 failing, unchanged with this branch, and the set-difference of failing test names shows **0 new failures**. The suite is not all-green on a plain checkout, hence the name diff rather than the raw count.

`npx eslint` clean on all three touched files.

### DIFF
--- a/open-sse/utils/modelMarkers.js
+++ b/open-sse/utils/modelMarkers.js
@@ -1,0 +1,20 @@
+// Claude Code appends a bracketed context marker to the model name when the
+// 1M-context beta is toggled on: `claude-opus-5` becomes `claude-opus-5[1m]`.
+// The marker is a client-side annotation, not part of any model id: it never
+// matches a combo name, an alias or a `provider/model` pair, so a request that
+// carries it dies at model resolution with "Invalid model format".
+//
+// The capability itself travels in the `anthropic-beta: context-1m-2025-08-07`
+// header, which is forwarded untouched — stripping the marker is enough to let
+// the request route normally and still reach the upstream as a 1M request.
+
+const CONTEXT_MARKER = /\[1m\]$/i;
+
+// Returns { model, contextMarker } — contextMarker is null when there is none.
+export function stripModelContextMarker(modelStr) {
+  if (typeof modelStr !== "string") return { model: modelStr, contextMarker: null };
+  const trimmed = modelStr.trim();
+  const match = trimmed.match(CONTEXT_MARKER);
+  if (!match) return { model: modelStr, contextMarker: null };
+  return { model: trimmed.slice(0, -match[0].length), contextMarker: match[0].slice(1, -1).toLowerCase() };
+}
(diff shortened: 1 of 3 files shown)


decolua:

PR #3695, opened 2026-09-01
fix: preserve upstream diagnostics and Responses telemetry

longdang193:
## Summary
- preserve request correlation IDs while redacting sensitive request data
- preserve translated Responses output and cache telemetry across streaming and non-streaming paths
- expose scalar upstream error diagnostics instead of returning raw parser JSON as the message

## Validation
- `npx vitest run tests/unit/fitcv-diagnostics.test.js --testNamePattern="keeps only scalar upstream diagnostics" --reporter=verbose`
- `git diff --check upstream/master...HEAD`

## Notes
- Provider-side Responses structured-output compatibility remains separate from gateway error projection.


longdang193:
Closing: correct target is fork master, not upstream master.

Issue #3701, opened 2026-09-01
[Feature] Quota tracking gaps: 12 connected providers report no usage/quota data


nitsuah:
## Summary

The Quota Tracker (`src/app/(dashboard)/dashboard/quota/`) shows no usage/quota data at all for a number of providers that are otherwise connected and working for inference. Consolidating what's currently missing into one tracking issue rather than filing 12 near-identical reports.

## Observed Gaps

| Provider | Symptom |
|---|---|
| Gemini CLI | Connected, but quota detection returns 403 (previously showed models & allocation; likely an endpoint/scope change) |
| Qoder | Connected, but usage/quota fetch returns 401 |
| Fireworks AI | Connected, not tracked |
| Cohere | Connected, not tracked |
| Groq | Connected, not tracked |
| Mistral | Connected, not tracked |
| OpenAI | Connected, not tracked |
| SiliconFlow | Connected, not tracked |
| Cerebras | Connected, not tracked |
| NVIDIA NIM | Connected, not tracked |
| Cloudflare | Connected, not tracked |
| OpenRouter | Connected, not tracked |
| Vercel AI Gateway | Connected, but no credit allocation detected/unfunded — should say so explicitly rather than surfacing as a generic 403/error |

## Why It Matters

9router's core pitch is routing across many providers to avoid hitting limits; if the Quota Tracker can't see a provider's remaining quota, users can't make an informed choice about which account/combo to lean on before it 429s. This is especially relevant now that the default quota-poll interval is being relaxed to 5 minutes (see companion Bugs issue) specifically to make room for tracking *more* providers without increasing backend load per-poll.

## Related

- #2126 is the long-running community "feature tracker" for provider quota support (June 2026, lists Antigravity/Anthropic/OpenAI/OpenCode/Ollama/xAI) — this issue is the current, narrower snapshot of what's connected-but-blind today; worth cross-linking rather than duplicating.
- Several of these (OpenAI, Groq, Mistral, Cerebras, Cohere, NVIDIA, SiliconFlow) already have a `validateUrl`-based connectivity check (see the Venice fix in the companion Bugs issue) — usage/quota is a separate concern from "is the key valid," but the same registry-driven pattern (declare the usage endpoint shape once in `open-sse/providers/registry/<id>.js`, consume it generically) is the right target architecture here too, rather than one bespoke usage parser per provider.

## Proposed Scope

[...]

PR #3703, opened 2026-09-01
fix(providers): config-driven test fallback for Venice AI, 5m quota-poll cadence

### LINKED ISSUE #3700: [Bug] Provider connectivity/quota cleanup: Venice test failure, deprecated Gemini CLI, stale-model demotion, quota-poll cadence
## Summary

Four related, but independently-fixable, provider-health bugs bundled into one tracking issue:

### 1. Venice AI — "Provider test not supported"
Re-testing an already-saved Venice AI connection (`src/app/api/providers/[id]/test/testUtils.js`) fails with `Provider test not supported`, even though Venice validates fine when *first added* (`/api/providers/validate` already has a config-driven fallback keyed off `PROVIDERS[id].validateUrl`, which Venice's registry entry declares). The re-test path (`testApiKeyConnection` in `testUtils.js`) lacked the equivalent generic fallback and fell through to a hardcoded `default: "Provider test not supported"`.

**Fix included below** (see PR link) — adds the same config-driven fallback (`validateUrl` → generic `/models` probe → chat probe, mirroring `validate/route.js`'s existing `default` branch) to `testApiKeyConnection`'s default case. This is provider-agnostic, so it also covers any *future* provider that only declares `transport.validateUrl`/`baseUrl` in the registry without a hand-written case in this file.

### 2. Gemini CLI — not functional
`open-sse/providers/registry/gemini-cli.js` already sets `display.deprecated: true` with a `RISK_NOTICE`, consistent with #1866 (sunset scheduled). Given that, "fix it" isn't the right framing — the open question for the maintainer is whether to (a) leave it visibly deprecated as-is, (b) actively surface the 403s users are hitting (e.g. #1852, #2742) as "deprecated, expect failures" in the UI rather than a raw 403, or (c) proceed with removal per #1866. This issue exists so the decision has one place to land instead of scattering across the ~59 open Gemini-CLI-adjacent issues.

### 3. Stale-model demotion
When a model fails after its retry budget (`retried 5 times... Last error: 502 [gemini-cli/gemini-2.5-pro] [403]`), it should be tracked as "stale" rather than silently retried at the same priority on every future combo/fallback attempt. Proposed: a lightweight stale-model registry (persisted, keyed by provider+model) populated when the retry budget is exhausted, surfaced as a dashboard panel with an action to de-rank or remove the model from active combos. This needs a small design pass (data model + where in the fallback loop to record failures) before implementation — first-slice PR to follow.

### 4. Quota poll cadence
The Providers/Quota panel (`src/app/(dashboard)/dashboard/usage/components/ProviderLimits/utils.js`) polled every provider's quota endpoint every 60s by default. With more providers gaining quota tracking (see the companion Quotas issue), that's needless backend load. **Fix included below** — default bumped to 5 minutes (300000ms); Claude's already-throttled interval is tick-ratio-derived from this constant so it stays at its existing 10-minute cadence unchanged. Also fixed a latent countdown-display bug this surfaced: the UI countdown timer was hardcoded to reset at `60` (seconds), which would have silently desynced from the real refresh cadence the moment the interval changed — now derived from the same constant.

## PR

Items 1 and 4 are fixed together in a small, self-contained PR (see linked PR). Items 2 and 3 remain open pending maintainer input (2) and a short design pass (3).

## Acceptance

- [x] Venice AI re-test uses the generic config-driven fallback and passes for any provider declaring `validateUrl`.
- [x] Quota panel default poll interval is 5 minutes; countdown display derives from the same constant instead of a hardcoded value.
- [ ] Maintainer decision recorded on Gemini CLI's UI treatment (surfaced-deprecated vs. removal per #1866).
- [ ] Stale-model demotion design doc + first-slice implementation.

nitsuah:
## Summary

Two small, unrelated-but-adjacent provider-health fixes from #3700, bundled because they're both one-line-of-intent changes touching quota/connection testing:

1. **Venice AI: "Provider test not supported" on re-test.** `testApiKeyConnection()` in `src/app/api/providers/[id]/test/testUtils.js` re-tests an already-saved connection through a hardcoded `switch`, falling through to `default: "Provider test not supported"` for any provider without an explicit case — Venice included. `/api/providers/validate` (used when a connection is first *added*) already has a generic, config-driven fallback for exactly this situation: it walks `PROVIDERS[id].validateUrl` (declared in `open-sse/providers/registry/venice.js` as `transport.validateUrl`) and probes it directly. This PR ports that same fallback logic into `testApiKeyConnection`'s `default` case, so re-testing a saved connection and validating a new one behave consistently — for Venice, and for any future provider that only declares `validateUrl`/`baseUrl` without a hand-written case in this file.

2. **Quota-poll cadence.** `REFRESH_INTERVAL_MS` in `ProviderLimits/utils.js` defaulted to 60s, auto-polling every connected provider's quota endpoint every minute. Bumped to 5 minutes to reduce backend load, especially as more providers gain quota tracking (see #3701). Claude's poll interval is computed as a tick-ratio of this constant (`claudeEvery = CLAUDE_REFRESH_INTERVAL_MS / REFRESH_INTERVAL_MS`), so its effective 10-minute cadence is unchanged by this bump.

   While making that change I found the countdown-timer display (`ProviderLimits/index.js`) was hardcoded to reset at `60` (seconds), independent of `REFRESH_INTERVAL_MS`. Left as-is, the visible countdown would have silently desynced from the real refresh cadence the moment the interval changed. Derived it from the same constant (`REFRESH_INTERVAL_S = REFRESH_INTERVAL_MS / 1000`) instead, at all 4 call sites.

## Why

Filed and explained in more detail in #3700.

## Validation

- `node --check` passes on all three changed files (Dockerized `node:22-alpine`, no host Node install used).
- Reviewed all other `REFRESH_INTERVAL_MS` usages (`ProviderTopology.js`'s `FE_ACTIVE_TICK_MS`, etc.) — no other hardcoded assumptions about the old 60s value found.
- Did not touch any of the ~40 already-cased providers in `testApiKeyConnection`'s switch — the new fallback only activates in the pre-existing `default` branch, so behavior for every provider with an explicit case is unchanged.

## Test plan

- [ ] Re-test a saved Venice AI connection from the Providers page and confirm it now validates instead of returning "Provider test not supported".
- [ ] Confirm the Provider Limits panel auto-refreshes every 5 minutes and the visible countdown counts down from 300s (not 60s) before looping.
- [ ] `cd tests && npx vitest run` — no new failures introduced (this PR doesn't touch anything in the existing 26-entry `known-fails.txt`).

Fixes #3700 (partial — Gemini CLI UI treatment and stale-model demotion remain open there).

### DIFF
--- a/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js
+++ b/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/index.js
@@ -43,6 +43,10 @@ import { ConfirmModal, EditConnectionModal } from "@/shared/components";
 import { USAGE_SUPPORTED_PROVIDERS } from "@/shared/constants/providers";
 import { useCopyToClipboard } from "@/shared/hooks/useCopyToClipboard";
 
+// Countdown display ticks in seconds; derive from REFRESH_INTERVAL_MS so it can't
+// drift out of sync with the actual auto-refresh cadence (was hardcoded to 60).
+const REFRESH_INTERVAL_S = REFRESH_INTERVAL_MS / 1000;
+
 // Maps the stored providerSpecificData.authMethod to a human label for Kiro.
 // Values come from the Kiro connect flows: builder-id/idc (device code),
 // google/github (social), imported (refresh-token paste), api_key (headless).
@@ -135,7 +139,7 @@ export default function ProviderLimits() {
   const [lastUpdated, setLastUpdated] = useState(null);
   const [hasHydratedAutoRefresh, setHasHydratedAutoRefresh] = useState(false);
   const [refreshingAll, setRefreshingAll] = useState(false);
-  const [countdown, setCountdown] = useState(60);
+  const [countdown, setCountdown] = useState(REFRESH_INTERVAL_S);
   const [connectionsLoading, setConnectionsLoading] = useState(true);
   const [deletingId, setDeletingId] = useState(null);
   const [togglingId, setTogglingId] = useState(null);
@@ -467,7 +471,7 @@ export default function ProviderLimits() {
     if (refreshingAll) return;
 
     setRefreshingAll(true);
-    setCountdown(60);
+    setCountdown(REFRESH_INTERVAL_S);
 
     // Throttle Claude: poll its quota every Nth auto-tick (manual force bypasses)
     const tick = (tickCountRef.current += 1);
@@ -646,7 +650,7 @@ export default function ProviderLimits() {
     // Countdown interval
     countdownRef.current = setInterval(() => {
       setCountdown((prev) => {
-        if (prev <= 1) return 60;
+        if (prev <= 1) return REFRESH_INTERVAL_S;
         return prev - 1;
       });
     }, 1000);
@@ -673,7 +677,7 @@ export default function ProviderLimits() {
         // Resume auto-refresh when tab becomes visible
         intervalRef.current = setInterval(() => refreshAll(), REFRESH_INTERVAL_MS);
         countdownRef.current = setInterval(() => {
-          setCountdown((prev) => (prev <= 1 ? 60 : prev - 1));
+          setCountdown((prev) => (prev <= 1 ? REFRESH_INTERVAL_S : prev - 1));
         }, 1000);
       }
     };
--- a/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/utils.js
+++ b/src/app/(dashboard)/dashboard/usage/components/ProviderLimits/utils.js
@@ -2,7 +2,7 @@ import { getModelsByProviderId } from "open-sse/config/providerModels.js";
 
 // ─── Constants ───────────────────────────────────────────────────────────────
 export const QUOTA_CACHE_KEY = "quotaCacheData";
-export const REFRESH_INTERVAL_MS = 60000;
+export const REFRESH_INTERVAL_MS = 300000;
 // Claude usage/quota endpoint rate-limits; poll it less often than other providers
 export const CLAUDE_REFRESH_INTERVAL_MS = 600000;
 export const DEPLETED_QUOTA_THRESHOLD = 5;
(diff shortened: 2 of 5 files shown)


nitsuah:
Pushed a follow-up commit closing a gap in the original PR: it only had `node --check` syntax validation, no new tests. Added:

- `tests/unit/provider-test-fallback.test.js` — exercises the actual fallback logic this PR adds: Venice AI against the real registry `validateUrl` (valid key → `valid:true`, 401 → reported as "Invalid API key" instead of "Provider test not supported"), an unrecognized provider still correctly reports "not supported", and the `noAuth` / openai-format `/models`-then-chat-probe / `x-api-key`-auth-header branches via a mocked `PROVIDERS` registry (so the tests don't depend on the real registry's current shape beyond the one real Venice assertion).
- `tests/unit/provider-limits-refresh-interval.test.js` — locks in `REFRESH_INTERVAL_MS = 300000` and the tick-ratio math that keeps Claude's effective poll cadence at 10 minutes unchanged.

9/9 passing (Docker, `node:22-alpine`).

nitsuah:
@decolua - friendly bump here for next version to increase default on refresh in prep for scaling more provider tracking.

> Quota-poll cadence - REFRESH_INTERVAL_MS in ProviderLimits/utils.js defaulted to 60s, auto-polling every connected provider's quota endpoint every minute. Bumped to 5 minutes to reduce backend load, especially as more providers gain quota tracking (see https://github.com/decolua/9router/issues/3701).

nitsuah:
Brought up to date with `master` (`39e36d3d`, v0.5.86; was 151 commits behind) via a merge commit. It merged cleanly, and the Venice test fallback and 5-minute quota-poll cadence are still not on master.

Verified in Docker (`node:22`, `CI=true`): `provider-limits-refresh-interval.test.js` and `provider-test-fallback.test.js` pass 9/9. Full suite: no new failures attributable to this branch.

Issue #3699, opened 2026-09-01
[Feature] Providers dashboard: status filter + local-network scan for custom OpenAI/Anthropic-compatible endpoints


nitsuah:
## Summary

Two small, independent dashboard UX gaps on the Providers page, bundled here since both are "nice to have" quality-of-life items rather than bug fixes:

1. **Status filter.** With 40+ registered providers, the Providers page has no way to filter by connection status (All / Active / Inactive / No connection). Everything renders in one long list regardless of whether it's configured.
2. **Local-network scan for custom providers.** When adding a Custom Provider (OpenAI/Anthropic-compatible), the user has to know and type the exact host:port. A "scan" action that probes common localhost/127.0.0.1/local-network addresses and ports (e.g. Ollama's `11434`, common Docker-exposed LLM-server ports) and lists reachable OpenAI/Anthropic-compatible endpoints would remove that friction.

## Item 2 status: fix for its blocker is up

**Update 2026-09-02:** [#3714](https://github.com/decolua/9router/issues/3714) documented that the SSRF guard (`src/shared/utils/ssrfGuard.js`) had confirmed bypasses, which mattered here because the local-network scan proposed below is itself an outbound-fetch-to-local-addresses feature that would have sat directly on that guard. [#3723](https://github.com/decolua/9router/pull/3723) fixes it (IPv6 representation handling, trailing-dot hostnames, DNS-resolving hostnames, and redirect-target re-validation). **Once #3723 merges, item 2 is unblocked** — still worth implementing it against the hardened guard (`fetchPublic`/`assertPublicUrlResolved`) rather than a fresh ad hoc check, for consistency.

Item 1 (status filter) had no such dependency and already shipped separately in PR #3704.

### 1. Status filter — done, see PR #3704
- Client-side filter (All / Active / Inactive / No connection) over the already-fetched provider+connection list — no new API.

### 2. Custom provider scan — unblocked once #3723 merges
- A "Scan for local servers" action in the Custom Provider add flow.
- Probe a short, explicit list of common local ports (documented, not exhaustive) against `127.0.0.1` and `localhost` — e.g. Ollama (`11434`), LM Studio (`1234`), common self-hosted OpenAI-compat gateways.
- For each reachable address, hit `/v1/models` (or the Anthropic-compat equivalent) with a short timeout and list what responds as a suggested endpoint, letting the user pick one to prefill the Custom Provider form.
[...]

PR #3704, opened 2026-09-01
feat(providers): add status filter to providers dashboard

nitsuah:
## What

Adds a client-side status filter (**All / Active / Inactive / No connection**) to the Providers dashboard page, applied over the already-fetched provider + connection list. No new API endpoint.

- `src/app/(dashboard)/dashboard/providers/utils.js` (new): `STATUS_FILTER_OPTIONS` constant plus two pure helpers, `getConnectionStatus(stats, isNoAuth)` and `matchesStatusFilter(statusFilter, stats, isNoAuth)`, mirroring the `ACCOUNT_FILTER_OPTIONS` pattern already used in `.../usage/components/ProviderLimits/utils.js`.
- `src/app/(dashboard)/dashboard/providers/page.js`: adds a `statusFilter` state + a compact `<select>` control (styled to match the existing filter `<select>` in `ProviderLimits/index.js`), and applies the filter across every existing provider-list filter chain (OAuth, Free, Free Tier, API Key, and both Custom/Compatible sections).

Status is derived from data the page already has per provider (`getProviderStats(...)`, which already returns `total` and `allDisabled`):
- **Active** — at least one connection exists and it isn't fully disabled.
- **Inactive** — connections exist but every one has been turned off (`allDisabled`).
- **No connection** — the provider is registered but has zero connections. `noAuth` providers (always-ready free proxies with no connection record) are treated as Active rather than "No connection".

## Why

Part of #3699 — local-network scan is tracked separately and intentionally out of scope here, to keep this PR small and reviewable.

## Test plan

- [x] `npx vitest run unit/providers-status-filter.test.js` — new focused unit test covering the pure classification/filter helpers (7 tests, all passing).
- [x] Manual read-through confirming the filter is applied consistently across all provider sections (OAuth, Free, Free Tier, API Key, Custom Compatible) and that the existing search filter still composes correctly with it.
- [ ] Manual UI smoke test in a running dashboard (not run in this environment — no live provider connections to exercise all three states).

### DIFF
--- a/src/app/(dashboard)/dashboard/providers/page.js
+++ b/src/app/(dashboard)/dashboard/providers/page.js
@@ -25,6 +25,7 @@ import { useNotificationStore } from "@/store/notificationStore";
 import { useHeaderSearchStore } from "@/store/headerSearchStore";
 import ModelAvailabilityBadge from "./components/ModelAvailabilityBadge";
 import AddCompatibleModal from "./components/AddCompatibleModal";
+import { STATUS_FILTER_OPTIONS, matchesStatusFilter } from "./utils";
 
 function getStatusDisplay(connected, error, errorCode) {
   const parts = [];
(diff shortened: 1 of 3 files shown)


decolua:

Issue #3702, opened 2026-09-01
[Testing] Expand core test coverage (routing, translation, combo fallback) beyond the CI/regression-gate fix


nitsuah:
## Summary

#2263 (open, with PR #2262) already tracks getting the existing no-regression gate green and running in CI. This issue is deliberately scoped *differently*: even once that gate is green, the underlying coverage is still thin relative to the surface area — `docs/ARCHITECTURE.md`'s own request-lifecycle description (parse → combo expansion → account-selection → translate → dispatch → retry/refresh → stream) has large stretches with no direct unit/integration tests, which is how PRs keep causing "constant churn" (per user reports) even when the known-fails baseline is respected.

## Why a Separate Issue From #2263 / #3100

- #2263 = make the *existing* 938-pass/64-fail suite reproducible + wire it into CI + fix the 31 accumulated pass→fail regressions. That's necessary but doesn't add new coverage.
- #3100 = a vague, unscoped "add more tests everywhere" wishlist with no concrete target.
- This issue = a concrete, prioritized list of *specific undercovered areas* to add tests for, once (or in parallel with) #2263 landing, so effort isn't duplicated.

## Proposed Priority Order (highest risk / lowest current coverage first)

1. **Combo/account-selection fallback loop** (`src/sse/handlers/chat.js` account-selection loop, `open-sse/handlers/chatCore.js` retry/refresh) — this is the code path most bug reports trace back to (e.g. #3619 "Combo routing fails with 'Sorry, no response was returned'"), and per `CLAUDE.md` it's genuinely central ("the thing to understand first").
2. **Translator direct-route pairs** (`open-sse/translator/`) — the fragile cases the project's own conventions call out (thinking blocks, tool ids, non-base64 images, `is_error`) are exactly where translation bugs keep surfacing (#3646 thoughtSignature dropped, #3622 tool name truncation collisions).
3. **Provider registry / alias regression baselines** (`tests/__baseline__/verify-providers.mjs`, `verify-alias.mjs`) — per `CLAUDE.md` these are already stale on `master` (qwen OAuth URL change, blackbox `/v1`, added kimchi/venice per #2263's findings); regenerating and keeping these current is cheap, high-leverage coverage.

## Proposed Scope (first slice)

Add a focused test suite for **one** of the above — recommend starting with the combo/account-selection fallback loop, given it's the most bug-report-heavy path — rather than attempting all three areas in one PR.

## Non-goals

[...]

Issue #3619, opened 2026-08-29
Combo routing fails with "Sorry, no response was returned" when using VS Code + GitHub Copilot with mixed VXP/OpenRouter/AG models


LingDev03:
### Bug Description
When using 9Router as a proxy for **GitHub Copilot in VS Code**, invoking a custom model combo containing mixed provider types (`vxp/*`, `openrouter/*`, `ag/*`) causes the chat request to fail completely, returning **"Sorry, no response was returned"** or an infinite loop/spam log.

However, calling each of these models **individually** works fine without any errors.

### Affected Combo Configuration
- **Combo Name:** `coding`
- **Models included in combo:**
  1. `vxp/qwen/qwen3-next-80b-a3b-thinking-maas`
  2. `vxp/deepseek-ai/deepseek-v3.2-maas`
  3. `vxp/qwen/qwen3-next-80b-a3b-instruct-maas`
  4. `openrouter/cohere/north-mini-code:free`
  5. `ag/gpt-oss-120b-medium`

### Environment
- **Client:** VS Code (GitHub Copilot Extension)
- **9Router Version:** v0.5.55
- **Endpoint:** `http://localhost:20128/v1`

### Steps to Reproduce
1. Configure 9Router with the model combo listed above.
2. Set VS Code / GitHub Copilot custom endpoint to point to 9Router (`http://localhost:20128/v1`).
3. Select the combo model in Copilot and send a prompt (e.g., coding request).
4. Observe the error: **"Sorry, no response was returned"** / streaming hangs or loops.
### Screenshots & Logs

<img width="733" height="663" alt="Image" src="https://github.com/user-attachments/assets/b3a0af93-17a5-4755-b8a5-4639d58b09cc" />

<img width="748" height="450" alt="Image" src="https://github.com/user-attachments/assets/3d789434-196e-4ef8-a133-56f51cb7edff" />

### Expected Behavior
Combo routing should successfully fall back or route through the mixed-provider combo without crashing or dropping responses

PR #3705, opened 2026-09-01
test(combo): add coverage for account-selection fallback loop

nitsuah:
## Summary

Part of #3702 — this is the first slice: real, focused test coverage for the **combo/account-selection fallback loop** (priority #1 in that issue). The translator direct-route and provider-registry-baseline priorities are separate fast-follow scope and are not touched here.

This is test-only. No runtime code is changed.

- `tests/unit/account-fallback-rules.test.js` — locks in the decision matrix of `checkFallbackError` (`open-sse/services/accountFallback.js`), the single function both the per-account loop and the per-combo-candidate loop call to decide "try the next account/model" vs. "stop here": status-based rules (401/402/403/404 fixed cooldown, 429 exponential backoff), text-based rules taking priority over status rules, the 30s default for unclassified 5xx, and backoff-level capping.
- `tests/unit/chat-account-selection-loop.test.js` — drives the real `handleChat` (`src/sse/handlers/chat.js`) with mocked credentials/`handleChatCore`/upstream responses (no live provider calls), covering:
  - per-account fallback on a retryable failure, including that the *excluded* connection set grows correctly and `clearAccountError` fires for the account that actually served the response;
  - the loop honoring `shouldFallback: false` by stopping immediately instead of trying another account;
  - the terminal `allRateLimited` path (`unavailableResponse` with `Retry-After`);
  - the "no credentials configured at all" 404 path;
  - that `handleChatCore` is dispatched with the **refreshed** credentials, and that its `onCredentialsRefreshed` callback persists against the *original* connection's `providerSpecificData`;
  - combo-candidate rollover once a candidate's own accounts are exhausted (real `open-sse/services/combo.js`, not mocked) — modeled directly on the mixed VXP/OpenRouter/AG failure pattern in #3619;
  - the full-exhaustion terminal case: every combo candidate fails, and the loop still returns a well-formed, non-empty JSON error with a real status code — the contract a client needs to render an actual message instead of the opaque "Sorry, no response was returned" symptom reported in #3619.

## Why this scope

`chat.js`'s account-selection loop + its recursion into `handleComboChat` is called out in this repo's `CLAUDE.md` as "the thing to understand first," but had no direct test coverage — the existing `combo-*.test.js` files cover round-robin rotation, capability-based reordering, and fusion, and one other test (`gemini-native-endpoint.test.js`) mocks `handleChat` itself rather than exercising its internals. `checkFallbackError` (the shared retry/fallback classifier) had no dedicated tests at all. This PR fills exactly that gap without touching the translator or provider-registry priorities from #3702.

## Findings for maintainer triage (not fixed here, by design)

Per this PR's test-only scope, these are flagged for you to decide on rather than fixed:

1. **No distinction between a retryable and a non-retryable upstream error.** `checkFallbackError` (`open-sse/services/accountFallback.js`) has no rule that ever returns `shouldFallback: false` for an error it classifies — a 400 "malformed request" (e.g. a translator bug producing an invalid payload, which will fail identically on every other account *and* every other combo candidate) is bucketed into the same 30s "transient" default as an unclassified 5xx. `tests/unit/account-fallback-rules.test.js` documents this as current behavior. This is directly relevant to #3619's failure class: a bad request to one candidate cascades through every account and every other combo model before surfacing a generic terminal error, rather than failing fast with a message that points at the real (translation) bug.
2. **Terminal combo status is the *first* failing candidate's status, not the last.** When every combo candidate fails, `handleComboChat`'s `lastStatus` is set only once (`if (!lastStatus) lastStatus = result.status`) and never updated afterward, so the status code returned to the client reflects whichever candidate failed first, not necessarily the most recent or most representative failure. Covered/asserted as current behavior in the new "full-exhaustion" test rather than changed.

Neither of these is a crash or a silent failure — they're both classification/reporting nuances worth a deliberate decision rather than an incidental fix bundled into a test PR.

## Test plan

- [x] `node --check` on both new test files — syntax verified clean.
- [x] Every assertion was hand-traced against the actual current source (`src/sse/handlers/chat.js`, `open-sse/services/combo.js`, `open-sse/services/accountFallback.js`) line-by-line, including exact call counts, argument shapes, and which `ERROR_RULES` entry each mocked error message/status would hit.
- [ ] **Not yet run against a live `vitest` install.** Three separate `docker run … npm install` attempts (root deps, then `tests/` deps, per this repo's Docker-first tooling convention) each stalled for 10+ minutes without completing in this sandbox, most likely due to contention from other concurrent containers on the same host rather than anything specific to this change. A local host run was also attempted as a fallback and hit a pre-existing, unrelated problem: the checked-out `tests/node_modules` has a platform-mismatched optional native binding for `@rolldown/binding-*` (vitest 4's bundler), which is a known npm optional-deps issue (https://github.com/npm/cli/issues/4828) and reproduces even for the pre-existing test suite, not just these new files.
- [ ] Given the above, **please run `cd tests && npx vitest run unit/account-fallback-rules.test.js unit/chat-account-selection-loop.test.js` (or let CI run the full suite) before merging** to confirm these pass for real. Happy to fix anything that surfaces — flagging this explicitly rather than claiming a green run I couldn't actually produce.
- [ ] Full `tests/` suite baseline comparison via `tests/__baseline__/verify-no-regression.mjs` once the above confirms — not expected to be affected, since no runtime files are touched.

### DIFF
--- a/tests/unit/account-fallback-rules.test.js
+++ b/tests/unit/account-fallback-rules.test.js
@@ -1,0 +1,78 @@
+import { describe, it, expect } from "vitest";
+
+import { checkFallbackError, getQuotaCooldown } from "../../open-sse/services/accountFallback.js";
+
+// checkFallbackError is the single decision point both the per-account loop
+// (src/sse/handlers/chat.js) and the per-combo-candidate loop (open-sse/services/combo.js)
+// call to decide "try the next account/model" vs "stop and surface this error".
+describe("checkFallbackError decision matrix", () => {
+  it("429 with no matching text triggers status-based backoff, escalating per attempt", () => {
+    const first = checkFallbackError(429, "Too Many Requests", 0);
+    expect(first).toEqual({ shouldFallback: true, cooldownMs: getQuotaCooldown(1), newBackoffLevel: 1 });
+
+    const second = checkFallbackError(429, "Too Many Requests", first.newBackoffLevel);
+    expect(second).toEqual({ shouldFallback: true, cooldownMs: getQuotaCooldown(2), newBackoffLevel: 2 });
+    expect(second.cooldownMs).toBeGreaterThan(first.cooldownMs);
+  });
+
+  it("text-based rate-limit phrasing triggers backoff even when the status code doesn't match (e.g. 200 wrapping an error body)", () => {
+    const r = checkFallbackError(200, "Rate limit exceeded, please retry later", 0);
+    expect(r.shouldFallback).toBe(true);
+    expect(r.newBackoffLevel).toBe(1);
+  });
+
+  it("text rules are matched before status rules (order = priority)", () => {
+    // 401 alone would hit the status-based long cooldown; "quota exceeded" text should win first.
+    const r = checkFallbackError(401, "quota exceeded for this key", 0);
+    expect(r.shouldFallback).toBe(true);
+    expect(r.newBackoffLevel).toBe(1); // backoff path, not the fixed long-cooldown 401 path
+  });
+
+  it("401/402/403/404 use a fixed long cooldown, not exponential backoff", () => {
+    for (const status of [401, 402, 403, 404]) {
+      const r = checkFallbackError(status, "some upstream message", 3);
+      expect(r.shouldFallback).toBe(true);
+      expect(r.newBackoffLevel).toBeUndefined();
+      expect(r.cooldownMs).toBe(2 * 60 * 1000);
+    }
+  });
+
+  it("falls back to a 30s transient cooldown for unclassified 5xx errors", () => {
+    for (const status of [500, 502, 503, 504]) {
+      const r = checkFallbackError(status, "upstream had a bad day", 0);
+      expect(r).toEqual({ shouldFallback: true, cooldownMs: 30 * 1000 });
+    }
+  });
+
+  // Documents current behavior rather than asserting a requirement: there is no rule
+  // anywhere in this file that ever returns shouldFallback:false. A malformed-request
+  // 400 (a client-side bug that will reproduce identically on every other account/model)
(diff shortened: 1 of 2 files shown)


nitsuah:
Follow-up on the open test-plan checkbox: got a real `vitest` run through this time (Docker, `node:22-alpine` — the previous attempt's stall was environment contention from other concurrent containers, not anything about this change).

```
✓ unit/account-fallback-rules.test.js (9 tests) 16ms
✓ unit/chat-account-selection-loop.test.js (8 tests) 31ms

Test Files  2 passed (2)
     Tests  17 passed (17)
```

17/17 passing, confirming the hand-traced assertions from the PR description hold up for real. No runtime files are touched by this PR, so I didn't re-run the full suite/baseline gate on top of this — that's already been separately confirmed clean on the same `master` base via #3706's validation.

nitsuah:
@decolua I would love to become the official tester for 9router!

nitsuah:
Brought up to date with `master` (`39e36d3d`; was 151 commits behind), plus one test update (`81ea19ad`). Master changed the behavior these characterization tests pin:

- **The flagged gap is fixed upstream.** The "malformed 400 is cooled down like a transient 503" finding from the description no longer holds: `checkFallbackError` now returns `{ shouldFallback: false, cooldownMs: 0 }` for request-scoped 4xx that match no rule. The test now asserts that, and the "never returns false" test became "falls back for account-scoped/transient failures, not for request-scoped 4xx".
- **All-accounts-rate-limited is now always 503.** Previously it reused the loop's last status (429). The message still carries the loop's last error; the test was updated accordingly.
- **New export mocked.** `chat.js` now calls `clearAntigravityStrikes` on success, and the mock provides it.

Both files pass 17/17 in Docker (`node:22`, `CI=true`), and no other suites are affected.

PR #3706, opened 2026-09-01
feat(quota): add usage tracking for Groq

nitsuah:
## What

Wires up quota/usage tracking for **Groq** — the first slice of #3701 (12 connected providers with no quota tracking).

## Why Groq

The issue suggests OpenAI and Groq as candidates. I investigated both:

- **OpenAI** exposes usage/costs only through an *organization admin* endpoint (`/v1/organization/usage/*`), which requires an admin API key most users won't have — the wrong fit for a per-connection dashboard quota check, and the "insufficient permissions" case would dominate.
- **Groq** has no dedicated quota endpoint either, but every API response carries `x-ratelimit-*` headers (request + token limits, remaining counts, reset windows) for free — the issue explicitly flags this as the cheapest provider to wire up, and that held up under investigation.

So this PR implements Groq only.

## Pattern (matches existing Claude/Codex/Kiro/DeepSeek usage handlers)

1. **`open-sse/services/usage/groq.js`** (new) — `getGroqUsage(apiKey, proxyOptions)` GETs the models-list endpoint (already declared as `transport.validateUrl`, so this costs zero tokens) and reads the rate-limit headers off the response instead of the body:
   - `x-ratelimit-limit-requests` / `x-ratelimit-remaining-requests` / `x-ratelimit-reset-requests`
   - `x-ratelimit-limit-tokens` / `x-ratelimit-remaining-tokens` / `x-ratelimit-reset-tokens`

   Groq's reset headers are Go-style duration strings (e.g. `"2m59.56s"`), not timestamps, so they're parsed and added to `Date.now()` rather than run through the existing `parseResetTime` helper (which expects an absolute time/timestamp).

2. **Graceful "not tracked" vs. real error**, per the issue's guidance (matching `getVercelAiGatewayUsage`'s "unfunded" handling):
   - Missing/invalid API key → a plain message, no network call.
   - `401`/`403` → "authentication failed" message.
   - Non-2xx → error message with the status code.
   - **2xx but no rate-limit headers present** (key valid, nothing to report yet) → `{ plan: "Groq", message: "...no rate-limit data...", quotas: {} }`, not an error.

3. **Registry** (`open-sse/providers/registry/groq.js`): added `transport.usage.url` (reusing the existing models URL, single source of truth) and `features: { usage: true, usageApikey: true }` — the same flags `deepseek`/`vercel-ai-gateway`/`ollama` already use to appear in `USAGE_SUPPORTED_PROVIDERS` / `USAGE_APIKEY_PROVIDERS`, which is what makes Groq show up in the Quota Tracker and Provider Limits dashboard pages with no further UI changes needed.

4. **Dispatch**: registered `groq: (c) => getGroqUsage(c.apiKey, c.proxyOptions)` in `open-sse/services/usage.js`'s `USAGE_HANDLERS` map (used by `GET /api/usage/[connectionId]`).

5. **Dashboard rendering** (`ProviderLimits/utils.js`): added a `"groq"` case to `parseQuotaData` that forwards `used`/`total`/`resetAt` as absolute counts (same style as `codex`/`kiro` — the UI's `calculatePercentage` derives the bar), rather than a manual `remainingPercentage` (which is only needed for ratio/credit-based providers like DeepSeek/Ollama).

## Fast-follow scope

This intentionally stops at one provider. The remaining ~10-11 providers from #3701 can each follow the exact same 4-touch-point pattern (usage handler file → registry `features` flags + `transport.usage` → `USAGE_HANDLERS` entry → `parseQuotaData` case) as separate PRs.

## Test plan

- New `tests/unit/groq-usage.test.js` (mocked `proxyAwareFetch`, matching the existing `deepseek-usage.test.js`/`ollama-usage.test.js` convention — no live API key used or required):
  - registry flags (`USAGE_SUPPORTED_PROVIDERS` / `USAGE_APIKEY_PROVIDERS` include `groq`)
  - GET to the models endpoint with `Bearer` auth
  - correctly parses request + token rate-limit headers into quotas, including resolving the duration-string reset headers to real future timestamps
  - **regression check**: when no rate-limit headers are present at all, returns a soft "not tracked yet" message with `quotas: {}` rather than misreporting `used:0/total:0` quotas (caught a bug during review: `Number(headers.get(missingHeader))` is `0`, a *finite* number, so the "were there any headers at all" check has to test for `null` explicitly before coercing)
  - missing API key / 401 handling
  - `parseQuotaData("groq", ...)` normalization
- Ran (inside a `node:22-alpine` Docker container per this repo's tooling convention, since the host's existing `node_modules` had a broken native binding unrelated to this change):
  - `unit/groq-usage.test.js` — 6/6 pass
  - `unit/groq-usage.test.js unit/deepseek-usage.test.js unit/usage-dispatch.test.js unit/usage-concern.test.js unit/ollama-usage.test.js` together — 27/27 pass, no cross-test interference
  - `tests/__baseline__/verify-providers.mjs` — still byte-for-byte equal (81 providers); the new `transport.usage` field is already in that script's `ADDED_FIELDS` allow-list alongside `vercel-ai-gateway`/`qoder`'s usage config, so no baseline snapshot update was needed
- `node --check` syntax validation on all changed/new files
- Not tested live — no Groq API key available in this environment; the header-parsing logic (including the duration-string reset parser) is exercised entirely through the mocked-response unit tests instead, per this task's guidance for untestable-live paths

### DIFF
--- a/open-sse/providers/registry/groq.js
+++ b/open-sse/providers/registry/groq.js
@@ -17,6 +17,12 @@ export default {
   transport: {
     baseUrl: "https://api.groq.com/openai/v1/chat/completions",
     validateUrl: "https://api.groq.com/openai/v1/models",
+    // No dedicated quota endpoint; rate-limit info rides on x-ratelimit-*
+    // response headers, always included. Reuse the models list (already
+    // used as validateUrl) so reading usage never costs tokens.
+    usage: {
+      url: "https://api.groq.com/openai/v1/models",
+    },
   },
   models: [
     { id: "llama-3.3-70b-versatile", name: "Llama 3.3 70B" },
@@ -33,5 +39,9 @@ export default {
     authType: "apikey",
     authHeader: "bearer",
     format: "openai",
+  },
+  features: {
+    usage: true,
+    usageApikey: true,
   },
 };
--- a/open-sse/services/usage.js
+++ b/open-sse/services/usage.js
@@ -14,6 +14,7 @@ import { getCodeBuddyCnUsage, getCodeBuddyIntlUsage } from "./usage/codebuddy-cn
 import { getGrokCliUsage } from "./usage/grok-cli.js";
 import { getKimiUsage } from "./usage/kimi.js";
 import { getDeepseekUsage } from "./usage/deepseek.js";
+import { getGroqUsage } from "./usage/groq.js";
 import { getZedUsage } from "./usage/zed.js";
 import { resolveQoderCredentials } from "./qoderModels.js";
 import { getGlmUsage } from "./usage/glm.js";
@@ -55,6 +56,7 @@ const USAGE_HANDLERS = {
   "grok-cli": (c) => getGrokCliUsage(c.accessToken, c.providerSpecificData, c.proxyOptions),
   kimi: (c) => getKimiUsage(c.accessToken, c.apiKey, c.proxyOptions, c.providerSpecificData),
   deepseek: (c) => getDeepseekUsage(c.apiKey, c.proxyOptions),
+  groq: (c) => getGroqUsage(c.apiKey, c.proxyOptions),
   zed: (c) => getZedUsage(c.accessToken, c.providerSpecificData, c.proxyOptions),
 };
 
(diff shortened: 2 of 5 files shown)


decolua:

PR #3382, opened 2026-08-16
feat(antigravity): add Gemini 3.7 Flash (High, Medium, Low) to MITM proxy defaultModels

alfep:
## Summary

- Adds `Gemini 3.7 Flash (High)`, `Gemini 3.7 Flash (Medium)`, and `Gemini 3.7 Flash (Low)` entries to `MITM_TOOLS.antigravity.defaultModels` in `src/shared/constants/cliTools.js`.
- Enables direct model mapping configuration for Gemini 3.7 Flash tiers in the 9Router Web Dashboard / MITM Proxy UI table.
- Added comprehensive unit tests for Gemini 3.7 Flash tier routing, upstream resolution, thinking level mapping, and MITM model extraction.

## Changes

1. **`src/shared/constants/cliTools.js`**:
   - Added `gemini-3.7-flash-high`, `gemini-3.7-flash-medium`, and `gemini-3.7-flash-low` to `defaultModels` so they appear in the MITM tool card and model mapping interface.
2. **`tests/unit/gemini-37-integration.test.js`**:
   - Added unit test suite validating Gemini 3.7 Flash tier transformations, `thinkingLevel` inclusion, MITM extraction for `gemini-3.7-flash-tiered`, default fallback handling, and CLI catalog synchronization.

## Verification

- Ran unit test suites with Vitest:
  - `tests/unit/gemini-37-integration.test.js` (PASSED)
  - `tests/unit/gemini-36-integration.test.js` (PASSED)
  - `tests/unit/antigravity-mitm.test.js` (PASSED)
  - `tests/unit/antigravity-retry-hook.test.js` (PASSED)
- All 39 tests passed cleanly without regressions.

PR #3715, opened 2026-09-02
fix: preserve Antigravity identity on model catalog requests

### LINKED ISSUE #3414: [Bug] Gemini 3.7 Flash missing when Antigravity MITM is enabled
## Description

When **Antigravity MITM** is enabled in 9router, the **Gemini 3.7 Flash** model is not available in Antigravity after launching it.

The issue is reproducible with the latest versions:

- **9router:** v0.5.55
- **Antigravity:** v2.8.1

Instead, the model list contains older Gemini models, including **Gemini 3.6 Flash**, but **Gemini 3.7 Flash is missing**.

When Antigravity is used **without MITM**, Gemini 3.7 Flash is available normally.

## Steps to reproduce

1. Install/update to **9router v0.5.55**.
2. Use **Antigravity v2.8.1**.
3. Enable **Antigravity MITM** in 9router.
4. Start/restart Antigravity.
5. Open the model selector in Antigravity.
6. Check the available Gemini models.
7. **Gemini 3.7 Flash is missing**, while older models such as Gemini 3.6 Flash are still available.

## Expected behavior

When using Antigravity through 9router MITM, the available model list should include the current models exposed by Antigravity, including **Gemini 3.7 Flash**.

## Actual behavior

With MITM enabled, Antigravity appears to receive an outdated or incomplete model list and **Gemini 3.7 Flash is not exposed**.

## Environment

- **9router:** v0.5.55
- **Antigravity:** v2.8.1
- **Antigravity MITM:** Enabled

## Additional information

The issue does **not** occur when Antigravity MITM is disabled.

This suggests that the model list exposed or intercepted by the **Antigravity MITM** may not be synchronized with the current models available upstream.

It would be useful if the MITM model list could be dynamically synchronized with the models currently available in Antigravity rather than relying on a static or outdated list.

mvanhorn:
Antigravity exposes Gemini 3.7 Flash when connected directly, but the model disappears from its selector when 9router MITM is enabled. Adding the Gemini 3.7 tiers to the dashboard mapping catalog in PR #3382 did not fix the native Antigravity catalog returned through the proxy. The MITM passthrough currently rewrites every Antigravity request to the hardcoded IDE version `1.23.2`, including non-chat model-catalog calls from newer Antigravity releases. This gives the issue a concrete local cause consistent with the reporter's reproduction on Antigravity 2.11.0.

## Summary

Restrict the legacy Antigravity IDE-version override in `src/mitm/antigravityIdeVersion.js` to generation endpoints that need the existing upstream compatibility workaround. Pass the request URL from `src/mitm/server.js` into that decision so model-catalog and other non-chat passthrough requests retain their original `User-Agent`, metadata, body bytes, and content length. Keep the existing chat-request rewrite behavior intact to avoid reopening the compatibility failure the override was introduced to handle.

## Tests

- A `fetchAvailableModels` or equivalent non-generation request with an Antigravity 2.11.0 identity is returned unchanged, preserving both the original `User-Agent` and `metadata.ideVersion` so current models remain eligible upstream.
- A `generateContent` and a `streamGenerateContent` request still applies the known-good legacy version override and returns the rewritten body for forwarding.
- A non-chat request with malformed or non-JSON content remains byte-for-byte unchanged and does not acquire a rewritten identity.
- A request without an Antigravity `User-Agent` or metadata identity remains unchanged instead of synthesizing identity fields.

Fixes #3414

### DIFF
--- a/src/mitm/antigravityIdeVersion.js
+++ b/src/mitm/antigravityIdeVersion.js
@@ -1,8 +1,8 @@
 "use strict";
 
-// Rewrite Antigravity IDE markers so upstream AG 2.x backend accepts the request.
-// User-Agent header (antigravity/<old>) and body.metadata.ideVersion are forced
-// to a known-good IDE version. Hardcoded MVP — toggle/version configurable later.
+// Rewrite Antigravity IDE markers on generation requests so upstream AG 2.x
+// backend accepts them. Catalog and other passthrough requests retain the
+// client's current identity. Hardcoded MVP — toggle/version configurable later.
 
 const ANTIGRAVITY_IDE_VERSION = "1.23.2";
 const ANTIGRAVITY_IDE_VERSION_OVERRIDE_ENABLED = true;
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @mvanhorn for the contribution! Reviewed and merged into master. 🙏

PR #3719, opened 2026-09-02
fix(claude): support Fable 5.1

LucasOl1337:
## Summary

- add `claude-fable-5-1` to the Claude Code model catalog
- centralize the spoofed Claude Code version and update both request and billing identities to `2.1.257`
- mark Fable 5.1 as a permanently adaptive 1M-context model and send `output_config.effort` without a redundant thinking switch
- add regression coverage for the catalog capabilities, headers, billing identity, and adaptive-effort payload

## Root cause

Direct CC requests advertised Claude Code `2.1.92`. Fable 5.1 rejects that identity with HTTP 400 because it requires Claude Code `2.1.251` or newer.

## Validation

- `git diff --check` passed
- tests were added but not run, per operator instruction

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@
 - **i18n**: pt-BR expanded to 1132 terms
 
 ## Fixes
+- **Claude Code**: add Claude Fable 5.1 and advertise Claude Code 2.1.257 in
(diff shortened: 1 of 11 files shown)


decolua:
Thanks @LucasOl1337 for the contribution! Reviewed and merged into master. 🙏

PR #3720, opened 2026-09-02
fix(translator): route Gemini thinking through reasoning_effort on OpenAI-compatible wire

### LINKED ISSUE #3718: [Bug] Custom OpenAI-compatible Gemini provider: 400 "Unknown name generationConfig" when Claude Code sends `thinking` param
## 9Router version
v0.5.59 (also observed on earlier version, ~1 month ago)

## Setup
- Custom provider (NOT the built-in "antigravity/" preset) pointed at Google's OpenAI-compatible Gemini endpoint: `https://generativelanguage.googleapis.com/v1beta/openai`
- Client: Claude Code, connected via `ANTHROPIC_BASE_URL` pointing at 9Router
- Combo/model routed through this custom provider

## Bug
Every request from Claude Code that includes Anthropic's `thinking` parameter (sent by default for any non-first-party/custom model) fails with:

    {
      "error": {
        "code": 400,
        "message": "Invalid JSON payload received. Unknown name \"generationConfig\": Cannot find field.",
        "status": "INVALID_ARGUMENT"
      }
    }

This makes the custom-provider path to Gemini unusable for Claude Code traffic, since `thinking` is essentially always present.

## Root cause (isolated via direct testing against Google's API)
Google's OpenAI-compatible endpoint (`/v1beta/openai/chat/completions`) does NOT recognize a top-level `generationConfig` field -- that's the *native* Gemini/Vertex API shape, not the OpenAI-compatible one. Confirmed directly:

    # This works (200 OK) -- correct OpenAI-compatible field:
    curl .../v1beta/openai/chat/completions -d '{"model":"gemini-3.5-flash-lite",
      "messages":[...], "reasoning_effort":"low"}'

    # This fails (400, exact same error 9Router surfaces) -- native-format field
    # sent to the OpenAI-compatible endpoint:
    curl .../v1beta/openai/chat/completions -d '{"model":"gemini-3.5-flash-lite",
      "messages":[...], "generationConfig":{"thinkingConfig":{"thinkingBudget":100}}}'

It looks like 9Router's translation layer for custom OpenAI-compatible providers maps Anthropic's `thinking` block to the *native* Gemini `generationConfig.thinkingConfig` shape instead of the OpenAI-compatible `reasoning_effort` field, even when the target provider is configured as `openai-compatible` rather than the native Gemini format.

## Suggested fix
When translating `thinking` for a provider configured as OpenAI-compatible (as opposed to the native Gemini/Vertex format used by the `antigravity/` preset), emit `reasoning_effort` instead of `generationConfig.thinkingConfig`.

## Impact
Blocks using Google AI Studio API keys (paid or free tier) as a custom provider for Claude Code traffic via 9Router -- currently the only workaround is to not send `thinking` at all, which isn't controllable from the Claude Code side.

mkhuda:
## Problem

Custom OpenAI-compatible providers pointed at Google's Gemini OpenAI-compat endpoint (`https://generativelanguage.googleapis.com/v1beta/openai`) fail whenever the client sends a `thinking` parameter (e.g. Claude Code sends this by default for any non-first-party model):

    {
      "error": {
        "code": 400,
        "message": "Invalid JSON payload received. Unknown name \"generationConfig\": Cannot find field."
      }
    }

`resolveFormat()` in `thinkingUnified.js` lets a model's capability-declared native thinking format (`gemini-level`/`gemini-budget`, from the `*gemini-3*`/ `*gemini-2.5*` capability patterns) win over the actual wire target, even when the connection's real `targetFormat` is `openai` (a custom/OpenAI-compatible provider).

The native format nests thinking under `generationConfig`, which Google's OpenAI-compat endpoint doesn't recognize.

## Fix

In `resolveFormat()`, only let a capability's native thinking format (`gemini-level`, `gemini-budget`, `claude-budget`, `claude-adaptive`, `kiro`) win when the wire target actually speaks that native protocol.
When the wire is OpenAI-compatible (`openai`/`openai-responses`), fall through to the OpenAI-shaped default (`reasoning_effort`) instead.

## Tests

Adds 5 cases to `tests/translator/thinking-unified.test.js` covering Gemini 3.x (`gemini-level`) and Gemini 2.5.x (`gemini-budget`) over a custom OpenAI-compatible provider, plus a control case confirming native-format routing (antigravity/gemini-cli/vertex) is unaffected.

- `npx vitest run tests/translator/thinking-unified.test.js` — 59/59 pass
- Full suite before/after diff — 0 regressions (85 pre-existing/environment fails, unchanged)
- Live verification: ran the real `translateRequest()` against an actual
  Google AI Studio API key — translated body now sends `reasoning_effort`
  and gets a real `200 OK` streamed response from Gemini (previously `400`)

Fixes #3718

### DIFF
--- a/open-sse/translator/concerns/thinkingUnified.js
+++ b/open-sse/translator/concerns/thinkingUnified.js
@@ -105,12 +105,16 @@ export function extractThinking(body) {
 // at the call-site where intent is snapshotted before format translation.
 export const captureThinking = extractThinking;
 
-// Resolve thinking format: provider override > capability > derive(targetFormat).
+const NATIVE_ONLY_FORMATS = new Set(["gemini-level", "gemini-budget", "claude-budget", "claude-adaptive", "kiro"]);
+
 function resolveFormat(targetFormat, model, provider) {
   const providerFmt = provider ? PROVIDERS[provider]?.thinkingFormat : null;
   if (providerFmt) return providerFmt;
   const caps = getCapabilitiesForModel(provider, model);
-  if (caps.thinkingFormat) return caps.thinkingFormat;
+  const isOpenAIWire = targetFormat === "openai" || targetFormat === "openai-responses";
+  if (caps.thinkingFormat && !(isOpenAIWire && NATIVE_ONLY_FORMATS.has(caps.thinkingFormat))) {
+    return caps.thinkingFormat;
+  }
   return FORMAT_TO_NATIVE[targetFormat] || "openai";
 }
 
--- a/tests/translator/thinking-unified.test.js
+++ b/tests/translator/thinking-unified.test.js
@@ -215,6 +215,26 @@ describe("applyThinking per provider format", () => {
     const out = apply("openai", "gpt-5.6-sol", { reasoning_effort: "max" }, "kiro");
     expect(out.reasoning_effort).toBe("xhigh");
   });
+  it.each([
+    ["gemini-3.5-flash-lite"],
+    ["gemini-3.7-flash"],
+    ["gemini-3-pro"],
+  ])("Gemini 3.x model %s (gemini-level) over a custom OpenAI-compatible provider → reasoning_effort, not generationConfig (regression: #3718)", (model) => {
+    const out = apply("openai", model, { reasoning_effort: "medium" }, "my-custom-gemini-openai");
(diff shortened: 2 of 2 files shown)


decolua:
Thanks @mkhuda for the fix! Reviewed and merged into master. 🙏

Issue #2263, opened 2026-07-01
Test suite: no CI, not reproducible outside Docker, and 31 pass→fail regressions


vuongtlt13:
## Problem

The `tests/` suite ships a no-regression gate (`tests/__baseline__/verify-no-regression.mjs` + `known-fails.txt`) that tolerates a fixed set of **26** known failures — but there are three gaps around it:

1. **No CI runs the gate.** Nothing enforces it on push/PR, so regressions land silently.
2. **Not reproducible outside the author's Docker image.** The gate hardcodes the `/app` path and the runner expects vitest at `/tmp/node_modules`, so a fresh clone can't run the suite as-is.
3. **31 real pass→fail regressions** have accumulated on top of the 26 tolerated ones — mostly environment drift and API-shape changes:
   - `security-audit` resolves source paths from the test CWD instead of the repo root (13)
   - `golden-url-header` snapshot embeds machine-specific version/platform/arch/node (3)
   - `codex-image-fetch` / `image-fetch-hardening` DNS mock returns the pre-`{ all: true }` object shape (3)
   - `force-stream-config` missing headroom mock exports (2)
   - `combo-autoswitch` asserts unshipped `web_search` behavior (2)
   - `executor-const-guard` expects antigravity 429 attempts = 6 (actual 3) (1)
   - `db-concurrent` same-millisecond rows collapsed by the dedupe key (3)
   - `db-benchmark` hard-depends on the legacy `lowdb` dep (1)
   - `*.live.test.js` (network) and cloud embeddings run even when `cloud/` is absent

## Expected

- CI runs the test suite + no-regression gate on every push/PR.
- A fresh clone can run `cd tests && npm install && npm run test:ci` with no `/tmp` or Docker-specific setup.
- The gate is green: 0 regressions, with the 26 `known-fails.txt` entries left untouched by design.

## Also surfaced (out of scope, needs a maintainer decision)

- The three config verifiers (`verify-oauth-urls` / `verify-providers` / `verify-alias`) are already red on `master` from stale baselines (qwen OAuth URL change, blackbox `/v1`, added kimchi/venice). Regenerating those baselines is a separate call.
- Two `known-fails.txt` entries look like genuine code regressions, not test bugs: antigravity model losing its `mandatory: true` flag, and Kiro missing an `auto` slot.

PR #3723, opened 2026-09-02
fix(security): close SSRF guard bypasses (#3714)

### LINKED ISSUE #3714: [Security] SSRF in /v1/search endpoint with multiple bypasses of URL guard
### Description

The `/v1/search` endpoint in 9Router allows clients to override the search provider's `baseUrl` via `provider_options.baseUrl`. The server then makes a direct `fetch()` request to this user-supplied URL.

Although the project implements an SSRF guard (`src/shared/utils/ssrfGuard.js`), it contains **multiple bypasses** that allow attackers to reach internal hosts and cloud metadata endpoints:

1. **No DNS resolution** – domain names that resolve to loopback addresses (e.g., `nip.io`, `localtest.me`, `sslip.io`) are not blocked because the guard only checks the hostname string, not the resolved IP.
2. **IPv6-mapped IPv4 bypass** – formats like `[::ffff:127.0.0.1]` are normalized by the URL parser to `[::ffff:7f00:1]`, causing the IPv6 block regex to fail.
3. **Trailing dot bypass** – `localhost.` does not match the `hostname.endsWith(".localhost")` check.
4. **Redirect following** – the `fetch()` implementation follows redirects by default, and the redirect target is **not re-checked** against the SSRF guard.

Furthermore, requests made by the server to loopback addresses are tagged with `x-9r-real-ip: 127.0.0.1` and a valid internal token. This causes the middleware to treat the request as originating from a trusted local source (`isLocalRequest=true`), effectively granting the attacker **authenticated access to local `/v1/*` endpoints without any API key**.

An authenticated attacker (or unauthenticated in `requireApiKey=false` deployments) can therefore:
- Exfiltrate upstream search provider API keys (which are sent in request headers) to an attacker-controlled server.
- Probe internal networks, local services, and cloud metadata endpoints.
- Access local `/v1/*` endpoints without credentials due to the `isLocalRequest` flag.

### Affected Versions

- All versions up to and including **0.5.55** (current public version).

### Expected Behavior

- The SSRF guard should block all internal/private IPs and loopback-equivalent domains, regardless of DNS resolution variants and URL encoding.
- `baseUrl` overrides should be strictly validated against a whitelist, or the feature should be disabled for non-admin users.
- Redirect targets should be re-checked against the SSRF guard.
- Local/internal requests should not be automatically trusted for authentication purposes.

### Actual Behavior

- Multiple bypasses exist in the SSRF guard, allowing access to internal networks.
- Upstream API keys are exposed during SSRF attacks.
- Requests to local endpoints are automatically authenticated.

### Impact

- **Critical** – internal network probing, local service enumeration, and potential data exfiltration.
- **API key disclosure** – upstream search provider credentials can be stolen.
- **Authentication bypass** – local `/v1/*` endpoints are accessible without valid API keys.
- In default deployments (`REQUIRE_API_KEY=false`), this vulnerability can be exploited completely unauthenticated.

### Root Cause Analysis

The SSRF guard (`src/shared/utils/ssrfGuard.js`) suffers from several design flaws:

```js
// No DNS resolution – only string matching
const BLOCKED_HOSTNAMES = new Set(["localhost", "ip6-localhost", "ip6-loopback"]);

// IPv6-mapped IPv4: URL parser normalizes [::ffff:127.0.0.1] to [::ffff:7f00:1]
// The regex ^::ffff:(\d+\.\d+\.\d+\.\d+)$ then fails to match

// Trailing dot: hostname.endsWith(".localhost") does not catch "localhost."

// Redirects: undici's fetch follows by default, no re-validation
```

Additionally, `custom-server.js` tags loopback requests as trusted, giving them `isLocalRequest=true` status in the middleware.

### Proposed Fix

- Implement DNS resolution in the SSRF guard and block all resolved IPs that are private or loopback.
- Use a robust IP range check (`isIP()` + `isPrivate()`) rather than string matching.
- Normalize URL hosts consistently before applying IPv6 checks.
- Disable automatic redirect following, or re-validate redirect targets.
- Do not automatically trust loopback requests for authentication purposes.
- Consider removing the `baseUrl` override feature entirely, or restrict it to administrators.
- Apply rate limiting to the `/v1/search` endpoint to reduce abuse potential.

### Additional Context

I have verified these bypasses in a local deployment. If you need further details, I'm happy to provide them privately. Thank you for maintaining this project!

nitsuah:
## Summary

Closes the four SSRF guard bypasses reported in #3714. Verified all four independently before fixing (the exact hex-form IPv6 normalization, the trailing-dot hostname, and the alternate IPv4 encodings were reproduced empirically against Node's own `URL` parser, not assumed from the report).

`src/shared/utils/ssrfGuard.js`'s checks matched specific **textual representations** of an address rather than the address itself — these are bugs in what was already being checked, not missing coverage:

1. **IPv6 representation mismatch.** The IPv4-mapped check only matched the dotted form (`::ffff:127.0.0.1`), but WHATWG URL parsing normalizes that to hex form (`::ffff:7f00:1`), which the old regex never matched — silently letting through localhost, RFC1918 ranges, and the `169.254.169.254` cloud metadata address whenever expressed that way. Replaced regex-matching with a proper IPv6-to-8×16-bit-groups parser and numeric range checks, so it's immune to which textual form a client (or the URL parser) picks. Also now catches NAT64 (`64:ff9b::/96`) and deprecated IPv4-compatible (`::a.b.c.d`) embeddings.
2. **Trailing-dot hostname** (`localhost.`) compared as a literal string against `localhost` — the semantically-identical FQDN form bypassed every hostname/suffix check. Now normalized before any check runs.
3. **No DNS resolution** — a hostname that merely *resolves* to a private/loopback address (wildcard-DNS services like nip.io/sslip.io, or an attacker domain with an A record pointed internally) passed every literal check. Added `assertPublicUrlResolved()`: same literal checks, plus `dns.promises.lookup` on non-literal hostnames with every returned address validated. Fails open on a lookup error (not an SSRF signal by itself) rather than blocking on transient DNS issues.
4. **Redirects** — the outbound `fetch()` followed redirects automatically with zero re-validation, so a validated public URL could 30x its way to an internal target. Added `fetchPublic()`: validates the initial URL, then manually chases redirects (bounded, default 5 hops), re-validating each `Location` before following it.

## What changed where

- `fetchPublic` wired into the actual outbound request in `open-sse/handlers/search/index.js` — the `/v1/search` dedicated-provider path #3714 reported.
- `src/sse/handlers/fetch.js`'s (`/v1/fetch`) pre-check upgraded from sync `assertPublicUrl` to the DNS-resolving `assertPublicUrlResolved`.
- `open-sse/handlers/search/callers.js`'s `resolveBaseUrl` **intentionally left on the existing sync `assertPublicUrl`** — it's a fast pre-check (immediate 400 on obviously-bad input) that doesn't need to become async, since the real protection against all four bypass classes now lives at the actual network call via `fetchPublic`. Making it async would have cascaded through its 13 provider request-builder functions for no additional safety.

## What I deliberately did NOT change

`src/app/api/provider-nodes/validate/route.js`'s `isLocalRequest`-gated bypass (self-hosted providers like `ollama-local` legitimately need to reach `localhost` from an actual local dashboard request). That trust decision is keyed off the real TCP peer address via `custom-server.js`/`trustedPeer.js`, which independently strips and regenerates any client-supplied trust headers from the real socket on every request — it isn't forgeable by a remote attacker directly. What the SSRF bypasses *did* enable was reaching that endpoint (or any other loopback service) via a server-side outbound request in the first place, which would then appear as a legitimate local request once it landed there. Closing the bypasses at the fetch layer closes that path at its root rather than touching the downstream trust decision, which is correct as designed for the legitimate local-provider use case.

## Test plan

- `tests/unit/ssrf-guard-hardening.test.js` (new, 15 tests) — every bypass class from #3714 reproduced and blocked, plus ones found while fixing it (NAT64, multi-A-record where only one resolved address is private, bounded redirect loops), against real registry data where applicable. `node:dns` and `fetch` mocked throughout — no live-network dependency.
- Existing `search-ssrf-guard.test.js` (8 tests, unchanged `resolveBaseUrl` sync path), `xquik-search-provider.test.js` (6 tests, exercises the real `fetchPublic` call path with a mocked `fetch`), and `fetch-success-clears-account.test.js` (updated its `ssrfGuard.js` mock to the new export name) — all still passing.
- `npx eslint` clean on all changed files.
- Ran the broader `search`/`fetch` test-file set for regression sanity; the only failures (`codex-image-fetch.test.js`, `image-fetch-hardening.test.js`, 3 tests) are pre-existing and already documented in #2263's own findings ("DNS mock returns the pre-`{ all: true }` object shape") — unrelated to this change, not touched by it.
- All 30 tests run via Docker (`node:22-alpine`) per this repo's tooling convention.

Manually verified the specific bypass strings from #3714 against Node's `URL` parser before writing the fix (not assumed from the report) — happy to share that verification script if useful for review.

Fixes #3714.

### DIFF
--- a/open-sse/handlers/search/index.js
+++ b/open-sse/handlers/search/index.js
@@ -10,6 +10,7 @@
 import { buildSearchRequest } from "./callers.js";
 import { normalizeSearchResponse } from "./normalizers.js";
 import { handleChatSearch } from "./chatSearch.js";
+import { fetchPublic } from "../../../src/shared/utils/ssrfGuard.js";
 
 const GLOBAL_TIMEOUT_MS = 15000;
 const NON_RETRIABLE = new Set([400, 401, 403, 404]);
@@ -100,7 +101,7 @@ async function tryDedicatedProvider({ provider, providerConfig, body, credential
   log?.info?.("SEARCH", `${provider.id} | "${params.query.slice(0, 80)}" | type=${params.searchType}`);
 
   try {
-    const resp = await fetch(url, { ...init, headers: sanitizeHeaders(init.headers), signal: controller.signal });
+    const resp = await fetchPublic(url, { ...init, headers: sanitizeHeaders(init.headers), signal: controller.signal });
     clearTimeout(timer);
     if (!resp.ok) {
       const errText = await resp.text().catch(() => "");
--- a/src/shared/utils/ssrfGuard.js
+++ b/src/shared/utils/ssrfGuard.js
@@ -1,4 +1,24 @@
 // SSRF guard: block internal/private/metadata targets for server-side fetch.
+//
+// Three layers, each closing a distinct bypass class documented in #3714:
+//   1. assertPublicUrl        - synchronous literal-IP/hostname checks (cheap, for
+//                                immediate rejection of obviously-bad input at request-build time).
+//   2. assertPublicUrlResolved - adds DNS resolution so a hostname that merely
+//                                *resolves* to a private/loopback address (e.g. a
+//                                nip.io/sslip.io wildcard-DNS domain, or an attacker's
+//                                own domain pointed at 127.0.0.1) is also rejected.
+//   3. fetchPublic             - wraps fetch() with manual redirect handling so a
+//                                validated public URL can't 30x its way to an
+//                                internal target without the redirect target being
+//                                re-validated through layer 2 first.
+//
+// Layer 1 alone previously had matching bugs, not just missing coverage: hostname
+// checks ran on the raw string without normalizing a trailing dot ("localhost."),
+// and the IPv6 check only recognized one textual representation of an IPv4-mapped
+// address (dotted "::ffff:a.b.c.d") while Node/WHATWG URL parsing can normalize the
+// same address to hex form ("::ffff:7f00:1") — a mismatch, not an oversight.
+
+import dns from "node:dns";
 
 const BLOCKED_HOSTNAMES = new Set(["localhost", "ip6-localhost", "ip6-loopback"]);
 const BLOCKED_SUFFIXES = [".internal", ".local", ".localhost"];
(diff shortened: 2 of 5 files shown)


decolua:

PR #3734, opened 2026-09-02
fix(kiro): preserve inline images in OpenAI MITM

OutisNyel:
## What

- Forward Kiro userInputMessage.images as OpenAI-compatible image_url content parts in the Kiro MITM request conversion path.
- Keep text before image parts and preserve the original user turn ordering, including historical turns.
- Emit image-only user turns even when tool results are present.
- Ignore malformed or unsupported image entries without changing text-only behavior.
- Ignore local .kiro/ workspace state in git.

## Why

Kiro carries attached images separately from userInputMessage.content. The MITM converter only forwarded text content, so downstream OpenAI-compatible vision models could receive a prompt that referenced an image without receiving the image payload.

## Tests

- npm --prefix tests test -- unit/kiro-image-forwarding.test.js

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -86,3 +86,6 @@ graphify-out/*
 .codegraph/
 .PR/
 .next-analyze/*
+
+# Kiro local workspace state
+.kiro/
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @OutisNyel for the contribution! Reviewed and merged into master. 🙏

PR #3732, opened 2026-09-02
feat(antigravity,gemini): add Gemini 3.8 Flash support and bump IDE fingerprint to 2.11.0

rafazafar:
Adds support for Gemini 3.8 Flash across Antigravity and Gemini providers.

Screenshot 2026-09-03 at 04 20 12

Screenshot 2026-09-03 at 05 51 56

### Notes on Upstream Behavior
- **IDE Version Bump (`2.1.1` → `2.11.0`)**: Google's Antigravity backend (`daily-cloudcode-pa`) gates 3.8 models by client version. Calls with the old `2.1.1` User-Agent return `404 NOT_FOUND`. Updating `ANTIGRAVITY_IDE_VERSION` to `2.11.0` resolves this.
- **Model IDs**: Unlike 3.6/3.7, Google does not use a `-tiered` slug for 3.8. The upstream endpoints expect `gemini-3.8-flash-high`, `gemini-3.8-flash-medium`, or `gemini-3.8-flash-low` directly.
- **Quota Discovery**: `fetchAvailableModels` on `daily-cloudcode-pa` serves the 3.8 models (the standard `cloudcode-pa` host does not expose them yet). Pointed `antigravity.usage.quotaApiUrl` to `ANTIGRAVITY_IDE_BASE_URL`.

### Changes
- Registered `gemini-3.8-flash-{high,medium,low}` in Antigravity provider registry, with `gemini-3.8-flash` aliasing to `medium`.
- Registered `gemini-3.8-flash` under direct Gemini provider.
- Added capabilities (1M context / 64k output) and pricing ($1.50 in / $7.50 out).
- Added 3.8 Flash models to MITM synonyms, tool configs, and CLI menus.
- Added 3.8 tiers to quota tracking (`importantModels`) and switched `quotaApiUrl` to daily host.
- Added unit and integration tests covering registration, thinking tier mapping, and quota checks.

### Testing
- `vitest run tests/unit/*3.8*`: All 15 tests pass.
- Verified live requests against `daily-cloudcode-pa.googleapis.com` with an Antigravity account (all three tiers return `200 OK` with thinking tokens and live quota tracking).

PR #3737, opened 2026-09-02
fix(antigravity): update IDE version and Gemini mappings

emi-ran:
## Summary

- Update Antigravity IDE client markers from stale `2.1.1`/`1.23.2` values to the current IDE version `2.5.5`.
- Add Gemini 3.8 Flash tier mappings to the Antigravity registry using tiered upstream IDs.
- Add regression coverage for public-to-upstream model mapping and update version assertions.

## Root cause

Antigravity IDE 2.5.5 exposes Gemini 3.8 models, while 9Router identified requests with older client-version markers and lacked registry mappings for the new public IDs. Requests could reach Antigravity with an unsupported model identity and return HTTP 404.

## Validation

- `tests/unit/antigravity-gemini-38-model-mapping.test.js` — 3 tests passed.
- `tests/unit/antigravity-retry-hook.test.js` — 12 tests passed.
- `tests/unit/antigravity-usage-headers.test.js` — 1 test passed.
- `git diff --check` passed.

Live Antigravity account requests are not included because credentials are environment-specific.

### DIFF
--- a/open-sse/providers/registry/antigravity.js
+++ b/open-sse/providers/registry/antigravity.js
@@ -45,6 +45,9 @@ export default {
     clientSecret: "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",
   },
   models: [
+    { id: "gemini-3.8-flash-high", name: "Gemini 3.8 Flash (High)", upstreamModelId: "gemini-3.8-flash-high(high)" },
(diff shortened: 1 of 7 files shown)


decolua:
Thanks @emi-ran for the contribution! Merged into master as co-author alongside #3732. 🙏

PR #3751, opened 2026-09-03
Update pricing.js

jalallinux:
add pattern: "claude-fable-*" to PATTERN_PRICING


Primexz:
#4068 this data should be correct.

PR #3757, opened 2026-09-03
feat(providers): add Meta AI (Muse Spark) via the Meta Model API

dasepmoch:
## What
Adds a first-class **Meta AI** provider (`api.meta.ai`, OpenAI-compatible) for Muse Spark reasoning models.

- New registry entry `open-sse/providers/registry/meta.js` (base URL, model list, `thinkingFormat: "meta"`, provider icon).
- Models: `muse-spark-1.1`, `muse-spark-1.2`, `muse-spark-1.2-contributor`, `muse-spark-1.3`.
- Capabilities (`capabilities.js`): Muse Spark always reasons, cannot disable (`thinkingCanDisable: false`), no `max`; a provider-scoped guard resolves any `muse-spark*` id — including suffixed and passthrough-fetched ids — to the `meta` format.
- Thinking: new `meta` thinking format (`thinkingLevels.js` + `thinkingUnified.js` `applyFormat` case) mapping `minimal/low/medium/high/xhigh`, clamping `max`/`ultra` → `xhigh`, and omitting `none` (which Meta rejects).
- **Reasoning level selectable via a dash model-id suffix** (`meta/muse-spark-1.3-xhigh`), handled by a dedicated `MetaExecutor` that strips the suffix for the upstream id (`muse-spark-1.3`) and sets `reasoning_effort`. The parenthesized `model(level)` form is still accepted as a fallback. The dashboard renders the level picker with the dash suffix for the meta provider only.
- Pricing (`pricing.js`) + provider icon (`public/providers/meta.png`).
- Regression baselines refreshed for the new provider + alias.

## Reasoning levels
`none` is rejected by the upstream (HTTP 400); accepted set is `minimal | low | medium | high | xhigh`. `max`/`ultra` clamp to `xhigh`.

## Tests
`tests/unit/meta-ai.test.js` (registry, capabilities, thinking mapping, dash-suffix executor) + existing `opencode-muse-spark-thinking.test.js` both pass; all three baseline verifiers (`verify-providers`, `verify-alias`, `verify-oauth-urls`) pass.


Twilight0:
@dasepmoch Can you also add support for muse image with a new commit and end up in the "media providers" section?

Screenshot_20260922_191221

PR #3767, opened 2026-09-03
feat(usage): surface Antigravity weekly quotas

Recoba86:
## Summary

Adds family-level Antigravity weekly quota visibility alongside the existing
per-model quota data.

Google exposes these limits through a separate Cloud Code RPC:

```text
v1internal:retrieveUserQuotaSummary
```

The response includes weekly buckets for:

* `Gemini Models`
* `Claude and GPT models`

9Router exposes them as:

* `gemini_weekly` → `Gemini Weekly`
* `claude_gpt_weekly` → `Claude & GPT Weekly`

The existing Provider Limits UI already renders quota entries generically, so
no frontend-specific change is required.

## Motivation

The existing Antigravity usage path reads per-model quota information from
`fetchAvailableModels`, which reflects the shorter quota window.

An account can therefore appear available in the existing quota view while its
family-level weekly pool is exhausted.

Displaying the weekly pools separately makes both limits visible without
changing routing or credential selection.

## Implementation

This PR:

1. Adds `quotaSummaryApiUrl` to the Antigravity usage transport config.

2. Adds a small weekly quota helper that requests:

```text
v1internal:retrieveUserQuotaSummary
```

3. Supports both:

```js
{ groups: [...] }
```

and the defensive alternate shape:

```js
{ quotaSummary: { groups: [...] } }
```

4. Uses:

```text
window: "weekly"
```

as the primary weekly bucket discriminator, with a conservative
`weekly` identifier fallback.

5. Maps:

```text
Gemini Models
-> gemini_weekly
-> Gemini Weekly

Claude and GPT models
-> claude_gpt_weekly
-> Claude & GPT Weekly
```

6. Merges these entries into the existing Antigravity `quotas` object.

## Response consumption ordering

`fetchAvailableModels` can return a relatively large response.

The main response body is therefore consumed immediately after its HTTP status
is checked, while the weekly summary request runs concurrently:

```js
const [data, weeklyQuotas] = await Promise.all([
  response.json(),
  fetchWeeklyQuotas(),
]);
```

This avoids delaying consumption of the existing quota response while waiting
for the additional network request.

Regression coverage includes concurrent multi-account usage refreshes.

## Failure behavior

The weekly quota request is best-effort and fail-open.

Failures such as:

* network errors
* 404 / 429 / 5xx
* malformed JSON
* unexpected response shapes

produce no weekly rows while preserving the existing per-model quota result.

No routing behavior changes.

## Caching

Weekly results use a small process-local cache to avoid repeatedly requesting
the quota summary during dashboard refreshes.

There is no database persistence or migration.

## Tests

Coverage includes:

* weekly group parsing
* explicit `window: "weekly"` detection
* compatibility fallback detection
* alternate response envelope
* missing quota families
* disabled weekly buckets
* malformed input
* remaining-fraction bounds
* reset-time parsing
* cache behavior
* HTTP/network fail-open behavior
* malformed JSON fail-open behavior
* main-response consumption ordering
* delayed response-consumption regression
* concurrent multi-account Antigravity usage

Relevant existing Antigravity and Gemini usage regressions also pass.

## Validation

Validated through the normal 9Router usage path with multiple real Antigravity
OAuth connections.

Both sequential and concurrent account refreshes completed successfully while:

* existing per-model quota rows remained intact
* both weekly family rows were returned
* no duplicate 5-hour family-summary rows appeared
* Antigravity Gemini inference remained unaffected

No account identifiers, credentials, or private infrastructure information are
included here.

## Scope

This PR intentionally does not:

* change credential selection
* change account fallback
* use weekly quota for routing decisions
* change Antigravity circuit-breaker behavior
* modify `/v1/models`
* modify Model Catalog Sync
* add a database migration
* add frontend-specific code
* modify the shared network transport

It only extends the existing Antigravity usage data with weekly quota
visibility.

## Prior art

The separate weekly quota RPC and family-level quota concept have also been
observed in OmniRoute:

* diegosouzapw/OmniRoute#4017
* diegosouzapw/OmniRoute#6818

This implementation is adapted to 9Router's existing usage transport,
proxy-aware request path, quota data shape, caching, and fail-open behavior.

### DIFF
--- a/open-sse/providers/registry/antigravity.js
+++ b/open-sse/providers/registry/antigravity.js
@@ -37,6 +37,7 @@ export default {
     },
     usage: {
       quotaApiUrl: `${ANTIGRAVITY_IDE_BASE_URL}/v1internal:fetchAvailableModels`,
+      quotaSummaryApiUrl: `${ANTIGRAVITY_IDE_BASE_URL}/v1internal:retrieveUserQuotaSummary`,
       loadProjectApiUrl: "https://cloudcode-pa.googleapis.com/v1internal:loadCodeAssist",
       tokenUrl: "https://oauth2.googleapis.com/token",
     },
--- a/open-sse/services/usage.js
+++ b/open-sse/services/usage.js
@@ -35,7 +35,7 @@ import {
 const USAGE_HANDLERS = {
   github: (c) => getGitHubUsage(c.accessToken, c.providerSpecificData, c.proxyOptions),
   "gemini-cli": (c) => getGeminiUsage(c.accessToken, c.providerDataWithProjectId, c.proxyOptions),
-  antigravity: (c) => getAntigravityUsage(c.accessToken, c.providerSpecificData, c.proxyOptions),
+  antigravity: (c) => getAntigravityUsage(c.accessToken, c.providerSpecificData, c.proxyOptions, { force: c.force }),
   claude: (c) => getClaudeUsage(c.accessToken, c.proxyOptions, { force: c.force }),
   codex: (c) => getCodexUsage(c.accessToken, c.proxyOptions),
   kiro: (c) => getKiroUsage(c.accessToken, c.providerSpecificData, c.proxyOptions),
--- a/open-sse/services/usage/antigravityWeeklyQuota.js
+++ b/open-sse/services/usage/antigravityWeeklyQuota.js
@@ -6,12 +6,37 @@
  * Best-effort, fail-open, and non-blocking.
  */
 
+import crypto from "crypto";
 import { ANTIGRAVITY_IDE_BASE_URL, ANTIGRAVITY_IDE_USER_AGENT, ANTIGRAVITY_IDE_VERSION } from "../../providers/shared.js";
 import { U, parseResetTime, fetchWithTimeout } from "./shared.js";
 
 const WEEKLY_QUOTA_TTL_MS = 60 * 1000;
 const weeklyQuotaCache = new Map();
 const inflightRequests = new Map();
+
+/**
+ * Generate a collision-resistant, leak-free cache key from projectId and full accessToken hash.
+ *
+ * @param {string} projectId
+ * @param {string} accessToken
+ * @returns {string}
+ */
+export function getWeeklyCacheKey(projectId, accessToken) {
+  const tokenHash = crypto
+    .createHash("sha256")
+    .update(String(accessToken || ""))
+    .digest("hex")
+    .slice(0, 16);
+  return `${projectId}:${tokenHash}`;
+}
+
+/**
+ * Reset in-memory cache (primarily for unit tests).
+ */
+export function _resetWeeklyQuotaCacheForTesting() {
+  weeklyQuotaCache.clear();
+  inflightRequests.clear();
+}
 
 /**
  * Parse raw retrieveUserQuotaSummary response into family-level weekly quotas.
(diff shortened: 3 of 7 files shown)


Recoba86:
image

vuuminhphuoc (commented):
Good direction - the upstream `retrieveUserQuotaSummary` RPC with per-family `remainingFraction` is a real signal, and namespacing weekly keys as `gemini_weekly`/`claude_gpt_weekly` avoids colliding with per-model quota keys. Three issues before merge:

1. **Fabricated absolute numbers.** `total = 1000; // Normalized base matching 9Router convention` invents a denominator upstream never provides. Dashboard will render e.g. `used 14/1000` when the real limit could be 50 or 5000 requests. Either expose only `remainingFraction`/`remainingPercentage`, or label the UI explicitly as percentage-based so users don't treat 1000 as a real quota.
2. **Cache key uses raw token prefix** (`projectId:token.slice(0,16)`). Two connections sharing a project with same-prefix tokens collide; also leaks token material into keys. Hash the full token (sha256, truncate) or key by `connectionId`.
3. **Stale cache on exhaustion.** 60s TTL with no bust path means a user who just hit the wall still sees green for up to a minute. Consider bypassing cache (`options.force` already exists) when the sibling per-model quota fetch sees a 429/empty, or when `remainingFraction` parses to 0 on a fresh fetch.

Verified against the diff; happy to re-review.

Recoba86:
Thanks for the thoughtful review and constructive feedback @vuuminhphuoc!

All three points have been addressed in commit `fcdae752`:

1. **Normalized percentage scale instead of fabricated 1000 base**:
   - Replaced the arbitrary `total = 1000` base with a normalized 0..100 percentage scale (`total = 100, used = 100 - remainingPct`), matching 9Router's `ratioQuota` convention (as seen in `open-sse/services/usage/misc.js`).
   - Also explicitly exposes `remainingFraction` and `remainingPercentage` on each weekly quota entry so callers can read raw percentages directly without mistaking the denominator for a literal request count.

2. **SHA-256 cache keying**:
   - Replaced the raw 16-character token prefix with `crypto.createHash("sha256").update(accessToken).digest("hex").slice(0, 16)`.
   - Prevents token material leakage and provides full 64-bit cryptographic entropy, eliminating collisions between distinct tokens sharing a project.

3. **Cache busting on force & exhaustion**:
   - Wired `options.force` (from `/api/usage/[id]?force=1`) through `USAGE_HANDLERS.antigravity` into `fetchAndParseAntigravityWeeklyQuotas` to immediately bust cache on manual refresh.
   - When any weekly quota reports exhaustion (`remainingPercentage <= 0`), the entry is immediately evicted from cache and never retained for the 60s TTL, ensuring stale green states are never shown to users.

4. **Rebased on latest upstream master (`v0.5.69`)**:
   - Cleanly rebased onto current `master` with zero conflicts; all 36 unit tests pass.

Ready for re-review!

vuuminhphuoc (approved):
Re-reviewed the updated diff - all three points are addressed and verified in the final state:

1. `total = 1000` -> `total = 100` with `used = 100 - remaining%`, plus raw `remainingFraction` exposed on each entry. No more fabricated request counts.
2. Cache key now `getWeeklyCacheKey()` via sha256 hex - no raw token material, no prefix collisions.
3. `options.force` wired from `/api/usage/[id]?force=1` through `USAGE_HANDLERS.antigravity`, and exhausted quotas (`remainingPercentage <= 0`) are never cached, so no stale green state.

One nit (non-blocking): exhausted quotas are now re-fetched from upstream on *every* poll since nothing is cached. Consider a short negative TTL (~5s) to avoid hammering the quota RPC when a user sits on an exhausted dashboard. Otherwise LGTM.

PR #3771, opened 2026-09-04
fix(errors): preserve the upstream status class instead of flattening it to 503

haumanto:
Three related defects in how upstream failures are reported to clients. They have to land together — each one alone is incomplete, and the third exists specifically to stop the first two going too far.

## 1. Every exhausted-account response is 503, including permanent ones

`chat.js` hardcodes `HTTP_STATUS.SERVICE_UNAVAILABLE` when all accounts are cooling down. But `checkFallbackError` cools an account for 30s on **any** error it has no rule for, client-fault 400s included, so this path is not only transient conditions.

Concrete: a client sends an over-long prompt, upstream answers `400 prompt is too long`, the account is locked, the client gets `503 + Retry-After: 30`. It waits, resends the identical prompt, gets 503 again. Forever. The prompt is never compacted, and the account stays cooled for a fault that was never the accounts.

`clientStatusForUpstream()` preserves the class instead: 4xx still means stop, 5xx still means retry.

Wrong-model errors are the one deliberate re-mapping. Providers report them as 400, as 404, and as **401 with a `ModelError` body** — passing that 401 through makes a client report "authentication failed" for perfectly good credentials, so they are normalised to 404.

## 2. A wrong model name cools down every account

`checkFallbackError` returns `shouldFallback: true` on every path, so one typo walks every account into cooldown and then answers "try again later" for something retrying can never fix.

Request-scoped failures are now marked `permanent`: they return the upstream status straight to the caller and leave account state untouched. Matching is on **text, not status**, because a bad model arrives as 400 as often as 404, and because one provider wraps both transient socket errors and permanent parameter errors in the same `Upstream request failed:` prefix — the bracketed class after that prefix is what separates them.

## 3. One requests error is replayed onto another

`lastError`/`errorCode` are flat **per-connection** fields, while model locks are per-model, so the stored error belongs to whichever model failed on that connection most recently. Serving it without checking leaks one requests failure into an unrelated one.

Observed: a probe for a bogus model produced a `401 ModelError`, which was then replayed to a live conversation **on a different model** for the whole backoff window, making good credentials look broken. On a shared router that is a cross-tenant fault. `lastErrorModel` now records ownership, and a stored error is only surfaced when it provably belongs to the model being requested.

This one is why 1 and 2 are safe. Preserving the upstream class without it would make the replay worse, not better — it would replay the *status* too.

## The guard rail: `clientStatusForBreakerOpen`

Preserving the class has a failure mode of its own, which we hit in production. A bodyless `404` from a gateway hop is not a model error, so it cools the account down for two minutes — and then the preserved 404 tells the client "this will never work" about a state that clears itself:

```
404  [codex/gpt-5.6-luna] [404]: HTTP 404 (reset after 1m 58s)
```

The model was in `/v1/models` the whole time and served the request two minutes later. A client classifying on status burned three retries in six seconds against a two-minute breaker.

So the breaker path remaps **exactly one** status — a 404 whose text does not name a model problem — and delegates everything else to `clientStatusForUpstream`. Deliberately not "we locked it, therefore its transient, therefore 503": that is defect 1 in reverse. `Retry-After` was already emitted and is now pinned by test alongside the status.

## Verification

Driven through the real functions rather than mocks:

| case | breaker status | cools account? |
|---|---|---|
| bodyless 404 cooldown | 503 | yes, 120s |
| `prompt is too long` (400) | **400** | yes, 30s |
| rejected parameter (400) | **400** | yes, 30s |
| wrong model name | 404 | **no — permanent** |
| `ModelError` on a 401 | 404 | **no — permanent** |
| upstream 500 | 500 | yes, 30s |

19 tests in `tests/unit/breaker-open-status.test.js`, shown non-vacuous in both directions by mutation: reverting the breaker remap fails 2, and widening it to a blanket 503 fails 12.

The suite is not green on a clean checkout, so I compared against a stashed clean tree in the same checkout: **237 failures before and after**, with only the 19 new tests added to the pass count.

https://claude.ai/code/session_01VnrdG9vY3xmA5PBB8gPPtM

### DIFF
--- a/open-sse/config/errorConfig.js
+++ b/open-sse/config/errorConfig.js
@@ -57,6 +57,22 @@ const COOLDOWN = {
  *   - backoff: true = use exponential backoff (rate limit)
  */
 export const ERROR_RULES = [
+  // --- Permanent, request-scoped failures (highest priority) ---
+  // The model name itself is wrong, so no other account can do better. Without
+  // these the default "fall back and cool down" applied: one typo walked every
+  // account into cooldown and then answered 503 "try again later" for something
+  // retrying can never fix. `permanent` returns the upstream 4xx straight to the
+  // caller and leaves account state untouched.
+  // Matched on text, not status, because providers report this as 400 as often
+  // as 404 — the observed message is "The 'x' model is not supported when using
+  // Codex with a ChatGPT account." on a 400.
+  { text: "model is not supported",   permanent: true },
+  { text: "model not supported",      permanent: true },
+  { text: "model not found",          permanent: true },
+  { text: "model does not exist",     permanent: true },
+  { text: "unknown model",            permanent: true },
+  { text: "invalid model",            permanent: true },
+
   // --- Text-based rules (checked first, order = priority) ---
   { text: "no credentials",           cooldownMs: COOLDOWN.long },
   { text: "request not allowed",      cooldownMs: COOLDOWN.short },
--- a/open-sse/config/errorConfig.js
+++ b/open-sse/config/errorConfig.js
@@ -56,22 +56,76 @@ const COOLDOWN = {
  *   - cooldownMs: fixed cooldown duration
  *   - backoff: true = use exponential backoff (rate limit)
  */
+/**
+ * Phrases that identify a permanently wrong model name, whatever status the
+ * provider chose to attach to it. Providers are wildly inconsistent here: the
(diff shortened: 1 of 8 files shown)


haumanto:
Supersedes #3221 (now closed), which carried defects 1-3 without the breaker guard rail, the combo path, or tests.

PR #3775, opened 2026-09-04
fix(proxy-pools): strip client-IP headers in relay forwarders

anojndr:
## Root cause

OpenCode free-tier 429 (`FreeUsageLimitError`) is a per-IP limiter. All three relay templates forwarded every inbound header upstream, including the platform edge's `x-forwarded-for` / `cf-connecting-ip` carrying the origin IP. opencode.ai (Cloudflare-fronted) recovers the origin IP from `X-Forwarded-For`, so Deno/Cloudflare relays leaked the real IP and the limit survived relaying. Only Vercel relays worked because Vercel's edge strips `X-Forwarded-For`/`X-Real-IP` on outbound fetch at the platform layer.

## Changes

- **`src/lib/network/relayHeaders.js` (new)** — single source of truth: exact + prefix strip lists, `sanitizeRelayHeaders()`, and `buildRelaySanitizeSnippet()` emitting equivalent JS for the remotely-deployed runtimes.
- **Vercel / Cloudflare / Deno deploy routes** — relay templates embed and call the snippet; POST body + `duplex: half` streaming preserved; `authorization`, `user-agent`, `x-opencode-*` untouched.
- **`proxy-pools/[id]/test`** — informational XFF-canary anonymity probe (`ipLeak` + redeploy `warning`); never flips liveness.
- **`tests/unit/relay-header-sanitize.test.js`** — 5 tests covering lists, local sanitize, template embedding, and snippet validity.

## Verification

- New test 5/5 pass; `node --check` parses all route files and all three interpolated relay payloads; old-vs-new demo proves leak then anonymity with opencode headers intact.
- eslint clean on touched files; `base-executor-retry` + `mimo-free` 35/35 pass (`executor-const-guard` antigravity-429 failure is pre-existing on clean tree).
- Code review verdict: correct (0.9), no blocking issues.

## Note

Already-deployed Deno/Cloudflare relays still run the old forwarder and must be **redeployed**; the pool test route now flags them with `ipLeak: true`.

### DIFF
--- a/src/app/api/proxy-pools/[id]/test/route.js
+++ b/src/app/api/proxy-pools/[id]/test/route.js
@@ -33,6 +33,35 @@ async function testVercelRelay(relayUrl, timeoutMs = 10000) {
   }
 }
 
+// Anonymity probe: a relay must not forward client-IP headers. Sends a canary
+// X-Forwarded-For through the relay to httpbin's header echo; if the canary
+// comes back, the relay leaks the origin IP and per-IP upstream rate limits
+// (e.g. OpenCode free-tier 429) survive relaying. Old Cloudflare/Deno relay
+// code forwards it; Vercel's edge strips it at the platform layer.
+async function testRelayIpLeak(relayUrl, timeoutMs = 10000) {
+  const canary = `9router-canary-${Date.now().toString(36)}`;
+  const controller = new AbortController();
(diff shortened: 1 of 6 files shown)


anojndr:
Closing — after a few more hours of testing, this doesn't actually fix the issue. I think it cannot be fixed; the cause is upstream's IP bucket applied to shared egress.

PR #3778, opened 2026-09-04
Fix Codex reset-credit errors

IEatCodeDaily:
## Summary

- preserve readable Codex reset-credit API errors when the upstream `error` field is an object
- prefer the nested error message and JSON-serialize other structured errors instead of displaying `[object Object]`

## Reproduction

The reset-credit endpoint can return a body such as:

```json
{"error":{"message":"Connector rate limit exceeded"}}
```

Passing that object directly to `new Error()` produces `[object Object]`.

## Verification

- added a regression test for structured upstream errors
- observed the test fail with `Received: "[object Object]"` before the fix
- focused Codex reset-credit tests pass after the fix
- live local endpoint now returns `{"error":"Connector rate limit exceeded"}` instead of `{"error":"[object Object]"}`

### DIFF
--- a/open-sse/services/usage/codex.js
+++ b/open-sse/services/usage/codex.js
@@ -21,6 +21,13 @@ function toIsoDate(value) {
   return Number.isFinite(time) ? date.toISOString() : null;
 }
 
+function errorMessage(value, fallback) {
+  if (!value) return fallback;
+  if (typeof value === "string") return value;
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @IEatCodeDaily for the contribution! Reviewed and merged into master. 🙏

PR #3783, opened 2026-09-04
fix(security): guard cowork-mcp-tools probe against SSRF

soroush5:
## Summary

`POST /api/cli-tools/cowork-mcp-tools` passed its `url` straight into three server-side fetches with no SSRF guard, so any caller could make the gateway reach loopback/internal URLs and read back the "tool list". Same `assertPublicUrl` + local-requester exception the sibling provider-nodes route already uses, so self-hosted MCP servers keep working.

## How to test

1. POST `{ "url": "http://127.0.0.1:18731/internal-admin" }` from a non-local origin.
2. Observe 400 `URL not allowed` with no outbound fetch now. Before the fix, the internal server saw all 3 JSON-RPC POSTs and its tool list came back.
3. Run `vitest run unit/cowork-mcp-ssrf-guard.test.js unit/ssrf-guard-hardening.test.js unit/search-ssrf-guard.test.js` in `tests/`. All 26 pass, including the 3 new regression tests (2 red without the fix, green with it).

### DIFF
--- a/src/app/api/cli-tools/cowork-mcp-tools/route.js
+++ b/src/app/api/cli-tools/cowork-mcp-tools/route.js
@@ -1,6 +1,8 @@
 "use server";
 
 import { NextResponse } from "next/server";
+import { assertPublicUrl } from "@/shared/utils/ssrfGuard.js";
+import { isLocalRequest } from "@/dashboardGuard";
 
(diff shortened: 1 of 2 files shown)


decolua:
Thanks @soroush5 for the contribution! Reviewed and merged into master. 🙏

PR #3790, opened 2026-09-04
fix(translator): sum Gemini non-streaming total when absent

### LINKED ISSUE #3789: Non-streaming Gemini usage reports total_tokens 0 when totalTokenCount absent
## Summary

The non-streaming Gemini→OpenAI path reports an internally inconsistent usage block when the upstream response has no `totalTokenCount`: `prompt_tokens` and `completion_tokens` are filled, but `total_tokens` is `0`.

Reproduced with the exact mapping logic: `{promptTokenCount: 10, candidatesTokenCount: 20}` (no `totalTokenCount`) yields `{prompt_tokens: 10, completion_tokens: 20, total_tokens: 0}`.

## Root cause

`nonStreamingHandler.js` computes `total_tokens: usage.totalTokenCount || 0` while the sibling branches in the same file sum the parts (`:136` uses `total || prompt + completion`). The streaming normalizer (`usageTracking.js`) also recomputes the total from prompt+completion, so the non-streaming Gemini branch is the only inconsistent path. With `totalTokenCount` present the output is correct.

## Expected

Fall back to the sum of the derived parts (including `thoughtsTokenCount`, which prompt already includes), same as the siblings.

soroush5:
Closes #3789.

The non-streaming Gemini path reported `{prompt_tokens: 10, completion_tokens: 20, total_tokens: 0}` when the upstream response has no `totalTokenCount` — internally inconsistent, and the only branch in the file that does this (the siblings and the streaming normalizer all sum the parts). Now falls back to the sum of the derived parts, including `thoughtsTokenCount` which the prompt count already includes. An explicit `totalTokenCount` still wins.

How to test:

1. Translate a Gemini non-streaming body with `usageMetadata` lacking `totalTokenCount` toward an OpenAI-shaped target — `total_tokens` now equals prompt + completion.
2. Run `vitest run unit/gemini-nonstream-usage-total.test.js` in `tests/` — all 3 pass (the 2 fallback cases fail without the fix).

### DIFF
--- a/open-sse/handlers/chatCore/nonStreamingHandler.js
+++ b/open-sse/handlers/chatCore/nonStreamingHandler.js
@@ -202,10 +202,13 @@ export function translateNonStreamingResponse(responseBody, targetFormat, source
     };
 
     if (usage) {
+      const promptTokens =
+        (usage.promptTokenCount || 0) + (usage.thoughtsTokenCount || 0);
+      const completionTokens = usage.candidatesTokenCount || 0;
       result.usage = {
-        prompt_tokens: (usage.promptTokenCount || 0) + (usage.thoughtsTokenCount || 0),
-        completion_tokens: usage.candidatesTokenCount || 0,
-        total_tokens: usage.totalTokenCount || 0
+        prompt_tokens: promptTokens,
+        completion_tokens: completionTokens,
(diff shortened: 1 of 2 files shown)


soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

PR #3791, opened 2026-09-04
fix(usage): track OpenCode Go quota

callmeusdt:
## Summary

OpenCode Go API-key connections now appear in the Quota Tracker and report rolling, weekly, and monthly subscription usage. Previously, the provider could serve model requests but was filtered out of the quota dashboard because its registry entry did not declare usage support.

The usage adapter reads the provider's existing `/zen/go/v1/usage` endpoint, normalizes consumed percentages and reset times, and distinguishes invalid credentials from a missing subscription entitlement. Malformed quota windows are ignored instead of being displayed as fully available.

## Test Plan

- `npx vitest run unit/opencode-go-usage.test.js unit/usage-dispatch.test.js --testTimeout=20000`
- `node tests/__baseline__/verify-providers.mjs`
- `node tests/__baseline__/verify-alias.mjs`
- `npx eslint open-sse/providers/registry/opencode-go.js open-sse/services/usage.js open-sse/services/usage/opencode-go.js tests/unit/opencode-go-usage.test.js tests/unit/usage-dispatch.test.js`

Also verified against a live OpenCode Go subscription: the upstream endpoint returned all three quota windows and the local Quota Tracker API included the API-key connection.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
OpenCode

### DIFF
--- a/open-sse/providers/registry/opencode-go.js
+++ b/open-sse/providers/registry/opencode-go.js
@@ -21,6 +21,9 @@ export default {
   transport: {
     baseUrl: "https://opencode.ai/zen/go/v1/chat/completions",
     headers: {},
+    usage: {
+      url: "https://opencode.ai/zen/go/v1/usage",
+    },
   },
   // Multi-endpoint: pick the transport matching the client sourceFormat to skip
   // translation. Guarded per-model by `supportedFormats` (see chatCore) because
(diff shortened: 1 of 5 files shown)


decolua:
Thanks @callmeusdt for the contribution! Reviewed and merged into master. 🙏

Issue #2894, opened 2026-07-29
[Bug] Adaptive thinking without an explicit effort always 400s on /v1/messages: claude-budget emits `enabled` with no `budget_tokens`, claude-adaptive emits `effort: "auto"`


wangsijie:
## Environment

- 9Router version: **v0.5.40** (Docker `decolua/9router:latest`, rev `79918c7`)
- Affected file: `open-sse/translator/concerns/thinkingUnified.js` (`applyFormat`), plus `open-sse/providers/capabilities.js`
- Affected route: `POST /v1/messages` passthrough to the `claude` provider (Claude Code OAuth)
- Trigger: any client that sends **adaptive thinking without an explicit effort** — e.g. OpenCode / Vercel AI SDK `@ai-sdk/anthropic`, which emits `thinking: { type: "adaptive" }` for modern Claude models

## Describe the bug

Two independent bugs in the unified thinking normalization make `/v1/messages` return `400` from Anthropic. They stack: fixing only the first one exposes the second.

### Bug 1 — `claude-budget` in "auto" mode emits `enabled` without `budget_tokens`

`thinkingUnified.js:244-245`:

```js
const budget = toBudget(eff, caps.thinkingRange);
body.thinking = budget === -1 ? { type: "enabled" } : { type: "enabled", budget_tokens: budget || 8192 };
```

`toBudget()` returns `-1` for `{ mode: "auto" }`, which is exactly what `extractThinking()` produces for `thinking: { type: "adaptive" }` with no `budget_tokens`. But the Anthropic Messages API requires `budget_tokens` whenever `thinking.type === "enabled"`:

```
400 invalid_request_error — thinking.enabled.budget_tokens: Field required
```

So there is no input for which the `budget === -1` branch produces a valid Anthropic body. `normalizeClaudePassthrough()` already uses `10000` for its Haiku downgrade — the same value would work here.

(The `hunyuan` branch at `:296` has the identical shape, but Hunyuan accepts `enabled` without a budget, so it is fine as-is.)

### Bug 2 — `claude-adaptive` sends `output_config.effort: "auto"`

`thinkingUnified.js:237-239`:

```js
body.thinking = { type: "adaptive" };
const level = toLevel(eff);
body.output_config = { effort: level === "xhigh" ? "high" : level };
```

`toLevel()` returns the literal string `"auto"` for `{ mode: "auto" }`, and it is passed straight through:

```
[...]

PR #3792, opened 2026-09-04
fix(claude): normalize adaptive auto effort

### LINKED ISSUE #3786: [Bug] claude-adaptive writes effort:"auto" into output_config -> 400, and the 400 backs the account off so the next valid request also fails
## Summary

## The code

```js
case "claude-adaptive": {
  let a = o(j);
  break
}
```

```js
function o(a) {
    : null
}
```

## Reproduction

```bash
# 1) adaptive thinking -> 400

```

| # | request | result |
|---|---|---|
| 1 | plain | **200 OK** |

## Suggested fix

```js
```

## Also still open from #2894

```js
```

## Secondary suggestion

Jordannst:
## Summary

Fixes #3786.

Claude adaptive requests without an explicit effort are normalized to `output_config.effort: "high"` instead of forwarding the unsupported literal value `"auto"`.

## Problem

A client request such as:

```json
{
  "thinking": {
    "type": "adaptive"
  }
}
```

is captured as unified thinking mode `auto`. For models using the `claude-adaptive` format, 9router currently serializes that mode as:

```json
{
  "output_config": {
    "effort": "auto"
  }
}
```

Anthropic rejects this value with HTTP 400. That failed request can then contribute to connection backoff, causing a following valid request to appear unavailable.

## Root cause

`toLevel()` intentionally preserves unified auto mode as the string `"auto"`. The `claude-adaptive` branch only normalizes `"xhigh"` before assigning `output_config.effort`, so `"auto"` reaches the provider unchanged.

Other discrete-level branches already resolve auto mode to a supported concrete level.

## Change

Treat `"auto"` like the existing `"xhigh"` compatibility case in the `claude-adaptive` formatter and emit `"high"`.

This preserves the existing adaptive-thinking behavior:

- Claude models that support an explicit adaptive switch still receive `thinking: { "type": "adaptive" }`.
- Permanently adaptive models still omit the redundant `thinking` switch.
- Explicit supported effort levels are unchanged.

## Regression coverage

Added tests for both Claude adaptive capability paths:

1. `claude-opus-4.7` maps adaptive auto mode to `effort: "high"` and retains the adaptive switch.
2. `claude-fable-5-1` maps adaptive auto mode to `effort: "high"` without adding a redundant switch.

The first regression test fails on `upstream/master` with:

```text
expected { effort: 'auto' } to deeply equal { effort: 'high' }
```

## Validation

```text
thinking-unified.test.js: 62 passed
related applyThinking suites: 99 passed across 5 files
ESLint on changed files: passed
git diff --check: passed
```

The full translator suite has the same 11 unrelated Kiro, reasoning-bridge, and snapshot failures on both this branch and a clean `upstream/master` checkout.

## Scope

This PR only fixes adaptive auto-effort serialization. It does not change general HTTP 400 backoff policy or the separate `claude-budget` behavior discussed in #2894.

### DIFF
--- a/open-sse/translator/concerns/thinkingUnified.js
+++ b/open-sse/translator/concerns/thinkingUnified.js
@@ -246,7 +246,7 @@ function applyFormat(fmt, body, cfg, caps, supportedLevels) {
       if (canDisable) body.thinking = { type: "adaptive" };
       else delete body.thinking;
       const level = toLevel(eff);
-      body.output_config = { effort: level === "xhigh" ? "high" : level };
+      body.output_config = { effort: level === "xhigh" || level === "auto" ? "high" : level };
       break;
     }
     case "claude-budget": {
--- a/tests/translator/thinking-unified.test.js
+++ b/tests/translator/thinking-unified.test.js
@@ -82,6 +82,16 @@ describe("applyThinking per provider format", () => {
     // Sonnet 5). Both fields together are the documented adaptive shape.
     expect(out.thinking).toEqual({ type: "adaptive" });
   });
+  it("claude adaptive thinking maps auto effort to a supported level", () => {
+    const out = apply("claude", "claude-opus-4.7", { thinking: { type: "adaptive" } }, "claude");
+    expect(out.output_config).toEqual({ effort: "high" });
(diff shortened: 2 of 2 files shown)


decolua:

PR #3800, opened 2026-09-05
fix(opencode-go): send stable session header

### LINKED ISSUE #3759: Missing OpenCode Go Header
recieved this email today from opencode since i added it to 9router : 

```
Hey there,

Some of your requests to OpenCode Go are missing an x-opencode-session header. If we don't have this we cannot properly optimize our service. Starting 09/06 requests missing this header may error.

Here are your useragents that are missing this header:

Hermes
Hermes merged the fix (https://github.com/NousResearch/hermes-agent/pull/101864) — update Hermes to the latest release.

curl
Add x-opencode-session (one stable ID per conversation) to your requests.

Python OpenAI client
If this is Hermes, update past 0.21.0. Otherwise open an issue with whatever Python tool sends this, or add the header where you construct the client.

JavaScript OpenAI client
Pass headers: {'x-opencode-session': '<stable-id-per-conversation>'} where you build the client.

Thank you.
```

looks like its also in litellm todos : https://github.com/Mintplex-Labs/anything-llm/issues/6283

turingcat:
## Summary

- add a dedicated OpenCode Go executor that always sends `x-opencode-session`
- preserve a valid caller-provided native OpenCode session header
- translate downstream Agent session IDs into opaque, stable, Agent-scoped IDs
- forward the original provider session seed and client tool on both initial and credential-refresh requests

## Session policy

Downstream session identities are translated as:

```text
ses_<first 32 hex chars of sha256("opencode-go\0<agent>\0<session>"))>
```

This keeps one stable ID per conversation without leaking the raw downstream ID. The Agent namespace prevents Claude, Codex, Hermes, and other clients that reuse the same raw ID from colliding. Headerless requests retain the existing connection-stable fallback. Credentials are cloned per request so concurrent conversations cannot overwrite shared executor state.

The header is applied to OpenCode Go Chat Completions, Claude Messages, and Responses transports.

## Validation

- `npx vitest run tests/unit/opencode-go-session.test.js tests/unit/session-manager.test.js tests/unit/opencode-go-models.test.js tests/unit/opencode-muse-spark-thinking.test.js` - 48 passed
- `npm run build` - passed
- `npm run cli:pack` - passed
- packed CLI installed globally and `/api/health` returned HTTP 200

The full upstream unit suite is not claimed green because it currently contains unrelated baseline failures; this change is covered by focused regression and neighboring session tests.

## Scope

This is intentionally narrower than #3780: it fixes the OpenCode Go session-header requirement before September 6, 2026 without adding or changing model support.

Fixes #3759

### DIFF
--- a/docs/superpowers/plans/2026-09-04-opencode-go-session-header.md
+++ b/docs/superpowers/plans/2026-09-04-opencode-go-session-header.md
@@ -1,0 +1,261 @@
+# OpenCode Go Session Header Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Send a stable, conversation-scoped `x-opencode-session` header on every OpenCode Go request and install the patched CLI locally.
+
+**Architecture:** Add a dedicated `OpenCodeGoExecutor` extending `DefaultExecutor`. `chatCore` passes the provider-scoped session resolved from the original request plus the detected client tool; the executor derives a request-local upstream session and delegates all existing transport, authentication, retry, and proxy behavior to `DefaultExecutor`.
+
+**Tech Stack:** Node.js ESM, Vitest, Next.js, npm CLI packaging, GitHub CLI.
+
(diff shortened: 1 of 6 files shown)


decolua:
Thanks @turingcat for the contribution! Reviewed and merged into master. 🙏

PR #3802, opened 2026-09-05
feat(providers): align codebuddy-cn catalog/capabilities with server config

zmf963:
Sync codebuddy-cn against the product-config payload the plugin fetches from `copilot.tencent.com`, and add the missing `glm-5-turbo` id to the GLM registries.

## CodeBuddy CN registry

Removed models (13 remaining, matching the server list 1:1):

| Model | Reason |
|---|---|
| `hy4-preview-x` | Endpoint returns `11102` — "model [hy4-preview-x] service info not found" (paid tier delisted) |
| `glm-5.0-turbo`, `minimax-m2.7`, `kimi-k2.5`, `hy3-preview`, `deepseek-v3-2-volc` | Absent from the server product-config list |
| `hy3-x` | Paid tier, unused here |

> Note: the five "absent from list" models still return **200** when called
> directly — the gateway does not reject them. They are removed because the
> published catalog is the contract for the model picker; keeping them offers
> ids the vendor no longer advertises.

`glm-5.0` / `glm-4.7` capability entries are **kept** even though they are not in the catalog: they intercept the `PATTERN_CAPABILITIES` fallback. Deleting them would flip `thinkingFormat` from `openai` to `zai` and change `maxOutput` to 128000 — verified by resolving capabilities with and without the entries.

## capabilities.js

Values now come from the server payload (`maxInputTokens` / `maxOutputTokens` / `supportsImages`), with the server table winning over the plugin-baked fallback where the two disagree.

| Model | Field | Before | After |
|---|---|---|---|
| glm-5v-turbo | maxOutput | 38000 | **64000** |
| glm-5.3-flash | maxOutput | 48000 | **32000** |
| minimax-m3 | maxOutput | 48000 | **128000** |
| glm-5.3-flash | contextWindow | 200000 | **1000000** |
| kimi-k3-1 | contextWindow | 256000 | **1000000** |
| glm-5.1 | vision | (absent → false) | **true** |

**`thinkingCanDisable` semantics fix** — it must map to the server's `reasoning.canDisableThinking`, which is *not* the inverse of `onlyReasoning`:

- `onlyReasoning` = thinking is **on by default**
- `canDisableThinking` = it **can** be turned off

`glm-5.3` and `glm-5.3-flash` are `onlyReasoning:true` **but** `canDisableThinking:true`, so their thinking is switchable. The `hy*` models carry `canDisableThinking:false` and stay forced-on. The previous code conflated the two and marked glm-5.3 / glm-5.3-flash as non-switchable.

## thinkingLevels.js

- Added the missing `glm-5.2` set (`high`, `xhigh`)
- Documented that the chat endpoint **accepts any** `reasoning_effort` value (probed `none/minimal/low/medium/high/xhigh/max` → all HTTP 200), but values outside a model's `supportedEfforts` are silently clamped. The declared set therefore stays authoritative for the picker.

## glm.js / glm-cn.js

- Added `glm-5-turbo` — present in the GLM `/v4/models` listing and confirmed answering 200 with a model echo.

## Test plan

- [x] `npx vitest run tests/__baseline__/ tests/unit/capabilities.test.js` → 5 passed
- [x] Resolved every codebuddy-cn model through `getCapabilitiesForModel` / `getThinkingLevels` and diffed field-by-field against the server payload → 0 conflicts
- [x] Hot-patched a running production instance and verified `/api/health` plus live chat calls through the gateway
- [x] `hy3` (kept) still resolves to ctx 192000 / `openai` format / `low,high`; `hy3-x` (removed) now falls through to the pattern fallback as expected

Notes: `tests/unit/providers-status-filter.test.js` fails both with and without this change (vitest is missing the `@/app/...` alias) — pre-existing, unrelated. Local Dockerfile mirror edits are deliberately excluded from this PR.

### DIFF
--- a/open-sse/providers/capabilities.js
+++ b/open-sse/providers/capabilities.js
@@ -178,33 +178,37 @@ export const PROVIDER_CAPABILITIES = {
   // CodeBuddy.cn — authoritative per-model metadata from the gateway's model
   // config (contextWindow=maxInputTokens, maxOutput=maxOutputTokens, vision=
   // supportsImages). Every model reasons via OpenAI-style reasoning_effort
-  // (see registry thinkingFormat). `onlyReasoning` models can't turn thinking
-  // off → thinkingCanDisable:false (clamped to minimal instead of disabled).
+  // (see registry thinkingFormat). For thinkingCanDisable use the server's
+  // reasoning.canDisableThinking flag — see the note in the codebuddy-cn block
+  // below; it is NOT the inverse of onlyReasoning.
   "codebuddy-cn": {
-    "glm-5.2":            { reasoning: true, thinkingFormat: "openai", thinkingCanDisable: false, contextWindow: 1000000, maxOutput: 48000 },
-    "glm-5.1":            { reasoning: true, thinkingFormat: "openai", thinkingCanDisable: false, contextWindow: 200000, maxOutput: 48000 },
+    "glm-5.2":            { vision: true, reasoning: true, thinkingFormat: "openai", thinkingCanDisable: true, contextWindow: 1000000, maxOutput: 48000 },
+    "glm-5.1":            { vision: true, reasoning: true, thinkingFormat: "openai", thinkingCanDisable: false, contextWindow: 200000, maxOutput: 48000 },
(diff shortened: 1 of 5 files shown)


decolua:

PR #3806, opened 2026-09-05
feat(codex): add GPT-5.6 Sol, Terra, and Luna image aliases

anndev-69:
## What changed

Codex already lists GPT-5.6 Sol, Terra, and Luna for chat, but its built-in image catalog stops at GPT-5.5. Add `cx/gpt-5.6-sol-image`, `cx/gpt-5.6-terra-image`, and `cx/gpt-5.6-luna-image` so they appear in the image dashboard and `/v1/models/image`.

The entries use the existing image adapter: it removes `-image` and calls the corresponding base model with the `image_generation` tool. No executor, authentication, pricing, or chat model behavior changes. Add usage documentation and regression coverage for discovery, non-streaming generation, reference images, and streamed partial/final image events.

## Validation

Validated commit: `eb101c9d`.

```text
npm --prefix tests test -- unit/codex-image-models.test.js unit/image-generation.test.js unit/thinking-levels-gpt56-sol.test.js unit/model-routing.test.js

 Test Files  4 passed (4)
      Tests  37 passed (37)
```

Before adding the registry entries, the new test file produced `3 failed | 3 passed (6)`: all three discovery cases failed because the aliases were absent; the existing adapter already handled their requests.

`npm run build` and `git diff origin/master --check` passed. ESLint on the three changed JavaScript files exited 0 with the existing `import/no-anonymous-default-export` warning on the Codex registry's default export.

Ran the production standalone server on loopback with an isolated `DATA_DIR` and no provider credentials:

```text
curl -sS -o image-catalog.json -w 'HTTP %{http_code}\n' http://localhost:20148/v1/models/image
HTTP 200
```

The response includes all three new `cx/gpt-5.6-*-image` IDs and the existing GPT-5.5/5.4/5.3 image IDs. `GET /v1/models` also returned HTTP 200, retained GPT-5.6 chat/review models, and did not include the image aliases.

Browser check at `/dashboard/media-providers/image/codex` (1280×720): all three aliases rendered; selecting each updated the example request's model; Run displayed `No credentials for provider: codex` on the isolated instance. No browser console errors or warnings were captured. A dashboard screenshot is available on request.

## Before marking ready

This is a draft because upstream image generation has only been exercised with mocked SSE responses. A maintainer with an eligible ChatGPT account should generate and edit an image with each alias to confirm current account/model entitlement. These are local routing aliases, not a claim that every account can access the underlying models or their image tool.

Rollback: remove the three registry entries (or revert this commit). Existing image and chat aliases are unchanged.


vuuminhphuoc (commented):
The `-image` generation aliases look correct, and the test proving the Codex adapter can already build `input_image` content for the Responses API is exactly the evidence for a bigger adjacent gap:

**Chat-time vision input on Codex models still doesn't work.** Verified live on 0.5.65: sending an OpenAI-style `image_url` part in `/v1/chat/completions` to `cx/gpt-6-astra` never reaches the model - the request is silently served by a different model (first entry of the vision capacityAdapter list), confirmed via the response `model` field across the whole SSE stream. Text-only requests to the same model serve correctly.

Since this PR proves the adapter knows how to construct `input_image`, the natural follow-up is mapping chat-completion `image_url` parts to Responses-API `input_image` content on the Codex path, instead of the current silent cross-model substitution (which once returned a *wrong* answer from the substituted model - worse than an error). At minimum, custom-model `vision:true` caps that only apply to the images endpoint shouldn't imply chat vision support.

Happy to open a separate issue with repro details if useful.

decolua:
Thanks @anndev-69 for the contribution! Reviewed and merged into master. 🙏

vuuminhphuoc (commented):
Correction to my earlier review: re-tested chat vision on 0.5.69 just now (blue-square PNG -> cx/gpt-6-astra, one-word color question). Response served by model gpt-6-astra with the correct answer Blue - no cross-model substitution. Whatever fixed it between 0.5.65 and 0.5.69 (adapter now maps chat image_url to input_image, presumably), the gap I described is closed. The -image generation aliases in this PR plus working chat vision make the Codex vision story complete.

Issue #2461, opened 2026-07-08
[ERROR] [403]: Pls help me !


Duongkhanhtool:
[11:11:52] ℹ️ [TOKEN] Antigravity refreshed
[11:11:52] ℹ️ [TOKEN] ANTIGRAVITY | refreshed
[11:11:52] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"550c097a-6cf7-453b-9ab8-104d399068ca","success":true}
[ProjectId] Onboarding user with tier: standard-tier
[11:11:52] ℹ️ [TOKEN] Antigravity refreshed
[11:11:52] ℹ️ [TOKEN] ANTIGRAVITY | refreshed
[11:11:52] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"ab1c51de-3256-4213-ba1e-01582ecbbd2c","success":true}
[ProjectId] onboardUser attempt 1 failed: onboardUser done but no project_id in response, retrying...
[11:11:52] ℹ️ [TOKEN] Antigravity refreshed
[11:11:52] ℹ️ [TOKEN] ANTIGRAVITY | refreshed
[11:11:52] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"c73ac4c7-1ffe-4c66-a2a2-017f110fda62","success":true}
[11:11:53] [PENDING] END (ERROR) | provider=antigravity | model=claude-sonnet-4-6
[ERROR] [403]: ���RAN�0��V�m�SO A�6E���Pe���:�e;���߱S�".�)���x����*���mlKD�;?9u���7�oA�Tk*8��S���h�(H*� � �Y=���O�N'I�����^([-�̗�z�y6��T`0e��(���Mz�l��0��ꄈ:��DI�d�F9AՓ`-�c_-��n���yq9�g}T%jL=j(����c� >d�Ԏo�0�T�Z���ƕ���纩��A2��q��(6'�-���t�v��@����w�azi0w�E����aPޜ0LS��/�P6l�&�JC����b�����4rO&�������iP�1��;У���1�O\�Fm�B���
❌ antigravity [403]: [403]: ���RAN�0��V�m�SO A�6E���Pe���:�e;���߱S�".�)���x����*���mlKD�;?9u���7�oA�T
Pls help me !

qolbudr:
same issues here

Duongkhanhtool:
[...]

Issue #1059, opened 2026-05-12
API ERROR 403 CLAUDE CODE & ANTIGRAVITY PROVIDER


amineelhailaa:
<img width="449" height="186" alt="Image" src="https://github.com/user-attachments/assets/fd3d3e50-d7b2-4cc0-a742-b30ff3eef3d5" />

<img width="174" height="230" alt="Image" src="https://github.com/user-attachments/assets/c2fd01fe-194d-45ba-976c-417f00e15849" />

```
i cant use antigravity accounts quota in claude code whats the issue ? 
[ProjectId] Onboarding user with tier: standard-tier
[ProjectId] onboardUser attempt 1 failed: onboardUser done but no project_id in response, retrying...
[ProjectId] onboardUser attempt 2 failed: onboardUser done but no project_id in response, retrying...
[ProjectId] onboardUser attempt 3 failed: onboardUser done but no project_id in response, retrying...
[ProjectId] onboardUser attempt 4 failed: onboardUser done but no project_id in response, retrying...
[ProjectId] onboardUser failed after 5 attempts: onboardUser done but no project_id in response
[ProjectId] could not fetch projectId for connection 21232eb6
[15:58:03] 🔍 [FORMAT] claude → antigravity | stream=true
[15:58:03] [PENDING] START | provider=antigravity | model=claude-opus-4-6-thinking
[15:58:03] 🔍 [REQUEST] ANTIGRAVITY | claude-opus-4-6-thinking | 5 msgs
[15:58:05] ℹ️ [TOKEN] Antigravity refreshed
[15:58:05] ℹ️ [TOKEN] ANTIGRAVITY | refreshed
[15:58:05] ℹ️ [TOKEN_REFRESH] Credentials updated in localDb {"connectionId":"21232eb6-e6b2-4cb4-8601-1e1d33c696f4","success":true}
[15:58:07] [PENDING] END (ERROR) | provider=antigravity | model=claude-opus-4-6-thinking
[ERROR] [403]: HTTP 403
❌ antigravity [403]: [403]: HTTP 403
[15:58:07] 🔍 [AUTH] antigravity | total connections: 4, excludeIds: 49ca60c3-5369-42a0-99bb-391a09af5653,7617a99d-88ad-4cd4-848e-346155cdaa7e,f3f24eb6-4523-4049-b66d-e7351127cf26,21232eb6-e6b2-4cb4-8601-1e1d33c696f4, model: claude-opus-4-6-thinking
[...]

PR #3813, opened 2026-09-05
fix(antigravity): prevent Google anti-abuse rate limits on multi-account refresh

hifzi:
### Problem & Context
When running 9router with multiple Google accounts (Antigravity / Gemini-CLI), background token refreshes trigger a massive burst of requests to Google Cloud Code internal APIs from the same IP address. This causes Google's automated anti-abuse systems to shadow-restrict project allocation and can lead to account suspensions.

Users experience errors like this repeatedly:
```text
[ProjectId] Onboarding user with tier: standard-tier
[ProjectId] onboardUser attempt 1 failed: onboardUser done but no project_id in response, retrying...
...
[ProjectId] onboardUser failed after 5 attempts: onboardUser done but no project_id in response
[ProjectId] could not fetch projectId for connection 709681e5
```
Related issues: #1059, #2461.

### Root Cause
1. **Parallel Burst:** `backgroundTokenRefresh.js` used `Promise.allSettled()` to refresh all due accounts concurrently without pacing or throttling.
2. **Eager Project ID Fetch:** Every background token refresh immediately called `_refreshProjectId()`, bombarding Google's `onboardUser` endpoint across all accounts simultaneously.
3. **Aggressive Retry Loop:** `onboardUser` in `projectId.js` retried 5 times per account with only a 2-second sleep (`50 accounts × 5 retries = 250 requests` in seconds).

### Key Changes
- **Sequential Queue in Background Refresh:** Replaced `Promise.allSettled` with a sequential loop in `backgroundTokenRefresh.js`. Added configurable throttle delay with random jitter (`BG_REFRESH_GOOGLE_DELAY_MS`, default 12s + jitter for Google; `BG_REFRESH_DELAY_MS`, default 1.5s for others).
- **Lazy Project ID Resolution:** Background token refresh now only refreshes credentials and invalidates cache. Project IDs are lazily resolved on demand when incoming requests actually require them (runtime handler already handles this). Backwards compatibility preserved via `EAGER_PROJECT_ID_REFRESH=true`.
- **Tamed Retry & Backoff Jitter:** Lowered `MAX_ATTEMPTS` from 5 to 2 (configurable via `ONBOARD_MAX_ATTEMPTS`) and increased retry backoff delay with randomized jitter (`ONBOARD_RETRY_DELAY_MS`, default 12s + jitter).
- **Log Visibility:** Included account `email` / identifier in `Connection refresh finished` logs for easier monitoring.

### Verification
- Tested locally with 40+ active Google accounts.
- Background refresh now progresses steadily one account at a time without triggering burst detection.
- Chat completions route normally through accounts without onboarding failures.

### DIFF
--- a/open-sse/services/projectId.js
+++ b/open-sse/services/projectId.js
@@ -203,7 +203,8 @@ async function onboardUser(accessToken, tierID, externalSignal, endpoints, provi
 
     const reqBody = { tierId: tierID, metadata: LOAD_CODE_ASSIST_METADATA };
     const headers = provider === "antigravity" ? ANTIGRAVITY_LOAD_CODE_ASSIST_HEADERS : LOAD_CODE_ASSIST_HEADERS;
-    const MAX_ATTEMPTS = 5;
+    const MAX_ATTEMPTS = Number(process.env.ONBOARD_MAX_ATTEMPTS) || 2;
+    const BASE_RETRY_DELAY_MS = Number(process.env.ONBOARD_RETRY_DELAY_MS) || 12_000;
 
     for (let attempt = 1; attempt <= MAX_ATTEMPTS; attempt++) {
         // Bail out immediately if the connection was removed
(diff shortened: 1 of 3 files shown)


decolua:
Thanks @hifzi for the contribution! Reviewed and merged into master. 🙏

PR #3819, opened 2026-09-05
feat(opencode-go): add muse-spark-1.3-contributor + fix parallel tool-call merging on Responses paths

sinasadeghi83:
## Summary

This PR does two things:

1. **Adds `ocg/muse-spark-1.3-contributor`** (OpenCode Go) as a Responses-only model on `https://opencode.ai/zen/go/v1/responses`, so 9router clients on the OpenAI (`/v1/chat/completions`) and Anthropic (`/v1/messages`) endpoints are translated correctly to the OpenAI Responses API.
2. **Fixes parallel tool calls being merged into a single call** on the Responses→client streaming path, which surfaced to agents as `InputValidationError: ... could not be parsed as JSON` with N complete JSON payloads concatenated into one tool input.

## Why this is necessary

**Model support.** `opencode-go` had transports for chat/messages/responses but no Muse Spark entry, and `getModelTargetFormat` only special-cased `oc`/`opencode`. Without this, `ocg/muse-spark-1.3-contributor` requests never routed to `/responses`.

**Correctness bug (the important one).** When a Responses upstream emits parallel `function_call` items, the SSE stream legally orders all `output_item.added` events before any `output_item.done`. `openaiResponsesToOpenAIResponse` attributed every `function_call_arguments.delta` to a positional counter that only advanced on `done` — so **all fragments landed on `index: 0`**. The downstream client concatenated them:

```json
{"file_path":"/docs/PRODUCT.md"}{"file_path":"/docs/ROADMAP.md"}{"file_path":"/docs/openapi.custom.yaml"}...
```

…and its tool-input validation rejected the blob (`InputValidationError`). Retries appeared to "fix" it only because the model then called sequentially. This affected every Responses-upstream → chat/Claude streaming client with parallel tool use, not just Muse Spark.

## What changed (4 commits)

1. `feat(opencode-go): add muse-spark-1.3-contributor as responses-only model` — registry entry (`targetFormat` + `supportedFormats: ["openai-responses"]`) so chatCore skips the chat/messages transports and always translates; muse-spark special-case extended to `ocg`/`opencode-go` (covers future versions).
2. `fix(translator): single-stringify tool coercions for chat→Responses` — shared `clampResponsesCallId` / `coerceResponsesArguments` / `coerceResponsesOutput` in `translator/formats/responsesApi.js` (valid JSON passes through untouched, fragments → `"{}"`, null output → `""`); nameless calls/tools dropped instead of sent as `"_unknown"`; array system content joined into `instructions`.
3. `feat(opencode-go): dedicated Responses executor for muse-spark` — new `OpenCodeGoExecutor` routes muse-spark to `/responses` even with a stale `runtimeTransport`, normalizes caps/reasoning, sanitizes tool items (also protects native Responses clients, which skip translation). Non-muse models keep `DefaultExecutor` behavior.
4. `fix(translator): key Responses→chat tool indices by item_id` — index assigned at `added`-time, deltas routed by event `item_id` with most-recent fallback (sequential output byte-identical to before); `done` is idempotent and emits args sent only in `done`. `sanitizeToolArgs` already failed soft — no change needed there.

No registry `index.js` regen needed (no new registry file); `*muse*spark*` capability pattern already covers the new model (verified by test).

## Test evidence

- New `tests/unit/opencode-go-muse-spark-responses.test.js` (11 tests): routing guard, executor sanitization, adversarial quote/newline/unicode/emoji round-trips on both hops.
- New `tests/unit/responses-parallel-tool-calls.test.js` (5 tests): hostile event orderings incl. the exact 4×`read_file` payloads from the bug report, plus Responses→Claude end-to-end. Verified 3/5 fail on the pre-fix code while legacy-behavior tests stay green.
- Targeted suites (13 files: all `openai-responses-*`, `openai-to-claude*`, `bugs-codexCli-responses`, `format-roundtrip`, `golden-response-stream`) green; `eslint` clean on touched files.
- Full suite (~2200 tests): failure set byte-identical to the clean-tree baseline run — zero regressions (remaining failures are pre-existing environmental ones: live-network, sqlite, audit-source tests).

### DIFF
--- a/open-sse/config/providerModels.js
+++ b/open-sse/config/providerModels.js
@@ -50,7 +50,7 @@ export function findModelName(aliasOrId, modelId) {
 }
 
 export function getModelTargetFormat(aliasOrId, modelId) {
-  if ((!aliasOrId || aliasOrId === "oc" || aliasOrId === "opencode") && isMuseSparkModel(modelId)) {
+  if ((!aliasOrId || aliasOrId === "oc" || aliasOrId === "opencode" || aliasOrId === "ocg" || aliasOrId === "opencode-go") && isMuseSparkModel(modelId)) {
     return FORMATS.OPENAI_RESPONSES;
   }
   const models = PROVIDER_MODELS[aliasOrId];
--- a/open-sse/executors/index.js
+++ b/open-sse/executors/index.js
@@ -10,6 +10,7 @@ import { CodexExecutor } from "./codex.js";
 import { CursorExecutor } from "./cursor.js";
 import { VertexExecutor } from "./vertex.js";
 import { OpenCodeExecutor } from "./opencode.js";
+import { OpenCodeGoExecutor } from "./opencode-go.js";
 import { GrokWebExecutor } from "./grok-web.js";
 import { GrokCliExecutor } from "./grok-cli.js";
 import { PerplexityWebExecutor } from "./perplexity-web.js";
@@ -40,6 +41,7 @@ const executors = {
   vertex: new VertexExecutor("vertex"),
   "vertex-partner": new VertexExecutor("vertex-partner"),
   opencode: new OpenCodeExecutor(),
+  "opencode-go": new OpenCodeGoExecutor(),
   "grok-web": new GrokWebExecutor(),
   "grok-cli": new GrokCliExecutor(),
   gcli: new GrokCliExecutor(), // Alias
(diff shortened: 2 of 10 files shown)


decolua:
Thanks @sinasadeghi83 for the contribution! Reviewed and merged into master. 🙏

PR #3830, opened 2026-09-06
fix(providers): clear stale locks after successful validation

### LINKED ISSUE #3810: [Bug] Connection testStatus goes stale: dashboard shows Valid/active while DB keeps old 403 and server uses cached credentials until restart
## Summary
After re-saving an OAuth connection (Antigravity) and getting Test=Valid + Passed in the dashboard, the persisted `testStatus` in SQLite stayed `unavailable` with the old 403 error, and live traffic kept failing over. Only a full server restart picked up the re-saved credentials.

## Steps observed
1. Connection fails with 403 on generate; dashboard marks it unavailable (correct).
2. Re-login OAuth via Edit dialog, Test Connection reports Valid, dashboard bulk test reports Passed 3/3.
3. DB row still shows old `testStatus=unavailable` + old `lastError`; live requests keep failing over to other accounts.
4. After server restart, status/traffic recover.

## Expected
- Saving credentials or passing a connection test should invalidate the in-memory cached session and reset backoff/locks for that connection without requiring a restart.
- Persisted `testStatus`/`lastError` should reflect the latest test result, not the historical failure.

## Environment
9Router 0.5.59-0.5.65, Windows, Antigravity provider with 3 OAuth connections, Round Robin + Sticky.

## Related
Similar staleness family as #3786 (a 400 backs the account off so the next valid request also fails) and #3794 (misreported 429 + 300s lock).

Jordannst:
## Summary

Clear stale connection health state whenever a connection is explicitly marked active after successful validation.

This fixes the split state reported in #3810, where the dashboard can show a revalidated connection as active while request routing still skips it because old model locks and backoff metadata remain persisted.

Closes #3810.

## Problem

Provider failures persist several fields on the connection record:

- `testStatus: "unavailable"`
- `lastError`, `lastErrorAt`, and `errorCode`
- `backoffLevel` and the legacy `rateLimitedUntil`
- dynamic per-model fields such as `modelLock_<model>`

A successful connection test later writes `testStatus: "active"`, while an OAuth re-login upserts fresh credentials with the same active status. Both paths use a shallow merge in `connectionsRepo`, so the old failure fields survive.

`getProviderCredentials()` reads the connection again and filters it through `isModelLockActive()`. The result is internally inconsistent:

1. the latest validation says the connection is active;
2. the old `modelLock_*` value still excludes it from live traffic;
3. requests continue to fail over until the stale lock expires or is otherwise cleared.

## Root cause

Connection activation was treated as a single-field status update instead of a health-state transition. The repository had no shared invariant tying `testStatus: "active"` to the routing metadata that controls eligibility.

This affected both repository write paths:

- `updateProviderConnection()`, used by connection tests;
- the existing-connection branch of `createProviderConnection()`, used by deduplicating OAuth credential re-saves.

## Fix

Add one repository-level normalization step for explicit activation updates:

- clear all existing `modelLock_*` fields;
- reset `backoffLevel`, `rateLimitedUntil`, and `errorCode`;
- clear historical error text and timestamps when the caller does not provide replacements;
- preserve an explicitly supplied active soft warning and timestamp, such as a valid credential with exhausted credits.

Keeping the invariant in the repository makes both activation paths consistent without provider-specific invalidation hooks.

### Before

```text
failed request -> modelLock_x set
successful test or OAuth re-login -> testStatus active
next request -> connection still skipped by modelLock_x
```

### After

```text
failed request -> modelLock_x set
successful test or OAuth re-login -> active status and stale routing state reset atomically
next request -> connection is eligible immediately
```

## Scope and safety

- The reset runs only when a write explicitly sets `testStatus` to `active`.
- Failed tests and unavailable transitions are unchanged.
- Credential values, provider selection strategies, and error classification rules are unchanged.
- Active soft warnings remain visible instead of being erased.

## Tests

Added regression coverage for:

1. successful validation clearing multiple model locks and historical error/backoff fields;
2. OAuth credential re-save clearing the same stale state on the deduplicating upsert path;
3. preservation of an explicitly supplied active soft warning.

Validation results:

- focused SQLite repository suite: 23/23 passed;
- deterministic related DB/auth suites: 35/35 passed;
- ESLint on changed files: passed;
- `git diff --check`: passed.

The broader concurrency run still has its three existing usage-aggregation failures. The same three failures reproduce on a clean `upstream/master` checkout, and this patch adds no new failure.

### DIFF
--- a/src/lib/db/repos/connectionsRepo.js
+++ b/src/lib/db/repos/connectionsRepo.js
@@ -10,6 +10,28 @@ const OPTIONAL_FIELDS = [
   "consecutiveUseCount", "idToken", "lastRefreshAt",
 ];
 
+const MODEL_LOCK_PREFIX = "modelLock_";
+
+function resetHealthStateOnActivation(existing, patch) {
+  if (patch?.testStatus !== "active") return patch;
+
+  const normalized = {
+    ...patch,
+    testStatus: "active",
+    lastError: Object.hasOwn(patch, "lastError") ? patch.lastError : null,
+    lastErrorAt: Object.hasOwn(patch, "lastErrorAt") ? patch.lastErrorAt : null,
+    errorCode: null,
+    rateLimitedUntil: null,
+    backoffLevel: 0,
+  };
+
+  for (const key of Object.keys(existing || {})) {
+    if (key.startsWith(MODEL_LOCK_PREFIX)) normalized[key] = null;
+  }
+
+  return normalized;
+}
+
 function rowToConn(row) {
   if (!row) return null;
   const extra = parseJson(row.data, {});
@@ -147,7 +169,8 @@ export async function createProviderConnection(data) {
     // access_token: never dedup — user manages duplicates manually
 
     if (existing) {
-      const merged = { ...existing, ...data, updatedAt: now };
+      const normalized = resetHealthStateOnActivation(existing, data);
+      const merged = { ...existing, ...normalized, updatedAt: now };
       upsert(db, merged);
       result = merged;
       return;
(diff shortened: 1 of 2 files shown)


decolua:

PR #3840, opened 2026-09-06
feat(providers): add Nous Portal OAuth provider (Hermes CLI device flow)

### LINKED ISSUE #3839: Provider request: Nous Research Portal (Hermes CLI) inference is not supported
## Problem

[Nous Research Portal](https://portal.nousresearch.com) offers an OpenAI-compatible inference API (`https://inference-api.nousresearch.com/v1/chat/completions`) with access to a large multi-vendor catalog (~390 model ids: Claude, GPT, Gemini, GLM, Kimi, DeepSeek, Grok, Qwen, Hunyuan, MiniMax, Muse Spark, Inkling, ...), but 9Router has no integration, so Nous accounts can't be used as an upstream.

## Auth details (verified live)

- **OAuth device-code flow** (same shape as Hermes CLI uses):
  - Device code: `POST https://portal.nousresearch.com/api/oauth/device/code` with `client_id=hermes-cli`, `scope=inference:invoke` — returns **200**
  - Token: `POST https://portal.nousresearch.com/api/oauth/token`
  - Account info: `GET https://portal.nousresearch.com/api/oauth/account` (Bearer token)
- **Non-standard refresh**: the refresh token rides in an `X-Nous-Refresh-Token` header instead of the request body — needs a dedicated refresh handler.
- Model catalog: `GET https://inference-api.nousresearch.com/v1/models` is public (**200**, no auth) and mirrors OpenRouter's shape, including `:free`-suffixed ids and `~vendor/model-latest` rolling aliases.
- Chat endpoint is plain OpenAI-compatible `/v1/chat/completions`, so the default executor should work as-is.

Happy to take a shot at a PR for this if there's interest.

AFG473319:
Closes #3839

## Summary

Adds **Nous Research Portal** as an OAuth provider. Login uses the Hermes CLI device-code flow against `portal.nousresearch.com`; inference runs through the OpenAI-compatible endpoint at `inference-api.nousresearch.com/v1`, so no custom executor is needed.

One non-standard detail: the portal returns the refresh token in an `X-Nous-Refresh-Token` header instead of the token body, so refresh has a dedicated handler.

## Models

Curated list of 21 ids: the `~vendor/model-latest` rolling aliases (which never go stale) plus pinned ids with no alias coverage (`deepseek-v4-pro`, `gpt-6-astra`, `minimax-m3`, etc.). The full ~390-id catalog stays reachable through the suggested-models fetcher (`:free` filter) and passthrough.

Per-model capabilities (context window, output limit, modalities, reasoning) mirror the gateway's public `/v1/models` data; every catalog model accepts OpenAI `reasoning_effort`, so that is the thinking format across the board.

Added/`:free` models can be enabled and disabled like built-in models.

## Verification

- Device-code endpoint: `POST /api/oauth/device/code` with `client_id=hermes-cli`, `scope=inference:invoke` → **200**
- `GET /v1/models` → **200**; all 21 curated ids exist in the live catalog
- Capabilities cross-checked against the same catalog (context lengths, input modalities, `reasoning.mandatory`)
- `tests/__baseline__/`: `verify-providers` / `verify-alias` / `verify-oauth-urls` all **byte-for-byte green**

## Notes

- `tests/__baseline__/providers-baseline.json` also picks up pre-existing `usage.url` drift for `groq`/`opencode-go` (stale on master) — unrelated to this feature, but required for the byte-for-byte check to pass.


AFG473319:
Follow-up commit `64ee018e` — corrected thinking format + gateway-exact capability tables for the curated Nous models.

**What was wrong**

1. **Thinking wire format.** The Nous gateway is an OpenAI-compatible aggregator, but the registry entry didn't force a thinking format, so models like `~z-ai/glm-latest`, `~moonshotai/kimi-latest`, `~deepseek/deepseek-v4-flash-latest`, `qwen/qwen3.8-*`, and `minimax/minimax-m3` resolved through the generic pattern tables to *native* thinking formats (`zai`/`kimi`/`deepseek`/`qwen`/`minimax`) — the translator would then inject GLM/Kimi/DeepSeek-native thinking fields into the OpenAI-compatible body sent upstream.
2. **Pattern-table misses.** `~openai/gpt-latest` matches no GPT glob → fell to the default floor: text-only, 200K context — images silently stripped. `tencent/hy4-preview` matches neither `hy3*` nor `*hunyuan*` → no reasoning flag. The `~latest` aliases got stale family windows (e.g. `*claude*opus*` → 200K vs the 1M actually served).

**Fix**

- `transport.thinkingFormat: "openai"` on `nous-portal` (same mechanism as `openrouter.js`; the override wins over the capability tables in `thinkingUnified.resolveFormat`).
- A `PROVIDER_CAPABILITIES["nous-portal"]` block with entries for all 21 curated ids, same pattern as the existing `codebuddy-cn`/`qoder` blocks.

**Evidence** (from the public, no-auth `https://inference-api.nousresearch.com/v1/models`, fetched 2026-09-06)

- Every model in the catalog lists `reasoning_effort` in `supported_parameters`, and none expose any native thinking parameter — `openai` is the correct wire format across the board.
- Every value in the new block mirrors the gateway's own per-model fields: `context_length` → `contextWindow`, `top_provider.max_completion_tokens` → `maxOutput`, `architecture.input_modalities` → vision/pdf/audio/video, `reasoning.mandatory` → `thinkingCanDisable: false`. Examples: `~z-ai/glm-latest` 1,310,720 ctx / 235,929 out; `~openai/gpt-latest` 1,050,000 ctx, image+file input; `~x-ai/grok-latest` 500,000 ctx / 450,000 out.
- No catalog model exposes `web_search`, so `search` stays `false` for Nous (the generic patterns would have set it `true` for claude/gemini/gpt/grok families).
- `tests/__baseline__`: verify-providers / verify-alias / verify-oauth-urls all green after regeneration; the only providers-baseline delta is the new `thinkingFormat` field on `nous-portal`.

AFG473319:
Rebased onto current master (17c4cc76, v0.5.75) — conflicts resolved, PR is mergeable again.

Two conflicts were:
- open-sse/providers/registry/index.js — re-applied the 
ous-portal import after upstream's provider list churn
- src/app/api/providers/suggested-models/filters.js — kept both upstream's irforce-free filter and the new 
ous free-model filter

Baselines re-verified after the rebase: providers / alias / oauth-urls all green. Ready for review — happy to adjust anything.

PR #3847, opened 2026-09-07
fix(usage): parse Fable weekly limit from limits[] instead of fabricating a row

b1nhm1nh:
## Problem

The "weekly fable (7d)" quota row always showed `0 / 100` with a full green `100%` bar, regardless of actual Fable usage.

## Root cause

`GET /api/oauth/usage` does **not** expose the Fable-scoped weekly window as a top-level `seven_day_fable` / `fable` key. It arrives in the `limits[]` array:

```json
{
  "kind": "weekly_scoped",
  "group": "weekly",
  "percent": 82,
  "severity": "warning",
  "resets_at": "2026-09-09T12:00:00.455637+00:00",
  "scope": { "model": { "id": null, "display_name": "Fable" }, "surface": null },
  "is_active": true
}
```

The lookup in `open-sse/services/usage/claude.js` never matched, so the fallback branch fabricated a row with `used: 0, remaining: 100`. An account that was actually **82% used (18% remaining, severity `warning`)** displayed as completely untouched.

## Fix

- Parse `limits[]` entries with `kind === "weekly_scoped"` into `weekly <model> (7d)` rows (`scope.model.display_name` lowercased, `percent` → used)
- Drop the fabricated fallback — a missing `limits` entry means the account has no such window, so the row is omitted instead of faked
- Remove the dead `seven_day_fable` / `fable` / `fable_5` / `fable_5_1` key mapping (these keys are never returned)

## Verification

Live-tested against the real endpoint on two Max accounts:

| Account | weekly (7d) | weekly fable (7d) from `limits[]` | severity |
|---|---|---|---|
| A | 63% used | 100% used | critical |
| B | 47% used | 82% used | warning |

Both previously rendered as `0 / 100`, `100%` remaining.


decolua:
Thanks @b1nhm1nh for the contribution! Reviewed and merged into master. 🙏

Issue #3846, opened 2026-09-07
[codex/gpt-5.6-sol-image] [400]: {"detail":"The 'gpt-5.6-terra' model requires a newer version of Codex. Please upgrade to the latest version."}


kietnguyenit357-creator:
[codex/gpt-5.6-sol-image] [400]: {"detail":"The 'gpt-5.6-terra' model requires a newer version of Codex. Please upgrade to the latest version."}

Locate: media providers -> text to image -> OpenAI Codex

Issue #3829, opened 2026-09-06
I connected to the codex subscription in 9router, but image2 can't be used.


rebron1900:
I performed a test operation on the model, and the following message was displayed.

HTTP 400: [codex/gpt-image-2] [400]: {"detail":"The 'gpt-image-2' model is not supported when using Codex with a ChatGPT account."} (reset after 28s)

HTTP 400: [codex/gpt-5.6-luna-image] [400]: {"detail":"The 'gpt-5.6-luna' model requires a newer version of Codex. Please upgrade to the latest app or CLI and try again."} (reset after 11s)

PR #3851, opened 2026-09-07
fix(codex): align image client version with model discovery

Jordannst:
## Summary

Align the existing Codex image adapter's advertised client version with model
discovery. This fixes the reproducible GPT-5.6 image request version rejection
without changing the requested model, image tool payload, or authentication.

Related to #3846 and the `requires a newer version of Codex` error in #3829.
This does **not** claim to fix the separate `gpt-image-2` account/model rejection.

## Root cause

On master (`eb712ca8`, v0.5.69), model discovery already requests the Codex catalog
with `client_version=0.144.6`. Image requests independently hardcode both
`version: 0.136.0` and `user-agent: codex_cli_rs/0.136.0`.

The official Codex manifest specifies `minimal_client_version: 0.144.0` for
`gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna`:
[pinned upstream manifest](https://github.com/openai/codex/blob/0df39752cbc4b88d0194ec62bdb0d56fbda4b014/codex-rs/models-manager/models.json).

Thus discovery advertises newer models, but generation identifies itself as an
older client. Updating a user's locally installed Codex CLI cannot change those
hardcoded 9router request headers.

## Changes

- Move the existing discovery version `0.144.6` to `open-sse/config/codexConstants.js`.
- Derive both image identity headers from that shared value.
- Import the constant into model discovery without changing its effective URL/version.
- Update the four parameterized image handler cases to check both headers;
  retain checks for bearer/account authentication, suffix removal, image tool
  payload, and SSE image extraction.

The version is the one master already uses for discovery, not an arbitrary
latest-version bump. A future update of the shared constant updates both paths.

## Live A/B verification

Tested on 2026-09-07 using two authenticated Codex accounts and six requests.
Within each pair, the account, model, prompt, body, and all other headers were
identical. Only `version` and `user-agent` changed. The baseline used the current
adapter with its headers overridden to master's values; the patched arm used
the new values. Credentials remained in process memory and are not included in artifacts.

Prompt: `Generate a solid green square.` Output format: PNG.

| Image alias | Baseline 0.136.0 | Patched 0.144.6 |
| --- | --- | --- |
| gpt-5.6-luna-image | HTTP 400, newer Codex version required | HTTP 200, decoded PNG 1254x1254 |
| gpt-5.6-sol-image | HTTP 400, newer Codex version required | HTTP 200, decoded PNG 1254x1254 |
| gpt-5.6-terra-image | HTTP 400, newer Codex version required | HTTP 200, decoded PNG 1254x1254 |

The image adapter's SSE parser extracted each image, then `sharp`
fully decoded it to raw pixels. Success was not inferred from HTTP 200 alone.
These probes exercised request construction and parsing directly, not the
dashboard, account failover, or the complete public HTTP route.

### Separate `gpt-image-2` check

With the patched `0.144.6` headers, `gpt-image-2` returned HTTP 400 on both tested Codex accounts: "not supported when using Codex with a ChatGPT account". A control request using `gpt-5.6-luna-image` on the same account returned HTTP 200 with a fully decoded 1254x1254 PNG. This confirms a separate upstream rejection for the tested model/endpoint/account combination, not the client-version error fixed here; no `gpt-image-2` remapping or support is claimed.

## Automated verification

```sh
./tests/node_modules/.bin/vitest run --config tests/vitest.config.js \
  tests/unit/image-generation.test.js \
  tests/unit/codex-image-models.test.js \
  tests/unit/provider-test-models-routing.test.js
```

- Before patch: updated image suite had 4 failures and 16 passes; failures showed
  the outbound `0.136.0` headers.
- After patch: all three suites pass, **27/27**.
- ESLint on changed files: zero errors, one existing anonymous-default-export warning.
- `git diff --check`: passed.

## Scope and related work

- #2714 already updated discovery; this reuses its version for the existing image path.
- #1618 introduced the now-stale `0.136.0` image identity.
- #2943 proposes a separate native transport; this patch does not add that transport.
- No change to chat executor headers, auth/refresh, account cooldown, retry,
  model mapping, or image payload options.
- No remapping of Sol to Terra. The different model name reported in #3846
  was not explained by this investigation and is not claimed as fixed here.
- No new support for `gpt-image-2` as a Codex Responses model.


kienmatu:
This is a critical fix, please prioritise it

kuravista:
go merge it

deadcoder0904:
Woud love this. Does this target gpt-image-2.5 too???

Jordannst:
This PR only fixes the Codex client version headers, it doesn’t explicitly select GPT-Image-2.5. Image generation works in my tests, but the responses don’t identify the image model, so I’m not able to confirm 2.5 support yet.

deadcoder0904:
@Jordannst #3943 supports it now

PR #3861, opened 2026-09-07
fix(translator): resolve streaming asymmetry, tool cloaking leaks, and prompt cache cap

### LINKED ISSUE #2433: [Bug] developer role silently dropped in openai→openai-responses request translation
## Bug

When translating an OpenAI Chat Completions request to OpenAI Responses format, a message with `role: "developer"` is silently dropped.

The same converter already maps `role: "system"` to top-level `instructions`, but `role: "developer"` is not handled by any branch.

This matters for GPT-5/Codex-style clients because `developer` is an instruction-level role. Dropping it means project/system instructions can be lost without an error.

## Reproduction

Run from a 9Router source checkout:

```bash
node --input-type=module <<'EOF'
import './tests/translator/registerAll.js';
import { translateRequest } from './open-sse/translator/index.js';
import { FORMATS } from './open-sse/translator/formats.js';

const out = translateRequest(FORMATS.OPENAI, FORMATS.OPENAI_RESPONSES, 'gpt-5.5', {
  messages: [
    { role: 'developer', content: 'Follow the project rules.' },
    { role: 'user', content: 'Hello' }
  ],
  stream: false
}, false, null, null);

console.log(JSON.stringify(out, null, 2));
EOF
```

Actual output before the fix:

```json
{
  "model": "gpt-5.5",
  "input": [
    {
      "type": "message",
      "role": "user",
      "content": [
        {
          "type": "input_text",
          "text": "Hello"
        }
      ]
    }
  ],
  "stream": true,
  "store": false
}
```

The developer instruction is gone.

## Expected behavior

The developer message should be treated as an instruction-bearing message, the same as system, and mapped to Responses API `instructions`:

```json
{
  "instructions": "Follow the project rules.",
  "input": [
    {
      "type": "message",
      "role": "user",
      "content": [
        {
          "type": "input_text",
          "text": "Hello"
        }
      ]
    }
  ]
}
```

## Evidence

Affected code:

`open-sse/translator/request/openai-responses.js`

`openaiToOpenAIResponsesRequest()` only handles `ROLE.SYSTEM` as instructions:

```js
if (msg.role === ROLE.SYSTEM) {
  if (!hasSystemMessage) {
    result.instructions = typeof msg.content === "string" ? msg.content : "";
    hasSystemMessage = true;
  }
  continue;
}
```

The later branches only handle `ROLE.USER`, `ROLE.ASSISTANT`, and `ROLE.TOOL`. `ROLE.DEVELOPER` matches none of them.

## Root cause

The direct `FORMATS.OPENAI -> FORMATS.OPENAI_RESPONSES` request translator bypasses `filterToOpenAIFormat()`, so the developer-role normalization added in #1011 does not run for this path.

`filterToOpenAIFormat()` only runs when `targetFormat === FORMATS.OPENAI`, while this path targets `FORMATS.OPENAI_RESPONSES`.

## Suggested fix

Treat `ROLE.DEVELOPER` as an instruction-bearing role in `openaiToOpenAIResponsesRequest()`:

```js
if (msg.role === ROLE.SYSTEM || msg.role === ROLE.DEVELOPER) {
  ...
}
```

Add regression coverage in `tests/translator/bugs-codexCli-responses.test.js`.

## Related

- Related to #1038 because both involve instruction/system prompt preservation for Codex-style routes.
- Related to #1011, which fixed developer-role normalization for OpenAI-format providers but does not cover this direct Responses conversion path.

### LINKED ISSUE #2458: 406 error when "stream" parameter is missing
## Phenomenon

When using any OpenAI-compatible plugin (e.g., OpenAI-API-compatible, new-api) in Dify, a 406 error occurs when adding the 9router endpoint.

## Analysis

Tcpdump packet analysis reveals that the request from the Dify plugin test connection is:

```shell
# Header
POST /v1/chat/completions HTTP/1.1
Host: <9router_server>
User-Agent: python-requests/2.34.2
Accept-Encoding: gzip, deflate
Accept: */*
Connection: keep-alive
Content-Type: application/json
Authorization: Bearer <my key>
Content-Length: 94

# request body
{"model": "9router-combo", "max_tokens": 5, "messages": [{"role": "user", "content": "ping"}]}
```

The request from 9router to the upstream LLM provider:

```shell
# Header
POST /api/v1/chat/completions HTTP/1.1
host: <upstream_server>
connection: keep-alive
Content-Type: application/json
Authorization: <upstream api key>
Accept: text/event-stream
accept-language: *
sec-fetch-mode: cors
user-agent: node
accept-encoding: gzip, deflate
content-length: 93

# request body
{"model":"Qwen/Qwen3.6-35B-A3B","max_tokens":5,"messages":[{"role":"user","content":"ping"}]}
```

The request body does not contain `"stream":true`, but the Header is set to `Accept: text/event-stream`. It is suspected that the handling here is inconsistent with the upstream provider's default.

Verification 1: Simulate the original request using curl:

```shell
curl -v -X POST http://<upstream_server>/v1/chat/completions \
  -H "Host: <upstream_server>" \
  -H "Connection: keep-alive" \
  -H "Authorization: <upstream api key>" \
  -H "accept-language: *" \
  -H "sec-fetch-mode: cors" \
  -H "Accept: text/event-stream" \
  -d '{"model":"Qwen/Qwen3.6-35B-A3B","max_tokens":5,"messages":[{"role":"user","content":"ping"}]}'
  
# Upstream returns:
  < HTTP/1.1 406 Not Acceptable
```

Verification 2: Remove `Accept: text/event-stream`:

```shell
curl -v -X POST http://<upstream_server>/v1/chat/completions \
  -H "Host: <upstream_server>" \
  -H "Connection: keep-alive" \
  -H "Authorization: <upstream api key>" \
  -H "accept-language: *" \
  -H "sec-fetch-mode: cors" \
  -d '{"model":"Qwen/Qwen3.6-35B-A3B","max_tokens":5,"messages":[{"role":"user","content":"ping"}]}'
```

Obtains a non-streaming response:
```json
{"id":"47084272d20c485fb55a9c7b13653a07","object":"chat.completion","created":1783480455,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"message":{"role":"assistant","reasoning_content":"用户突然发了个“"},"finish_reason":"length"}],"usage":{"prompt_tokens":13,"completion_tokens":5,"total_tokens":18}}
```

Verification 3: Retain `Accept: text/event-stream` and add `"stream":true` to the request:

```shell
curl -v -X POST http://<upstream_server>/v1/chat/completions \
  -H "Host: <upstream_server>" \
  -H "Connection: keep-alive" \
  -H "Authorization: <upstream api key>" \
  -H "accept-language: *" \
  -H "sec-fetch-mode: cors" \
  -H "Accept: text/event-stream" \
  -d '{"model":"Qwen/Qwen3.6-35B-A3B","max_tokens":5,"messages":[{"role":"user","content":"ping"}],"stream":true}'
```

Obtains a streaming response:

```json
data:{"id":"ea3ec88f44144e43a38a55f742f2007e","object":"chat.completion.chunk","created":1783480655,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"delta":{"role":"assistant","content":"","reasoning_content":""}}]}

data:{"id":"ea3ec88f44144e43a38a55f742f2007e","object":"chat.completion.chunk","created":1783480655,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"delta":{"content":"","reasoning_content":"用户"}}]}

data:{"id":"ea3ec88f44144e43a38a55f742f2007e","object":"chat.completion.chunk","created":1783480655,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"delta":{"content":"","reasoning_content":"发了"}}]}

data:{"id":"ea3ec88f44144e43a38a55f742f2007e","object":"chat.completion.chunk","created":1783480655,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"delta":{"content":"","reasoning_content":"一个"}}]}

data:{"id":"ea3ec88f44144e43a38a55f742f2007e","object":"chat.completion.chunk","created":1783480656,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"delta":{"content":"","reasoning_content":"\""}}]}

data:{"id":"ea3ec88f44144e43a38a55f742f2007e","object":"chat.completion.chunk","created":1783480656,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"delta":{"content":"","reasoning_content":"ping"}}]}

data:{"id":"ea3ec88f44144e43a38a55f742f2007e","object":"chat.completion.chunk","created":1783480656,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"delta":{"content":"","reasoning_content":""},"finish_reason":"length"}]}

data:{"id":"ea3ec88f44144e43a38a55f742f2007e","object":"chat.completion.chunk","created":1783480656,"model":"Qwen/Qwen3.6-35B-A3B","choices":[{"index":0,"delta":{"role":"assistant","content":""}}],"usage":{"prompt_tokens":13,"completion_tokens":5,"total_tokens":18}}

data:[DONE]
```

Thus, it is confirmed that when `"stream":true` is missing from the request body, the provider usually treats it as non-streaming, which conflicts with `Accept: text/event-stream` in the Header, resulting in the 406 error.

## Fix

Analyzing the code (v0.5.18), it was found that the judgment logic for `body.stream` in lines 70 and 72 of `open-sse/handlers/chatCore.js` is inconsistent:

```ts
# Line 70
const clientRequestedStreaming = body.stream === true || sourceFormat === FORMATS.ANTIGRAVITY || sourceFormat === FORMATS.GEMINI || sourceFormat === FORMATS.GEMINI_CLI;
const providerRequiresStreaming = PROVIDERS[provider]?.forceStream === true;
# Line 72
```

In 9router, the upstream providers are managed as "OpenAI compatible," so `providerRequiresStreaming` is `false`. The expression in line 72 causes `stream` to be `true` when `body.stream` is missing (undefined), which leads to the `Accept: text/event-stream` header being set.

Modify line 72 to:

```ts
let stream = providerRequiresStreaming ? true : (body.stream === true);
```

After packaging and testing:

```shell
curl --request POST \
  --url http://9router/v1/chat/completions \
  --header 'Authorization: Bearer <9router_api_key>' \
  --header 'content-type: application/json' \
  -H "Accept: */*" \
  --data '{"messages":[{"role":"user","content":"ping"}],"model":"9router-combo","max_tokens":5}'
```

It correctly returns a non-streaming response. Adding the 9router endpoint in Dify is successful, and the workflow functions normally.

## Notes

Since I am unsure if the author had other considerations for line 72, I have not provided a PR and am reporting this as an issue for reference.

### LINKED ISSUE #2489: Antigravity fails with 400 INVALID_ARGUMENT due to invalid function declaration schema

### LINKED ISSUE #2582: Support Anthropic Advisor server-tool passthrough

### LINKED ISSUE #2595: ❌ antigravity [400]: [400]: { "error": { "code": 400, "message": "Invalid value at 'request.tools[0].function_d

### LINKED ISSUE #2894: [Bug] Adaptive thinking without an explicit effort always 400s on /v1/messages: claude-budget emits `enabled` with no `budget_tokens`, claude-adaptive emits `effort: "auto"`

### LINKED ISSUE #3567: 400 on claude provider: cache_control set on last tool with defer_loading=true

### LINKED ISSUE #3764: TypeError in ensureToolCallIds on malformed content/tool_calls crashes every translation path
## Summary

Any client POSTing a malformed OpenAI body crashes every translation path with a raw `TypeError`, because `ensureToolCallIds` runs unconditionally on the request body with no null guards:

- `content: [null]` → `TypeError: Cannot read properties of null (reading 'type')`
- `tool_calls: [null]` → `TypeError: ... (reading 'id')`
- numeric `function.name` with a bad id → `TypeError: toolName.replace is not a function`

All three reproduced against the real `ensureToolCallIds` (no mocks). Sibling code (`formats/claude.js`) already guards null blocks — this site is the unguarded outlier.

## Expected

Skip non-object entries instead of throwing, and only call string methods on strings, so a malformed body degrades instead of failing the gateway request. No behavior change for well-formed bodies.

### LINKED ISSUE #3784: Crash on malformed chat body: reasoning injector maps non-array messages
## Summary

`injectReasoningContent` crashes with `TypeError: body.messages.map is not a function` when the chat body carries a non-array `messages` (string, number, object). The guard only checks truthiness, and the throw escapes uncaught past the executor — a malformed-but-JSON-valid body becomes an unhandled 500 instead of a 400.

Real runs (`injectReasoningContent`, deepseek path):

- `messages: 'hi'` → throws `TypeError`; `messages: 42` → throws; `messages: {role:'user'}` → throws
- well-formed array body → fine

## Root cause

`applyRule` maps `body.messages` without an `Array.isArray` check, unlike every sibling transform in the same pipeline (`stripContinuityFields`, `stripOpenAI`, `stripClaude`, `applyJsonSchemaFallback`), which all guard.

## Expected

Same one-word guard as the siblings: non-array messages pass through untouched for upstream validation to 400.

### LINKED ISSUE #3786: [Bug] claude-adaptive writes effort:"auto" into output_config -> 400, and the 400 backs the account off so the next valid request also fails

### LINKED ISSUE #3794: [Bug] NVIDIA kimi-k3: a 400 for reasoning_effort=medium triggers account fallback, locks the model for 300s, and is reported to the client as 429
## Bug

With two NVIDIA API-key connections configured, a request to `nvidia/moonshotai/kimi-k3` with `reasoning_effort: "medium"` is answered by 9router with:

```
{"error":{"message":"[nvidia/moonshotai/kimi-k3] [429]: {\"status\":429,\"title\":\"Too Many Requests\"} (reset after 30s)"}}
```

NVIDIA never returned a 429 for that request. The real upstream answer is a **400 with a useful message**, which 9router hides:

```
{"message":"Unsupported Kimi K3 thinking_effort=\"medium\"; supported values are low, high, and max","type":"Bad Request","code":400}
```

Three problems compound here:

1. A **400 for an invalid request parameter** is treated as an account failure: the model is locked on account 1 (30 s) and the same request is retried on account 2.
2. Account 2 answers a genuine 429 (rapid retry of the same request), so the model is locked on **account 2 for 300 s** — a five-minute outage for kimi-k3 caused by one bad parameter value.
3. The client is told the **last** error (429) instead of the **first, meaningful** one (400). `providerConnections.data.lastError` still holds the 400, so the information was there.

The practical effect: any client whose default effort is `medium` (Hermes Agent, for one) sees kimi-k3 as "rate limited" and there is nothing in the client-facing error pointing at the actual fix (`low`, `high` or `max`).

## Reproduction

9router **0.5.65**, npm install, Linux, Node 22. Two NVIDIA API-key connections (`nvidia`, `nvidia-2`). Token Saver / RTK off.

```bash
curl -s localhost:20128/v1/chat/completions \
  -d '{"model":"nvidia/moonshotai/kimi-k3","messages":[{"role":"user","content":"Say OK"}],"max_tokens":20,"reasoning_effort":"medium"}'
```

Server log (`--log`):

```
▶ POST nvidia/moonshotai/kimi-k3 → nvidia/moonshotai/kimi-k3 · FMT: openai→openai · STREAM · 1 MSG · THINK:medium · ACC:nvidia-2 (acct 2)
✗ ERROR 429 · nvidia/moonshotai/kimi-k3 · 182ms
    [429]: {"status":429,"title":"Too Many Requests"}
⚠️  [AUTH] nvidia-2 (acct 2) locked modelLock_moonshotai/kimi-k3 for 300s [429]
⚠️  [FALLBACK] ⇄ ACC:nvidia-2 (acct 2) UNAVAILABLE (429) → NEXT ACCOUNT
⚠️  [AUTH] nvidia | all 2 accounts locked for moonshotai/kimi-k3 (reset after 30s) | lastError=[400]: {"message":"Unsupported Kimi K3 thinking_effort=\"medium\"; supported values are low, high, and max" ...
⚠️  [CHAT] [nvidia/moonshotai/kimi-k3] [429]: {"status":429,"title":"Too Many Requests"} (reset after 30s)
```

With `ENABLE_REQUEST_LOGS=true`, `4_req_target.json` shows the body sent upstream is identical to the client body (model rewritten to `moonshotai/kimi-k3`, `reasoning_effort: "medium"` passed through), so this is not a translation issue.

Control cases, same setup:

| request | direct to `integrate.api.nvidia.com` | via 9router |
|---|---|---|
| no effort field | 200 | 200 |
| `reasoning_effort: "high"` | 200 | 200 |
| `reasoning_effort: "medium"` | **400** (message above) | **429** "reset after 30s" |

## Expected

- A 4xx that names an invalid client parameter (400 `Unsupported ... thinking_effort`) should be returned to the client as-is, without account fallback and without locking the model on any account. Retrying an invalid request on another account cannot succeed and only burns the second account's rate limit.
- When fallback does happen and every account fails, surface the **first** (or the most specific) upstream error rather than the last one, or at least include both.

## Suggested fix

In the fallback classifier (`shouldFallback` / cooldown logic in `open-sse`), treat `400` with a body matching `Unsupported`, `invalid`, `not supported`, `unrecognized` as **non-retryable client errors**: return immediately, no `modelLock_*`, no backoff. Optionally normalise `medium` → `high` for models whose capability entry is `thinkingFormat: "kimi"`, the way `auto`/`minimal`/`xhigh` are already mapped in the kimi case.

Related: #2690 (kimi-k3 capability matching) and #2427 (kimi effort normalisation) — this is the error-handling side of the same feature.

### LINKED ISSUE #3795: 400 "A maximum of 4 blocks with cache_control may be provided. Found 5." — anchorClaudeCache adds a 5th marker when the client already sent 4
Related to #3567 — same function, different symptom. That one is about
`defer_loading` and `cache_control` colliding on a single tool; this is the
block-count limit, and it happens with no `defer_loading` involved.

## What happens

Client sends 4 `cache_control` markers, which is Anthropic's maximum.
9Router adds one more while translating to the Anthropic Messages API, and
the request is rejected:

```json
 "message":"A maximum of 4 blocks with cache_control may be provided. Found 5."}}
```

Setup: Hermes Agent → 9Router `/v1/chat/completions` → `cc/` slot →
`api.anthropic.com/v1/messages?beta=true`. Image
`decolua/9router@sha256:3eb104460661900d0c387e2cce910fe348201e28dd6b31067f019a23ec04f95d`,
built 2026-09-03, dashboard reports `0.5.65`.

## Measurement

I put a counting proxy on the client → 9Router hop — bodies forwarded
byte-for-byte, one log line per request — and lined its timestamps up
against 9Router's own log:

```
15:51:20   client sent:      4   [system×2 tool×2]
15:51:21   Anthropic got:    5   Found 5.
15:51:22   Anthropic got:    5   Found 5.   ← same body, second account
```

Four such pairs in one window. The client caps itself at 4 and the proxy
confirms it never exceeded that — the count it sends is 3 or 4, never 5.

Turn by turn within a single task:

| turn | markers the client sent | outcome |
| ---: | ---: | --- |
| 1 | 3 | passes — 3 + 1 = 4, still within the limit |
| 2+ | 4 | rejected — 4 + 1 = 5 |

None of the client's markers are on the tools array; they sit on the system
message and on messages (`[system×2 tool×2]`, `[assistant×1 system×2
tool×1]`, and so on). So the added one is the last-tool anchor.

## Where it comes from

#3567 reads the code and names it:

> It sets `cache_control` on the last tool **unconditionally**
> — `openai-to-claude.js` (line 181), `anchorClaudeCache()`

The `defer_loading` collision reported there is one consequence of that
line. The block-count rejection is the other, and it needs no
`defer_loading` at all: if the client has already spent all four markers,
one more is always five.

The layout of the client's markers turns out to be irrelevant — only the
count matters.

## Suggested fix

The `if (!tool.defer_loading)` guard proposed in #3567 fixes that collision
but not this one. Counting first covers both, and any future anchor
position:

```js
// before adding any re-anchor marker
if (countCacheControlBlocks(body) >= 4) return;
```

At 4 the client has spent the whole budget, so a 5th guarantees a
non-retryable 400 — while skipping it costs nothing, since a marker the
client placed is still a valid breakpoint.

## Second, separable problem: a 400 burns the account pool

A `400 invalid_request_error` is currently treated as an account problem.
Each rejection is followed by

```
[FALLBACK] ⇄ ACC:Account N UNAVAILABLE → NEXT ACCOUNT
```

and the same malformed body is retried across the pool until the slot
reports `[AUTH] claude | all N accounts locked for claude-sonnet-5`. The two
rejections one second apart above are exactly that. In one 3,000-line window
there were 52 rejections and 52 `ACC:… UNAVAILABLE` lines — one per
rejection.

A 4xx describes the request, so retrying it on another credential cannot
succeed, and it costs the whole pool. From the outside the result is
indistinguishable from rate limiting: the first turn or two of a task
succeed, then the combo silently walks down to a different model, and
nothing in the client's view suggests a malformed request.

Happy to split that into its own issue if you would prefer.

## Reproducing

Any client that sends 4 `cache_control` markers on
`/v1/chat/completions` for a Claude model routed through a `cc/` slot. With
Hermes Agent that is any `base_url` containing `nousresearch`, which is how
it decides to emit envelope-layout `cache_control` for Claude models on the
OpenAI wire; the count reaches 4 from the second turn of an agent loop.

### LINKED ISSUE #3799: 🇺🇸 [Bug] Vercel AI SDK compatibility: Invalid Tool Schemas and forced SSE streaming | 🇧🇷 [Bug] Compatibilidade com Vercel AI SDK: Schemas de Tools inválidos e stream (SSE) forçado
🇺🇸 Hello! I'm using 9router as a custom endpoint for Quill (which uses the Vercel AI SDK), and I encountered two major compatibility issues that prevent it from working properly.

  1. Tool Schema Transformation Bug (HTTP 400 from Gemini)
  When translating OpenAI tool schemas to Gemini's functionDeclarations, the recursive schema converter (the Yr
  function) blindly injects type: "object" into any dictionary named properties if it lacks a type key.
  Since properties in a JSON Schema is meant to be a dictionary mapping field names to schemas (not a schema
  object itself), injecting type: "object" as a literal field causes the Gemini API to reject the request with:
  INVALID_ARGUMENT: Invalid value at '...properties[x].value' ... "object".

  Proposed Fix: The schema transformer should ignore maps that are specifically the properties map, or apply a
  cleanup pass to delete type: "object" strings from properties maps before sending to Gemini.

  2. Forced Server-Sent Events on missing Accept header
  When Vercel AI SDK tests the connection (e.g., calling generateText), it expects a standard JSON response.
  However, if the client sends an empty Accept header, the current 9router fallback logic forces stream = true
  and returns Server-Sent Events (text/event-stream), even if the request body didn't pass stream: true. This
  causes the SDK to crash with an Unexpected token 'd', data: ... is not valid JSON error when trying to parse
  the response.

  Proposed Fix: The stream fallback evaluation (aJ=aI.includes("application/json")||aI.includes("*/*")) should
  also evaluate to true if the Accept header is empty (aI === ""), so it correctly defaults to a standard JSON
  response when stream: true is not requested in the payload.

-----------------------------------

🇧🇷 Olá! Estou usando o 9router como um endpoint customizado para o Quill (que utiliza a Vercel AI SDK) e
  encontrei dois problemas críticos de compatibilidade.

  1. Bug na transformação de Tool Schemas (HTTP 400 do Gemini)
  Ao traduzir os schemas das tools do formato OpenAI para o functionDeclarations do Gemini, a função recursiva
  de conversão (a função Yr) injeta cegamente type: "object" em qualquer dicionário chamado properties se ele
  não tiver a chave type.
  Como properties em um JSON Schema deve ser um mapa de propriedades (e não um schema em si), a injeção da
  chave literal type: "object" faz com que a API do Gemini rejeite a requisição com o erro: INVALID_ARGUMENT:
  Invalid value at '...properties[x].value' ... "object".
  
Sugestão de Correção: O conversor deveria evitar injetar a propriedade de tipo nos mapas properties ou fazer
  uma varredura de limpeza removendo chaves .type de formato string de dentro de dicionários properties antes
  de enviar ao Gemini.

  2. Stream (SSE) forçado quando o Accept header está vazio
  Quando a Vercel AI SDK testa a conexão, ela espera um JSON simples de resposta. Porém, se o cliente enviar a
  requisição sem o header Accept (ou vazio), a lógica atual do 9router força o stream = true e devolve a
  resposta em Server-Sent Events (text/event-stream), mesmo que o body da requisição não tenha enviado stream:
  true. Isso quebra a SDK, que tenta dar JSON.parse nos chunks e falha com o erro Unexpected token 'd', data:
  ... is not valid JSON.
  
Sugestão de Correção: A validação de fallback (aJ=aI.includes("application/json")||aI.includes("*/*"))
  deveria considerar o header Accept vazio (aI === "") como válido para retornar um JSON comum quando a flag
  stream: true não for enviada no payload.

### LINKED ISSUE #3825: [Bug] /v1/embeddings rejects multimodal input for VL embedding models
## Observation

`/v1/embeddings` only accepts a plain string (or an array of strings) as `input`. Vision-language embedding models available in the catalog (e.g. `openrouter/nvidia/llama-nemotron-embed-vl-1b-v2:free`) natively support image input, but 9router's validation layer rejects any multimodal payload.

## Reproduction (2026-09-05)

```bash
curl -s http://192.168.0.52:20128/v1/embeddings \
  -H "Authorization: Bearer ***" \
  -d '{
    "model": "openrouter/nvidia/llama-nemotron-embed-vl-1b-v2:free",
    "input": [
      {"type": "text", "text": "Describe this image"},
    ]
  }'
```

**Response:**

```json
{
  "error": {
    "message": "[openrouter/nvidia/llama-nemotron-embed-vl-1b-v2:free] [400]: [
      {
        \"code\": \"invalid_union\",
        \"errors\": [
          [ { \"expected\": \"string\", ... } ]
        ]
      }
    ]"
  }
}
```

**Control:** the same model with plain string `input` returns **HTTP 200**, 2048 dims, cost 0. The upstream provider is healthy; the rejection comes from 9router's OpenAI wrapper.

## Affected model

`openrouter/nvidia/llama-nemotron-embed-vl-1b-v2:free` — listed in `/v1/models/embedding` (14 models).

## References

- #427 — image format bug in `/v1/chat/completions`
- #1419 — image fallback feature request
- `tests/unit/embeddingsCore.test.js` : only covers `string | string[]` for `input`.

### LINKED ISSUE #3828: Tool name corrupted with _ide suffix on repeated tool_use in same-format Claude OAuth streaming
## Summary

When a tool is called a **second time (or more) with the same name** within a growing conversation (multi-turn, with prior `tool_use`/`tool_result` history in `messages`), 9Router returns the `tool_use` block with a **corrupted name** — a `_ide` suffix appended to the real tool name.

Example:
- Real tool name: `esp32-datalogger__esp32_device_summary`
- Name returned by 9Router on the second call: `esp32-datalogger__esp32_device_summary_ide`

Since no tool with that name exists, the call fails (`Tool X not found`). The model (tested with Claude Haiku 4.5) then retries the same (still-wrong) name repeatedly — we observed 10–20+ consecutive failed attempts in a single reply — before giving up and answering without the tool.

In worse cases, once a session hits this once, the corrupted tool_use/tool_result history stays in context and **every subsequent message in that session** (even ones that need no tool at all) triggers the same failed retry loop. The built-in `sessions compact` (from our agent framework, OpenClaw) does not reliably fix this, since the corrupted turns are at the *end* of the transcript, not the part that gets summarized away.

## Environment

- **9Router version:** confirmed present in **0.5.55, 0.5.65, and 0.5.69** (latest as of 2026-09-06) — this is not a stale/already-fixed report, we re-tested after each update.
- **Platform:** macOS (Apple Silicon / arm64), Node v25
- **Provider connection affected:** Claude via **OAuth subscription** (Claude Pro/Max login), `authType: "oauth"`, `scope: "user:inference user:profile"` — not a plain API-key connection
- **Client used for reproduction:** both a real agent framework (OpenClaw) with real tools, and a minimal `curl` request directly to 9Router with no client framework involved at all (see below)

## Minimal reproduction (no client framework needed)

```bash
curl -sS -X POST http://127.0.0.1:20128/v1/messages \
  -H "x-api-key: <9ROUTER_API_KEY>" \
  -d '{
  "model": "claude/claude-haiku-4-5",
  "max_tokens": 512,
  "tools": [
    {
      "name": "get_weather",
      "description": "Get current weather for a city",
      "input_schema": {
        "type": "object",
        "properties": {"city": {"type": "string"}},
        "required": ["city"]
      }
    }
  ],
  "messages": [
    {"role": "user", "content": "Check the weather for Jakarta and Bandung, calling get_weather for each city one at a time."},
    {"role": "assistant", "content": [
      {"type": "tool_use", "id": "toolu_manual_001", "name": "get_weather", "input": {"city": "Jakarta"}}
    ]},
    {"role": "user", "content": [
      {"type": "tool_result", "tool_use_id": "toolu_manual_001", "content": "Jakarta: 31C, sunny"}
    ]},
    {"role": "assistant", "content": [
      {"type": "tool_use", "id": "toolu_manual_002", "name": "get_weather", "input": {"city": "Bandung"}}
    ]},
    {"role": "user", "content": [
      {"type": "tool_result", "tool_use_id": "toolu_manual_002", "content": "Bandung: 22C, light rain"}
    ]},
    {"role": "user", "content": "Now check Jakarta again with the same tool, make sure the data is still the same."}
  ]
}'
```

**Response received (name field corrupted):**

```json
{
  "model": "claude-haiku-4-5-20251001",
  "id": "msg_011CeKcj1Mo9mGptqeu2kH6p",
  "type": "message",
  "role": "assistant",
  "content": [
    {
      "type": "tool_use",
      "id": "toolu_01DsaLesGcmeDHaU7sHZXneR",
      "name": "get_weather_ide",
      "input": {"city": "Jakarta"},
      "caller": {"type": "direct"}
    }
  ],
  "stop_reason": "tool_use",
  ...
}
```

Two things worth noting in that response:
1. `"name": "get_weather_ide"` — this is the **second** `get_weather` call in the conversation; the first one (`toolu_manual_001`) already used that exact name correctly.
2. There's a `"caller": {"type": "direct"}` field on the `tool_use` block that isn't part of the documented Anthropic Messages API — indicating 9Router is actively reconstructing/rewriting the `tool_use` block rather than passing it through untouched, which is presumably where the corruption is introduced.

This is 100% reproducible: every repeated tool name after the first occurrence gets corrupted, every time, across three separate 9Router versions we tested (0.5.55, 0.5.65, 0.5.69).

## What we ruled out

- **RTK (Rust Token-Saver)** — toggled `rtkEnabled` off in `/api/settings`, bug persisted. Restored to `true`.
- **Other compression/transform toggles** (`headroomEnabled`, `cavemanEnabled`, `ponytailEnabled`, `pxpipeEnabled`) — all `false` by default in our install, so the bug happens with none of these active.
- Not something introduced by our own commit changes to that update-window — the maintainer's own commit log between 0.5.55 and 0.5.69 has no entry addressing this that we could find (closest we saw was `fix(claude): decloak tool names in same-format streaming passthrough`, which sounded related but did not fix our repro).

## Where we suspect the bug lives (from reading the minified/built npm package — we don't have access to the un-minified source)

Trace through the compiled `app/.next-cli-build` bundle for `/v1/messages` (webpack module numbers are specific to the versions we tested, likely to shift between releases, but the call chain shape is probably still similar):

1. `POST /api/v1/messages` route → dispatcher that resolves provider/model and calls the core request executor (an async function, param-destructured with `rtkEnabled`, `headroomEnabled`, `cavemanEnabled`, `ponytailEnabled`, `pxpipeEnabled`, `sourceFormatOverride`, etc.)
2. For streaming responses, that hands off to a response-streaming handler that picks between a "translate" mode and a "passthrough" mode based on whether `sourceFormat === targetFormat`
3. Inside "translate" mode, there's a per-SSE-event converter that: `if (sourceFormat === targetFormat) return [event]` (untouched) — otherwise it round-trips the event through an **OpenAI-format intermediate representation** (`sourceFormat → OpenAI → targetFormat`) even when both sides are nominally "Claude"
4. We believe the OAuth/subscription Claude connection uses an internal target-format label that's *different* from the literal wire format the client sent (something like a `claude-cli`-flavored variant, distinct from a plain API-key "anthropic" connection) — which would explain why this specific connection type goes through the double-hop conversion at all, and why the `caller` field gets added along the way.
5. We spotted a synthetic tool-call ID generator in the OpenAI→Anthropic direction of that converter: `id: a.id || \`toolu_${Date.now()}_${e.length}\`` (array-index-based) — a plausible place for a duplicate/second tool_use with the same name to get mis-tracked against per-request state (`toolNameMap` / `customToolNames`), but we were not able to confirm this is the exact faulty line without the real source.

We'd guess the fix is somewhere in that streaming SSE→SSE per-event tool_use reconstruction path, specifically the branch that handles a *second* occurrence of an already-seen tool name within one request's conversation history — but we're working blind against minified code, so take our code pointers as a lead, not a diagnosis.

## Suggested things to check

1. Does this also happen on a plain **API-key** Claude connection (not OAuth/subscription)? We only had an OAuth connection to test with.
2. Whatever per-request state tracks "have I seen this tool name before" during streaming reconstruction (`toolNameMap`/`customToolNames`-like structures) — check what happens on a second occurrence of the same name.
3. Is the `"caller": {"type": "direct"}` field on `tool_use` blocks intentional? If so, is its injection logic sharing code with whatever mangles the name?

## Impact

Any tool-calling agent using a Claude OAuth/subscription connection through 9Router is unreliable as soon as it needs to call the same tool more than once per conversation — which is extremely common (e.g., "check sensor A and sensor B", "list items then get details on one of them twice"). We've had to route Claude traffic directly to Anthropic (bypassing 9Router) to work around this, which defeats the purpose of using 9Router for that connection.

---
*Reported after reproducing on 9Router v0.5.55 (2026-08-23), re-confirmed still present on v0.5.65 (2026-09-04) and v0.5.69 (2026-09-06), each time via a clean install/update + the same curl repro above.*

### LINKED ISSUE #3836: 9Router strips vision capability from model
I run Qwen 3.8 27b with vision enabled. If I host it using vLLM, Unsloth Studio, LM Studio and so on vision works just fine. As soon as I route the communication through 9Router (tested on multiple backends) vision no longer works.

It's kind of funny because I can set up two models in OpenCode that point to the same actual running model. On directly to the backend it is hosted on and one through 9Router. The response if I ask it to explain an image through 9Router is "sorry, I don't have vision capabilities... bla blah blah".

Then I right after switch to the same model but directly against the backend (so no reload or anything) and vision works just fine and it happily explains the image.

### LINKED ISSUE #3844: Qoder SSE passthrough emits delta.role:"" — breaks every AI SDK v5 (@ai-sdk/openai-compatible) client
## 问题描述

Qoder executor 把上游 SSE chunk 原样透传给客户端。Qoder 上游（`api3.qoder.sh/algo/...`）返回的 `choices[0].delta.role` 是**空字符串**，而不是 `"assistant"`。这不符合 OpenAI Chat Completions 规范，Vercel AI SDK v5 的 `@ai-sdk/openai-compatible` 会在 Zod schema 层直接拒绝，导致第一个内容 chunk 就把整个流打断。

结果是：任何基于 AI SDK v5 的客户端（opencode，以及所有用 `@ai-sdk/openai-compatible` 构建的工具）都完全无法使用 `qoder` provider，与选哪个 Qoder 模型无关。

## 复现步骤

1. 连接一个 Qoder 账号（OAuth 或 `pt-` 开头的 PAT）。
2. 让 AI SDK v5 客户端把 9router 当作 OpenAI 兼容 provider：

```json
{
  "provider": {
    "9router": {
      "npm": "@ai-sdk/openai-compatible",
      "options": {
        "baseURL": "http://127.0.0.1:20128/v1",
        "apiKey": "sk-..."
      },
      "models": { "qd/ultimate": { "name": "qd/ultimate" } }
    }
  }
}
```

3. 向 `qd/ultimate` 发起任意流式对话请求（`qd/auto` 同样复现）。
4. 流在第一个内容 chunk 处中断。

### 实际结果

```
Type validation failed: Value: {"choices":[{"delta":{"content":"你好","role":""},"index":0}],"created":1788747926,"id":"resp_061e3eee7774c31d016a9e2091fac481969af491cb9633bae3","model":"auto","object":"chat.completion.chunk"}.
Error message: [
  {
    "code": "invalid_union",
    "errors": [
      [
        {
          "code": "invalid_value",
          "values": ["assistant"],
          "path": ["choices", 0, "delta", "role"],
          "message": "Invalid input: expected \"assistant\""
        }
      ],
      [
        {
          "expected": "object",
          "code": "invalid_type",
          "path": ["error"],
          "message": "Invalid input: expected object, received undefined"
        }
      ]
    ],
    "path": [],
    "message": "Invalid input"
  }
]
```

注意 chunk 里的 `"role":""`。按规范，`delta.role` 出现时必须是 `"assistant"`，续传 chunk 则应完全省略该字段。`@ai-sdk/openai-compatible` 把它实现为字面量类型，所以空字符串是硬性 schema 失败——客户端侧无法通过任何配置绕过。

### 期望结果

9router 在转发前对 chunk 做归一化，让 AI SDK v5 客户端能像使用其他 provider 一样正常消费 Qoder。

## 根因

Qoder 的 SSE 中继会解开上游信封（`statusCodeValue` / `body`）、在非 200 时合成一个错误 chunk、去掉换行——然后就把 payload **原样**吐出，完全没有做 chunk 归一化。构建产物中对应位置（`app/.next-cli-build/server/chunks/318.js`；源码中应为 Qoder executor 的流转换逻辑，即 `open-sse/executors/qoder.js`）：

```js
if (!j) return;
if ("[DONE]" === j) { c.enqueue(h.encode(k.$U)); i = !0; return }
let l = j.replace(/\r?\n/g, "");
c.enqueue(h.encode(`data: ${l}\n\n`));   // ← 原样透传，role 从未被校验
```

由于 `body` 被无条件转发，上游**任何**不合规字段都会直接泄漏到客户端；`delta.role: ""` 只是最先触发 schema 报错的那一个。

## 版本信息

- 在 **0.5.65**（npm 全局安装）上复现。
- 我另外解包了 **0.5.69**（当前最新版）并对比了这段代码路径，透传逻辑逐字节相同，因此升级无效。
- 我搜索过是否有现成开关可绕过（`normalize*`、`strict*`、`compat*`、`sanitize*`），唯一命中的是 `strictProxy`，属于 HTTP 代理配置，与 chunk 结构无关。也就是说目前没有任何配置项能规避。

## 修复建议

在入队前对解析后的 chunk 做归一化，而不是转发原始字符串。大致如下：

```js
let parsed;
try { parsed = JSON.parse(j) } catch { /* 解析失败则保留现有原样透传行为 */ }

if (parsed?.object === "chat.completion.chunk" && Array.isArray(parsed.choices)) {
  for (const choice of parsed.choices) {
    const delta = choice?.delta;
    if (delta && "role" in delta && delta.role !== "assistant") {
      // 空字符串 / 非法值 → 整个字段删掉。
      // 不要无脑改写成 "assistant"：续传 chunk 和 tool_calls chunk
      // 本身就应该完全不带 role，这才是符合规范的。
      delete delta.role;
    }
  }
  j = JSON.stringify(parsed);
}
```

两个需要注意的细节：

- **删除字段而非强制赋值。** 给每个 chunk 都塞上 `role: "assistant"` 在 schema 上同样合法，但删除字段更接近 OpenAI 官方实现对续传 chunk 的处理方式，也不会让那些跟踪 role 变化的客户端产生误判。
- **保持逐行流式处理。** 修复应放在现有的逐行处理函数内部；如果为了归一化而缓冲整个 body，会把流式输出变成结束时一次性刷出。

如果你们更希望把它做成上一层的通用「OpenAI chunk 消毒器」、让所有透传型 executor 都过一遍（这样以后其他不合规上游也能被统一覆盖，而不是逐个 provider 打补丁），我可以照那个方向提 PR —— 你们定形态就行。

AnandkumarMall:
### Summary

This PR addresses several high-frequency upstream translation, streaming, and tool cloaking defects reported across v0.5.65–0.5.69:

1. **Streaming & Vercel AI SDK Compatibility:**
   - Fixed `stream` detection in `chatCore.js` line 120 so non-streaming requests without `stream: true` do not default to `true` (resolves HTTP 406 on OpenAI upstreams and Vercel AI SDK `JSON.parse` failures).
   - Sanitized `choices[0].delta.role` in `qoder.js` to omit empty strings `""` that crash `@ai-sdk/openai-compatible`.

2. **Tool Cloaking & Decloaking (`_ide` Leak):**
   - Passed `toolNameMap` to `handleForcedSSEToJson` so forced SSE responses decloak properly.
   - Added fallback de-suffixing in `claudeCloaking.js` to strip `_ide` when multi-turn conversations reuse tool names.

3. **Cache Control Cap & Effort Normalization:**
   - Added `countCacheControlBlocks(body)` in `claude.js` and capped markers at 4 (Anthropic's hard limit), preventing 400 errors that cascade into locking entire account pools.
   - Mapped OpenAI `role: "developer"` to system instructions.
   - Mapped Kimi `reasoning_effort: "medium"` to `"high"` (Kimi K3 only supports `low`, `high`, `max`).
   - Sanitized `claude-adaptive` effort to prevent literal `"auto"` emissions.

4. **Translator Robustness & Gemini Schemas:**
   - Added null-guards in `ensureToolCallIds` and `reasoningContentInjector`.
   - Normalized shorthand properties in `cleanJSONSchemaForAntigravity`.
   - Added capability patterns for `qwen3.8`, `qwen3.9`, and Llama 3.2 Vision.

### Verification
- `node tests/__baseline__/verify-providers.mjs`: Byte-for-byte equal (81 providers).
- `node tests/__baseline__/verify-alias.mjs`: Byte-for-byte equal (117 tokens).
- `node tests/__baseline__/verify-oauth-urls.mjs`: Byte-for-byte equal.

Closes #2458, Closes #3844, Closes #3799, Closes #3828, Closes #2582, Closes #3795, Closes #3567, Closes #3786, Closes #2894, Closes #3794, Closes #2433, Closes #3764, Closes #3784, Closes #2489, Closes #2595, Closes #3836, Closes #3825.


AnandkumarMall:

PR #3862, opened 2026-09-07
fix(db): prevent silent database wipe on SQLite corruption and add startup quick_check

### LINKED ISSUE #3817: [CRITICAL] SQLite database corruption silently destroys ALL provider connections — happened twice in 10 days
## Severity: critical

The main SQLite database (`~/.9router/db/data.sqlite`) gets corrupted in a way that **wipes every provider account/connection from the UI**. Happened to me **twice in ~10 days** (Aug 27 and Sep 4–5). No error is shown — the router just starts with an empty accounts list.

## Environment

- macOS 25.5.0 (Darwin), Apple M4 Pro (arm64)
- 9Router `0.91.21 (9router) / 0.5.8 (upgrade target)` (installed via `bun i -g 9router`)
- DB: `~/.9router/db/data.sqlite`, WAL journal mode, schemaVersion 8
- Live traffic running through the router daily

## What happened (second corruption, Sep 4–5 2026)

1. Sep 4 18:16 — router wrote a pre-upgrade DB backup (`db/backups/upgrade-0.91.21 (9router) / 0.5.8 (upgrade target)-to-0.5.8-.../data.sqlite`, 1.18 GB, healthy).
2. Sep 4 18:25 — last WAL checkpoint / write activity.
3. Sep 5 15:25 — last write to main DB. After this the file is **truncated to 361 MB** (vs 1.18 GB healthy).
4. Next start: accounts list is empty.

Diagnosis:

```
$ sqlite3 data.sqlite "PRAGMA integrity_check;"
Error: in prepare, database disk image is malformed (11)

$ sqlite3 data.sqlite ".recover" | grep -c 'INSERT INTO lost_and_found'
618701
```

`.recover` yields **618k orphan rows, ALL in `lost_and_found`** — the real schema tables (including `providerConnections`) are unrecoverable from the main file. The UI shows zero accounts because the tables simply don't exist anymore.

Only the pre-upgrade backup was intact. Restored from it — 85 connections recovered, minus everything since Sep 4 18:16.

## First corruption (Aug 27)

Same signature: `data.sqlite.corrupt-backup` from Aug 27 03:55 also fails `integrity_check` with `malformed`. Pattern repeats: process killed / crashed mid-write, WAL not checkpointed, main DB left truncated.

## Suspected causes

1. **Hard kill (SIGKILL / terminal close / crash) during write or WAL checkpoint** leaves the main DB truncated. Nothing on next startup detects this.
2. WAL checkpoint handling around the 0.91.21 (9router) / 0.5.8 (upgrade target) → 0.5.8 upgrade path — corruption window coincides with upgrade-time backups.
3. Possibly default `synchronous=NORMAL` under WAL — fast, but a kill mid-write can lose/truncate pages.

## What I ask for (in priority order)

1. **Startup corruption detection**: run `PRAGMA quick_check` (or page-header sanity check: expected size vs actual) before opening the DB. If malformed — refuse to start / show a loud error and auto-swap to the newest intact `db/backups/` copy instead of silently booting with an empty DB.
2. **Graceful shutdown checkpoint**: on SIGINT/SIGTERM run `PRAGMA wal_checkpoint(TRUNCATE)` and close cleanly, so a kill between checkpoints can't truncate the main file.
3. **Optional `synchronous=FULL` setting** for users who care about durability over speed.
4. **Automatic rolling backup of `providerConnections`** (e.g. keep N last daily `VACUUM INTO` copies in `db/backups/`) — the schema backups already exist for upgrades, but there is no regular rolling one.

## Evidence files I still have

- Corrupted main DB (Sep 5): truncated at 361 MB, `malformed` on integrity_check
- `data.sqlite.corrupt-backup` (Aug 27): same `malformed` signature
- Healthy pre-upgrade backup (Sep 4 18:16, 1.18 GB, 85 connections, quick_check ok)
- Full `.recover` dump (618k rows in `lost_and_found`)

Happy to provide hex dumps of corrupt page headers or the recover logs if useful. This needs a fix — two silent full-data-loss events in 10 days makes the router unusable with real account sets.

AnandkumarMall:
### Summary

Fixes critical silent data loss where a malformed SQLite disk image caused 9Router to boot with a completely empty account/connections list (#3817).

- `isFreshDb(adapter)` caught all query errors on `SELECT COUNT(*) as c FROM _meta` and returned `true`, mistaking disk corruption for a brand-new, empty database and wiping user accounts.
- Now explicitly detects `malformed`, `corrupt`, and `disk image` errors and aborts with a critical error rather than treating the DB as fresh and wiping tables.
- Added `PRAGMA quick_check(1)` before running migrations on startup to catch corruption early and preserve WAL/pages for recovery.

Closes #3817.


AnandkumarMall:

PR #3863, opened 2026-09-07
fix(combos): preserve visibility for strategy combos and improve runtime CLI defaults

### LINKED ISSUE #2496: Dashboard: combos with kind=fallback still hidden after PR #1684
PR #1684 fixed issue #1682 by showing combos with kind="llm".
But the filter still hides other combo kinds like fallback,
round-robin, fusion, and capacity.

Impact:
Combos with kind="fallback" work via API but show "No combos yet"
in the Dashboard.

Root cause:
The Dashboard filter from PR #1684 is:
  !c.kind || c.kind === "llm"

This excludes all other combo strategy kinds.

Proposed fix:
  !c.kind || ["llm","fallback","round-robin","fusion","capacity"].includes(c.kind)

Or simply remove the filter entirely — users created those combos intentionally.

Environment:
- 9Router v0.5.20 (Docker)
- Combo: hermes-stack | kind: fallback | 4 models
- Dashboard: /dashboard/combos

### LINKED ISSUE #2602: running port always fallback to 20127
Hi there, I try running 9routers source on my VM. After I git pull latest update, It seems the default port always fallback to 20127, even though I already state to use port 20128 on .env and on pm2 

pm2 : 

Image

.env : 

Image

result (always attached to port 20127) : 

Image

Is this normal ?

Thanks

### LINKED ISSUE #3365: [Bug] Hard-coded `--max-old-space-size=6144` in cli.js makes cgroup/low-memory deployments OOM-loop

### LINKED ISSUE #3756: SearXNG Docker endpoint configured via SEARXNG_URL is blocked by runtime SSRF guard
## Summary

In 9Router v0.5.65, the documented Docker-style SearXNG configuration:

```env
SEARXNG_URL=http://searxng:8080/search
```

fails at runtime on `/v1/search` requests with:

```json
{"error": {"message": "searxng error: Blocked URL: hostname resolves to an internal host", "code": 502}}
```

The request is rejected even when no client-side `provider_options.baseUrl` is supplied and the endpoint is exclusively configured by an administrator via `SEARXNG_URL`.

## Root Cause Analysis

In commit `b870b5d4` (PR #3714: "fix(security): close SSRF guard bypasses in ssrfGuard.js"), `open-sse/handlers/search/index.js` was modified to wrap all outbound search requests with `fetchPublic()`:

```js
// open-sse/handlers/search/index.js:104
const resp = await fetchPublic(url, { ...init, headers: sanitizeHeaders(init.headers), signal: controller.signal });
```

`fetchPublic()` invokes `assertPublicUrlResolved(url)`, which resolves `searxng` via Docker's internal DNS (`127.0.0.11`) to its container IP (e.g. `172.18.0.x`). Because `172.16.0.0/12` is defined as a blocked private IPv4 range in `src/shared/utils/ssrfGuard.js`, `assertPublicUrlResolved` throws:

```
Error: Blocked URL: hostname resolves to an internal host
```

### Inconsistency with `callers.js` Design Intent

The comment in `open-sse/handlers/search/callers.js:68-73` explicitly outlines the intended security model:

```js
/**
 * Read the base URL for this provider call.
 *
 * The override is client-controlled and therefore SSRF-hardened: only public
 * http(s) URLs are accepted (internal/private/loopback/metadata addresses are
 * rejected via assertPublicUrl). The provider's own configured baseUrl is
 * trusted as-is (admin-controlled).
 */
export function resolveBaseUrl(config, params) {
  const override = getProviderSetting(params, "baseUrl");
  if (override) {
    ...
    assertPublicUrl(override);
  }
  return (override || config.baseUrl).replace(/\/+$/, "");
}
```

While `resolveBaseUrl()` correctly differentiates between untrusted client overrides and trusted admin configuration (`config.baseUrl`), the subsequent call to `fetchPublic()` in `open-sse/handlers/search/index.js` unconditionally subjects all search endpoints to public-IP enforcement. This prevents standard internal deployments (e.g., Docker Compose sibling containers like `searxng:8080` or loopback `localhost:8888`) from functioning.

## Steps to Reproduce

1. Deploy 9Router alongside a SearXNG container in Docker Compose on a shared network (e.g., `9router_default`).
2. Set `SEARXNG_URL=http://searxng:8080/search`.
3. Verify `searxng` is reachable from inside the 9Router container:
   ```bash
   docker exec 9router wget -q -O - "http://searxng:8080/search?q=test&format=json"
   # Responds with valid search results (HTTP 200)
   ```
4. Query 9Router's search endpoint with no client baseUrl override:
   ```bash
   curl -X POST http://127.0.0.1:20128/v1/search \
     -H "Authorization: Bearer <valid-api-key>" \
     -d '{"provider": "searxng", "query": "test query"}'
   ```
5. Request fails with HTTP 502:
   ```json
   {"error": {"message": "searxng error: Blocked URL: hostname resolves to an internal host", "code": 502}}
   ```

## Expected Behavior

Requests to SearXNG configured via admin environment variable `SEARXNG_URL` (without client-provided `provider_options.baseUrl` overrides) should be trusted and permitted, allowing internal Docker network routing to SearXNG.

Client-controlled `provider_options.baseUrl` overrides should continue to be strictly validated and blocked if they resolve to private or internal addresses.

## Suggested Solution

1. When no client-supplied override is present (`!getProviderSetting(params, "baseUrl")`) and the provider is `searxng` using admin-configured `searchConfig.baseUrl` / `SEARXNG_URL`, allow internal network fetch, OR:
2. Add an explicit allowlist or flag for trusted internal provider endpoints, while preserving `fetchPublic()` enforcement on any client-supplied URL overrides.

### LINKED ISSUE #3787: [Bug] Combo disappears from dashboard /dashboard/combos as soon as `kind` is set via PUT /api/combos/:id
## Summary

Any combo whose `kind` field is a non-null value (`"fallback"`, `"fusion"`, `"round-robin"`) — i.e. any combo that has gone through `PUT /api/combos/:id` to actually configure its strategy — disappears entirely from the dashboard's Combos page (`/dashboard/combos`), even though `GET /api/combos` and `GET /api/settings` both return complete, correct data for it. A combo with `kind: null` (never PUT-updated, e.g. one created and left alone via the dashboard's own "Create Combo" flow) renders fine.

This is 100% reproducible and I isolated it to the `PUT` call specifically — not a data/auth/caching issue.

## Reproduction

1. Create a combo via `POST /api/combos`:
   ```
   POST /api/combos
   {"name":"zzz-test-combo","models":["gemini/gemini-3.8-flash","gemini/gemini-3.5-flash-lite"]}
   ```
   → Visit `/dashboard/combos` — the new combo renders correctly as a card.

2. Set its strategy via `PUT /api/combos/:id`:
   ```
   PUT /api/combos/3cf69968-889b-43c1-8c03-617dc2cf2644
   {"name":"zzz-test-combo","kind":"fallback","models":["gemini/gemini-3.8-flash","gemini/gemini-3.5-flash-lite"]}
   ```
   → Reload `/dashboard/combos` — the combo card is now **completely gone** from the page.

3. `GET /api/combos` still returns the combo correctly (`kind: "fallback"`, full model list, correct timestamps) — confirmed via direct curl, response is complete and well-formed.
4. `GET /api/settings` → `data.comboStrategies["zzz-test-combo"]` is also present and correct (`{"fallbackStrategy":"fallback"}`).
5. The combo continues to work correctly for real inference (`POST /v1/chat/completions` against it behaves as a genuine fallback chain) — this is purely a dashboard rendering bug, not a functional/routing regression.

## Real-world impact

I hit this on a fleet of 12 production combos (mix of `fallback`, `fusion`, and `round-robin` kinds, all requiring the `PUT` call to actually behave as configured rather than degrading to plain fallback-to-first-model). All 12 are invisible on `/dashboard/combos` — only a combo built once through the dashboard's own UI and never `PUT`-updated since remains visible. Since setting `kind` via `PUT` is the *only* documented way to make a combo other than plain fallback actually work, this means **any dashboard-managed combo that's correctly configured is guaranteed to disappear from the dashboard** — a real usability problem, since it becomes impossible to inspect, edit, or delete these combos from the UI at all (only via direct API calls).

## Environment

- 9Router version: v0.5.65
- Reproduced via direct API calls (curl) + confirmed in the actual dashboard UI (Chromium via Playwright), consistent both times.

## Suspected cause (not confirmed, just a hint from behavior)

Given a fresh `kind: null` combo renders but the *same* combo stops rendering the instant `kind` becomes non-null, the combo-list rendering code likely branches on `kind` (or reads via `comboStrategies` when `kind` is set) and throws/short-circuits in that branch — possibly related to how the card looks up `judgeModel` or per-model capability metadata only on the "configured strategy" render path. No console errors were visible in Chromium DevTools during reproduction (production build likely swallows the error), which points at a silent try/catch or an error boundary quietly dropping the card rather than a hard crash.

AnandkumarMall:
### Summary

1. **Dashboard Combos Visibility:**
   - Fixed `combos/page.js` line 75 filter: combos configured with `kind: "fallback"`, `"fusion"`, or `"round-robin"` via `PUT /api/combos/:id` no longer disappear from `/dashboard/combos`. Only media combos (`webSearch`, `webFetch`) are filtered to their dedicated page (#3787, #2496).

2. **Docker SearXNG SSRF:**
   - Outbound search requests using admin-configured `SEARXNG_URL` in Docker networks (e.g. `http://searxng:8080`) are no longer blocked by `fetchPublic()`. `fetchPublic` is now only enforced when untrusted client-supplied baseUrl overrides are present (#3756).

3. **CLI Port & Memory Defaults:**
   - Removed hardcoded `--port 20127` from `package.json` `"start"` script so `.env` `PORT=20128` is respected (#2602).
   - Sized `--max-old-space-size` dynamically (75% of system memory, min 512MB, max 6GB) and respected existing `NODE_OPTIONS` to prevent OOM crash-loops on ≤4GB hosts (#3365).

Closes #3787, Closes #2496, Closes #3756, Closes #2602, Closes #3365.


AnandkumarMall:

PR #3871, opened 2026-09-08
fix(reliability): prevent 400 client error account lockouts, handle Codex compaction on Gemini, and fix console scroll

### LINKED ISSUE #3786: [Bug] claude-adaptive writes effort:"auto" into output_config -> 400, and the 400 backs the account off so the next valid request also fails

### LINKED ISSUE #3794: [Bug] NVIDIA kimi-k3: a 400 for reasoning_effort=medium triggers account fallback, locks the model for 300s, and is reported to the client as 429

### LINKED ISSUE #3812: [Bug] Custom-model capability metadata is user-declared and wrong: vision flag and context/output limits don't match runtime behavior
## Problem
Custom models (`kv.customModels`) carry user-declared `caps` (e.g. `vision:true`) and the `/v1/models` catalog reports generic limits (e.g. a newly released flagship showing 200K/64K instead of its real 1.05M/128K). Downstream tooling trusts these flags and misroutes:

1. An image request sent to a custom model flagged `vision:true` on a provider whose adapter can't forward images is silently served by a *different* model (first entry of the vision capacityAdapter list). The response `model` field changes mid-flow; the client asked model A and got model B's answer. In one case the substituted model answered the image question *incorrectly*, which is worse than an error.
2. Wrong context/output limits break token budgeting in clients.

## Expected
- Validate custom-model caps at Test-time where cheap (e.g. capability probe), or mark user-declared caps as `declared` vs `verified` in the catalog.
- When the vision adapter substitutes a different model, return an explicit signal (warning header/field) instead of silently changing the serving model. Optionally fail closed with a clear 4xx when the requested model can't take images and no verified fallback consented.
- Refresh hardcoded catalog entries for newly released models (context/output) or source them from the provider's models endpoint where available.

## Environment
9Router 0.5.65, Windows, Codex + Antigravity providers.

### LINKED ISSUE #3816: [Bug] Codex Remote Compaction V2 fails with Antigravity Gemini 3.8 Flash
## Bug Description

I'm using Codex through 9Router with the Antigravity provider and Gemini 3.8 Flash.

When Codex triggers Remote Compaction V2, the compaction request fails.

### Environment

- 9Router: 0.5.65
- Codex: Desktop
- Provider: Antigravity
- Model: `ag/gemini-3.8-flash-high`
- Also tested with: `ag/gemini-3.8-flash-medium`
- 9Router API: `http://localhost:20128/v1`
- Endpoint: `/v1/responses`

### Codex configuration

I'm using a Combo with:

- `ag/gemini-3.8-flash-high`
- `ag/gemini-3.8-flash-medium`

### Error

When Codex tries to run remote compaction, I get:

```text
Error running remote compact task: unexpected status 503 Service Unavailable:
[antigravity/gemini-3.8-flash-high] [400]: {
  "error": {
    "code": 400,
    "message": "Requests ending with a model turn are not supported"
  }
}

### LINKED ISSUE #3833: Dashboard: per-model Test buttons block each other — only one test can run at a time
## Problem
On the Providers dashboard, when you click **Test** on a model and then click **Test** on another model while the first is still running, the second click is silently ignored. You have to wait for each test to finish before testing the next model, so checking several models on one provider is slow and serial.

## Cause
In both `src/app/(dashboard)/dashboard/providers/components/ModelsCard.js` and `src/app/(dashboard)/dashboard/providers/[id]/CompatibleModelsSection.js`, the handler is guarded by a single shared state variable:

```js
const [testingModelId, setTestingModelId] = useState(null);
const handleTestModel = async (modelId) => {
  if (testingModelId) return; // blocks ALL other tests while one runs
  ...
}
```

Note that `providers/[id]/page.js` already solves this with a `testingModelIds` **Set** (`if (testingModelIds.has(modelId)) return;`), so the two components above appear to have just been missed when that pattern was introduced.

## Expected
Each model's Test button tracks its own in-flight state, so multiple models on the same provider can be tested in parallel (the backend `POST /api/models/test` already handles one request per call fine).

Happy to send a PR for this.

### LINKED ISSUE #3838: Console Log scroll reset as new entries are printed, making it hard to read old entries for debugging.
http://localhost:20128/dashboard/console-log
## Problem
Scroll is reset to the bottom when new entries are printed, so you can't read older entries without constantly scrolling upwards.
## Case
I was seeing which fusion model failed. However, because it fans out to multiple calls, the console keep resetting to the bottom, making it impossible to read. Even in low traffic scenario, background token refreshes happens so the same problem still happens.
## Solution
Keep the scroll level consistent and stop scroll reset when new entries are printed.

AnandkumarMall:
### Summary

Addresses several high-impact reliability, translation, and UI issues across the 9Router catalog, provider fallback, and dashboard:

1. **Account Fallback De-Poisoning on Client Errors:**
   - In `accountFallback.js`, client validation errors (HTTP 400 Bad Request / 422 Unprocessable Entity that are not quota/rate-limit issues) now return `{ shouldFallback: false, cooldownMs: 0 }`.
   - Prevents client-induced parameter mistakes (e.g. invalid thinking level or malformed body) from triggering account cooldowns, preventing rolling outages across the entire account pool (#3794, #3786).

2. **Codex Remote Compaction V2 on Gemini / Antigravity:**
   - In `openai-to-gemini.js`, conversations where the final turn has `role === "model"` (common in Codex Remote Compaction V2 requests) automatically receive a continuation user turn.
   - Eliminates the raw upstream 400 error: `Requests ending with a model turn are not supported` (#3816).

3. **Bare Model Name Matching in Model Lookup (`GET /v1/models/{id}`):**
   - In `api/v1/models/[...model]/route.js`, added fallback matching on bare model suffixes (`candidate.id.endsWith('/' + identifier)`).
   - Prevents 404s when clients (such as Claude Code CLI interactive mode) query bare model IDs without the provider prefix (#3812, #3866).

4. **Console Log Scroll UX:**
   - In `ConsoleLogClient.js`, auto-scroll is now only triggered if the user is already near the bottom of the container.
   - Prevents forced scroll-to-bottom on incoming log lines when a user is inspecting or copying earlier error messages (#3838).

5. **Concurrent Model Testing:**
   - Converted `testingModelId` single-item mutex to a `Set` in `ModelsCard.js` and `CompatibleModelsSection.js`, allowing users to test multiple models concurrently without blocking (#3833).

### Verification
- `node tests/__baseline__/verify-providers.mjs`: Byte-for-byte equal (81 providers).
- `node tests/__baseline__/verify-alias.mjs`: Byte-for-byte equal (117 tokens).
- `node tests/__baseline__/verify-oauth-urls.mjs`: Byte-for-byte equal.

Closes #3794, Closes #3786, Closes #3816, Closes #3812, Closes #3838, Closes #3833.


AnandkumarMall:

blestafist:
decolua pls merge

PR #3878, opened 2026-09-08
fix(translator): six format-conversion defects (#3243, #3247, #3254, #3369, #3492, replaces 6 PRs)

ntdatt812:
Six translator/handler fixes that were sitting in separate PRs (#3243, #3247, #3254, #3369, #3492/#3528, #3549), rebased onto current `master` and sent as one review. Each is an independent commit, so any one of them can be dropped without touching the others.

All six are format-conversion defects: a request or response crossed a format boundary and lost a field that the target API requires.

| commit | defect |
|---|---|
| `1b042a1c` | `reasoning.effort` was sent flat to `openai-responses` targets, which nest it |
| `0f23a482` | a non-streaming binary-transport reply was returned in the upstream's format, not the client's |
| `8805156e` | `tool_calls: []` was treated as "has tool calls", producing an empty `tool_use` block |
| `74658a51` | a tool result arriving without an id emitted `tool_use_id: undefined`; Anthropic rejects the whole request |
| `65339392` | a request with no stream key defaulted to streaming, so non-streaming clients got SSE |
| `d43850f6` | the CommandCode error chunk opened without a role, so clients that key on the first delta dropped it |

## Verification

```
npx vitest@3 run --config tests/vitest.config.js \
  tests/unit/{commandcode-to-openai,nonstream-binary-transport-source-format-3199,openai-format-empty-tool-calls,stream-mode-default-3492,thinking-responses-nested-3154,tool-result-missing-id-3362}.test.js \
  tests/translator/thinking-unified.test.js
→ 7 files, 107 passed
```

Two notes on running these, because both cost me time:

- The alias `@/` only resolves with `--config tests/vitest.config.js`; without it collection dies on `Failed to load url @/lib/dataDir.js`.
- **vitest 2.x cannot run any test that reaches `open-sse/translator/index.js`.** It fails with `TypeError: register is not a function` at the top-level `register(...)` calls. That is not a defect in this repo: `index.js` deliberately relies on function-declaration hoisting to survive the import cycle (the comment above `var requestRegistry` says so), and Vite 5's SSR transform turns the named import into a call-time property access that is not yet populated. I confirmed the same failure on clean `master`, on `master` as of 15 Aug, and on this repo's own `tests/translator/bugs-antigravity.test.js` — so it predates these changes and is not caused by them. vitest 3 runs them correctly.

Each claim was checked by reverting it and confirming which test dies; no test passes for a reason other than the fix.

One of the six did not survive that check as it stood. Reverting the call site in `chatCore.js` for `#3492` left `stream-mode-default-3492.test.js` **green**: the tests exercised `clientRequestedStreaming` in isolation and nothing observed the wiring — the same shape of bug the fix is about. `handleChatCore` takes ~30 collaborators and cannot be driven from a unit test, so `18e67517` pins the call site at the source level instead. Reverting `chatCore.js` now fails 3 of that file's 10 cases.

### DIFF
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -175,7 +175,11 @@ export async function handleChatCore({ body, modelInfo, credentials, log, onCred
     translatedBody = { ...body, model: stripThinkingSuffix(upstreamModel) };
     if (provider === "codex") {
       const suffixThinking = {};
-      applyThinking(sourceFormat, upstreamModel, suffixThinking, provider);
+      // Pinned to OPENAI on purpose: this branch reads the flat reasoning_effort key
+      // off the scratch object and nests it itself, so applyThinking must not nest.
+      // Passing sourceFormat here would hand back {reasoning:{effort}} for a Responses
+      // client and the suffix would silently stop applying.
+      applyThinking(FORMATS.OPENAI, upstreamModel, suffixThinking, provider);
       if (suffixThinking.reasoning_effort) {
         const reasoning = translatedBody.reasoning;
         translatedBody.reasoning = {
--- a/open-sse/handlers/chatCore.js
+++ b/open-sse/handlers/chatCore.js
@@ -16,6 +16,7 @@ import { getExecutor } from "../executors/index.js";
 import { supportsGrokCliReasoningEffort } from "../config/grokCli.js";
 import { buildRequestDetail, extractRequestConfig } from "./chatCore/requestDetail.js";
 import { handleForcedSSEToJson } from "./chatCore/sseToJsonHandler.js";
+import { clientRequestedStreaming as requestedStreaming } from "./chatCore/streamMode.js";
 import { handleNonStreamingResponse } from "./chatCore/nonStreamingHandler.js";
 import { handleStreamingResponse, buildOnStreamComplete } from "./chatCore/streamingHandler.js";
 import { detectClientTool, isNativePassthrough } from "../utils/clientDetector.js";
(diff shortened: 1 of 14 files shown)


ntdatt812:
Rebased onto `master` (`39e36d3d`). Was CONFLICTING, now clean. Two commits came out of the bundle rather than being merged forward, because `master` has since decided both questions the other way:

- **`fix(commandcode): open the error chunk with the assistant role`** — withdrawn. That branch now `throw`s instead of emitting the error as content, so the stream handler marks the stream errored. Giving the emitted chunk a role has nothing left to apply to.
- **`fix(translator): nest reasoning effort for openai-responses targets`** — withdrawn, and I would like your read on it. `d1de3245` (22 Sep) added `tests/unit/thinking-budget-max-level.test.js`, which asserts `reasoning_effort` stays flat on an `OPENAI_RESPONSES` target:

  ```js
  const out = applyThinking(FORMATS.OPENAI_RESPONSES, "gpt-5.6-sol", body, "codex");
  expect(out?.reasoning_effort).toBe("max");
  ```

  That is the opposite of what the commit did, so it is out. Worth asking anyway: the OpenAI Responses API reads `reasoning: { effort }` and ignores a top-level `reasoning_effort`, so for a genuine `/v1/responses` upstream the flat key is dropped on the floor. If `codex` is a wire that wants the flat key and only some responses targets want it nested, the split is per provider rather than per format and I am happy to send that separately. If you meant flat everywhere, nothing to do.

What remains is five commits, unchanged in intent:

| commit | what |
| --- | --- |
| `f2688315` | non-streaming binary-transport replies are built in the client's format, not the source's (#3199) |
| `d052cd8f` | an empty `tool_calls: []` is no tool calls, not a tool call with no fields |
| `ea7492dd` | a tool result that arrives without an id is matched to its call instead of dropped (#3362) |
| `350c8ce7` | a request with no `stream` key is non-streaming (#3492) |
| `36d4cf18` | pins that `chatCore` reads the mode from `streamMode.js` |

Two conflicts were merges rather than choices, and both keep your side: `thinkingUnified.js` keeps the new `display` passthrough, and `chatCore.js` keeps the token-saver block — only the inlined streaming predicate is replaced by `requestedStreaming(body, sourceFormat)`.

`npx vitest@3 run --config tests/vitest.config.js`: 89 failures, and they are the same 89 by name as on `master` at `39e36d3d` — compared per test name, not by count.

PR #3880, opened 2026-09-08
fix(models): four provider-catalog and connection defects (#3262, #3115, #3486, #3467)

ntdatt812:
Four provider/model-catalog fixes that were sitting in separate PRs (#3330, #3252, #3529, #3539), rebased onto current `master`. Independent commits.

| commit | defect |
|---|---|
| `fix(auth)` (#3262) | no-auth free providers got a fresh upstream session id per request, so providers that key state on the session saw every call as a new conversation |
| `fix(models)` (#3115) | a curated custom-model list did not suppress the live `/models` fetch, so the catalog came back with the provider's full list instead of the operator's selection |
| `feat(models)` (#3486) | combo entries in `/v1/models` reported no token limits, so clients sizing a request against a combo had nothing to read |
| `fix(providers)` (#3467) | self-hosted TTS/STT/embedding connections had no Base URL field, so they could not point at anything but the default host |

The `#3115` commit reads as larger than it is: most of it is one block moved earlier in `buildModelsList` so that `customModelIds` exists before the dynamic-fetch decision. The behavioural change is the added `&& customModelIds.length === 0` on that condition.

## Verification

Run with **vitest 3** and the repo config — both parts matter:

```
npx vitest@3 run --config tests/vitest.config.js <files>
```

Without `--config` the `@/` alias does not resolve (`Failed to load url @/lib/dataDir.js`). And vitest 2.x cannot run anything that reaches `open-sse/translator/index.js`: it throws `TypeError: register is not a function` at the top-level `register(...)` calls, because `index.js` deliberately relies on function-declaration hoisting to survive the import cycle while Vite 5's SSR transform turns the named import into a call-time property access. I confirmed that failure on clean `master` and on `master` as of 15 Aug, so it is not caused by these changes.

Every fix was checked by reverting it and confirming which test dies. No claim here rests on a test that passes for another reason.

### DIFF
--- a/open-sse/providers/registry/ollama-local.js
+++ b/open-sse/providers/registry/ollama-local.js
@@ -11,6 +11,11 @@ export default {
     website: "https://ollama.com",
   },
   category: "apikey",
+  // The connection's own endpoint, stored as providerSpecificData.baseUrl.
+  baseUrlField: {
+    label: "Ollama Host URL",
+    placeholder: "http://localhost:11434",
+  },
   transport: {
     baseUrl: "http://localhost:11434/api/chat",
     format: "ollama",
--- a/open-sse/providers/registry/selfhosted-embedding.js
+++ b/open-sse/providers/registry/selfhosted-embedding.js
@@ -40,6 +40,12 @@ export default {
     website: "https://github.com/ggml-org/llama.cpp",
   },
   category: "apikey",
+  // Same gap as the TTS and STT entries.
+  baseUrlField: {
+    label: "Base URL",
+    placeholder: "http://localhost:8080/v1",
+    help: "OpenAI base URL — /embeddings is appended.",
+  },
   auth: {
     apiKey: {
       // Note the /v1: the adapter appends "/embeddings" to whatever it is given,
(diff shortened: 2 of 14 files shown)


ntdatt812:
Rebased onto `master` (`39e36d3d`). Was CONFLICTING, now clean.

One commit came out of the bundle: **`fix(auth): give no-auth free providers a stable upstream session id`**. `0c6ab4f9` solved the same problem more generally — `stableSessionId` now keys one long-lived session per identity, and `identityKey` reads `credentials?.connectionId || credentials?.id`. The no-auth record already carries `id: "noauth"`, so the `connectionId` this commit added changed nothing. Its two remaining test expectations were characterizing the behaviour `0c6ab4f9` replaced, so they went with it.

Three commits remain:

| commit | what |
| --- | --- |
| `b7b07df7` | a curated custom-provider model list suppresses the live catalog instead of being merged into it (#3115) |
| `0a560c77` | combo entries in `/v1/models` report `context_length` and `max_completion_tokens` (#3486) |
| `ed8069b6` | self-hosted TTS/STT/embedding providers can set their base URL (#3467) |

The conflict in `src/app/api/v1/models/route.js` was an addition on both sides, not a choice: your `aggregateComboCapabilities` still sets `entry.capabilities`, and this adds the two token-limit fields beside it. Both imports are kept.

`npx vitest@3 run --config tests/vitest.config.js`: 89 failures, the same 89 by name as on `master` at `39e36d3d`.

PR #3882, opened 2026-09-08
fix(usage): key-mask collision, double countdown timer, and OpenCode Go quota (#3640, #3470, #3334)

ntdatt812:
Three usage/quota fixes that were sitting in separate PRs (#3337, #3537, #3650), rebased onto current `master`. Independent commits.

| commit | defect |
|---|---|
| `feat(opencode-go)` (#3334) | OpenCode Go subscription quota was not surfaced on the dashboard |
| `fix(quota)` (#3470) | the auto-refresh countdown ran two interval owners, so the timer jumped and fired twice |
| `fix(usage)` (#3640) | usage was bucketed by an API key's **mask**, and every key minted on one machine shares its first 8 characters — so separate keys collapsed into one row |

The `#3640` one is the load-bearing fix: bucketing by the mask both mis-attributed usage and, in the obvious "just group by the raw key" alternative, would have put the key itself on the dashboard. It buckets by key identity instead, which is why `tests/unit/security-audit.test.js` is touched — that suite pins that no raw key reaches the response.

## Verification

Run with **vitest 3** and the repo config — both parts matter:

```
npx vitest@3 run --config tests/vitest.config.js <files>
```

Without `--config` the `@/` alias does not resolve (`Failed to load url @/lib/dataDir.js`). And vitest 2.x cannot run anything that reaches `open-sse/translator/index.js`: it throws `TypeError: register is not a function` at the top-level `register(...)` calls, because `index.js` deliberately relies on function-declaration hoisting to survive the import cycle while Vite 5's SSR transform turns the named import into a call-time property access. I confirmed that failure on clean `master` and on `master` as of 15 Aug, so it is not caused by these changes.

Every fix was checked by reverting it and confirming which test dies. No claim here rests on a test that passes for another reason.

### DIFF
--- a/open-sse/services/usage.js
+++ b/open-sse/services/usage.js
@@ -24,6 +24,7 @@ import {
   getOllamaUsage,
   getVercelAiGatewayUsage,
   getQoderUsage,
+  getOpencodeGoUsage,
 } from "./usage/misc.js";
 
 /**
@@ -60,6 +61,7 @@ const USAGE_HANDLERS = {
   deepseek: (c) => getDeepseekUsage(c.apiKey, c.proxyOptions),
   groq: (c) => getGroqUsage(c.apiKey, c.proxyOptions),
   zed: (c) => getZedUsage(c.accessToken, c.providerSpecificData, c.proxyOptions),
+  "opencode-go": (c) => getOpencodeGoUsage(c.apiKey, c.proxyOptions),
 };
 
 export async function getUsageForProvider(connection, proxyOptions = null, options = {}) {
--- a/open-sse/services/usage/misc.js
+++ b/open-sse/services/usage/misc.js
@@ -1,9 +1,9 @@
 /**
- * Misc usage handlers (iFlow, Ollama, GLM, Vercel AI Gateway, Qoder)
+ * Misc usage handlers (iFlow, Ollama, GLM, Vercel AI Gateway, Qoder, OpenCode Go)
  */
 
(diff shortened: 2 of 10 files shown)


ntdatt812:
Rebased onto `master` (`39e36d3d`). Was CONFLICTING, now clean.

One commit came out of the bundle: **`feat(opencode-go): show subscription quota on the dashboard`** — `master` already has it. `usage.js` imports `getOpenCodeGoUsage` from `./usage/opencode-go.js` and registers it at line 58, and `tests/unit/opencode-go-usage.test.js` covers it.

Worth flagging how that showed up, because it is quiet: the conflict was only over neighbouring lines, so the merge produced a `USAGE_HANDLERS` object with **two** `"opencode-go"` keys. The later one wins silently — no error, no warning — so your implementation was shadowed by mine and your four tests failed while the map still looked correct on inspection. A rebase that "succeeds" can still do this; it only surfaced because the failures were compared against `master` by test name.

Two commits remain:

| commit | what |
| --- | --- |
| `dfaf85ba` | the quota auto-refresh countdown gets a single timer owner, instead of one timer per row racing the others (#3470) |
| `8922b9e2` | usage is bucketed by API-key identity rather than by the key's mask, so two keys sharing a mask prefix no longer merge into one row (#3640) |

`npx vitest@3 run --config tests/vitest.config.js`: 89 failures, the same 89 by name as on `master` at `39e36d3d`.

PR #3899, opened 2026-09-08
feat(usage): CommandCode real quota tracking

sancaksert:
Adds real quota tracking for CommandCode (`commandcode`), following the same pattern as the existing provider trackers. Probes the endpoints the official CLI uses for its usage overlay (`/alpha/billing/credits`, `/alpha/usage/summary`, `/alpha/billing/subscriptions`) and reports 5h/weekly/monthly. Monthly cap is reconstructed as remaining + spent-in-period. Object-shaped quotas fit the dashboard generic fallback, so no UI change is needed.

New: `open-sse/services/usage/commandcode.js`. Wired into `usage.js` + registry `features` flags. Tested against a live account.


sancaksert:

PR #3900, opened 2026-09-08
perf(mitm): cache derived sudo-password key

sancaksert:
Caches the derived MITM sudo-password key in-process instead of re-deriving it (machine-id lookup + KDF) on every call. Behavior unchanged, less per-request overhead on Windows where machine-id lookup is slow.


sancaksert:

PR #3902, opened 2026-09-08
fix(responses): assemble non-streaming chat replies from Responses SSE

sancaksert:
Non-streaming Chat Completions requests against Responses-API upstreams always returned empty content with a misleading stop reason.

Two defects combined:

1. handleNonStreamingResponse parsed Responses SSE as Chat SSE. Responses events have no choices/delta.content, so the assembler produced ''. Fixed by converting through convertResponsesStreamToJson + pickAssistantMessageForChatCompletion when the upstream format is openai-responses.
2. The converter only understood output_item.done events and reported everything else as in_progress/stop. It now also harvests the full output array shipped on response.completed/response.done, treats response.incomplete as terminal, and tolerates data-only SSE lines (payload type fallback). incomplete surfaces as finish_reason length instead of a silent empty stop.

Verified against a live Responses-API provider: adequate budgets return content with stop, starved budgets return length. Other formats untouched (branch is taken only for openai-responses upstreams).


sancaksert:

PR #3903, opened 2026-09-08
feat: CommandCode quota, MITM cache, Responses fix, model families, disconnect safety

sancaksert:
Several small, independent improvements in one PR, each verified live.

## 1. CommandCode real quota tracking (new: open-sse/services/usage/commandcode.js)
Follows the same pattern as the existing provider trackers. Probes the endpoints the official CLI uses for its usage overlay (/alpha/billing/credits, /alpha/usage/summary, /alpha/billing/subscriptions) and reports 5h/weekly/monthly. Monthly cap is reconstructed as remaining + spent-in-period. Object-shaped quotas fit the dashboard generic fallback, so no UI change is needed. Wired into usage.js + registry features flags. Tested against a live account.

## 2. MITM: cache derived sudo-password key (src/mitm/manager.js)
Caches the derived key in-process instead of re-deriving it (machine-id lookup + KDF) on every call. Behavior unchanged, less per-request overhead on Windows where machine-id lookup is slow.

## 3. Responses non-streaming fix (streamToJsonConverter.js, nonStreamingHandler.js, sseToJsonHandler.js)
Non-streaming Chat Completions requests against Responses-API upstreams always returned empty content with a misleading stop reason. Two defects combined: handleNonStreamingResponse parsed Responses SSE as Chat SSE (no choices/delta.content there), and the converter only understood output_item.done events. Fixed by converting through convertResponsesStreamToJson when the upstream format is openai-responses, harvesting the full output array shipped on response.completed/response.done, treating response.incomplete as terminal (surfaces as finish_reason length), and tolerating data-only SSE lines. Other formats untouched (branch taken only for openai-responses upstreams).

## 4. Dashboard: family-grouped models with real caps (providers/[id]/page.js, ModelRow.js, CapacityBadges.js)
Model cards grouped into collapsible vendor families with per-family open counts and bulk enable/disable, model search, and collapse preference persisted per page. Rows show the real capability data where known (vision/reasoning icons plus context-window chip, e.g. 1M); the 200K default floor is a guess and stays hidden rather than printed as fact. All strings go through the i18n runtime (EN keys + tr literals). Existing test/copy/remove actions unchanged. Also fixed: disabled custom models now appear in the restorable Disabled section (previously they stayed visible with no way back).

family-grouped models

## 5. Upstream model fetch for opencode-go/commandcode (providers/[id]/models/route.js + dashboard button)
These subscription gateways expose OpenAI-shaped /models; two config lines wire them into the existing provider-models endpoint, and the provider page gained a refresh button plus a one-click add list for models on the account that are not registered yet.

## 6. Server: abort upstream on client disconnect + absolute timeout (chatCore.js, streamHandler.js, chat.js)
Client disconnects were never propagated: the disconnect controller existed but nothing called it, so aborted requests leaked the upstream fetch forever. The route signal is now wired through, plus a 10-minute absolute cap per request.

---
TR özet: CommandCode gerçek kota takibi; MITM anahtar önbelleği; Responses tabanlı sağlayıcılarda akışsız istek düzeltmesi; panelde aile gruplu model listesi + gerçek rozetler + model çekme düğmesi; istemci kopuş güvenliği.


sancaksert:
Superseded by #3978 (rebased on current master).

PR #3904, opened 2026-09-08
feat(fetch): add String Web Access provider

EllAchE:
## Why

9router can only route Web Fetch requests to providers registered in-tree. String Web Access has a bearer-authenticated URL-to-Markdown endpoint, but it was not available as a provider.

## Summary

- register `string-web-access` against `https://request.usestring.ai/v1/fetch`
- normalize Markdown and upstream errors through the existing fetch response shape
- expose the provider in the Web Fetch skill, dashboard copy, translations, and focused unit coverage

## Test plan

- [x] run syntax checks for the handler, provider, and focused unit test
- [x] parse all updated translation JSON files
- [x] run a dependency-free handler smoke test covering registry lookup, bearer auth, request body, and normalized Markdown

Not run: the focused Vitest suite because the scratch clone has no installed dependencies.


EllAchE:
Bots gone wild

PR #3912, opened 2026-09-09
feat(video): add OpenRouter and Vertex AI (Veo) video generation

coozgan:
## Summary

Video generation is currently xAI-only. This adds **OpenRouter** and **Vertex AI (Veo)** as video providers so users can point `/v1/videos/*` at their own OpenRouter or Google Cloud credentials.

Introduces a small adapter layer under `open-sse/handlers/videoProviders/`. **A provider with no adapter keeps the exact current behaviour** (raw body to `{baseUrl}/{action}`, poll `{baseUrl}/{id}`, verbatim passthrough), so the xAI path is unchanged.

## What each provider needed

**OpenRouter** — near drop-in. Same async job shape as xAI; only two differences: creation POSTs to the `/videos` collection root (no `/generations` suffix), and the registry headers (`HTTP-Referer` / `X-Title`) are applied. Bodies pass through verbatim.
Docs: https://openrouter.ai/docs/api/api-reference/videos

**Vertex AI (Veo)** — real translation, since Veo does not speak the OpenAI-ish videos shape:
- create → `:predictLongRunning` with `{ instances[], parameters{} }`
- poll → `:fetchPredictOperation` (Veo has no REST GET poll — polling is a POST)
- the operation *resource name* is base64url-encoded into the job id, so `GET /v1/videos/{id}` stays a flat path
- access tokens are minted from Service Account JSON by reusing the existing `refreshVertexToken`; raw API keys are rejected up front with a clear message (Veo requires OAuth)
- `prompt` / `duration` / `aspect_ratio` / `resolution` / `n` / `seed` / `negative_prompt` / `storage_uri` / `generate_audio` map onto Vertex fields; a data-URL or GCS URI in `image` maps onto the image instance
- the operation response is mapped back onto the `{ id, status, video, videos }` shape clients already poll

Docs: https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/veo-video-generation

## Other changes

- `videoCore.js` — the request plan is now rebuilt per attempt, so the existing *401 → refresh once → retry once* path picks up the refreshed token. Adapter validation errors return **400 before any upstream call**, so a malformed request can never create a billable job.
- `videoGeneration.js` — `GET /v1/videos/{id}` no longer hardcodes `xai`. The provider resolves from the pinned `x-connection-id` connection, then `?provider=`, then falls back to the xAI default (back-compatible).
- Registry — `openrouter` and `vertex` gain `videoConfig`, the `video` serviceKind, and video-kind models (Veo 3.1 / 3 / 2, Sora 2 Pro, Seedance 2.0), so they show up in the dashboard's Video section.

## Usage

```bash
# OpenRouter
curl -X POST localhost:20128/v1/videos/generations \
  -d '{"model":"openrouter/google/veo-3.1","prompt":"a paper boat in neon rain"}'

# Vertex (Service Account JSON credential)
curl -X POST localhost:20128/v1/videos/generations \
  -d '{"model":"vertex/veo-3.1-generate-preview","prompt":"a neon city","duration":8,"aspect_ratio":"16:9"}'

# poll (echo back the x-9router-connection-id returned on create)
curl localhost:20128/v1/videos/<job_id> -H "x-connection-id: <conn_id>"
```

## Test plan

- [x] `tests/unit/video-providers.test.js` — 11 new cases: registry wiring, OpenRouter create/poll/unsupported-action, Vertex body translation, image mapping, `fetchPredictOperation` polling, completed/failed mapping, and pre-flight rejections (no project id, raw API key, invalid job id) asserting **no upstream call is made**.
- [x] `unit/xai-video-core.test.js` (21) and `unit/xai-video-handler.test.js` (12) still pass — xAI behaviour unchanged.
- [x] Registry baselines pass: `verify-providers.mjs`, `verify-alias.mjs`, `verify-oauth-urls.mjs` all byte-for-byte equal.
- [x] **No regressions:** full `vitest run unit/` gives **103 failures on a clean checkout and 103 on this branch** (the documented pre-existing baseline), with 11 additional passing tests.

Not covered: live calls against either provider — the adapters are unit-tested against mocked upstreams only.

## Notes / scope

- OpenRouter exposes `generations` only; `edits` / `extensions` return 400 (no such endpoint upstream today).
- Vertex likewise routes through `generations` — Veo image-to-video and extend are driven by `image` / `video` in the body, not separate endpoints.
- Both are marked with `ponytail:` comments where the ceiling is deliberate.


decolua:
Thanks @coozgan for the contribution! Reviewed and merged into master. 🙏

Tested locally: the 11 new adapter cases plus the existing xAI video suites all pass (44/44), the three registry baselines are byte-for-byte equal, and a full `unit/` run shows no regressions versus a clean `upstream/master` checkout. The adapter design also keeps the xAI path on the untouched default plan, which is what made this safe to merge.

PR #3921, opened 2026-09-09
feat(xiaomi-desktop): 新增小米 MiMo Desktop provider 支持

cc-claws:
## Summary
- 新增小米 MiMo Desktop provider：支持本地 Desktop 独占模型（mimo-x-pro-preview / mimo-x-flash-preview）与云端稳定模型双通道路由
- 实现自定义 ECDH 加密回调 OAuth 流程（非标准 PKCE），支持本地凭证一键自动导入
- 增加周配额用量查询与 engine token 自动 mint / 进程缓存 / 401 重铸

## Test plan
- [x] `node --check` 语法校验通过（executor / mimoEngine / registry / usage）
- [ ] 本地启动 9Router，连接 xiaomi-desktop 账号并验证云端模型对话
- [ ] 启动 MiMo Desktop 后验证 `mimo-x-pro-preview` / `mimo-x-flash-preview` 本地引擎路由
- [ ] 验证一键导入本地凭证与 OAuth 兜底流程
- [ ] 验证周配额用量展示


decolua:
Thanks @cc-claws for the contribution! Merged into master. 🙏

A few notes on what changed during review, since the merge is not a verbatim copy:

**Merged into the existing `xiaomi-mimo` provider** instead of adding a new
`xiaomi-desktop` one. The two shared the same cloud endpoint
(`api.xiaomimimo.com`) and five identical models, which would have put three
near-identical MiMo entries in the dashboard. `xiaomi-mimo` now does dual auth
(the same pattern as `kimi`) and routes Preview models to the account-service
endpoint per-model (the same pattern as `opencode-go`). `xiaomi-tokenplan`
stays separate — it is a genuinely different tier.

**Fixed four issues found while merging:**

- The OAuth browser flow was unreachable: `poll-status` cleared the session
  before the modal could POST `/exchange`, so every exchange returned 400.
- A Claude-format client was sent to `/v1/chat/completions` rather than the
  declared `/anthropic/v1/messages` transport — `buildUrl` ignored
  `runtimeTransport`, so this affected cloud models too, not just Preview.
- `stopXiaomiMimoProxy()` left every pending session (each holding an X25519
  private key) in the map for the process lifetime.
- The OAuth exchange did not persist the Desktop passToken, so Preview models
  could not work after a browser sign-in.

**Removed unused code** so it does not have to be maintained: the local engine
token minting (`mimoEngine.js`, never reached on the request path), the
`model-catalog` and `usage` routes (no caller), the `engineToken`/`engineUrl`
plumbing, and a top-level `usage` block no consumer reads. Also kept the
`flattenContent` rule but scoped it to Preview models — unscoped it would have
stripped images from `mimo-v2-omni`.

`src/lib/db/driver.js` (the Node >= 24 `better-sqlite3` guard) was kept as-is.

**Added tests** — `tests/unit/xiaomi-mimo-{executor,oauth-session,oauth-proxy}.test.js`.
The provider had none before.

The `X-Mimo-Source: mimocode-cli` header was deliberately NOT added to the
cloud models, so existing `xiaomi-mimo` traffic is unchanged; the baseline
snapshot confirms all 81 other providers are byte-for-byte identical.

If you are using the Desktop Preview models locally, they should now work
end-to-end — please open an issue if not.

cc-claws:
@decolua Just tried it on master — working great! I'm successfully using `mimo-x-pro-preview` through the dual-auth setup now. Fun fact: this very reply is generated by `mimo-x-pro-preview` itself, routed via Claude Code → 9Router → MiMo Desktop account session. Full loop works end-to-end. Thanks again for the clean integration! 🙏

Issue #3644, opened 2026-08-30
z-ai/glm-5.3-flash returns "no completion choices" via Cline provider — confirmed working fine in official Cline VS Code extension


ashen-induwara:
Related: #3573 (closed by original reporter based on outdated info)

Description

Reopening this problem as a new issue. #3573 was closed by the original reporter based on a secondhand report that GLM-5.3-Flash was also failing in the official Cline VS Code extension, which was assumed to mean this was a Cline-side rollout issue rather than a 9router problem.

I'm still hitting this issue on 9router, but GLM-5.3-Flash works fine for me directly in the official Cline VS Code extension — same model, no errors. That contradicts the reasoning used to close #3573 and points this back at 9router's integration rather than an upstream Cline/Z.ai problem.

Steps to reproduce
Go to Providers → Cline
Add Custom Model with ID z-ai/glm-5.3-flash
Click "Test"
Result: Provider returned no completion choices for this model
Expected behavior

Should return a normal completion. the model works correctly for me through the official Cline VS Code extension, so 9router should be able to route it successfully too.

Actual behavior
Model ID resolves correctly through 9router (no 404) — routing/mapping is fine
Request reaches Cline's backend but 9router reports no completion choices
Same model works without issue directly in the official Cline VS Code extension on my end
Why this points back at 9router, not Cline

#3573 was closed on the assumption this was a broad Cline-side rollout stability issue, based on one other user's report of the same failure in the extension itself. I'm not seeing that. the extension works fine for me with this exact model. That leaves 9router's handling of this model's response as the more likely cause:

Possible gap in 9router's response parsing for this model's completion payload shape
Possible difference in how 9router constructs/sends the request compared to the official extension
Environment
9router version: v0.5.55
Provider: Cline (OAuth connection, active)
Model: z-ai/glm-5.3-flash (Z.ai, released ~Aug 26–27, 2026)

Issue #3230, opened 2026-08-11
Cline connection fails with HTTP 401 Unauthorized when using DeepSeek provider


ridhoarmand:
# Bug: Cline connection fails with HTTP 401 Unauthorized when using DeepSeek provider

## Description

When connecting Cline through 9router using the `deepseek-v4-flash` model, the request consistently fails with an HTTP 401 Unauthorized error.

## Error

```text
HTTP 401: [cline/deepseek/deepseek-v4-flash] [401]:
{"error":"Unauthorized: Please make sure you're using the latest version of Cline and re-authenticate your Cline account."}
(reset after 2m)
```

## Steps to Reproduce

1. Configure Cline to use 9router.
2. Select the `cline/deepseek/deepseek-v4-flash` model.
3. Attempt to connect or send a prompt.
4. Observe the HTTP 401 response.

## Expected Behavior

The request should be authenticated successfully and the model should respond normally.

## Actual Behavior

The request fails with HTTP 401 Unauthorized, indicating that Cline should be updated and re-authenticated, even though the request is being routed through 9router.

## Additional Information

* Model: `cline/deepseek/deepseek-v4-flash`
* Error persists across retries.
* The error message includes a reset timer (`reset after 2m`).
* It is unclear whether this originates from Cline authentication, the DeepSeek provider, or how 9router forwards authentication.

Any guidance on whether additional authentication headers or configuration are required when using Cline via 9router would be appreciated.

Issue #2333, opened 2026-07-03
ClinePass OAuth flow in master is broken — uses IDE extension flow, not API consumer auth


adentdk:
## Problem

The ClinePass implementation currently in `master` (from #2261) uses OAuth `client_type=extension` — this is the Cline **IDE extension** auth flow, not the ClinePass API consumer flow. The API rejects tokens from this flow with **401**.

Per [ClinePass docs](https://docs.cline.bot/getting-started/clinepass), ClinePass authenticates with a plain **API key** from `app.cline.bot/settings/api-keys`. No OAuth needed.

## What's wrong in master

| Component | Current master | Should be |
|---|---|---|
| Auth | OAuth `authorization_code` (`client_type=extension`) | API key (`category: "apikey"`) |
| Model alias | `clinepass/cline-pass/<model>` (double prefix) | `clinepass/<model>` |
| Response envelope | Not handled — raw `{success,data}` passed through | Unwrapped to inner `data` |
| Empty response | No retry | Retry once for transient errors |
| `max_tokens` for thinking models | No floor | At least 4096 to avoid empty content |
| `reasoning.effort: max` | Sent as-is → Vercel rejects | Map `max` → `xhigh` |

## Related

- Fix PR: #2332

PR #3923, opened 2026-09-09
fix(cline/clinepass): stop workos:-prefixing ClinePass API keys; add clinepass token refresh

izzzzzi:
## Summary

Fixes Cline/ClinePass HTTP 401 (and resulting HTTP 503 from 9Router) reported in #3230 / #2333 / #3644:

> `[cline/clinepass/cline-pass/<model>] [401]: {"error":"Unauthorized: Please make sure you're using the latest version of Cline and re-authenticate your Cline account."}`

### Root cause

`getClineAccessToken()` in `open-sse/shared/clineAuth.js` unconditionally prefixed **every** token with `workos:`. That is correct for Cline **OAuth** access tokens (WorkOS JWTs, `eyJ…`), but **ClinePass API keys** (category `apikey`, e.g. `clp_…`) are opaque strings and must be sent verbatim. Sending `workos:clp_…` makes `api.cline.bot` reject the request with the 401 above.

I verified against `api.cline.bot` directly:
- `Bearer eyJ…` (raw OAuth JWT) → 200
- `Bearer workos:eyJ…` → 200
- ClinePass API keys must be sent without the `workos:` scheme (the official client does exactly this).

### Changes

1. **`open-sse/shared/clineAuth.js`**: only add the `workos:` prefix to tokens that look like WorkOS JWTs (`eyJ…`). API keys & opaque tokens pass through untouched. Also no double-prefixing when the token already starts with `workos:`.
2. **`open-sse/services/tokenRefresh.js`**: register a `clinepass` entry in `REFRESH_HANDLERS` (reuses the shared Cline WorkOS refresh endpoint) so background token refresh actually runs for ClinePass connections — without it, expired ClinePass OAuth tokens were never rotated and every request 401s.
3. **`open-sse/providers/registry/clinepass.js`**: list `apikey` before `oauth` in `authModes` — ClinePass is meant to be used with an API key from `app.cline.bot/settings/api-keys` (see #2333), not the extension OAuth flow.

### Tests

Added `tests/unit/cline-auth.test.js` (4 cases, pass with `node --test`):
- already-`workos:` tokens are untouched
- bare JWTs get `workos:` prefixed
- ClinePass API keys / opaque tokens are NOT prefixed
- `Authorization` header built without double prefixing

### Manual validation

With a valid Cline OAuth token the direct `api.cline.bot` POST `/api/v1/chat/completions` returns 200 with the 9Router-style headers, so 401 was never about headers — it was about the token scheme (this PR) and expired tokens not being refreshed (this PR).

### DIFF
--- a/open-sse/providers/registry/clinepass.js
+++ b/open-sse/providers/registry/clinepass.js
@@ -14,7 +14,11 @@ export default {
     },
   },
   category: "oauth",
-  authModes: ["oauth", "apikey"],
+  // ClinePass authenticates with a plain API key from app.cline.bot/settings/api-keys
+  // (category "apikey"). The OAuth extension flow used by Cline does not issue
+  // tokens that the ClinePass API consumer endpoint accepts (HTTP 401) — see #2333,
+  // so API key is listed first.
+  authModes: ["apikey", "oauth"],
   hasOAuth: true,
   transport: {
     baseUrl: "https://api.cline.bot/api/v1/chat/completions",
--- a/open-sse/services/clinepassModels.js
+++ b/open-sse/services/clinepassModels.js
@@ -48,7 +48,7 @@ export async function resolveClinepassModels(credentials) {
     if (!Array.isArray(rawList)) return null;
 
     const models = rawList
-      .filter((m) => typeof m?.id === "string" && m.id.startsWith("cline-pass/"))
+      .filter((m) => typeof m?.id === "string" && !m.id.startsWith("cline-pass/"))
       .map((m) => ({
         id: m.id,
         name: m.name || m.id,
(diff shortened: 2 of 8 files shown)


decolua:
Thanks @izzzzzi for the contribution! Reviewed and merged into master. 🙏

The `workos:` prefix fix and the `clinepass` refresh handler are exactly right. I did drop one part of the diff: the inverted filter in `resolveClinepassModels` (which made ClinePass list everything *except* `cline-pass/*`). That overlapped with another PR merged the same day (#3950) which splits the resolver into `resolveClineModels` (full catalog) and `resolveClinepassModels` (`cline-pass/*` only), so keeping #3950's version avoids two resolvers doing the same job. I also trimmed the `authModes` comment — array order is never read (`authModes` is only used via `.includes()`), so the reorder is a no-op; the value was kept. Your auth fix, refresh handler, and the Import button are all in as-is.

PR #3926, opened 2026-09-09
fix(tools): scope Claude tool `type` defaulting to gateways that need it (#3905)

### LINKED ISSUE #3905: Unconditional `type: "custom"` on Claude-format tools breaks the DeepSeek Anthropic endpoint (HTTP 400)
## Summary

Since `e08ac6da` (2026-08-28), `defaultClaudeToolType()` is applied unconditionally to every Claude-format request that carries tools. DeepSeek's Anthropic-compatible endpoint does not implement the `custom` tool variant and rejects the payload with HTTP 400.

The result is that **every Claude Code request routed to a DeepSeek connection fails**, since Claude Code always sends its tool definitions. In a combo the leg silently drops out and traffic falls through to the next provider.

## Environment

- 9router `0.5.69`, installed with `npm i -g`, running under systemd with `--no-browser --log --skip-update`
- Node 22, Debian 13 (LXC container)
- Client: Claude Code `2.1.263` via `ANTHROPIC_BASE_URL` → `/v1/messages`
- Provider: `deepseek` (API key), models `ds/deepseek-v4-pro` and `ds/deepseek-v4-flash`

## Symptom

```
POST /v1/messages  model=ds/deepseek-v4-pro
→ 400 invalid_request_error

Failed to deserialize the JSON body into the target type:
tools[0]: unknown variant `custom`,
expected `web_search_20250305` or `web_search_20260209`
```

`requestDetails` shows the whole cascade collapsing on every turn:

```
provider   model                    status
deepseek   deepseek-v4-pro          error
deepseek   deepseek-v4-flash        error
nvidia     deepseek-v4-pro-0813     ok
```

## Reproduction

Two requests to `https://api.deepseek.com/anthropic/v1/messages`, identical except for the `type` field:

**Without `type` → HTTP 200, tool is called**

```json
{
  "model": "deepseek-v4-flash",
  "max_tokens": 32,
  "tools": [{
    "name": "read_file",
    "description": "reads a file",
    "input_schema": {
      "type": "object",
      "properties": {"path": {"type": "string"}},
      "required": ["path"]
    }
  }],
  "messages": [{"role": "user", "content": "read /tmp/a.txt"}]
}
```

**With `"type": "custom"` → HTTP 400**

```json
{
  "model": "deepseek-v4-flash",
  "max_tokens": 32,
  "tools": [{
    "type": "custom",
    "name": "read_file",
    "description": "reads a file",
    "input_schema": {
      "type": "object",
      "properties": {"path": {"type": "string"}},
      "required": ["path"]
    }
  }],
  "messages": [{"role": "user", "content": "read /tmp/a.txt"}]
}
```

```
{"error":{"message":"Failed to deserialize the JSON body into the target type: tools[0]: unknown variant `custom`, expected `web_search_20250305` or `web_search_20260209`","type":"invalid_request_error","code":"invalid_request_error"}}
```

Claude Code talking **directly** to `api.deepseek.com/anthropic` (no gateway) works fine and executes tools, which confirms the client does not send `type` — the field is added by 9router.

## Root cause

`open-sse/translator/concerns/toolCall.js`:

```js
export function defaultClaudeToolType(tools) {
  if (!Array.isArray(tools)) return tools;
  return tools.map(tool => tool?.type ? tool : { ...tool, type: "custom" });
}
```

Called from `open-sse/handlers/chatCore.js` whenever the final format is Claude and `tools` is an array — with no check on which provider is being dispatched to.

The commit message and the code comment both state the motivation was MiniMax (error 2013). That requirement is real, but it is **provider-specific and is being applied globally**.

This also contradicts `prepareClaudeRequest()` in `open-sse/translator/formats/claude.js`, which strips `type` for non-Anthropic providers. The two paths disagree; `chatCore.js` runs later and wins.

There is no way to opt out: the only related header (`X-9Router-Token-Saver`) is read after the call site, and no setting exposes this behaviour.

## Impact beyond the 400

DeepSeek's Anthropic endpoint does **automatic prefix caching** — [`kv_cache`](https://api-docs.deepseek.com/guides/kv_cache), enabled by default, no client markers needed. Measured here with two identical calls and no `cache_control` at all:

| call | `input_tokens` | `cache_read_input_tokens` |
|---|---|---|
| 1st | 4085 | 0 |
| 2nd | 117 | 3968 |

Because the DeepSeek leg always 400s, Claude Code traffic falls through to OpenAI-compatible providers, which have no prefix caching. Claude Code re-sends ~21,500 tokens of system prompt plus tool definitions every turn, so the prefill is reprocessed from scratch each time.

Measured on this instance, same task, three turns:

| | turn 1 | turn 2 | turn 3 | cache hit |
|---|---|---|---|---|
| DeepSeek leg failing (fallback to NVIDIA) | 33.1 s | 15.4 s | 41.0 s | 0 |
| DeepSeek leg reachable | 15.0 s | 10.9 s | 12.6 s | 22528 / 22764 (98.9%) |

So the regression costs roughly 3x latency and full input billing on every turn, which makes it easy to mistake for a slow provider rather than a rejected request.

## Suggested fix

Gate the normalisation with the quirks mechanism the project already has, rather than applying it to every Claude-format dispatch:

```js
if (finalFormat === FORMATS.CLAUDE
    && Array.isArray(translatedBody.tools)
    && PROVIDERS[provider]?.quirks?.requireClaudeToolType) {
  translatedBody.tools = defaultClaudeToolType(translatedBody.tools);
}
```

with `requireClaudeToolType: true` set on MiniMax only. That keeps `e08ac6da` working for the case it was written for and restores every endpoint that only accepts the legacy tool shape.

`open-sse/utils/claudeToolTypeSelfCheck.mjs` covers the shape of the transform but not provider selection, so a case asserting that a non-MiniMax Claude provider is left untouched would catch a regression like this.

## Workaround in the meantime

Patching the built bundle so the mapper is an identity function restores the DeepSeek leg. It has to be reapplied after every `npm i -g 9router`.

Happy to open a PR for the quirk-gated version if that approach looks right.

galiehneh:
## Summary

`defaultClaudeToolType()` has stamped `tools[].type = "custom"` onto **every** Claude-format request that carries tools since `e08ac6da`. That satisfied MiniMax (error 2013) but broke every other Anthropic-compatible endpoint that only accepts the **legacy typeless tool shape**.

Concrete victim: DeepSeek's Anthropic-compatible endpoint (`api.deepseek.com/anthropic/v1/messages`). Its tool `type` enum only contains the `web_search_*` variants, so it answers:

```json
{"error":{"message":"Failed to deserialize the JSON body into the target type: tools[0]: unknown variant `custom`, expected `web_search_20250305` or `web_search_20260209`","type":"invalid_request_error","code":"invalid_request_error"}}
```

Since Claude Code always sends its tool definitions, **every** Claude Code request routed to a DeepSeek connection fails. Seen in the wild as persistent `503` (gateway relays the upstream 400 after locking the account) — full write-up in #3905.

## Change

- **`open-sse/handlers/chatCore.js`** — only run the defaulting when the target provider declares the new quirk.
- **`open-sse/translator/concerns/toolCall.js`** — new tiny helper `shouldDefaultClaudeToolType(provider, finalFormat, tools, PROVIDERS)` so the gate is unit-testable (no handler harness needed).
- **`open-sse/providers/registry/minimax.js` / `minimax-cn.js`** — declare `quirks.requireClaudeToolType: true`, i.e. the gateways the original change was written for.
- **`tests/translator/bugs-3905-deepseek-tool-type.test.js`** — regression coverage: MiniMax keeps the explicit type; DeepSeek and Anthropic stay on the typeless shape; non-Claude formats and tool-less requests never default.

Effectively a no-op for MiniMax, a bug fix for DeepSeek, and a restore of the pre-`e08ac6da` behaviour for every other Claude-format provider. If another gateway turns out to be strict like MiniMax, it now just needs the same one-line quirk instead of a global behavioural change.

## Verification

- New regression test passes (`vitest run translator/bugs-3905-deepseek-tool-type.test.js`).
- Full translator suite exercised locally (192 passing; remaining failures in this checkout are pre-existing missing root deps such as `undici`/`open`, unrelated to this change).
- Real-world E2E on a production gateway (9router `0.5.69`, Docker):
  - Before: Claude Code → `ds/deepseek-v4-flash` with tools → upstream `400 unknown variant 'custom'` → client `503`, 11 retries, all failed.
  - After (same behaviour as this PR: no `type` stamped for `deepseek`): tool-bearing Claude Code sessions run multi-turn tool calls successfully (`HTTP 200`, tool use + cache hits).
  - Direct upstream probes: tools without `type` → `200`; same tools with `"type":"custom"` → `400`.

Closes #3905


decolua:
Thanks @galiehneh for the contribution! Reviewed and merged into master. 🙏

The provider-scoped quirk is the right fix — it restores the pre-`e08ac6da` shape for every Claude-format endpoint while keeping MiniMax working, and it leaves the door open to add a one-line quirk for any future strict gateway instead of another global behaviour change.

PR #3934, opened 2026-09-10
Hide inactive and hidden providers from the model selector

eachann1024:
## Summary
- Filter `ModelSelectModal` so inactive connections and `AI_PROVIDERS.hidden` entries do not appear in the picker or group list.

## Change graph
PR split overview

```mermaid
flowchart LR
  picker[ModelSelectModal] --> active[keep isActive connections]
  picker --> hidden[drop AI_PROVIDERS.hidden]
```

## Test plan
- [x] `node tests/unit/model-selector-active-providers.check.mjs`
- [ ] Open a combo or adapter model picker and confirm disabled/hidden providers are gone.


eachann1024:
Closing at the author's request.

PR #3936, opened 2026-09-10
fix(claude): repair trailing assistant prefill instead of dropping it (#2302)

anhtran-ai:
## Problem

Vertex-backed Claude routes reject any conversation whose last message is an assistant turn:

```
400 invalid_request_error
"This model does not support assistant message prefill.
 The conversation must end with a user message."
```

Reported in #2302. Agentic clients (Claude Code, Cowork) legitimately produce that shape when resuming a turn, so the request should be repaired rather than passed through to a guaranteed 400.

## Why not just strip the trailing assistant turn

Popping trailing assistant messages does restore the invariant, and #2321 takes that approach as part of a larger multi-fix PR. It is lossy in three cases:

| Trailing content | Unconditional pop | This PR |
|---|---|---|
| Unresolved `tool_use` | pairing lost → next request structurally invalid | closed with matching `is_error` `tool_result` |
| Signed `thinking` | provider-owned history destroyed | kept, user boundary appended |
| `redacted_thinking` | opaque blob unrecoverable | kept, user boundary appended |
| `server_tool_use` | dropped | kept, user boundary appended |
| text-only | dropped | kept, user boundary appended |
| empty / unsigned thinking | dropped | dropped (nothing recoverable) |

The `tool_use` case is the practically important one: dropping an assistant turn that contains an unresolved `tool_use` leaves a dangling call with no result, which is itself rejected on the next request.

Unsigned `thinking` is deliberately **not** preserved — Anthropic rejects it and the existing cleanup passes already discard it, so keeping it would reintroduce a different 400.

## Change

Adds `open-sse/translator/concerns/assistantPrefillPolicy.js`, applied on the Antigravity request path where the rejection was reported. The policy is a pure function over the request body and is a no-op for conversations that already end with a user turn.

## Testing

13 new unit tests in `tests/unit/assistant-prefill-policy.test.js` cover every branch, multi-tool pairing, string-form content, malformed bodies, and the invariant that output always ends with a user turn.

Full suite before and after this change on `master` @ `eb712ca8`: **36 failing tests in both runs** — no regression. Passing count goes 1247 → 1260 (+13, this PR's tests). The pre-existing failures are unrelated (missing optional deps such as `crypto` mocks in `xai-oauth-service`).

## Relationship to #2321

#2321 is an open multi-fix PR that includes a trailing-assistant strip for the same issue alongside Kiro model-id, TTS chunking, Gemini schema, and MITM changes. This PR is scoped to only the prefill contract and preserves the tool/reasoning cases described above. If maintainers prefer #2321's broader batch, the policy here can be applied on top of it — the two touch the same function and would conflict textually, but the intent is complementary.

## Provenance

Extracted from a downstream deployment where this policy has been running in production against Claude and Codex traffic. Verified there with direct probes and end-to-end gateway traffic: `invalid_stream=0`, and the prefill rejection no longer reaches clients.


anhtran-ai:
Pushed a follow-up commit to this branch: `61428678`.

Live verification against a Claude account exposed an edge case the original commit still failed. The policy inspected `messages.at(-1)` exactly once, so dropping a contentless trailing assistant could expose *another* assistant turn underneath it, and the request still reached Anthropic ending in an assistant message:

```
400 invalid_request_error
This model does not support assistant message prefill.
```

Consecutive assistant turns do reach this pass in practice — an interrupted turn followed by an empty one. A second failure mode: a body whose only turn was a contentless assistant became `messages: []`, which the caller then dereferences.

The fix re-checks the invariant after every drop instead of once, and leaves a minimal user turn when the whole conversation was contentless prefill (an empty array is itself a 400, "at least 1 message").

Six regression tests added, all exercising the policy directly:
- 4 parameterized stacked-drop cases (empty content array, empty text block, whitespace-only text, missing `content`)
- an interrupted `tool_use` exposed by dropping an empty assistant above it — still gets its paired error `tool_result`
- the never-leave-`messages`-empty case

Evidence: the 6 new tests fail on `5793cec5` and pass on `61428678` (`tests/unit/assistant-prefill-policy.test.js`, 19/19). Broader run of `translator` + this file, before vs after: same 16 pre-existing failures, passing count 382 → 388. No regression introduced.

PR #3648, opened 2026-08-31
feat(codebuddy-cn): refresh model catalog — add hy4-preview/hy3/glm-5.3/kimi-k3, drop EOL glm-5.0/glm-4.7

zmf963:
## Problem

The codebuddy-cn model catalog is stale against the live gateway (verified 2026-08-30):

**Dead models still advertised** — `glm-5.0` and `glm-4.7` answer `11102 service info not found` on every request. They show up in `/v1/models` and silently consume combo fallback slots.

**Missing models** — the gateway has shipped a new generation that 9router can't route:

| Model | Status | Notes |
|---|---|---|
| `hy4-preview` | live | Tencent's new flagship, 770B MoE, **1M context**, released 2026-08-28 |
| `hy4-preview-x` | live | paid twin of hy4-preview (see below) |
| `hy3` | live | Hy3 GA (preview graduated 2026-07-06) |
| `hy3-x` | live | paid twin of hy3 |
| `glm-5.3` | live | GLM-5.3, 1M context |
| `glm-5.3-flash` | live | GLM-5.3 flash tier |
| `kimi-k3-1` | live | canonical id per the client's server-delivered table (bare `kimi-k3` currently resolves but the client sends `-1`) |

**The `-x` suffix convention:** the official client's server-delivered model table (extracted from CodeBuddy IDE v4.11.2 logs) lists pairs like `hy4-preview` + `hy4-preview-x` with identical display names — the `-x` id bills credits directly while the plain id rides the limited-time free quota (hy3 free until 2026-08-31, hy4-preview until 2026-09-10). Routing both lets users pick free-vs-paid explicitly instead of silently burning promo quota.

## Change

**`open-sse/providers/registry/codebuddy-cn.js`**
- Add `hy3`, `hy3-x`, `hy4-preview`, `hy4-preview-x`, `glm-5.3`, `glm-5.3-flash`, `kimi-k3-1`
- Remove `glm-5.0`, `glm-4.7` (EOL, 11102)

**`open-sse/providers/capabilities.js`**
- Matching `PROVIDER_CAPABILITIES["codebuddy-cn"]` entries: `hy4-preview*` 1M context (official release notes), `glm-5.3` 1M, `glm-5.3-flash` conservative 200K (window unverified), `hy3` family 192K (matches existing `hy3-preview`), `kimi-k3-1` 256K

**`open-sse/providers/thinkingLevels.js`**
[...]

PR #3940, opened 2026-09-10
feat(codebuddy-cn): replace deepseek-v4-flash with deepseek-v4.1-flash

zmf963:
## What

Replace `deepseek-v4-flash` with `deepseek-v4.1-flash` in the codebuddy-cn provider:

- `open-sse/providers/registry/codebuddy-cn.js` — models[] entry swapped; drop log extended
- `open-sse/providers/capabilities.js` — per-model capability entry swapped

`deepseek-v4.1-flash` shows up in the gateway's product-config payload (the config the
IDE plugin fetches from copilot.tencent.com) while `deepseek-v4-flash` is gone from it —
same pattern as the previous catalog refreshes (#3648, #3802): the published list is the
contract, so the old id is removed even though its endpoint still answers.

## Server table values (product-config snapshot)

| field | value |
|---|---|
| maxInputTokens (contextWindow) | 1000000 |
| maxOutputTokens (maxOutput) | **128000** (up from 50000 on v4-flash) |
| supportsImages | true |
| reasoning.canDisableThinking | true |
| reasoning.supportedEfforts | low, high, xhigh |

## No changes needed elsewhere

- `thinkingLevels.js` — the existing `deepseek-v4*` pattern already matches the new id and
  publishes the same `low/high/xhigh` set the server declares for v4.1-flash.
- `pricing.js` — no per-model entry for the flash tier; the `deepseek-v*` glob fallback
  yields the same rates the old id resolved to.
- `alias-baseline.json` — modelKeys record provider-level keys only, unaffected.

## Verification

- Live endpoint probe (`stream:true`): `deepseek-v4.1-flash` → HTTP 200, model echoed back
  in the SSE deltas; `deepseek-v4-pro` (control) → 200. The retired `deepseek-v4-flash`
  also still returns 200 — removed per the published list, as with previous EOL drops.
- Runtime resolution on this branch:
  - `getCapabilitiesForModel('codebuddy-cn','deepseek-v4.1-flash')` → contextWindow 1000000,
    maxOutput 128000, thinkingCanDisable true
  - `getThinkingLevels(...)` → `["low","high","xhigh"]` (matches server supportedEfforts)
  - `getPricingForModel(...)` → flash-tier rates via the `deepseek-v*` fallback
- Cross-check scripts against the server snapshot: 0 field mismatches for all models
  published by the server; `deepseek-v4.1-flash` OK on every field.
- `npx vitest run tests/unit/zed-usage.test.js` fails identically with and without this
  change (`Cannot find package 'open-sse/providers/registry/index.js'` — pre-existing
  vitest alias issue in this checkout, unrelated to the diff).


decolua:

PR #3947, opened 2026-09-10
fix(deepseek): strip unsupported tool types when forwarding to /anthropic/v1/messages

kimono381:
# fix(deepseek): strip unsupported tool types when forwarding to /anthropic/v1/messages

## Symptom

When Claude Code is routed through a DeepSeek account via 9router, every request
that carries MCP / function tools fails at the upstream with HTTP 400:

```
[AUTH] Key N locked modelLock_deepseek-v4-pro-max for 30s [400]
[FALLBACK] ⇄ ACC:Key N UNAVAILABLE (400) → NEXT ACCOUNT
[FMT: claude→claude · STREAM · 2 MSG · 38 TOOL · THINK:max · ACC:Key …]
ERROR 400 · deepseek/deepseek-v4-pro-max · 774 ms
URL: https://api.deepseek.com/anthropic/v1/messages
{"error":{"message":"Failed to deserialize the JSON body into the target type:
  tools[0]: unknown variant `custom`, expected `web_search_20250305` or `web_search_20260209`",
  "type":"invalid_request_error","code":"invalid_request_error"}}
```

The 400 cascades through `auto-fallback`, every DeepSeek account gets locked,
and Claude Code sessions effectively lose DeepSeek as a usable backend.

## Root cause

9router's `prepareClaudeRequest()` (in `open-sse/translator/formats/claude.js`)
runs a single generic tool filter for all non-`claude` providers:

```js
.filter(tool => !tool.type || tool.type === "function")
```

That filter drops **both** `type: "custom"` (correct for OpenAI-compatible
upstreams) **and** `type: "web_search_*"` (wrong — DeepSeek's Anthropic-
compatible endpoint **accepts** web_search_* but only those). For DeepSeek,
this is exactly the wrong direction:

| Tool type from Claude Code | Old behaviour | What DeepSeek expects |
| --- | --- | --- |
| `type: "custom"` (MCP, Read, Bash, …) | dropped (but the `.map` ran on no-type tools and the body still leaked — see below) | dropped |
| `type: "web_search_20250305` / `…60209` | dropped | **kept** |
| no type (legacy) | kept, normalised to `{name, description, input_schema}` | depends on upstream |

Worse, the subsequent `.map` unconditionally strips the `type` field from any
tool that survives the filter:

```js
const { type, ...rest } = tool;
return rest;
```

So even after the filter is fixed to keep `web_search_*`, the `.map` would
remove the discriminator and DeepSeek would reject the request with the same
"unknown variant" error against an empty discriminator.

The filter also did not know that `prepareClaudeRequest()` is being run for
the Claude-format DeepSeek transport (`https://api.deepseek.com/anthropic/v1/messages`)
in the first place — the existing `targetFormat === "claude"` branch is taken
for any provider whose transport matches the Claude format (DeepSeek, GLM,
Kimi, etc.).

## Fix

Introduce a per-provider `quirks.claudeSupportedToolTypes` whitelist on the
provider transport. When declared, `prepareClaudeRequest()` uses it as a
strict allow-list for Anthropic-format tool `type` values, and stops
stripping the `type` field from surviving tools so the upstream gets the
discriminator it needs.

```js
// open-sse/translator/formats/claude.js  (extract)
const supportedTypes = PROVIDERS[provider]?.quirks?.claudeSupportedToolTypes;
const hasWhitelist = Array.isArray(supportedTypes);
body.tools = body.tools
  .filter(tool => {
    const t = tool?.type;
    if (!t || t === "function") return true;
    if (hasWhitelist) return supportedTypes.includes(t);
    return false;                       // prior behaviour
  })
  .map(tool => {
    if (tool.function) { /* normalize OpenAI-style */ }
    if (hasWhitelist) return tool;      // keep `type` so upstream can route
    const { type, ...rest } = tool;
    return rest;
  });
```

```js
// open-sse/providers/registry/deepseek.js
transport: {
  …
  quirks: {
    claudeSupportedToolTypes: ["web_search_20250305", "web_search_20260209"],
  },
},
```

## Backwards compatibility

- Providers without `quirks.claudeSupportedToolTypes` (GLM, Kimi, every
  current and future Anthropic-compatible transport that has not opted in)
  keep the existing filter and normalisation semantics byte-for-byte.
- The OpenAI-compatible DeepSeek transport (`format: "openai"`) is
  unaffected — `prepareClaudeRequest()` is only called when `targetFormat
  === "claude"`, which is not the case for OpenAI-format targets.
- No public API change. The quirk is added under an existing optional
  `transport.quirks` field already used by `cerebras`, `mistral`,
  `claude`, etc.

## Tests

`tests/unit/deepseek-claude-tools.test.js` covers:

1. The quirk is present on `PROVIDERS.deepseek`.
2. `type: "custom"` tools (the regression) are stripped.
3. `type: "web_search_20250305"` and `type: "web_search_20260209"` survive.
4. The `type` field is preserved on surviving `web_search_*` tools (the
   second regression caught by the e2e harness during local development).
5. Mixed `custom` + `web_search_*` payload — only `web_search_*` survives.
6. Body without `tools` does not crash.
7. Future / unknown tool types are rejected (defence in depth).
8. Backwards-compat: providers without the quirk (any current GLM / Kimi /
   future Anthropic-compatible transport) keep prior behaviour.

## Files

- `open-sse/providers/registry/deepseek.js` (+15 lines)
- `open-sse/translator/formats/claude.js` (+18 / -1 lines)
- `tests/unit/deepseek-claude-tools.test.js` (new, 144 lines)

## Manual verification (out-of-tree)

Standalone e2e harness loads the patched `prepareClaudeRequest` with mocked
siblings (`PROVIDERS`, `defaultThinkingSignature`, schema, capabilities,
`sessionManager`, `claudeCloaking`, …) and runs representative payloads
through it:

```
ok  1 - DeepSeek: original body contains type:'custom' tools
ok  2 - DeepSeek: prepareClaudeRequest removes all custom tools
ok  3 - DeepSeek: no tool with unknown variant survives
ok  4 - DeepSeek: web_search_20250305 survives
ok  5 - DeepSeek: web_search_20260209 survives
ok  6 - provider=claude: filter is NOT applied (Claude native keeps everything)
ok  7 - provider=glm-cn (no quirk): backwards-compat drops custom

7/7 end-to-end tests passed
```

Plus an isolated smoke harness over the filter logic alone: **12/12** pass.


decolua:
Thanks @kimono381 for the contribution! Reviewed and merged into master. 🙏

Verified: the new test passes, the full `unit/` suite shows no regressions vs upstream/master (102 known fails unchanged), and providers/alias/OAuth baselines stay byte-for-byte equal. Also confirmed empirically that the behaviour change is limited to the `deepseek` provider — the other nine Claude-format providers produce identical output.

PR #3948, opened 2026-09-10
fix(claude): cap re-anchored cache_control at the 4-marker budget and keep single-object content turns

### LINKED ISSUE #3795: 400 "A maximum of 4 blocks with cache_control may be provided. Found 5." — anchorClaudeCache adds a 5th marker when the client already sent 4

n1klaus:
Fixes two request-translation defects on the Claude legs. Fixes #3795 (cache budget leg).

## Problem 1: cache_control budget overflow (#3795)

Anthropic's Messages API allows **at most 4 blocks carrying `cache_control` per request**. When the client has already spent its marker budget, 9Router's re-anchoring added one more breakpoint — the last system block, the last cacheable tool, the last assistant turn — and the request was rejected:

```json
 "message":"A maximum of 4 blocks with cache_control may be provided. Found 5."}}
```

As #3795 documents, the 400 is non-retryable, but the failure path treats it as an account problem, so the same malformed body is retried across the whole pool (`ACC:… UNAVAILABLE → NEXT ACCOUNT`) until every account locks. In one 3,000-line window the reporter counted 52 rejections and 52 account-unavailable lines.

### Fix

`anchorClaudeCache` (passthrough, runs last after every reshaping step) now, in order:

1. **Strips invalid markers** — a `defer_loading: true` tool carrying `cache_control` is rejected by Anthropic outright (#3567); it is cleaned regardless of budget.
2. **Prunes over-budget bodies** — a client body already carrying more than 4 markers is a guaranteed 400; the router now keeps the last 4 breakpoints (tail-most dominate for cache hits once the conversation moves past the earlier ones) and strips the rest, instead of forwarding it untouched.
3. **Guards re-anchoring** — at exactly 4 the client has spent the whole budget, so a 5th marker guarantees a 400 while skipping re-anchoring costs nothing: every marker the client placed is itself a valid cache breakpoint. The count runs after single-object normalization so a bare-object marker can never be missed or survive into it.

## Problem 2: single-object content turns are dropped

The Messages API content field accepts `string | array of blocks`; some clients send a **single bare block object** (`content: {type: "text", …}`) instead of a one-element array. On every leg that reads or reshapes messages, such a turn was dropped or corrupted:

- `hasValidContent()` returned `false` for bare objects → `prepareClaudeRequest` filtered the turn out of the conversation entirely
- the mid-conversation-system fold in `normalizeClaudePassthrough` assumes array-shaped neighbors → a bare-object message folded with a system reminder was zeroed to `content: []`
- `convertClaudeMessage` (claude→openai) skipped bare-object content — including **bare-object system reminders**: the system-role branch ran before the normalization, so `systemReminderText` read the bare object as empty and the turn vanished

Result: silent conversation-turn loss — history holes the client never sees.

### Fix

New `normalizeMessageContent(msg)` wraps bare-object content as a one-element array and strips any client-placed `cache_control` from it (a bare-object marker must never survive normalization). Applied in order:

1. in `normalizeClaudePassthrough`, **before** the mid-conversation-system fold (the fold assumes arrays)
2. in `anchorClaudeCache`, **before** the budget guard (so the count sees real array-block markers only)
3. in `prepareClaudeRequest`, before the cache_control-removal pass
4. in `convertClaudeMessage` (claude→openai leg), **before the role branch** so system reminders fold correctly

Messages already using array content keep their structure and ordering. `hasValidContent` also gained a bare-object branch so a single-object turn is kept wherever messages are filtered.

## Combined behavior

A request combining both defects — the marker budget spent AND a single-object turn carrying a client marker — keeps the turn, normalizes it to a one-element array, removes its marker, and adds no marker past the four-marker limit. An over-budget body (5 markers) is pruned to the last four; a deferred tool never keeps a marker under any budget.

## Tests

`tests/unit/claude-cache-budget-single-object.test.js` — 11 unit tests, 27 assertions, covering: budget-exhausted body never emits >4 markers; single-object turn normalized + anchored in passthrough; single-object turn kept with its marker stripped in `prepareClaudeRequest`; no turn dropped for single text objects; re-anchoring the last assistant turn when it uses single-object content; marked single-object turn kept while budget is spent; single-object turns kept on the claude→openai leg (incl. image blocks); bare-object user turn folded correctly with a mid-conversation system message; **5-marker body pruned to 4**; **deferred tool marker stripped at full budget**; **bare-object system reminder kept on the claude→openai leg**.

Run: `cd tests && npx vitest run unit/claude-cache-budget-single-object.test.js` → 11/11 pass. The wider suite's failing tests were verified identical at clean master (pre-existing rot, e.g. 4 in `tests/__baseline__/known-fails.txt`) — zero regressions from this change.

## Relationship to open PR #3861

#3861 also guards against the 4-marker overflow (#3795, cache-cap section) with per-site `countCacheControlBlocks(body) < 4` checks. This PR additionally: normalizes single-object content (not covered by #3861 or #3795), cleans invalid defer_loading markers, prunes over-budget bodies instead of passing them through, and keeps the budget guard as a single early return after normalization. Happy to rebase whichever lands second.


decolua:
Thanks @n1klaus for the contribution! Reviewed and merged into master. 🙏

Reviewed the two fixes and verified both against the current tree. The cache-budget guard reproduces cleanly on the passthrough leg (client sends 4 markers, the re-anchor emitted a 5th), and the single-object turn loss is confirmed on the claude→openai leg — a bare-object assistant turn and a bare-object system reminder were both dropped.

Two adjustments went in with the merge:

- The budget guard ran before the head anchors, so a spent budget left the last system block and last cacheable tool on the 5m default rather than the 1h TTL the rest of the file uses. The anchors now run first, preserving the existing cache policy on exactly the requests that already cache hardest.
- The over-budget trim kept the last four markers in document order, which drops the head anchors first — they lead document order. The trim now holds the anchors and fills the remaining slots from the tail of the messages.

Two tests were added covering the 1h anchors at a spent and an over-budget state; both fail without the change. Full suite shows no regressions.

Issue #3845, opened 2026-09-07
❌ kiro [400]: [400]: {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}


congncif:
🔴 ✗ ERROR 400 · kiro/claude-opus-5 · 24756ms URL: https://runtime.us-east-1.kiro.dev/generateAssistantResponse [400]: {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}

NoReplyUI5:
> 🔴 ✗ ERROR 400 · kiro/claude-opus-5 · 24756ms URL: https://runtime.us-east-1.kiro.dev/generateAssistantResponse [400]: {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}

use 0.5.55 and enjoy.. last 2 updates are useless

```
npm i -g 9router@0.5.55
```

agusm:
yes, in v0.5.69 for kiro and github copilot useless.

congncif:
> > 🔴 ✗ ERROR 400 · kiro/claude-opus-5 · 24756ms URL: https://runtime.us-east-1.kiro.dev/generateAssistantResponse [400]: {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}
> 
> use 0.5.55 and enjoy.. last 2 updates are useless
> 
> ```
> npm i -g 9router@0.5.55
> ```

0.5.55 doesn't work for me. I tested OK with 0.5.30. The updates following Kiro are practically useless; I don't understand why they keep releasing them.

NoReplyUI5:
> > > 🔴 ✗ ERROR 400 · kiro/claude-opus-5 · 24756ms URL: https://runtime.us-east-1.kiro.dev/generateAssistantResponse [400]: {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}
> > 
> > use 0.5.55 and enjoy.. last 2 updates are useless
> > 
> > ```
> > npm i -g 9router@0.5.55
> > ```
> 
> 0.5.55 doesn't work for me. I tested OK with 0.5.30. The updates following Kiro are practically useless; I don't understand why they keep releasing them.


I'm not sure, but 55 works for me. Anyway, I don't think the repository's main author tests anything before approving those shitty pull requests.

haiueom:
v0.5.55 still good

[...]

Issue #3749, opened 2026-09-03
can't request to model with provider kiro


DucAnh28:
HTTP 503: [kiro/auto] [400]: {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"} (reset after 1m 57s)

<img width="947" height="167" alt="Image" src="https://github.com/user-attachments/assets/5318ad2b-cf8b-4ab2-be84-5bfdbced5b84" />

haiueom:
Same problem here

<img width="772" height="150" alt="Image" src="https://github.com/user-attachments/assets/85491b35-3665-4d62-ada2-86679399b40e" />

renefoster:
<img width="1328" height="280" alt="Image" src="https://github.com/user-attachments/assets/0c9c260f-3a97-42cf-a5eb-df258f8dc49d" />

v0.5.30 works well.

haiueom:
> v0.5.30 works well.

yeah, work well after downgrade version to v0.5.30

<img width="1241" height="255" alt="Image" src="https://github.com/user-attachments/assets/0ec5b34f-7fd7-481a-8262-d252e98e0bd4" />

NoReplyUI5:
> > v0.5.30 works well.
> 
> yeah, work well after downgrade version to v0.5.30
> 
> <img width="1241" height="255" alt="Image" src="https://github.com/user-attachments/assets/0ec5b34f-7fd7-481a-8262-d252e98e0bd4" />

you must try 0.5.55 that's also working as well. but latest.. nahh that's still broken

Issue #3459, opened 2026-08-21
Kiro: Codex CLI returns REQUEST_BODY_INVALID due to top-level systemPrompt (0.5.55) — verified workaround


EVControl:
Summary

I reproduced a Kiro compatibility issue with 9router 0.5.55 and OpenAI
Codex CLI 0.149.0.

A simple direct request through 9Router succeeds, but the actual Codex
CLI payload fails against Kiro with:

    400 REQUEST_BODY_INVALID
    Improperly formed request.

I then applied a minimal local workaround to the installed 9Router
0.5.55 bundle.

After removing only the two assignments that serialize systemPrompt as a
top-level property of the Kiro request, the exact same Codex CLI request
succeeds.

Importantly, I did not remove the internal systemPrompt handling used by
the session/replay/cache logic.

This appears consistent with the Kiro compatibility work previously
discussed in #2989, #3091 and closed PR #3166.

------------------------------------------------------------------------

Environment

-   OS: Windows 10
-   PowerShell: 7.6.5
-   9router: 0.5.55
-   OpenAI Codex CLI: 0.149.0
-   API endpoint: /v1/responses
-   Provider/model: kr/claude-sonnet-4.5
-   Codex wire_api: responses

9Router provider configuration:

    [model_providers.9router]
    name = "9Router"
    base_url = "http://127.0.0.1:20128/v1"
    wire_api = "responses"
    env_key = "OPENAI_API_KEY"
    requires_openai_auth = false

------------------------------------------------------------------------

Reproduction

1. Direct /v1/responses request works

A simple request directly to 9Router:

    POST /v1/responses

    model: kr/claude-sonnet-4.5
    input: Responde solamente OK
    stream: false

returns successfully:

    model: claude-sonnet-4.5
    content: OK

So:

    Client
      ↓
    9Router /v1/responses
      ↓
    Kiro / Claude Sonnet 4.5
      ↓
    OK

Authentication, provider connectivity and the model itself are therefore
working.

------------------------------------------------------------------------

2. Codex CLI fails with the same model

Starting Codex directly with:

    codex --model "kr/claude-sonnet-4.5"

and sending:

    responde solamente OK

returned:

    [kiro/claude-sonnet-4.5] [400]:
    {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}

Codex also displayed:

    Model metadata for `kr/claude-sonnet-4.5` not found.
    Defaulting to fallback metadata; this can degrade performance and cause issues.

[...]

PR #3238, opened 2026-08-11
fix(kiro): comment out systemPrompt field causing REQUEST_BODY_INVALID

chmichae1:
# Fix: Kiro provider HTTP 400 REQUEST_BODY_INVALID errors

## 🐛 Problem

Multiple users are experiencing HTTP 400 errors when using Kiro models (particularly `kr/claude-sonnet-4.5`) through 9router:

```
HTTP 400: [kiro/claude-sonnet-4.5] [400]: 
{"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}
```

**Impact:**
- Affects all users trying to use `kr/claude-sonnet-4.5` and potentially other Kiro models
- Models work when accessed directly via Kiro API but fail through 9router
- Related to multiple open issues (see below)

### Step 1: Initial Diagnosis
Captured failing request and compared with working direct Kiro API calls:

**Failing 9router request:**
```json
{
  "conversationState": {...},
  "systemPrompt": "You are a helpful assistant...",  // ← PROBLEM
  "profileArn": "...",
  "additionalModelRequestFields": {...}
}
```

**Working direct API request:**
```json
{
  "conversationState": {...},
  "profileArn": "...",
  "additionalModelRequestFields": {...}
  // No systemPrompt field
}
```

### Step 2: Isolated the Problematic Field
Systematically removed fields from the request payload:
1. ❌ Removed `reasoning_effort` → Still failed
2. ❌ Removed `extra_body.think` → Still failed  
3. ❌ Removed `tools` → Still failed
4. ✅ Removed `systemPrompt` → **SUCCESS**

### Step 3: Confirmed Root Cause
The `systemPrompt` top-level field is being rejected by Kiro's backend for certain models (confirmed with `claude-sonnet-4.5`).

**Why this happens:**
- 9router translators send `payload.systemPrompt` based on AWS CodeWhisperer API spec
- Kiro's current implementation doesn't accept this field for some/all models
- The field was likely deprecated or never fully supported by Kiro

### Step 4: Located Source Code
Found the problematic lines in both translators:
- `open-sse/translator/request/openai-to-kiro.js` line 411
- `open-sse/translator/request/claude-to-kiro.js` line 318

```javascript
if (systemPrompt) payload.systemPrompt = systemPrompt;
```

## ✅ Solution

**Minimal fix:** Comment out the `payload.systemPrompt` assignment in both translator files.

#### File 1: `open-sse/translator/request/openai-to-kiro.js` (line 411)
```diff
[...]

PR #3949, opened 2026-09-10
fix(kiro): never send a top-level systemPrompt (400 REQUEST_BODY_INVALID)

### LINKED ISSUE #3641: Kiro free-tier models broke between v0.5.55 and v0.5.59 with REQUEST_BODY_INVALID
## Description
Free-tier Kiro models that were working fine through v0.5.55 started
failing with REQUEST_BODY_INVALID starting in v0.5.59. Thinking intensity
was never manually configured — left at default the entire time.

(Loosely related to #2716, but that issue covers a different set of
suffixed models with a longer, still-unresolved history — filing this
separately since the affected models and version range here are distinct.)

## Affected models
- kr/claude-sonnet-4.5
- kr/claude-haiku-4.5
- kr/deepseek-3.2
- kr/qwen3-coder-next

## Error message
HTTP 400:
[kiro/qwen3-coder-next]
[400]: {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}

Same error across all four models listed above.

## Version bisection
- v0.5.55 → all four models work fine
- v0.5.59 → same account, same models, all fail with the error above
- Downgrading back to v0.5.55 resolves it immediately, zero config changes

## Steps to reproduce
1. Run 9router v0.5.59 (or later, untested)
2. Connect a Kiro account
3. Send a request to any plain model, e.g. kr/claude-sonnet-4.5
4. Request fails with REQUEST_BODY_INVALID
5. Downgrade to v0.5.55, same request succeeds

f-liva:
## Problem

Every `kr/` model fails with

```
400 {"message":"Improperly formed request.","reason":"REQUEST_BODY_INVALID"}
```

as soon as an RTK system-injection plugin (`caveman`, `ponytail`) is enabled — regardless of model, payload size or client. Token refresh succeeds right before the 400, so it is not a credentials problem.

## Root cause

`kiro.dev` rejects any body carrying a **top-level `systemPrompt`**. `claude-to-kiro.js` / `openai-to-kiro.js` stopped emitting that field in v0.5.59 for exactly this reason (the prompt travels in the first user turn via `contentPrefix`), but two paths kept writing it back **after** the translator ran:

- `open-sse/rtk/systemInject.js::injectKiroSystem()` → `body.systemPrompt = next`
- `open-sse/executors/kiro.js::appendRepairInstruction()` → `repaired.systemPrompt = ...` on the tool-call repair retry

Because the mutation happens downstream of the translator, instrumenting the translator shows a clean payload — only a wire-level capture shows the offending field.

## Evidence

Captured the real outgoing body from a running 0.5.65 instance and replayed 7 variants against `https://runtime.us-east-1.kiro.dev/generateAssistantResponse` with the same credentials:

| variant | result |
|---|---|
| captured body, unmodified | **400** REQUEST_BODY_INVALID |
| minus top-level `model` | **400** |
| `conversationId` normalized to 36 chars | **400** |
| minus `agentMode` / `agentTaskType` / `agentContinuationId` | **400** |
| **minus `systemPrompt`** | **200**, streaming |
| minimal hand-built body | **200** |

End-to-end: with a temporary `fetch` hook stripping only `systemPrompt`, `kr/claude-haiku-4.5` and `kr/glm-5` both returned 200 streaming through the normal 9router pipeline.

## Fix

- `injectKiroSystem()` appends the RTK prompt to the first history user turn's content (else `currentMessage`), never to `systemPrompt`. Idempotency reuses the existing `dedupStringAppend`/`hasPrompt` helpers, so retries do not duplicate the prompt.
- `isKiroBody()` no longer requires `typeof body.systemPrompt === "string"` — that marker is gone from the wire shape, so the gate would stop dispatching. It now sniffs the conversation turn shape (`history[].userInputMessage` / `currentMessage.userInputMessage`), which also keeps the existing "stray `conversationState` must not hijack a chat body" guard passing.
- `appendRepairInstruction()` appends its repair instruction to `currentMessage.userInputMessage.content`.
- Stale comments in both kiro translators corrected: the `systemPrompt` local is only a session-replay cache key, not a wire field.

Behaviour note: with two injectors active the prompts now land at the tail of the user turn in injection order (caveman, then ponytail) instead of being mirrored into the leading system prefix. Same content, different position — worth a look if anyone depends on the old ordering.

## Tests

Two test files asserted the old contract and were updated:

- `tests/unit/system-inject.test.js` — kiro cases rewritten: no top-level write, content append, idempotency, frozen-message fail-open, non-enumerable `_kiroUpstreamModel` preserved.
- `tests/unit/kiro-terminal-integrity.test.js` — "keeps model-controlled parser detail out of the retry system prompt" now reads the retry's user turn instead of `systemPrompt`; the actual guarantee (no model-controlled text echoed back) is asserted unchanged, and the fixture body got a real `currentMessage`.

`tests/unit/rtkKiro.test.js` passes unchanged.

Full suite (`cd tests && npx vitest run`): the set of failing **test names** is identical before and after this branch — 118 pre-existing failures, live/network cases.

## Closes / relates

Fixes #3641, #3845, #2890, #2901, #2939, #3109, #3459, #3749.

Related PRs #2911 and #3238 both remove the field at the translator level only — that part already landed in v0.5.59; this PR closes the two remaining paths (RTK injector, repair retry) that write it back downstream of the translator.


decolua:

The root-cause writeup (translator clean, injection downstream of it) and the replayed wire variants made this one easy to verify. Confirmed the patch applies cleanly on top of #3776 — that one only touched executor routing, so both fixes compose. Net -52 lines in the injector is a nice bonus.

PR #3950, opened 2026-09-10
fix(cline,airforce): unwrap {success,data} envelope, add live catalog, and refresh airforce free models

### LINKED ISSUE #3644: z-ai/glm-5.3-flash returns "no completion choices" via Cline provider — confirmed working fine in official Cline VS Code extension

n1klaus:
Closes #3644
Related: #3387 (narrower envelope unwrap, single file), #3626 (live catalog, different implementation)

## Problem

1. **Cline free-model envelope** — `api.cline.bot` wraps non-stream chat completions in `{"success":true,"data":{...choices...}}`. Both the dashboard model-test ping (`src/app/api/models/test/ping.js`) and the proxy non-stream path (`open-sse/handlers/chatCore/nonStreamingHandler.js`) read `choices` at top level, so enveloped responses produce "Provider returned no completion choices for this model" (#3644).

2. **No live Cline catalog** — Free Cline models (e.g. `z-ai/glm-5.3-flash`) are invisible in the model picker and `/v1/models` because only `cline-pass/` prefixed models were fetched. New free models require manual registry updates.

3. **Stale airforce free models** — The three hardcoded api-airforce model IDs (`anthropic/claude-3.7-sonnet`, `moonshot/kimi-k2.6`, `google/gemini-2.5-flash`) are dead. Live free models are `gpt-oss-120b`, `gpt-oss-20b`, `kimi-k2.7-code`.

### Envelope unwrap (commits 1 & 2)
- **`nonStreamingHandler.js`**: Unwrap `responseBody.data` when `responseBody.success === true && responseBody.data` is a non-array object. Runs before usage extraction and response translation, so usage tracking and client response both see bare OpenAI-shaped body. Error envelope (`success: false`) passes through untouched.
- **`ping.js`**: Same guard in `pingModelByKind` so dashboard "Test" button works with enveloped responses.

### Live Cline catalog (commit 2)
- **`clinepassModels.js`**: Extract shared `fetchClineRawModels()` from `resolveClinepassModels`. Add `resolveClineModels()` returning ALL models (no `cline-pass/` prefix filter). Both handle `{success,data}` envelope on the `/models` endpoint itself.
- **`v1/models/route.js`**: Register `cline` in `LIVE_MODEL_RESOLVERS`.
- **`providers/[id]/models/route.js`**: Add `cline` and `clinepass` entries in `PROVIDER_MODELS_CONFIG` with `customResolver` and static fallback.
- **`ModelSelectModal.js`**: Generalize the Cursor-only `useLiveProviderModels` hook to also serve `cline` and `clinepass`. Static catalog stays as fallback when live fetch fails.
- **`cline.js` registry**: Add `authModes: ["oauth"], hasOAuth: true`.
- **`cliTools.js`**: Update Cline brand color to `#5B9BD5`.

### Airforce catalog refresh (commit 3)
- **`api-airforce.js`**: Replace three dead model IDs with `gpt-oss-120b`, `gpt-oss-20b`, `kimi-k2.7-code`. Add `passthroughModels: true` and `forceStream: true`. Add `modelsFetcher` for suggested-models discovery.
- **`suggested-models/filters.js`**: Add `airforce-free` filter selecting `tier === "free"` chat models.

## Tests

- `tests/unit/cline-free-models-envelope.test.js` (270 lines): Covers ping envelope unwrap (success, reasoning-only, error passthrough, bare body), nonStreamingHandler envelope unwrap (usage extraction verification, bare body passthrough), and `resolveClineModels` vs `resolveClinepassModels` filtering/shape/error/null-token behavior.
- `tests/unit/api-airforce-free-models.test.js` (36 lines): Verifies new model IDs registered, dead IDs dropped, passthrough flag set, PROVIDER_MODELS exposure, and capabilities resolution.

## Note on related PRs

- **#3387** covers only `nonStreamingHandler.js` (+6 lines) with a narrower guard (`!choices && data.choices`). This PR uses a broader guard (`success === true && data is object`) that also covers edge cases where `data` contains fields other than `choices` (e.g. `usage`), and additionally fixes `ping.js`. The approaches are compatible but this PR supersedes #3387's scope.
- **#3626** adds a separate `clineModels.js` service file. This PR instead extends the existing `clinepassModels.js` with a shared `fetchClineRawModels()` to avoid duplicating the fetch/timeout/auth logic, and additionally wires per-connection resolvers and generalizes ModelSelectModal.


decolua:
Thanks @n1klaus for the contribution! Reviewed and merged into master. 🙏

Envelope unwrap, the shared `fetchClineRawModels` split, and the airforce catalog refresh are all in. I verified the airforce claim against the live endpoint — the three new ids exist and the three removed ones are gone, and the `airforce-free` filter matches the real field names.

One change on merge: the envelope unwrap was applied to every provider in `nonStreamingHandler`/`ping`. I scoped it behind `transport.quirks.clineEnvelope` (a shared `unwrapClineEnvelope()` helper) with `cline` and `clinepass` opting in, so no other provider's response body can be rewritten — including one that legitimately returns `{success:true,data:...}`. `ping.js` resolves the provider via `resolveProviderId()` first, since the model string there is an alias (`cl/...`). Tests were added pinning both the opt-in and the non-opt-in path.

PR #3982, opened 2026-09-11
fix(updater): avoid killing unrelated next-server/cloudflared processes during shutdown

Ali-Shaikh:
The `collectAppPids()` function in `src/lib/appUpdater.js` used overly broad patterns (`next-server`, `cloudflared`) that could match and kill unrelated processes (e.g., other Next.js apps running on localhost). 

This fix narrows the matching to only 9router-specific processes by checking for:
- `9router` in command line
- `/bin/app/` path (9router's install structure)
- `9router-app` (standalone server binary name)
- `cli.js` (CLI wrapper)
- Tray binaries (`tray_darwin`, `tray_linux`, `tray_windows_release`)

Generic `next-server` and `cloudflared` matches are removed since they could belong to other applications.


Ali-Shaikh:
Closing in favor of existing PRs #2133 and #2292 which address the root cause properly by renaming the process at startup.

PR #3986, opened 2026-09-11
fix(antigravity): omit requestType="agent" to prevent false upstream 429 quota exhaustion

Recoba86:
### Summary
Omit `requestType: "agent"` from the Antigravity request envelope in both the translator (`openai-to-gemini.js`) and executor (`antigravity.js`).

### Problem
Google's Cloud Code internal endpoint (`daily-cloudcode-pa.googleapis.com`) recently enforces an extremely low, separate quota bucket whenever `requestType: "agent"` is present in the request envelope.

When agent clients (such as Oh My Pi / OMP, Claude Code, etc.) send requests that include their full system prompt and tool definitions (~25k-30k tokens), Google immediately rejects the request with:
```json
{
  "error": {
    "code": 429,
    "message": "Resource has been exhausted (e.g. check quota).",
    "status": "RESOURCE_EXHAUSTED"
  }
}
```
This occurs even when the Google account has 99%+ remaining quota. 9Router's internal quota manager then marks the account as 429, cascades through all configured accounts, triggers `STRIKE_429: 3x 429; CACHE_BLOCK 15m`, and locks the accounts.

### Empirical Testing & Root Cause
We tested the exact same 28,119 token payload (system prompt + 13 tool declarations) directly against Google Cloud Code API:
- **With `requestType: "agent"`:** Google immediately fails with `429 RESOURCE_EXHAUSTED`.
- **Without `requestType: "agent"`:** Google accepts the request and returns `200 OK` in ~2.5s (`promptTokenCount: 28119`).

### Changes
1. `open-sse/translator/request/openai-to-gemini.js`:
   - Omit setting `envelope.requestType = "agent"` for Antigravity requests.
   - Omit `requestType: "agent"` in `wrapInCloudCodeEnvelopeForClaude`.
2. `open-sse/executors/antigravity.js`:
   - Omit `requestType: "agent"` from the returned request body in `transformRequest`.

### Verification
- Tested live with real agent workloads (OMP client sending large system prompt + 13 tools to `ag/gemini-3.8-flash-high`).
- Response generated immediately with zero 429 errors and no account lockouts.


jarjut:
Works for me. Fixes the false 429s I was getting even with quota left.

PR #3991, opened 2026-09-12
feat(quota): add opt-in configurable stagger groups

matebuteler:
## Motivation

Immediate auto-pings can keep multiple accounts on the same reset schedule. This adds opt-in stagger groups for explicitly selected connections, plus an independent account-routing preference that consumes quota with the earliest reset first.

## Stagger groups

- Create, reorder, enable, and disable explicit connection groups in Quota Tracker. Groups default off; session and weekly policies have independent toggles.
- Plan phase slots from the reference member and provider-reported window duration. Existing active windows remain usable until expiry; routing protection prevents premature restarts while awaiting reserved slots.
- Preserve reservations through the first zero-usage observation, cached samples, and restart. Bounded 150-second catch-up avoids an accidental full-cycle delay without starting before actual expiry.
- When both policies are inactive, the longer-duration policy governs. Session alignment must not override an enabled weekly reservation.
- Verify sliding weekly resets before scheduling them; unknown weekly windows receive a bounded 180-second verification guard. Distinguish fixed rollover from transient zero usage.
- Require completed Codex SSE responses before recording successful pings; preserve failure cooldowns.
- Start the scheduler through Node instrumentation, including headless startup. Persist groups and observations in existing SQLite records with v3 stagger signatures.

## Prefer earliest quota reset

- New per-provider **Prefer earliest quota reset** toggle, disabled by default, for Codex, Claude, Antigravity, and Gemini CLI.
- For the same requested model, prefer the eligible account whose applicable session or weekly quota resets first. Keep Codex normal/review/Spark pools and model-scoped provider limits separate.
- Locks, exhausted applicable quotas, exclusions, and stagger protection remain hard gates. Eligible connection pins take precedence; equal reset times retain existing fill-first or round-robin behavior.
- Refresh quota in the background every five minutes when this preference alone is enabled, without sending warm-up messages. Reuse existing quota polling when auto-ping is separately enabled; never fetch quota in the request-selection path.
- Store normalized snapshots separately from credentials and provider-specific configuration. Transactional writes reject older/equal observations and recheck opt-in. Missing or stale metadata does not block routing; snapshots expire after ten minutes.
- Preserve unrelated provider options and existing override-clearing semantics in both provider UIs and the settings API.

## Limitations

- Stagger warming supports Codex and Claude. Antigravity remains unsupported for warming and excluded from stagger phase calculations; its earliest-reset routing preference is independent.
- External account activity, missing observations, and incompatible session/weekly calendars limit exact alignment guarantees.
- Historical production observations verified a planned stagger deadline, not sustained live convergence. The latest fixes and earliest-reset feature have not been deployed by this update.

## Validation

- **323 tests passed across 15 focused suites**, covering scheduling, routing, model-specific quota mapping, atomic persistence, background refresh, settings/UI behavior, and startup:

```sh
# From tests/
npx vitest run unit/quota-reset-routing.test.js unit/quota-routing-persistence.test.js unit/quota-routing-usage.test.js unit/provider-quota-routing-ui.test.js unit/quota-stagger.test.js unit/quota-auto-ping.test.js unit/quota-stagger-routing.test.js unit/settings-stagger-groups.test.js unit/startup-autoping.test.js unit/instrumentation-autoping.test.js unit/provider-quota-visibility.test.js unit/codex-spark-quota-tracking.test.js unit/antigravity-quota-routing.test.js unit/antigravity-weekly-quota.test.js unit/usage-dispatch.test.js
```

- Production Next.js build passed. Isolated production smoke passed for health, login, dashboard, default-off settings, preference-only scheduler start/stop, validation, and override clearing.
- Backend and test ESLint checks passed. The two existing provider UI files report four hook errors and one warning, independently reproduced on the pre-feature commit.
- Independent review approved after fixing settings-clear compatibility and snapshot ordering/atomicity. No live-provider traffic was required for these tests.


matebuteler:
This has been great for my workflow, effectively halving the amount of time I have to wait for the 5h session limit to reset by just having two different Codex accounts. Still testing this, though - will mark it ready for review if a week goes by with no quota reset errors.
Best

PR #3999, opened 2026-09-12
fix(gemini): expand shorthand string subschemas in tool declarations

coozgan:
## Summary

Claude Code against any Gemini-family model fails with a 400 from Google, and every leg of a fallback combo fails identically:

```
Invalid value at 'tools[0].function_declarations[104].parameters.properties[4].value'
(type.googleapis.com/google.cloud.aiplatform.v1.Schema), "object"
INVALID_ARGUMENT
```

Observed on `vertex`, `antigravity`, `github`, and `openrouter` (Google upstream) — four different transports, one shared cause, so a fallback chain cannot route around it.

## Cause

Agent and MCP tool definitions routinely use the shorthand

```json
{ "properties": { "value": "object" } }
```

for `{ "properties": { "value": { "type": "object" } } }`. Gemini's `Schema` proto has no union accepting a bare string, so **one** occurrence anywhere in ~154 tool declarations rejects the whole request.

`cleanJSONSchemaForAntigravity()` already normalises far more exotic cases, but every pass recurses with

```js
for (const v of Object.values(obj)) if (v && typeof v === "object") walk(v);
```

A string subschema is not `typeof "object"`, so it was invisible to *all* passes and reached the wire untouched.

Reproduced against the current sanitizer before the fix:

```
in : {"type":"object","properties":{"value":"object"}}
out: {"type":"object","properties":{"value":"object"}}     <-- unchanged
```

## Fix

New `expandStringSchemas()` pass, run as **phase 0** — before any pass that would skip strings. It rewrites the *parent's* slot (a primitive cannot be mutated in place) across schema-valued slots (`items`, `contains`, `if`/`then`/`else`, `not`, `propertyNames`, ...) and schema-valued maps (`properties`, `patternProperties`, `$defs`, `definitions`, `dependentSchemas`).

After the fix the expanded object also picks up the existing placeholder treatment, so it satisfies Gemini's "object schemas need properties" rule:

```
out: {"type":"object","properties":{"value":{"type":"object",
      "properties":{"reason":{...}},"required":["reason"]}}}
```

Deliberately narrow: only **string** values in schema position are touched. `additionalProperties: false` stays a boolean and is still stripped later; `required` and `enum` are arrays of literal strings, not subschemas, and are asserted untouched.

## Test plan

- [x] New `tests/unit/gemini-string-subschema.test.js` — **12 tests**, covering the exact reported shape, bare `items`, all four scalar type names, nesting inside arrays/objects, `$defs`, placeholder insertion, and negative cases (`required`, `enum`, `additionalProperties: false`, plain object schemas).
- [x] Full `unit/` suite, branch: **49 failed / 1125 passed**.
- [x] Full `unit/` suite, clean `master` (stashed, re-run): **49 failed / 1113 passed**.
- [x] Same 49 pre-existing failures, +12 newly passing -> **zero regressions**.
- [x] Registry baselines byte-for-byte equal (81 providers, 117 alias tokens, OAuth URLs).

Not verified live: I have no Claude Code session against this deployment. The reproduction and fix are exercised against the real sanitizer, but an end-to-end Claude Code run is still worth doing before closing.


coozgan:
## Update: real payload capture found a second, distinct bug

A captured 155-tool Claude Code request (the exact configuration that reproduces the report) contained **zero** shorthand string subschemas — yet still failed. Running it through the sanitizer reproduced the violation, which means **9router was creating the invalid value itself**, not passing one through.

### Second cause: schema maps mistaken for schemas

Every pass recursed with a blind walk:

```js
for (const v of Object.values(obj)) if (v && typeof v === "object") walk(v);
```

This cannot distinguish a schema **map** (`properties`, `$defs`, `patternProperties`, `definitions`, `dependentSchemas` — keyed by *user-chosen names*) from a schema **node**. Any tool declaring a property named `properties` has its map walked as if it were a schema.

`mcp__pascal__set_zone` does exactly that:

```json
{"type":"object","properties":{
  "levelId":{"type":"string"},
  "polygon":{"type":"array","items":{"type":"array","items":{"type":"number"}}},
  "label":{"type":"string"},
  "properties":{"type":"object","propertyNames":{"type":"string"},"additionalProperties":{}}}}
```

`ensureObjectType` saw the map, found no `type`, and stamped one **into the map**:

```json
"properties": { "levelId": {...}, "properties": {...}, "type": "object" }
                                                      ^^^^^^^^^^^^^^^^^
```

A bare string lands where Gemini's proto requires a `Schema`, and Google rejects the entire request — which is why the field path in the error points at a *property index* rather than any genuinely malformed input. Indices differ between the client's tool list and the translated one, so `[104]` in the log and `set_zone` at `[105]` in the capture are the same class of failure.

### Fix

`forEachChildSchema()` descends map **values** rather than the map itself, and all nine recursion sites route through it — `convertConstToEnum`, `convertEnumValuesToStrings`, `mergeAllOf`, `flattenAnyOfOneOf`, `flattenTypeArrays`, `convertPrefixItems`, `ensureObjectType`, `ensureArrayItems`, `expandStringSchemas`, `cleanupRequired`, `addPlaceholders`.

`ensureArrayItems` had the same latent flaw for a property named `items`; `expandStringSchemas` (commit 1) did too. Both are covered.

Before / after on the real tool:

```
before: properties: { levelId, polygon, label, properties, type: "object" }   <-- invalid
after : properties: { levelId, polygon, label, properties }                   <-- correct
```

### Verification

- Real 155-tool payload through the sanitizer: **0 post-sanitizer violations** (was 1).
- `tests/unit/gemini-schema-maps.test.js` — 6 tests: the real `set_zone` shape, properties named `items`/`type`, and an explicit guard that traversal still descends into genuine nested schemas (a fix that simply stopped recursing would pass the others).
- Full `unit/` suite: **49 failed / 1131 passed**, against a clean-`master` baseline of **49 failed / 1113 passed** measured in the same session — same 49 pre-existing failures, +18 newly passing across both commits. No regressions.

Still not verified end-to-end against a live Claude Code session; the fix is proven against the real captured payload, not a live request.

coozgan:
## Verified live

Built from upstream `master` with this branch merged, running locally, against a real Vertex-backed `gemini-3.8-flash`.

Request: an Anthropic Messages call carrying the `set_zone` tool — a schema with a property literally named `properties`, the shape that previously drew `INVALID_ARGUMENT` from every provider.

```
HTTP 200
event: message_start
data: {"type":"message_start","message":{"id":"FkSmatD0POzGsvsP87CjgQ4","model":"gemini-3.8-flash",...}}
event: content_block_delta
data: {"type":"content_block_delta","index":0,"delta":{"type":"text_delta","text":"ok"}}
```

Accepted and streamed normally. Before the fix the same payload failed at the schema-conversion step with:

```
Invalid value at 'request.tools[0].function_declarations[N].parameters.properties[M].value'
(type.googleapis.com/google.cloud.aiplatform.master.Schema), "object"
```

All four merge cleanly onto current `master` (#3964, #3965, #3966, #3999) and were verified together in the same working tree.

PR #4002, opened 2026-09-12
fix(stream): stop blocking an Ollama upstream's NDJSON as a non-SSE body

### LINKED ISSUE #3985: 9router: Fix Ollama support by allowing application/x-ndjson content type
When using Ollama with 9router from Codex, requests to /v1/chat/completions fail with 503 "Provider error (reset after 5s)". The root cause is 9router's streaming content-type
validation rejecting Ollama's "application/x-ndjson" responses. 9router only allows "text/event-stream" and "application/json".

The fix requires adding "application/x-ndjson" to the content-type allowlist in `chunks/8895.js` (line ~14775):

```diff
- !L.includes("application/json")
+ !L.includes("application/json") && !L.includes("application/x-ndjson")
    ```                                                                                                                                                                        
    This matches Ollama's standard streaming response format. Patching this file allows 9router to correctly translate between OpenAI API and Ollama.                          
    Note: This patch is required for all Ollama providers. The 9router build system doesn't currently handle this format, so the patch must be applied manually until upstream 
    fixes it.                                                                                                                                                                  
    Relevant log line from 9router:                                                                                                                                            
    [21:47:47] 🟣 ✗ BLOCKED 200 · ollama-local/qwen3:30b · non-SSE (application/x-ndjson)

### PR DESCRIPTION
**Pull request by L4XB**
Fixes #3985. The report is right that Ollama is blocked, and right about the line; the fix here is narrower than the proposed allowlist, for a reason.

## What is actually happening

`ollama-local` dials Ollama's **native** endpoint:

```js
// open-sse/providers/registry/ollama-local.js
transport: { baseUrl: "http://localhost:11434/api/chat", format: "ollama" },
```

`/api/chat` answers `application/x-ndjson` on a **successful** stream, and the router already knows how to read it: `open-sse/translator/response/ollama-to-openai.js` is registered `(OLLAMA → OPENAI)` with the docstring "Convert Ollama NDJSON response to OpenAI SSE format", and `tests/unit/ollama-stream-tail.test.js` already drives that transform.

The blocker is `handleStreamingResponse`, which runs **before** the transform is built:

```js
if (upstreamContentType && !upstreamContentType.includes('text/event-stream') && !upstreamContentType.includes('application/json')) {
```

That guard exists for HTML/text error pages (its own comment names a Cloudflare 5xx page) which crash the SSE pipe. It compares the content type alone, with no awareness of what format the provider declared, so the one provider whose successful stream is NDJSON can never get past it. Hence `BLOCKED 200 · ollama-local/qwen3:30b · non-SSE (application/x-ndjson)` and the 503 the reporter saw.

## Why not the allowlist in the issue

Adding `application/x-ndjson` to the shared condition would also accept NDJSON from a provider declared `format: "openai"`, where nothing translates it: the guard's refusal is the correct answer there, and widening it would turn a clean error into an empty stream. So the guard now consults the declared format:

```js
const UPSTREAM_STREAM_CONTENT_TYPES = {
  [FORMATS.OLLAMA]: ['application/x-ndjson'],
};
```

A content type an upstream format legitimately streams is not an error page. Everything else is unchanged, and a new format that streams something unusual adds one line here rather than another `includes`.

The issue points at `chunks/8895.js:14775`, which is the built bundle; the source is `open-sse/handlers/chatCore/streamingHandler.js`, so this needs no manual patching after a build.

## Tests

`tests/unit/ollama-ndjson-not-blocked-3985.test.js`, four cases against the real `handleStreamingResponse`:

- an Ollama-format upstream streaming `application/x-ndjson` is not blocked, and the client is served `text/event-stream`
- an HTML error page is **still** blocked, with the same `[200]: 502 Bad Gateway` body, which is what the guard is for
- `application/x-ndjson` from a provider whose format is **not** Ollama is still blocked
- `text/event-stream` and `application/json` still pass for any provider

Pre-fix control: with only `streamingHandler.js` reverted, exactly the first case fails (`expected false to be true`) and the other three pass, so each asserts its own property.

## No regression, measured

`tests/__baseline__/verify-no-regression.mjs` cannot classify anything outside the upstream container: it keys each test on `f.name.split("/app/")[1]`, which is `undefined` in a normal checkout, so every baseline red reads as a regression. I compared the failing sets of two full runs instead:

| | passed | failed |
| --- | --- | --- |
| clean tree | 2220 | 116 |
| with this change | 2221 | 115 |

**Newly failing: 0.** Newly passing: one, my own Ollama case. The 115 reds are the documented baseline (CLAUDE.md notes the suite is not green on a plain checkout). `tests/unit/ollama-stream-tail.test.js` and `ollama-usage.test.js` pass 9/9, and `npx eslint` is clean on both changed files.

## Not verified

I have no Ollama instance here, so the end-to-end path (Codex → 9router → `/api/chat` → back) is not something I ran. What is measured is that the guard no longer rejects the response and the client is handed an SSE stream; the translation itself was already covered by the existing test. @jayarajjg, if you can run this branch against your Ollama, that would close the last gap.


L4XB:
Force-pushed to `6ad36f84`: the branch had been cut from my own open #3973 rather than `master`, so the first push carried that commit plus a snapshot file one of my local test runs had written. The PR is now the two files it should be:

```
open-sse/handlers/chatCore/streamingHandler.js    | 16 +++-
tests/unit/ollama-ndjson-not-blocked-3985.test.js | 96 ++++++++++++++
```

Re-verified at that head: `unit/ollama-ndjson-not-blocked-3985` and `unit/ollama-stream-tail` pass 8/8.

PR #3482, opened 2026-08-23
fix(capabilities): mark DeepSeek V4 Flash Vision as vision-capable

Spoon94:
## Summary
- `*deepseek-v4*` matched first without `vision:true`, so `stripUnsupportedModalities` replaced images with `image omitted: model has no vision support` on **both** opencode-go and commandcode.
- Register `deepseek-v4-flash-vision-exp` on OpenCode Go and `deepseek/deepseek-v4-flash-vision-exp` on Command Code (plus a bare-id alias that remaps to the official GOAT id).
- Pass image blocks through the commandcode translator instead of dropping them as `[image omitted]`.

## Test plan
- [x] `getCapabilitiesForModel` reports vision for `deepseek-v4-flash-vision-exp` (ocg + cmc) and not for `deepseek-v4-flash`
- [x] OpenCode Go catalog + `supportedFormats` include the vision model
- [x] commandcode translator keeps `data:image/png;base64,BBBB`
- [ ] Send an image via `ocg/deepseek-v4-flash-vision-exp` — console must not show `image omitted`
- [ ] Send `cmc/deepseek-v4-flash-vision-exp` — upstream model id is `deepseek/deepseek-v4-flash-vision-exp`, no 403 `Model/provider not recog`

PR #4004, opened 2026-09-12
fix(commandcode): forward images instead of "[image omitted]"

### LINKED ISSUE #1330: AI SDK-style image parts are dropped by multimodal OpenAI translators
## Summary

Several multimodal OpenAI-input translators accept OpenAI `image_url` data URLs, but drop or ignore AI SDK-style image parts shaped as:

```json
{ "type": "image", "image": "data:image/png;base64,..." }
```

Any OpenAI-compatible client that sends image content in this shape loses the image before the request reaches providers that otherwise support image input. OpenCode is one affected client, but the compatibility gap is generic.

## Affected paths

- Kiro: should map to `userInputMessage.images[].source.bytes`
- Gemini / Antigravity: should map to Gemini `inlineData`
- Claude / Anthropic-compatible: should map to Claude `{ type: "image", source: { type: "base64" } }`
- OpenAI Responses: should map to `{ type: "input_image", image_url }`
- Ollama vision models: should map to `message.images[]` raw base64

CommandCode currently converts images to `[image omitted]`; that should remain unchanged unless upstream image support is confirmed.

## Reproduction

Send an OpenAI-compatible chat request to any affected multimodal provider with an AI SDK-style image content part:

```json
{
  "model": "kr/claude-sonnet-4.6",
  "messages": [
    {
      "role": "user",
      "content": [
        { "type": "text", "text": "Describe this image" },
        { "type": "image", "image": "data:image/png;base64,<base64>" }
      ]
    }
  ]
}
```

## Expected

The image is translated into each provider's existing image input shape, same as equivalent OpenAI `image_url` data URLs.

## Actual

Some translators only handle:

```json
{ "type": "image_url", "image_url": { "url": "data:..." } }
```

and sometimes Claude-style:

```json
{ "type": "image", "source": { "type": "base64", ... } }
```

They do not read `contentPart.image`, so AI SDK-style image parts are ignored or serialized as text.

## Related

This generalizes the remaining image-shape gap after #1026. PR #1331 broadens the fix across affected OpenAI-input translators and adds regression coverage.

tariq951:
## Problem

`open-sse/translator/request/openai-to-commandcode.js` replaced every OpenAI `image_url`/`image` content block with the literal text `[image omitted]` before sending to `https://api.commandcode.ai/alpha/generate`:

```js
} else if (part.type === OPENAI_BLOCK.IMAGE_URL || part.type === OPENAI_BLOCK.IMAGE) {
  blocks.push({ type: OPENAI_BLOCK.TEXT, text: "[image omitted]" });
}
```

So CommandCode models that support vision never see the image and truthfully answer "I can't see the image" — even though `/v1/models` advertises `vision: true` for them. This was catalogued as a known bug in `tests/translator/bugs-gemini-cursor-commandcode.test.js` (`it.fails("image content is preserved")`).

## Fix

Emit a real Anthropic-style image block that `/alpha/generate` accepts:

- data URI → `{ type:"image", source:{ type:"base64", media_type, data } }`
- `http(s)` → `{ type:"image", source:{ type:"url", url } }`

`toImageBlock()` accepts both the OpenAI shape (`{type:"image_url", image_url:{url}}`) and the AI SDK shape (`{type:"image", image:"data:..."}`, see #1330). Opaque/unfetchable URLs still fall back to the placeholder rather than producing an invalid block.

## Verification

Unit tests updated: the former `it.fails` case now passes, plus new cases for remote URLs, AI SDK parts, and the placeholder fallback.

Also verified **live** against `api.commandcode.ai` through a running 9router instance — `cmc/deepseek/deepseek-v4.1-flash`, `cmc/moonshotai/Kimi-K2.6` and `cmc/Qwen/Qwen3.6-Plus` correctly identify red/blue/green test images.

> Note: there is an open conflicting attempt at this in #3482 which used `{type:"image", image:url}` and had a strict-mode runtime bug. This patch uses the shape confirmed against the live endpoint.

Fixes #1330 (CommandCode path).


tariq951:
Related: #1330 (AI SDK-shaped image parts). This PR covers the CommandCode path for both the OpenAI and AI SDK shapes. Also supersedes the CommandCode translator change in #3482, which targeted a different model id and used a different block shape.

Issue #3977, opened 2026-09-11
Cursor provider returns silent empty responses (HTTP 200, OUT 0) on all cu/* models since ~Sep 10 — soft-block; PR #3161 resolves it


w1ndhunter:
## Summary
Since ~Sep 10 2026, every request through the Cursor provider (`cu/*`) returns a silently empty completion: HTTP 200, ~430 ms, `OUT 0`, no error frames, no 429. This is a new Cursor server-side behavior (soft-reject) — it is **not** the known "429 Update Required" retirements from #2685, and **not** a token problem.

## Environment
- 9Router: 0.5.75 (Docker `decolua/9router:0.5.75`) and 0.5.69 (Docker) — both affected
- macOS arm64 (Apple Silicon), container via Rancher Desktop
- Cursor auth: token import via `/api/oauth/cursor/import` (`cursorAuth/accessToken` + `storage.serviceMachineId` from `state.vscdb`)
- Client: opencode via OpenAI-compatible `/v1`, but reproducible with plain curl

## Timeline
- 2026-09-09 ~13:44 — `cu/composer-2.5` works (`'OK'`, OUT 1) on 0.5.69, token imported ~1 h earlier
- 2026-09-10 — upgraded to 0.5.75, everything else works
- 2026-09-11 ~11:47 — all `cu/*` requests return silent empty

## What we ruled out
1. **Token staleness** — re-imported a token Cursor issued minutes before the test (`time` claim verified in the JWT). Still silent empty.
2. **9Router version** — rolled back to 0.5.69: same failure. Full matrix:

   | build | token | plain text | tools + history |
   |---|---|---|---|
   | 0.5.69 | stale | OUT 0 silent | OUT 0 silent |
   | 0.5.75 | stale | OUT 0 silent | OUT 0 silent |
   | 0.5.75 | fresh | OUT 0 silent | OUT 0 silent |

3. **Account/quota** — the same Cursor account still works fine through the official Cursor client (`cursor-agent` CLI) at the same time, so the account is not quota-blocked. (Caveat: model-specific pools can't be fully excluded.)

## Repro (minimal)
```bash
# on a build from master (0.5.69/0.5.75), with a Cursor connection imported:
curl http://localhost:20128/v1/chat/completions \
  -H "Authorization: Bearer <9router-key>" -H "Content-Type: application/json" \
  -d '{"model":"cu/composer-2.5",
       "messages":[{"role":"user","content":"Reply with exactly: OK"}],
[...]

Issue #3300, opened 2026-08-14
cursor ide not working


clockworkchen:
[429]: {"error":{"message":"Update Required","type":"rate_limit_error","code":"ERROR_GPT_4_VISION_PR

mattecnology:
this rate limit

MagonYT:
same error/issue

AMSeify:
This looks like the same Cursor AgentService empty-reply / hang cluster. Tracking the framing fix in https://github.com/decolua/9router/issues/4016 and PR https://github.com/decolua/9router/pull/4017. cc @decolua @anuragg-saxenaa

PR #4017, opened 2026-09-13
fix(cursor): stop AgentService empty turns and silent tool hangs

### LINKED ISSUE #4016: fix(cursor): AgentService empty replies (OUT 0) and silent hangs with tools on Composer/Grok/Auto
## Summary

Cursor-hosted models (`cu/composer-2.5`, `cu/cursor-grok-*`, `cu/default`) often return **HTTP 200 with `OUT 0`** (~400–600ms) or **hang until abort** when the client sends tools. The same account works in Cursor itself. This is a 9Router AgentService framing bug, not quota.

cc @decolua @anuragg-saxenaa

## Root causes (verified live on 0.5.75)

1. **Empty turn (`OUT 0`)** — AgentService `RunRequest` field 8 (`custom_system_prompt`) makes Cursor return an empty turn. Caveman/Ponytail inject a system message that was sent there. Composer/Grok also put the visible answer in `thinking_delta` (interaction_update field 4); the executor only read `text_delta` (field 1). Thinking variants need `ModelDetails` (field 3), not only `RequestedModel` (field 9).
2. **Silent hang (`chunks=0`, abort after 60–148s)** — MCP tools were sent on the run request **and again** on the request-context ack. Cursor AgentService then stalls with HTTP 200 and zero SSE bytes. IDE builtins (shell/read/…) aborted the whole turn instead of being rejected so the model could continue.
3. **RTK no-op on Cursor** — `openai-to-cursor` rewrites `role:tool` into user XML `<tool_result>`. RTK only compresses source-format `tool_result` / `role:tool`, so it ran too late.

## Fix

PR follows this issue. In short:

- Fold system text into the current user message; **never send field 8**
- Send `ModelDetails` field 3; map Composer `</think>` / Grok thinking-as-content
- Empty request-context ack (do not echo tools); reject IDE execs and continue
- Encode MCP tools on the run request; surface field-11 execs as OpenAI `tool_calls`
- Run RTK on the **source-format** body **before** translation (matches `open-sse/AGENTS.md`)

## Related issues / PRs

Please look at these together — they are the same Cursor empty/hang/tool cluster:

- #3977 — silent empty responses (`OUT 0`) on all `cu/*` since ~Sep 10
- #3161 — AgentService MCP bridge (overlapping; this issue is the empty-turn + hang framing)
- #1077 — Cursor doesn't respond with default / Auto
- #2282 — Cursor Auto model unused / empty
- #1811 — Cursor models: no tool calls and no response
- #1490 — Composer blocked by tool calls
- #1747 — Composer 2.5 via Cursor provider
- #3300 — Cursor IDE not working
- #2782 — Cursor provider tool call error
- #1981 — Cursor proxy invalid
- #3276 — AgentService HTTP/2 proxy tunneling
- #2685 — earlier ChatService `429 Update Required` retirement
- #2149 — tool-name translation on the legacy path
- #3463 — combo failover treats empty HTTP 200 streams as success

## Repro

```bash
curl http://localhost:20128/v1/chat/completions \
  -H "Authorization: Bearer <9router-key>" -H "Content-Type: application/json" \
  -d '{"model":"cu/composer-2.5",
       "messages":[{"role":"system","content":"hi"},{"role":"user","content":"Reply with exactly: OK"}],
       "max_tokens":30,"stream":false}'
# broken: 200, content null, OUT 0
# fixed:  "OK" / "Hi! …", OUT > 0
```

With ~37 client tools, `cu/default` previously hung until abort; after the empty request-context ack it answers in a few seconds.

## Test plan

- [ ] `npx vitest run unit/cursor-agent-proto.test.js unit/cursor-agent-exec-request.test.js unit/cursor-composer-thinking.test.js unit/rtk-cursor-pretranslate.test.js`
- [ ] Live `cu/composer-2.5` + system prompt → non-empty assistant text
- [ ] Live `cu/default` + many tools → no hang, `AgentService request_context ack`
- [ ] Token savers on: `[RTK] saved …` + `⚙ RTK:… · CAVEMAN:full · PONYTAIL:full`

AMSeify:
## Summary
- Stop Cursor-hosted models (`cu/composer-2.5`, `cu/cursor-grok-*`, `cu/default`) returning **HTTP 200 / `OUT 0`** or hanging when clients send tools.
- Fold system prompts into the current user message — AgentService field 8 (`custom_system_prompt`) produces an empty turn (this also unblocks Caveman/Ponytail).
- Send `ModelDetails` (field 3); map Composer `</think>` / Grok `thinking_delta` into visible content.
- Ack request-context **without echoing MCP tools** (double-advertise stalls the HTTP/2 stream); reject IDE builtins so the model can continue.
- Run RTK on the source-format body **before** `openai-to-cursor` rewrites `role:tool` into user XML.

Closes #4016
Related: #3977, #3161, #1077, #2282, #1811, #1490, #1747, #3300, #2782, #1981, #3276, #2685, #2149, #3463

cc @decolua @anuragg-saxenaa

This is a focused framing/codec fix. It does **not** replace #3161 (session-resume MCP bridge); it addresses the empty-turn + hang path that still reproduces on master/`0.5.75`.

## Test plan
- [x] `npx vitest run unit/cursor-agent-proto.test.js unit/cursor-agent-exec-request.test.js unit/cursor-composer-thinking.test.js unit/rtk-cursor-pretranslate.test.js unit/headroom-chat-core.test.js` (52 passed)
- [x] Live `cu/composer-2.5` + system prompt → non-empty text (`OUT > 0`)
- [x] Live `cu/default` + many tools → answers in a few seconds (`AgentService request_context ack`)
- [x] Token savers on: `[RTK] saved 23221B / 39096B (59.4%) via [git-diff,grep,ls,tree,dedup-log]` and `⚙ RTK:5 · CAVEMAN:full · PONYTAIL:full`
- [ ] Re-check Docker image / region-restricted Cursor accounts (`403 not supported in your region` is independent)

Made with [Cursor](https://cursor.com)


kopach:
Not sure if anything got changed on Cursor side again, but I can't make it work on this branch. Here is a simple test:


```sh
curl http://localhost:20128/v1/chat/completions \
  -H "Authorization: Bearer sk-<token>" -H "Content-Type: application/json" \
  -d '{"model":"cu/composer-2.5",
       "messages":[{"role":"user","content":"Reply with exactly: OK"}],
       "max_tokens":30,"stream":false}'
```
I'm getting back `null` content



```json
{
  "id": "chatcmpl-msg_0000000000000",
  "object": "chat.completion",
  "created": 1789542658,
  "model": "composer-2.5",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": null // << - no content (NULL)
      },
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 4870,
    "completion_tokens": 0,
    "total_tokens": 4870,
    "estimated": true
  }
}

```

decolua:
Thanks @AMSeify for the contribution! Reviewed and merged into master. 🙏

PR #4023, opened 2026-09-13
feat(commandcode): record prompt-cache reads and add quota/usage tracking

cielhaidir:
## What

Command Code (`cmdc`) is already a provider here, but two things were missing:

1. **Prompt-cache reads were dropped**, so cached tokens were billed at the full input rate.
2. **No quota/usage card**, even though the API exposes the same credit data the CLI shows.

## 1. Cache accounting

Upstream's NDJSON usage block reports cache reads:

```json
{"inputTokens":22044,"inputTokenDetails":{"noCacheTokens":156,"cacheReadTokens":21888},
 "outputTokens":16,"totalTokens":22060,"cachedInputTokens":21888}
```

`commandcode()` in `open-sse/translator/concerns/usage.js` returned only
`promptTokens / completionTokens / totalTokens`, so `cached_tokens` never reached
`extractUsageFromResponse` and the cache portion was charged at the input rate.
For `deepseek/deepseek-v4-flash` that is `$0.14` vs `$0.0028` per 1M.

**Important:** `inputTokens` is cache-inclusive
(`noCacheTokens + cacheReadTokens == inputTokens`), so the fix emits
`cachedTokens`/`reasoningTokens` as details and deliberately leaves `promptTokens`
untouched — adding them would double-bill the cache portion.

Measured on real traffic after the fix: 3.36M cached tokens billed at the cached
rate instead of input rate, **~17x cheaper** for those requests.

## 2. Quota / usage dashboard

New `open-sse/services/usage/commandcode.js` reads the alpha API the provider
already talks to for `/alpha/generate` — no CLI shelling:

| Endpoint | Feeds |
| --- | --- |
| `/alpha/whoami` | optional `orgId` scope |
| `/alpha/billing/credits` | 5-hour + weekly windows, credits |
| `/alpha/billing/subscriptions` | plan id, current period end |
| `/alpha/usage/summary` | billing-period spend |

Rendered rows:

- **5-hour** — `used` / `cap` credits, `resetAt` per window
- **Weekly** — same
- **Monthly credits** — the plan allowance for the billing period, derived as
  `remaining monthlyCredits + billing-period spend` (`monthlyCredits` is the
  plan's *remaining* allowance, not the cap)

Wiring: registered in `USAGE_HANDLERS`, `features.usage`/`usageApikey` added to
the registry (dashboard skips providers without the flag), plus an explicit
`commandcode` case in the ProviderLimits parser.

Quotas carry `remainingPercentage` only — the table reads `remaining` as a 0–100
percentage, so an absolute credit amount there would render incorrectly.

### Note on the CLI identity headers

Cloudflare fronting this API fingerprints requests and answers
`403 error code: 1010` when the CLI headers are absent. The handler sends
`x-command-code-version`, `x-cli-environment` and a `user-agent`; without them
the endpoints look unreachable but the key is fine.

## Verification

```
tests/unit/usage-concern.test.js    (9 tests)  ✓  incl. new cache-detail case
tests/unit/commandcode-usage.test.js (5 tests) ✓  mapping + dashboard rendering
```

Live against a real account through the handler:

```
plan: Command Code (individual-pro-v1)
  5-hour           1.10 / 16        remaining  93%   reset 2026-09-13T10:25:34Z
  Weekly          12.35 / 40        remaining  69%   reset 2026-09-17T11:56:59Z
  Monthly credits 12.38 / 80.02     remaining  85%   reset 2026-10-10T11:39:53Z
```

`npm run build` passes.


cielhaidir:
Superseded by #4025, which is the same change opened from my own fork (cielhaidir/9router-fork) and includes an additional fix: credit amounts are now formatted with a dot decimal, since the quota table applies the viewer's locale and rendered 1.261 as "1,261" (read as 1261 against a cap of 16). Closing this one to avoid duplicate PRs — please review #4025.

PR #4034, opened 2026-09-13
feat(models): merge member capabilities into combo /v1/models entries

ItachiSan:
Expose context window and feature set for combos instead of bare stubs

## Problem
Combos in GET /v1/models were emitted as bare stubs ({ id, object, owned_by: "combo" }) with no capabilities, context_length, or max_completion_tokens. The capability/context-window decoration only runs in the per-provider loop, which combos never enter. Clients reading the list (e.g. trajectory-compaction thresholds) therefore see no context window for a combo and guess high, so the combo never reaches its real compaction point and can hard-fail upstream.

## Solution
Added mergeComboCapabilities(memberStrings, comboByName) which aggregates each member's getCapabilitiesForModel output: boolean features are OR'd, contextWindow/maxOutput are min'd (a request can route to any member, so the combo is bounded by its smallest window), and format scalars (thinkingFormat, thinkingRange) take the first non-null. Nested combos are flattened by recursion over a name->combo map built from the already-fetched combos list, with a seen-set cycle guard. LLM combos now receive capabilities + context_length + max_completion_tokens; webSearch/webFetch combos stay bare.

## Details
- Imports: DEFAULT_CAPABILITIES from open-sse/providers/capabilities.js, parseModel from @/sse/services/model.js.
- Merge depth is unbounded (recursion) but cycle-safe.
- Added tests/unit/combo-models-capabilities.test.js: flat, single-nested, two-layer-nested (plan -> tier -> selection-provider -> model), and web-combo cases -- all green.

💘 Generated with Crush

Co-Authored-By: Crush:code-fast


kuehnberger:
Independent verification of this approach on live combos (v0.5.75, GET /v1/models before/after a local min-window patch in the same buildModelsList combo loop):

- Before: combo entries carried only id/object/owned_by — OpenAI-compatible clients fell back to small default context sizes.
- After: combo-free-9r-spark -> context_length 1048576, combo-free-9r-1m -> 1000000, combo-free-9r -> 128000 (min contextWindow across members, fail-open).

This matches the min-window rationale here, and the recursion + cycle guard + boolean-OR in this PR goes beyond what I validated — supporting this direction over a flat min-only variant. One question: should webSearch/webFetch combos stay bare permanently, or gain a marker so clients can distinguish "intentionally bare" from "missing"? Happy to test this branch against live combos if a rebase or test extension would help.

ItachiSan:
Hope this can be merged soon! This helps a lot local coding agents.

Regaring the WebSearch / WebFetch combos, I am unsure how they work since I've never used them; having guidance would help.

kuehnberger:
Hi ItachiSan — good question about webSearch/webFetch combos. After tracing through the code, the PR handles them correctly. Here's what we found:

**How the combo capability merge interacts with web combos**

Web combos use bare provider aliases as members (e.g. "tavily", "firecrawl", "searxng") — the `models` array stores the provider name directly, not `provider/model`. This is confirmed at `src/sse/handlers/search.js:33` where the comment says "provider IS the model for webSearch."

If `mergeComboCapabilities` were to run on a web combo, its members would fall through to `getCapabilitiesForModel(null, "tavily")` → no catalog match → `DEFAULT_CAPABILITIES` floor (fabricated `contextWindow: 200000`). That's exactly the kind of fake number this PR is trying to eliminate — just inverted.

**The guard is correct and sufficient**

The `if (combo.kind === "webSearch" || combo.kind === "webFetch")` check at line 377 means the merge is never invoked for web combos. They stay as `{ id, object, owned_by, kind }` only — the `kind` field already signals "this is a web tool, not an LLM" to any client. No additional marker is needed to distinguish "intentionally bare" from "missing" — `kind` carries that.

**Retest against latest master**

Pulled the PR (commit `6793b2e76a0e`) and confirmed it applies cleanly against the latest `master` (`17c4cc76877b`). All 4 unit tests in `tests/unit/combo-models-capabilities.test.js` pass:

- Flat LLM combo merges min context window + OR'd features ✓
- Nested combos flatten correctly ✓
- Two-layer nested combos flatten ✓
- Web combo stays bare — `capabilities`, `context_length`, `max_completion_tokens` all `undefined` ✓

This is a valuable fix for OpenAI-compatible clients that read `/v1/models` to size context windows. Hope this can be merged soon!

PR #4039, opened 2026-09-13
refactor: remove retired providers and media services

johnitvn:
Closes #4038

## Scope

- Remove Qoder and the direct NVIDIA NIM provider.
- Preserve NVIDIA-named models catalogued by independent providers.
- Remove standalone media services and their dashboard surface while retaining LLM chat routing.
- Clean persisted retired records safely through a SQLite migration.

## Verification

Pending implementation.

Issue #4038, opened 2026-09-13
refactor: remove Qoder, NVIDIA NIM, and standalone media services


johnitvn:
## Summary

Remove the Qoder provider and all Qoder-only logic, remove the direct NVIDIA NIM provider (`id: `nvidia``), and remove standalone media services from 9Router.

## Scope

- Delete Qoder registry, executor, OAuth/PAT flow, usage/model catalog, UI, tests, assets, and documentation references.
- Delete the direct NVIDIA NIM provider and its NIM-only behavior while preserving NVIDIA-named model IDs catalogued by other providers (for example Kilo Gateway).
- Delete standalone image, video, music, TTS, STT, embeddings, web-search, and web-fetch APIs, handlers, dashboard pages, provider configurations, and tests.
- Retain LLM/chat routing and chat-level multimodal/native-search capabilities.
- Migrate SQLite data so retired connections and standalone-media records cannot leave stale runtime state.

## Expected behavior

- Qoder and NVIDIA NIM no longer appear as configurable providers or routable models.
- NVIDIA model IDs supplied by a different provider remain available through that provider.
- Removed standalone endpoints and `/dashboard/media-providers/*` are no longer served.
- Existing non-media LLM connections, combos, aliases, and history remain usable after migration.

## Verification

Run focused Vitest coverage, provider/alias regression baselines, lint, production build, and an app smoke test for retained LLM chat routing.

## Duplicate check

Searched open and closed issues for Qoder, NVIDIA, and media-provider removal; no matching issue found.

PR #4044, opened 2026-09-13
refactor!: remove Qoder provider (registry, executor, OAuth/PAT, usage, UI, tests)

johnitvn:
Implements the Qoder-only unit of #4038 (parent: #4039). Branched from `decolua/master` (17c4cc76), not the parent branch.

## What is removed
- Registry entry `open-sse/providers/registry/qoder.js` (alias `qd`, uiAlias `qd`) + its line in the generated `registry/index.js` (81 → 80 providers)
- `QoderExecutor` (`open-sse/executors/qoder.js`) and its export/registration in `executors/index.js`
- Shared transport libs `open-sse/shared/qoder/` (COSY RSA/AES/MD5 signing, WAF-bypass `Encode=1` body encoding, `{statusCodeValue, body}` SSE coalescing, attachments, context tiers, constants)
- OAuth: `src/lib/oauth/providers/qoder.js`, `src/lib/oauth/services/qoder.js`, `QODER_CONFIG` / `PROVIDERS.QODER` constants, the device-code poll branch and `noPkceDeviceProviders` entry in the OAuth route, the `qoder` branches in `OAuthModal` (`_qoderNonce/_qoderMachineId/_qoderVerifier` extraData)
- Usage: `getQoderUsage` + `resolveQoderCredentials` (`open-sse/services/qoderModels.js`) and the `qoder` USAGE_HANDLERS entry
- Model catalog: `resolveQoderModels` resolvers in `/api/providers/[id]/models` and `/api/v1/models`; validate-route PAT-exchange `case "qoder"` and its `OAUTH_TEST_CONFIG` entry
- Dashboard: provider-page "Import Qoder models" flow + PAT label, AddApiKeyModal `pt-…` placeholder / bulk-PAT copy, ProviderLimits color + usage formatting `case`
- Assets/docs: `public/providers/qoder.png`, i18n literals in 5 locales, CHANGELOG mentions, `open-sse/AGENTS.md` shared-dir listing
- Tests/baselines: `tests/unit/qoder{,-billing,-context-tier}.test.js`, `"qoder"` in the golden-url-header SPECIALIZED set and usage-dispatch SUPPORTED list, `qd/qoder` in `verify-alias.mjs` ALIAS_TOKENS; `providers-baseline.json` + `alias-baseline.json` regenerated

## What is deliberately NOT touched
- Generic OAuth / device-code / PKCE / usage / provider machinery (all qoder-specific branches removed; shared fallback paths now simply have one fewer consumer)
[...]

PR #4045, opened 2026-09-13
feat(db): migration 009 cleans retired Qoder/NVIDIA provider and media data

johnitvn:
Data-compatibility unit for #4038 (Qoder / NVIDIA NIM / standalone media removal). Parent draft PR: #4039.

## What

- **Migration 009** (`src/lib/db/migrations/009-cleanup-retired-provider-media-data.js`, transactional, no DDL):
  - Deletes `providerConnections` for retired providers — Qoder under **both** persisted ids: registry id `qoder` and model-string alias `qd`; direct NVIDIA id `nvidia` only.
  - Deletes media provider nodes (`type = 'custom-embedding'`).
  - Combos: drops all media kinds (`embedding/image/tts/stt/webSearch/webFetch/video/music`); strips retired-provider members (`qd/auto`, `nvidia/nim`, …) from kept LLM combos; deletes combos left with zero members.
  - `modelAliases` kv: removes entries whose **key or value** is a retired-provider model string (both writer routes persist opposite directions: `alias→model` via `PUT /api/models/alias`, `model→alias` via `PUT /api/models`).
  - `customModels` kv: removes entries with a retired `providerAlias` or a media `type` (absent type defaults to `llm` → kept).
  - `disabledModels` / `pricing` kv rows keyed by a retired provider are removed.
  - Settings JSON: strips `providerStrategies` / `quotaVisibility` keys for retired providers (reliably identifiable); media `comboStrategies` entries are left in place (inert, not reliably identifiable without reading the combos table).
  - **Usage history is never touched.**
  - NVIDIA scope guard: matching is **first URL segment only** — `kgw/nvidia/nemotron-…` (kilo-gateway catalog) is preserved; never a mid-string `nvidia/` match.
- **`src/lib/db/helpers/retiredData.js`** — single source of truth for the removal rules, reused by all import paths.
- **Re-import safety** (coordinator requirement):
  - `importDb()` sanitizes the full-DB payload **before writing**, so importing a pre-removal export cannot reintroduce retired connections, media nodes/combos, aliases, custom models, pricing, or settings references.
  - Legacy JSON import (`db.json` via `importLegacyMain`, `disabledModels.json` via `importLegacyDisabled`) runs through the same sanitizers; filtering happens **before** the `importWithAssertion` row-count checks so the abort-on-mismatch invariant still holds.
  - `usage.json` history import is untouched — historical rows (including retired-provider rows) are preserved verbatim.
- `SCHEMA_VERSION` 1 → 9 (pre-migration lightweight backup triggers per existing convention; version 9 slots after the parent branch's 2/3/8 chain to avoid collisions at integration).

## What is intentionally NOT here

No runtime/feature removal — this unit only cleans persisted data. Provider-registry, executor, translator and dashboard removals live in the sibling units of #4039.

## Tests

`tests/unit/db-cleanup-retired-provider-media.test.js` — 5 cases:
1. migration 9 registered, `latestVersion() === 9`.
2. Full clean pass: retired connections/nodes/combos/aliases/custom models/disabled/pricing/settings removed; LLM rows preserved; `qd/…` alias shapes, member-stripped LLM combo, `kgw/nvidia/…` preserved; usage history (incl. a qoder row) preserved; schemaVersion stamped.
3. Idempotency on reboot after the migration ran.
4. `importDb()` filters an old export payload (asserts sanitized return value + that retired disabled keys are not written back).
5. Legacy `db.json` / `disabledModels.json` import filters retired rows while `usage.json` history (qoder row) is kept.

Result: `5 passed` for the new suite; `db-migration-chain`, `db-sqlite-vs-lowdb`, `db-driver-chain`, `embedding-usage-persistence`, `dashboard-guard` all green. `db-concurrent` has 3 failures that reproduce identically on clean HEAD (pre-existing, unrelated — verified in a detached temp worktree at the base commit).

## Integration notes for #4039

- Migration **version 9** and `SCHEMA_VERSION = 9` must be reconciled when stacking with the parent branch's migrations 2/3/8 (expected final chain: 1,2,3,8,9 renumbered sequentially if needed).
- This branch is the single owner of the data-cleanup migration; the Qoder-unit peer confirmed no competing migration.


johnitvn:
Residual note for review (from the sibling Qoder unit, verified here): frozen baseline files under `tests/__baseline__/` still mention retired providers, with one distinction:

- `current.json` / `baseline-results.json` — historical vitest-run records; **no verify script reads them** (verified: `verify-no-regression.mjs` reads `known-fails.txt` + a results file passed via argv). `known-fails.txt` itself has no retired mentions. Safe to leave as historical data.
- `alias-baseline.json` and `providers-baseline.json` — **do have consumers**: `verify-alias.mjs` (which also hardcodes `qd`/`qoder`/`nvidia` in its expected-alias lists) and `verify-providers.mjs`. Once the provider-registry removal units land, these baselines + the hardcoded lists will need regeneration, otherwise `verify-alias.mjs` / `verify-providers.mjs` will report drift.

This PR touches no registry/alias code, so those verifies are unaffected by this branch alone — flagging so regeneration is assigned to the registry-removal unit rather than falling between units.

johnitvn:
Follow-up correcting my note above: the baseline regeneration for the **Qoder scope already landed** on `refactor/remove-qoder` (c6849500, #4044) — verified against that committed tree: `verify-alias.mjs` no longer lists `qd`/`qoder`, and both `alias-baseline.json` / `providers-baseline.json` contain no qoder/qd entries. The only retired mentions left in those files are `nvidia` (registry + baseline + verifier list), which correctly belongs to the NVIDIA-removal unit: if that unit regenerates the baselines, expect trivial textual conflicts there, plus the usual `CHANGELOG.md` `# Unreleased` heading overlap.

This branch's conflict surface is none of those — its 7 changed files are entirely `src/lib/db/**` plus one new test file (no CHANGELOG, no baselines, no verify scripts).

johnitvn:
Closing per maintainer decision: this work continues as local commits only, no PRs.

Issue #4031, opened 2026-09-13
/v1/responses proxied as chat/completions → function tools + reasoning unusable on OpenAI reasoning models (gpt-6-astra, gpt-5.6-*)


maziyarvex:
### Version
9router 0.5.75 (npm), self-hosted, OpenAI-compatible upstream.

### Summary
OpenAI "reasoning" models (e.g. `gpt-6-astra`, `gpt-5.6-luna`, `gpt-5.6-sol`) cannot be used with **function tools** through 9router. Both the `/v1/chat/completions` and `/v1/responses` paths fail.

### Details

**1. `/v1/chat/completions` + function tools** — upstream returns:
```
400: Function tools with reasoning_effort are not supported for gpt-6-astra in
/v1/chat/completions. To use function tools, use /v1/responses or set
reasoning_effort to 'none'.
```

**2. The suggested workaround (`reasoning_effort: "none"`) also fails.** Even a request that does **not** set `reasoning_effort` at all returns:
```
400: Unsupported value: 'reasoning_effort' does not support 'none'
```
→ 9router appears to **inject `reasoning_effort: "none"`** for these models, but the upstream rejects `none` for reasoning models (valid values are `low`/`medium`/`high`). So they are unusable on `chat/completions`.

**3. `/v1/responses` returns the same error.** Calling `/v1/responses` directly (the path the error message points to) yields the *same* "Function tools ... use /v1/responses" error — which indicates 9router is **internally translating `/v1/responses` into `/v1/chat/completions`** before contacting the upstream, rather than proxying the Responses API natively.

### Expected
- `/v1/responses` proxied natively to the upstream Responses API, so function tools + reasoning work on these models; and/or
- Do not inject `reasoning_effort: "none"` for models whose upstream doesn't accept it.

### Repro (OpenAI-compatible upstream)
```bash
curl $BASE/v1/chat/completions \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" -d '{
  "model": "gpt-6-astra",
  "messages": [{"role":"user","content":"2+2? use the tool"}],
  "max_completion_tokens": 300,
  "tools": [{"type":"function","function":{"name":"calc",
    "parameters":{"type":"object","properties":{"expr":{"type":"string"}}}}}]
}'
```

### Note
Anthropic models (e.g. `claude-sonnet-5`) with the same `tools` payload work fine through 9router — the issue is specific to the OpenAI reasoning-model path.

PR #4048, opened 2026-09-14
fix(capabilities): OpenAI reasoning models cannot disable thinking

L4XB:
Refs #4031 — this fixes point **2** of that issue, not points 1 and 3. See "What this does not fix" below.

## The defect

`applyThinking` writes `reasoning_effort: "none"` for any openai-format model whose capabilities do not say otherwise (`thinkingUnified.js:237`). OpenAI's reasoning models reject that value outright:

```
400 Unsupported value: 'reasoning_effort' does not support 'none'
```

which is exactly what the reporter saw — including on requests that **never set the field**, which is what made it look like an injection.

Measured on `master` before the change:

```
codex/gpt-6-astra   -> reasoning_effort "none"   levels [none, minimal, low, medium, high, xhigh, max]
codex/gpt-5.6-luna  -> reasoning_effort "none"   levels [none, minimal, low, medium, high, xhigh, max]
kiro/gpt-5.6-sol    -> reasoning_effort "none"   levels [none, minimal, low, medium, high, xhigh]
```

and after:

```
codex/gpt-6-astra   -> reasoning_effort "minimal"   levels [minimal, low, medium, high, xhigh, max]
```

## The change is data, not logic

The mechanism already existed; these models simply did not declare it. `thinkingCanDisable: false` makes `applyFormat` clamp `"none"` to the minimum instead of disabling, and makes `getThinkingLevels` stop advertising a value it cannot honour. `minimaxai/minimax-m2.7` and `minimax-m3` under `nvidia` carry the flag for the same upstream reason, so this follows the existing precedent rather than inventing one.

Four entries gain the flag: `gpt-6-astra`, `CODEX_GPT_56_SOL_CAPS`, `CODEX_GPT_56_DEFAULT_CAPS`, `KIRO_GPT_5_6_CAPABILITIES`.

## One existing test changes, deliberately

`tests/unit/thinking-levels-gpt56-sol.test.js` pinned the advertised level lists with `"none"` as the first element. Those expectations lose that entry — advertising a level the upstream refuses is the bug. Everything the file is actually pinning (Codex levels differ from Kiro's; sol/terra carry `ultra`, luna does not; other codex models get no `max`) is untouched, and I left a note in the file saying why the value went.

## What this does not fix

- **Point 1** (`/v1/chat/completions` + function tools → "use /v1/responses") is upstream's own restriction on that endpoint; nothing here changes which endpoint is used.
- **Point 3** (`/v1/responses` internally translated to chat/completions rather than proxied natively) is the architectural half, and much larger than a capability flag.

With this change the reporter's *stated workaround path* stops failing for the wrong reason: a request that omits `reasoning_effort` no longer has `"none"` put in it. Whether that is enough to make function tools work end to end I could not verify — it needs a live OpenAI-compatible upstream and a key, which I do not have.

**One thing I could not check:** that `minimal` is accepted by these models. The reporter's error only proves `none` is rejected, and their note says the valid values are `low`/`medium`/`high`. `minimal` is this project's existing clamp target for openai-format models that cannot disable, so I followed it rather than inventing a different one — but if your upstream also refuses `minimal`, the clamp in `applyFormat` needs `low` instead and I will change it.

## Tests

`tests/unit/thinking-openai-reasoning-cannot-disable.test.js`, 19 cells over 8 provider/model pairs. Two accept controls:

- **an explicit level still passes through** — clamping `none` must not flatten every request to the minimum.
- **a model that CAN disable still disables** — `openai/gpt-5` keeps its `none`, because the flag is per model.

Mutations:

| mutation | result |
| --- | --- |
| revert the four capability flags (the bug) | 16 of 19 fail |
| filter `none` for every model, ignoring the flag | `a model that CAN disable still disables` fails |

```
$ npx vitest run tests/unit tests/translator
```

130 failures remain, and the failure set is **identical on `master`** with this branch stashed (146 there, of which 16 are this PR's own new cells failing for lack of the fix). No test fails on this branch that passes on master.


L4XB:
I updated this branch from master. It had been cut from the branch behind #4030, so it was also carrying that PR's commit. That change now stays in #4030 only. This PR's diff against master is just the capabilities change and its two test files.

Re-checked on current master: `thinking-openai-reasoning-cannot-disable.test.js` fails with master's `capabilities.js` (e.g. `codex/gpt-6-astra` still gets `reasoning_effort: "none"`) and passes with this change. Across the 48 thinking, capabilities, codex and kiro test files, master and this branch fail the same 26 tests, all in kiro and codex files this change does not touch.

PR #4053, opened 2026-09-14
feat(models): declare real deepseek-v4.1-flash effort levels (low..max)

raflyazf:
Probed `deepseek-v4.1-flash` live on Alibaba Cloud MaaS (both routes of the same workspace):

| Probe | Result |
|---|---|
| `output_config.effort` = low / medium / high / xhigh / max (anthropic route) | all `200` |
| `output_config.effort` = `none` | `400 Invalid value 'none' for output_config.effort. Supported values are: low, medium, high, xhigh, max.` |
| `reasoning_effort: "max"` (compatible-mode openai route) | `200`, `reasoning_content` present |
| `thinking: {"type":"disabled"}` | `200`, no reasoning output |

The current metadata undersells the model: `*deepseek-v4*` caps leave `thinkingEffortSupported` at the `false` default, and the format-derived levels (`none|high|max`) hide the fact that all five effort levels are accepted and `none` is actually the invalid one.

## Changes

- `capabilities.js`: `*deepseek-v4*` pattern caps now declare `thinkingEffortSupported: true`.
- `thinkingLevels.js`: new `*deepseek-v4.1*` pattern with levels `["none", "low", "medium", "high", "xhigh", "max"]` — scoped to v4.1 so older v4 models keep the conservative `none|high|max` set. `none` stays in the list as disable-thinking (the anthropic route rejects it as an effort *value*, but disabling thinking entirely is a 200).

Note: `thinkingEffortSupported` is currently only read by the `zai` case in `thinkingUnified.js`, so this is metadata/dashboard correctness today; the deepseek wire format already sends `reasoning_effort` unconditionally.


decolua:
Thanks @raflyazf for the contribution! Reviewed and merged into master. 🙏

PR #4059, opened 2026-09-14
fix(auth): don't cool down an account for a request-scoped 4xx

GimingRao:
## Problem

`checkFallbackError()` returns the transient 30s cooldown for any error that
matches no rule. A 400 caused by the *request* therefore put the account on
cooldown: the chat handler excluded that connection, found no other one, and
answered every later request in the window with a copy of the first error.

In practice one session's oversized request produced:

    [400]: This model's maximum context length is 1048576 tokens.
    However, you requested 1186139 tokens
    ...
    [AUTH] deepseek | all 1 accounts locked for deepseek-v4-flash (reset after 30s)
           | lastError=[400]: This model's maximum context length ...
    [CHAT] [deepseek/deepseek-v4-flash] [400]: { ...same body... }

so unrelated, healthy sessions appeared to fail with the same token count, and
the caller lost the real cause of its own failure.

## Change

Rule-less 4xx statuses no longer fall back or cool the account down; the upstream
error is returned to the client that caused it. Statuses that do describe the
account keep their existing rules (`401`, `402`, `403`, `404`, `429`), and the
text rules still win for rate-limit / quota / capacity wording, so a 400 whose
body says `rate limit` keeps its current behaviour.

## Tests

```bash
cd tests && npx vitest run unit/account-fallback-4xx.test.js
```

Covers the 400 "maximum context length" case, the account-scoped statuses, the
text-rule precedence and the unchanged transient cooldown for server errors.
`unit/github-monthly-usage-lock.test.js` and `unit/multimodal-drop-lock.test.js`
still pass.


decolua:
Thanks @GimingRao for the contribution! Reviewed and merged into master. 🙏

PR #4061, opened 2026-09-14
fix(opencode): strip prior reasoning items on Muse Spark Responses models

anojndr:
### Problem
When using `muse-spark-1.3-contributor-free` (or other Muse Spark contributor models) in multi-turn conversations or tool-calling loops under the Responses API, requests on subsequent turns fail with:
```
URL: https://opencode.ai/zen/v1/responses [400]: {"model":"muse-spark-1.3-contributor-free","error":{"param":null,"type":"invalid_request_error","message":"Error from provider (Console): Upstream request failed: [invalid_request_error] reasoning \`encrypted_content\` was not issued to this caller"}}
```

### Root Cause
1. Stateless mode (`store: false`) in OpenAI Responses API emits `encrypted_content` on reasoning items by default.
2. `encrypted_content` is encrypted with keys bound to the caller/account identity that generated it.
3. OpenCode Free uses public pooled credentials (`Bearer public`) routing to an internal upstream account pool on Console (and requests frequently rotate across proxy pool relays).
4. When subsequent turns replay prior reasoning items with `encrypted_content` to a different upstream caller, OpenAI rejects the request with `reasoning encrypted_content was not issued to this caller`.
5. Under `store: false`, omitting `encrypted_content` while keeping `type: "reasoning"` causes OpenAI to reject it as `Referenced reasoning item was not found or was deleted`.

### Solution
- Strip prior-turn `type: "reasoning"` items and continuity properties (`encrypted_content`, `reasoning_encrypted_content`) from `body.input` before dispatching in `OpenCodeExecutor` and `OpenCodeGoExecutor`.
- Bring `OpenCodeExecutor` into parity with `OpenCodeGoExecutor` by adding `normalizeResponsesInput`, `normalizeResponsesTools`, and `sanitizeResponsesItems`.
- Add unit tests verifying reasoning items carrying `encrypted_content` are stripped and tool definitions are normalized.


decolua:
Thanks @anojndr for the contribution! Reviewed and merged into master. 🙏

PR #4065, opened 2026-09-14
feat(codex): add per-account fast mode

Primexz:
### Problem

Codex accounts cannot consistently control priority processing because client-supplied service tiers may vary, and the account-level Fast mode setting is not available from the quota dashboard.

### Cause

The Codex executor previously derived service-tier behavior from each incoming request instead of a persisted per-account setting. The dashboard-safe provider response also omitted that setting, preventing the quota card from displaying or updating it.

### Fix

- Persist and validate a per-account `codexFastMode` setting.
- Force `service_tier: "priority"` when Fast mode is enabled and remove `service_tier` when it is disabled.
- Expose the saved setting through the dashboard-safe provider response.
- Add Fast mode controls to the provider-detail connection row and quota dashboard card.

### Verification

- Ran `npm run build` successfully.
- Ran `git diff --check` successfully.
- Confirmed the feature branch is pushed to `Primexz/9router`.


Primexz:
Closing this PR because the branch was based on a stale local master. A replacement branch and PR will be created from the current upstream master.

PR #4069, opened 2026-09-14
fix: preserve optional tool parameters across Responses and Codex tra…

felixfoertsch:
## Summary

Preserve Chat Completions function-tool optionality when routing through Responses, including Codex (`cx/*`). This targets the OpenCode v2 `subagent.sessionID` failure: omitting the argument creates a child, while supplying it requests a resume.

## Two concrete defects found in 0.5.75

1. `open-sse/translator/request/openai-responses.js` already preserves `parameters.required`, but forwards an omitted `function.strict` as `undefined`. JSON serialization removes that field. Chat Completions is non-strict by default, whereas Responses can normalize omitted strictness into strict mode, making optional fields required.
2. `open-sse/executors/codex.js::normalizeCodexTools` rebuilds each function declaration and discards `strict`, including explicit `false` and `true`. A translator-only fix is therefore insufficient for `cx/*`.

OpenAI documentation: https://developers.openai.com/api/docs/guides/function-calling#strict-mode

The user-reported direct-versus-routed OpenCode result is consistent with these defects. This PR does not claim a live reproduction of model behavior or OpenCode child creation.

## Changes

- Emit `strict: tool.function.strict ?? false` when translating Chat Completions function declarations.
- Capture and restore strictness while rebuilding Codex function tools.
- Default nested Chat-shaped tools reaching the executor to non-strict, while retaining the existing defaults for native flat Responses tools.
- Leave parameter schemas and existing `required` arrays unchanged; do not derive required fields from property names or silently repair explicitly strict schemas.
- Do not inject, remove, or replace `sessionID` in returned arguments. Legitimate resume IDs remain intact.

## Regression coverage

Adds `tests/unit/openai-responses-optional-params.test.js` with 26 cases covering:

- Optional `sessionID`; absent and empty `required`; nested objects, arrays, unions, and definitions; parameter-less functions.
- Explicit `strict` values, Chat default handling, native Responses defaults, and nested Chat tool normalization.
- Serialized outbound tool declarations, idempotence, input-schema preservation, and hosted/custom/namespaced tools.
- Replayed arguments and synthetic streamed arguments, both fragmented deltas and done-only payloads, for omitted and supplied child IDs.

The synthetic streaming fixtures verify transport fidelity, not actual model argument selection or OpenCode execution.

## Validation performed

In the available offline environment, the same 26 test assertions were run using Node 22's native test runner against the fetched production translator/executor source and the extracted production response-conversion function. External I/O and unrelated provider configuration dependencies were replaced by isolated adapters; helper behavior was copied from the inspected source.

| Version | Passed | Failed |
| --- | ---: | ---: |
| Original 0.5.75 | 8 | 18 |
| Translator change alone | 8 | 18 |
| Both changes | 26 | 0 |

`node --check` and `git diff --check` passed. Final GitHub blob hashes match the locally validated files.

**Not run:** the normal Vitest dependency graph, full repository suite/build, live Codex inference, or the OpenCode acceptance probe. Dependencies could not be installed in this environment. This PR remains a draft pending normal integration validation.

From a full checkout with dependencies installed, run:

```sh
cd tests
npm test -- unit/openai-responses-optional-params.test.js
```

## Live acceptance still required

Using OpenCode v2 2.0.3, rerun the original identical probe through `9router/cx/gpt-5.6-sol` after building and deploying this branch. Retain the direct-provider control.

- [x] Captured final provider request has `strict: false` for the translated subagent function, and its original `required` array does not include `sessionID`.
- [x] The new-child call's captured `rawInput` lacks `sessionID` entirely, not merely `null` or an empty string.
- [x] A real child session is created; the output contains its returned session ID and `ROUTED-SUBAGENT-OK`.
- [x] An explicit valid child `sessionID` can still resume that child.

No changes to OpenCode's parent/child validation are proposed. No automatic merge, deployment, or upstream issue/PR is included.


forggy0113:
Hi, just following up on this PR. I’m experiencing the same issue, and I’ve tested this PR locally — the fix works for me.

Would it be possible to review and merge this when you have a chance? This fix would really help with my use case. Thanks!

Issue #2918, opened 2026-07-30
Usage by API key merges distinct keys for today/24h; aggregate periods serialize raw keys


hiennguyenle:
`getUsageStats("today")` / `getUsageStats("24h")` group API-key usage by the
masked key prefix instead of a unique key identity, and aggregate periods
serialize raw API keys as JSON property names.

### Bug 1: distinct keys merge (today/24h)

In `src/lib/db/repos/usageRepo.js` the live-history branch builds the
grouping key from `maskApiKey()`, which keeps only the first 8 characters:

    const apiKeyMasked = maskApiKey(r.apiKey);
    const akKey = `${apiKeyMasked}|${r.model}|${r.provider || "unknown"}`;

Generated keys are `sk-{machineId}-{keyId}-{crc8}`
(`src/shared/utils/apiKey.js`), so the first 8 characters are `sk-` plus the
first 5 characters of the installation's machine id — **identical for every
key the installation creates**. Any two keys used with the same
model+provider in the window collapse into one row: requests, tokens and
cost are summed, and the row shows whichever key's name created the bucket
first. On one live deployment, 7 distinct keys shared 1 distinct first-8
prefix across >9,000 requests in 24h.

**Repro:** create two API keys with different names, send one request
through each (same model/provider), open Usage → Today: one row contains
both keys' traffic. `7d` shows two rows (that path keys by full key).

### Bug 2: raw keys serialized in responses (7d/30d/60d/all)

The aggregate branch copies persisted `usageDaily.data.byApiKey` entries into
`stats.byApiKey` under their persisted keys — `${fullRawKey}|model|provider` —
and three routes return the object verbatim (`NextResponse.json(stats)`):
`/api/usage/stats`, `/api/usage/stream` (always period `"all"`, so every
dashboard SSE update carries them), and `/api/usage/history`. Full API keys
(and, with `requireApiKey` off, any forwarded bearer secret) appear as JSON
property names in responses. The dashboard never renders them, but they cross
the wire into browser devtools and proxy logs — including unauthenticated
when `requireLogin` is false.

Note: `tests/unit/security-audit.test.js` ("byApiKey object keys should use
masked key, not raw key") enforces the masked template for the live branch,
but the masked prefix cannot serve as identity (bug 1) and the aggregate
branch never satisfied the test's intent (bug 2). The fix needs a third kind
of key: a stable identity that is neither raw nor masked.

I have a fix ready (PR to follow): key `stats.byApiKey` by the `apiKeys`
[...]

PR #4073, opened 2026-09-15
feat & fix (keys,models): per-key model access & token limits, fix OpenCode and compatible provider models

miqonee:
## Summary

This pull request addresses two provider catalog issues and introduces per-API-key model access control and token quotas:

### 1. Per-Key Model Access Control & Token Limits
- **Database Schema (`SCHEMA_VERSION: 2`)**:
  - Added `tokenLimit`, `usedTokens`, `allowedModels`, and `updatedAt` to the `apiKeys` table with additive auto-sync and pre-change safety backup.
- **Token Quota Tracking & Enforcement**:
  - Atomic `usedTokens` increment in SQLite transaction during `saveRequestUsage()`.
  - Rejection with HTTP 429 (`Too Many Requests`) when a key's `tokenLimit` is reached.
  - Support for resetting the used tokens counter from the dashboard without regenerating the key.
- **Model Restrictions Whitelist**:
  - Rejection with HTTP 403 (`Forbidden`) when a request targets a model not permitted for that key.
  - Supports exact model IDs (`gpt-4o`), provider-prefixed IDs (`oc/muse-spark-1.2-contributor-free`), provider wildcards (`oc/*`, `openai/*`), and glob patterns (`*claude*`).
  - `/v1/models` automatically filters models for authenticated API keys so clients only see models they are permitted to use.
- **Dashboard UI (`EndpointPageClient.js`)**:
  - Create Key modal: configure key name, optional token quota, and allowed models (via `ModelSelectModal` or pattern input).
  - Key list: progress bar and token usage stats (`used / limit`), quota exceeded badge, and allowed models summary.
  - Edit Key modal: modify key name, token limit, allowed models list, and reset token usage counter.

### 2. Provider Bug Fix: OpenCode Free models missing from `/v1/models`
- In `buildModelsList()`, `activeConnectionByProvider` only populated connections from SQLite `providerConnections`. Since `opencode` is a free no-auth provider (`noAuth: true`), it has no database connection row and was omitted from `/v1/models` as soon as any other provider was configured.
- Added virtual connection registration for `noAuth` free providers in `buildModelsList()`, making OpenCode Free models (`oc/*`) discoverable alongside other providers.

### 3. Provider Bug Fix: Compatible provider exposing all upstream models (~400 LiteRouter) instead of user-selected models (5-8)
- In `src/app/api/v1/models/route.js`, `rawModelIds.length === 0` previously triggered an upstream `/models` query even when user-configured custom models already existed, merging all 400 upstream models into `/v1/models`.
- Skipped dynamic upstream fetch if the provider already has configured custom models, aliases, or `enabledModels`.
- Added interactive model selection modal with search, multi-checkbox selection, and select-all in `CompatibleModelsSection` instead of blindly importing all upstream models.
- Fixed `isDisabled` check to inspect `providerId` as well.

### Verification & Tests
- Added unit tests in `tests/unit/per-key-and-provider-fixes.test.js` covering model pattern matching, key validation rules, token quota limits, model whitelist enforcement, and `/v1/models` filtering.
- Verified all baseline snapshots byte-for-byte (`verify-providers.mjs`, `verify-alias.mjs`, `verify-oauth-urls.mjs`).
- Production build passes cleanly (`npm run build`).

#### 1. API Keys Management (Quotas, Usage & Model Restrictions)
API Keys List

#### 2. Create API Key with Token Quota & Custom Model Rules
Create Key Modal

#### 3. Edit API Key, Quota Reset & Permissions
Edit Key Modal

#### 4. Model Catalog Selector
Select Models Catalog Modal

#### 5. OpenAI/Anthropic Compatible Provider Model Selection
Compatible Provider Models


miqonee:
Hey @decolua!

Huge thanks for 9router. I've been using it for quite a while now and honestly love the tool, it makes local routing and provider setups so much easier.

Decided to contribute a bit and add a couple of small things that I felt were really missing for daily use:

Per-key token limits and allowed models list. Super handy when sharing an endpoint with colleagues or friends so they don't burn through your quota or call the wrong models.
OpenCode Free models fix: they were disappearing from /v1/models once any other provider was configured.
Compatible providers model selector: added a nice modal with search and checkboxes instead of automatically dumping 400 upstream models into the catalog (literouter provider(( ).
All tests pass, build is clean, and I added UI screenshots to the description so you can see how it looks.

Hope you find this useful! Take a look whenever you have time, and let me know if anything needs to be tweaked.

Thanks again for the great work!

inyund:
Follow-up from an operator running 9router 0.5.75 (asgard, https://9r.inyund.xyz) — confirming the OpenCode Free half of this PR.

Reproduced on 0.5.75: with 3 real provider connections present (tokenrouter, commandcode, openrouter), `GET /v1/models` returns **6** ids and no `oc/*` at all, while the dashboard lists all four OpenCode Free models. Looking at `buildModelsList()`, the else-branch iterates `activeConnectionByProvider`, which is built only from `providerConnections` rows — and `opencode` is `category: "free"` / `noAuth: true`, so it can never have one. `noAuth` is only honoured at request time in `getProviderCredentials()` (which injects the `{ id: "noauth", connectionName: "Public" }` virtual connection); `buildModelsList()` never consults `FREE_PROVIDERS`.

Worth pointing out as a regression shape: because the empty-connections branch returns the whole static catalog, *adding* a provider makes `/v1/models` **smaller**. On this instance it went from 605 ids (no connections) to 6. Any client that discovers models from `/v1/models` silently loses the free providers at exactly the moment they configure their first paid one.

`/v1beta/models` does list 2 of the 4 (`muse-spark-1.2/1.3-contributor-free`) since it reads `PROVIDER_MODELS` directly, but it misses kv-custom models, so it isn't a workable substitute. `oc/big-pickle` and `oc/mimo-v2.5-free` here are hand-seeded via `POST /api/models/custom` (`kv` scope `customModels`) — they also disappear from `/v1/models` for the same reason.

Independent of discovery, all four route fine: `oc/big-pickle` → "PONG", `oc/mimo-v2.5-free` → "PONG" (incl. an image turn), `oc/muse-spark-1.2/1.3-contributor-free` → "PONG" over `/zen/v1/responses`. So the fix is purely catalog visibility.

The `noAuth` virtual-connection registration in `buildModelsList()` is exactly what's needed. Thanks for putting it together — the per-key `allowedModels` filtering in the same PR composes nicely with this, since a key scoped to `oc/*` currently sees an empty list.

miqonee:
I’ve been using and testing, it works great and helps bring complete order to the models.

miqonee:
Just pushed a quick follow-up commit with a couple of quality of life tweaks while testing this in daily use:

First, in the provider models import modal, I removed the browser alert popup that popped up every time after adding models. Also, previously added models had their checkboxes locked and disabled, so if you wanted to clean up models you had to delete them manually one by one from the main list. Now all checkboxes stay interactive, meaning you can uncheck models right there to remove them. The modal clearly tags what is being added and what is marked for removal, and applies everything in one click with Save Changes.

Second, in the key modal for selecting allowed models, you can now simply click on any provider title to toggle a wildcard rule like provider/* (such as openai/* or oc/*). It saves a lot of time when you want to grant access to an entire provider without having to select dozens of models individually or type the pattern by hand. Along with that, the wildcard matching logic was updated so it reliably recognizes both provider aliases and provider IDs.

miqonee:
By the way, I also built and pushed a Docker image with all the latest changes to Docker Hub (`miqone/9router:latest`) if anyone wants to test this functionality out directly without building it locally. Would really appreciate any feedback, thoughts, or critique!

miqonee:
## Updates in this revision:
- **Rebased & clean merge with v0.5.85**: Resolved all conflicts with the latest release, verified all baseline snapshots (`verify-providers.mjs`, `verify-alias.mjs`, `verify-oauth-urls.mjs`), and clean production build.
- **Interactive Model Selector for OAuth Providers**: Extended the interactive model selection modal (search filter, multi-select checkboxes, batch apply) to OAuth and live providers (`cline`, `clinepass`, `qoder`, `codex`, `cursor`, `zed`, etc.) with proactive token refresh before querying `/models`.

miqonee:
Found and fixed another bug while testing multiple keys today.

On the Usage page, different keys kept merging into a single row or swapping stats with each other. Looked into it and saw that maskApiKey was truncating keys down to just the machine ID prefix, so all keys generated on the same machine shared the exact same identifier (issues #3640 and #2918).

Fixed the masking so each key keeps its unique identity and stats track separately, and also fixed an issue where token counts were dropping to zero for models returning input_tokens/output_tokens.

PR #4076, opened 2026-09-15
fix(deepseek,model-catalog): vision for the V4.1-Flash ids, and make the synced catalog reach requests

izzzzzi:
### What

`deepseek-v4.1-flash` and `deepseek-flash` had no exact entry in `MODEL_CAPABILITIES`. They fell through to the `*deepseek-v4*` / `*deepseek*` patterns, which set reasoning/thinkingFormat/limits but **not** `vision`.

Both ids name the multimodal V4.1-Flash model, and each is the spelling used by a different source of truth:

- the registry id for V4.1-Flash on `opencode-go` (`opencode-go.js`, name "DeepSeek V4.1 Flash")
- the id models.dev lists for `opencode-go`, `openrouter`, `vercel`, `kilo`, `requesty`, `nano-gpt`, `deepseek` and others

So the table declares both spellings, the same way it already carries `claude-opus-4.6` and `claude-opus-4-6`.

The evidence is provider-scoped for the primary id: models.dev lists `opencode-go/deepseek-v4.1-flash` **itself** as `["text","image"]` (`opencode-go` currently exposes 36 ids, 22 of them image-input). `opencode-go` has no id normalising to `deepseek-flash`; that spelling's evidence is DeepSeek's own API record (`deepseek/deepseek-flash` -> `["text","image"]`, likewise `302ai` and `vancine`) together with this registry naming the entry "DeepSeek V4.1 Flash" (upstream `c712641` puts it first in the catalog as *the* V4.1 id). So neither entry rests on a cross-vendor id collision.

The text-only id is the superseded `deepseek-v4-flash`, which most providers list as `["text"]`; the table keeps it text-only and the test asserts that for the registry ids. It does not follow that every gateway keeps it text-only: models.dev records `deepseek/deepseek-v4-flash` as `["text","image"]`, and part 2 lets a gateway's own upstream record win, so that one gateway may raise it. The claim in this part is about the table, which is the layer that has to be right without a catalog. DeepSeek's release notes say the retired `deepseek-v4-flash` / `deepseek-v4-flash-vision-exp` ids route to V4.1-Flash, and to set the API model to `deepseek-flash`.

Because the flag was missing, image requests broke twice over:

1. `chatCore` replaced the image with its `[image omitted: model has no vision support]` placeholder before translation.
2. The capacity adapter -- vision enabled with an empty pool, which falls back to `oc/mimo-v2.5-free` -- floated a vision model ahead of the requested one, so image requests were answered by mimo instead of DeepSeek (global pool, so this happened across every provider).

### Change

- `open-sse/providers/capabilities.js`: canonical exact entries for `deepseek-v4.1-flash` and `deepseek-flash`.
- The other fields repeat the pattern-derived deltas, because an exact entry short-circuits the pattern table -- a vision-only entry would have silently dropped `reasoning` and the 1M context window down to the 200K default (verified: `reasoning=false, ctx=200000`).
- `tests/unit/capabilities.test.js`: new case asserting vision **and** reasoning/context survive, checked through the canonical path and registry/documented providers, plus that `deepseek-v4-flash` stays `vision: false`.

Two aggregators list `deepseek-v4.1-flash` as text-only in models.dev (`digitalocean`, `llmgateway-providers/consensusprotocol`); neither exists in this registry, so nothing here over-declares for them. If a provider ever needs the opposite, `PROVIDER_CAPABILITIES` is the place.

### What

Two defects in the models.dev catalog layer, both measured on a stock 0.5.75 install:

- **The source never reached request handling.** `capabilities.js` is bundled into every route chunk that needs it: the 0.5.75 server bundle carries six copies of the module, `setCatalogSource` is present in two of them, and neither of those two is among the chunks that serve `/v1/models`. `installCatalogSource()` at startup therefore wrote a slot that no request ever read, so `refine()` saw `catalogSource === null` for its whole life. Served capabilities for the registry are identical to a run of the tables with no catalog at all, while installing the source in-process flips them (`deepseek-flash`, `deepseek-v4.1-flash` -> `vision: true`); that is the proof, not a log.
- **Modalities were tallied across gateways and keyed by bare model id.** Short ids are not unique across vendors: `auto`, `free`, `efficient`, `frontier` are router modes in one catalog and model names in another, so a router mode inherited a stranger's vision. Of the 36 registry entries the catalog moved, 30 drew the modality from a provider other than the one being asked -- 14 of those are the same model under another vendor, 10 are image/video generators, and 6 are router aliases (`qoder/auto`, `qoder/efficient`, `bazaarlink/auto:free`, `kilo-gateway/kilo-auto/*`). The 6 are the dangerous ones: `vision: true` on a mode means an image is forwarded to it instead of being stripped, and the capacity adapter stops diverting it.

### Change

- `open-sse/providers/capabilities.js`: the source slot lives on `globalThis` (the local binding is the fast path), so an install in one bundle copy is visible to the others; `getModalities` is called with the gateway.
- `open-sse/providers/catalogOverride.js`: `getCatalogModalities(provider, model)` reads `provider:model` keys in the local provider id space -- the namespace the `providers` half has always used -- and the file carries `CATALOG_VERSION = 2`. The reader takes no alias map: attribution happens while building.
- `src/lib/modelCatalog/sync.js`: `build()` files each gateway's modalities under **every** local id that maps to its upstream provider (one upstream name can back more than one, e.g. `zhipuai` behind both `glm-cn` and `zhipu`) and, when the upstream name is not among them, under that as well -- a custom provider node can carry an upstream name without appearing in the registry snapshot the sync walks, and nothing else would resolve for it. The cross-vendor tally and `MIN_SHARE` are gone; the file is stamped `v: 2`; and a file from an older writer is rebuilt instead of sending `if-none-match`.
- `tests/unit/model-catalog-scope.test.js`: builder output shape and the absence of a model-only key, router modes staying text-only, every aliased local id getting its keys, the gateway reaching the reader, the source being visible to a second copy of the module, plus the migration path (a v1 file is inert, an older-schema file is refetched and rewritten at v2, a current file still sends `if-none-match` and accepts a 304).

### Trade-off, stated explicitly

Scoping is the conservative half of the choice, and it is worth being clear about what it costs:

- **Gained:** the only sources of a raised capability are records upstream makes about *that* gateway, so no router mode and no image generator can inherit anything, and no cross-vendor guess can turn a modality on.
- **Given up:** a gateway upstream does not list at all (`github`, `cursor`, `codex`, custom provider nodes such as `burngate`) no longer inherits a modality from another vendor's record for the same short id. Those keep whatever the tables resolve, which is the same fallback every other unknown model gets. This is a coverage limit, not a defect: those gateways were never described by upstream, they only looked like they were because the key dropped the provider.
  The lever, if more coverage is wanted, is `PROVIDER_ALIASES` in the sync, not the aggregation, which cannot tell "another gateway serving the same model" from "another model with the same short name". An alias may only be added for the *same* gateway, and the way to establish that is model-id overlap with the registry provider plus the gateway's own docs -- `codex -> openai`, `github -> github-copilot` and similar guesses look plausible by name and would silently hand a gateway someone else's modalities, which is the failure this change removes. Nothing was added on name resemblance.
- `refine()` only ever turns capabilities on, so a stale or text-only upstream record cannot strip an image from a model the tables marked multimodal: it simply contributes nothing.

### Migration

The file on disk is `v: 1` with flat keys, which this reader does not look up, so an un-rebuilt file contributes nothing (no half-applied state). The rebuild happens within a minute of the next boot, because the sync always runs at startup and a file whose `v` is older than the writer is refetched rather than answered with a 304; `POST /api/models/catalog-sync` triggers it immediately. The window is at most the next daily run for a process that never restarts.

One shared-state note: detaching for `collectEntries()` (which the sync must do, or deltas would be computed against the previous catalog and the file would erase itself over two runs) now nulls the slot for every copy in the process, not just the syncing one. That is a sub-second window once a day, and the existing `finally { installCatalogSource() }` puts it back.

### Verification

- `vitest run --config tests/vitest.config.js unit/model-catalog-scope.test.js unit/capabilities.test.js` -> 16 passed (9 + 7), which is exactly what the branch ships.
- On the pre-change modules the file cannot pass either way: it fails its own contract guard (the builder it drives is not exported there), and reverting just the reader and the source slot while keeping the new builder fails 6 of the 9 cases (alias attribution per local id, router-mode leak, namespace, gateway-to-reader, shared source, older-schema rebuild), so the cases pin the fixed behavior rather than describing it.
- End to end with the live models.dev catalog, through `build()` -> catalog file -> reader -> `getCapabilitiesForModel`, feeding the builder the same registry snapshot the sync builds (`{provider, model, current: getCapabilitiesForModel(p, m)}` over all 804 entries) and comparing against the tables with no catalog installed: **46 modality changes** and 81 limit changes, every one of them sourced from that gateway's own upstream record -- per-gateway `pdf`/`videoInput`/`audioInput` that the cross-vendor majority used to suppress. `qoder/auto`, `qoder/efficient`, `bazaarlink/auto:free` and `kilo-gateway/kilo-auto/*` stay text-only. 4387 gateway-scoped modality keys replace 1228 shared ones. (The limit half is pre-existing behaviour that the sync already wrote; it only shows up here because this run feeds the builder a real snapshot, which the earlier measurement did not.)

## Why the DeepSeek entry stays in the table

Gateway-scoped catalog keys fix the collisions, but they do not give the V4.1 ids a capability source: a gateway upstream does not list at all (`burngate`, or any custom provider node) has no record to read, and a gateway that upstream lists under a different model id for the same weights is not matched either (`opencode-go` lists this model as `deepseek-v4.1-flash`; this registry calls it `deepseek-flash`). An exact entry in the capability table is the layer for exactly that fact, and it applies whatever provider the request comes through.


decolua:
Thanks @izzzzzi for the contribution! Reviewed and merged into master. 🙏

PR #4083, opened 2026-09-15
fix(translator): keep tool-result images, restore Kiro tool names, preserve thinking display

Manan-Santoki:
Agent clients (Claude Code, and any client speaking the Anthropic Messages API with MCP tools) hit three gaps when routed to non-Anthropic upstreams. Each was reproduced against a live 9router 0.5.75 instance.

## 1. Images inside `tool_result` never reach OpenAI-format upstreams

`claude-to-openai` (and `claude-to-kiro`) keep only `text` blocks from a `tool_result`, so a screenshot returned by a browser tool is silently dropped and the model only sees "Successfully captured screenshot…". OpenAI's `tool` role is text-only, so the image is now attached to the user message that follows the tool messages, tagged `[Image from tool result <tool_use_id>]`; Kiro receives it as a user image. A tool result that carried only an image no longer dumps its base64 into the tool message via `JSON.stringify`.

## 2. Kiro tool calls come back with sanitized names

`normalizeKiroToolSpecs` collapses `__` to `_` (`mcp__browser__computer` → `mcp_browser_computer`) but the response path copies Kiro's name straight back, so the client rejects every tool call ("tool not found") and the model loops. The Kiro request translators now return the reverse map as `_toolNameMap`, which `chatCore` already lifts into the stream state for Claude cloaking, and `kiro-to-claude` / `kiro-to-openai` restore the client's name (streaming and non-streaming, including chunks passed through as-is).

## 3. `thinking.display` is dropped

`applyThinking` resets `body.thinking` to `{ type: "adaptive" }` (or `enabled` + budget), discarding the client's `display: "summarized" | "omitted"`. Anthropic returns signature-only thinking blocks when the field is missing on some routes, so clients that asked for summaries get empty thinking. The client's `display` is now preserved on adaptive and budget thinking.

## Tests

`tests/translator/agent-client-fixes.test.js` covers all three (11 cases). `thinking-unified`, `bugs-kiro`, `kiro-image-forwarding` and `kiro-conversation-canonicalization` still pass. The 9 pre-existing failures in `claude-kiro-direct.test.js` are unchanged and fail identically on `master`.


Manan-Santoki:
Two more commits on this branch, both verified live against a self-hosted 9router build:

- `e7839d6` claude: send `redact-thinking-2026-02-12` only when the client did not ask for `thinking.display: "summarized"`. Clients requesting summaries got signature-only thinking; Sonnet 5 / Opus 5 now return thinking text when asked.
- `ed28e04` translator: deliver tool-result images to Command Code (real `{type:"image"}` blocks instead of "[image omitted]"; `COMMANDCODE_IMAGES=off` restores the placeholder) and to Anthropic-compatible endpoints other than Anthropic by hoisting a tool result's images into the same user turn. Verified: Muse Spark (cmc + ocg), Kimi K2.6 (cmc) and DeepSeek (ocg) now describe a screenshot returned by a tool.

decolua:
Thanks @Manan-Santoki for the contribution! Reviewed and merged into master. 🙏

PR #4084, opened 2026-09-16
fix(opencode): route every responses-only OpenCode Go model (incl. thinking variants) to /responses

ridhoperdana:
## Problem

`opencode-go/gpt-5.6-luna` (and thinking variants like `gpt-5.6-luna(high)`) fails with HTTP 500 / 400:

```
URL: https://opencode.ai/zen/go/v1/chat/completions [500]: Internal server error
...
[400]: Unsupported parameter: 'messages'. In the Responses API, this parameter has moved to 'input'.
```

## Root cause (two parts)

1. `gpt-5.6-luna` is registered as Responses-only, but `OpenCodeGoExecutor.buildUrl()` only force-routed Muse Spark models to `/responses`. A chat-completions request for `gpt-5.6-luna` therefore hit `/chat/completions` with a Responses-shaped body.
2. With thinking suffixes (`gpt-5.6-luna(high)`), `findModel()` in `providerModels.js` looked up the literal ID, missed the registered `gpt-5.6-luna` entry, so `chatCore` fell back to the provider default format (`openai`) and sent `messages` — the executor change alone then produced `Unsupported parameter: 'messages'` against `/responses`.

## Fix

- `open-sse/executors/opencode-go.js`: `isResponsesModel()` also matches `gpt-5.6-luna` (base ID, so suffix variants are covered).
- `open-sse/config/providerModels.js`: `findModel()` strips a trailing `(...)` thinking suffix before the registry lookup, so `(high)`/`(xhigh)` resolve to the base model entry and get `targetFormat: openai-responses` (the `messages` → `input` translation then happens in `chatCore`, which runs before the executor).

Resulting flow: `gpt-5.6-luna(high)` → registry hit → `FMT: openai→openai-responses` → `input` → `https://opencode.ai/zen/go/v1/responses`.

## Tests

- Unit: `tests/unit/opencode-go-muse-spark-responses.test.js` (routing incl. `runtimeTransport` override) and `tests/unit/opencode-go-models.test.js` (format resolution for `(high)`/`(xhigh)`).
- Live: `POST /v1/chat/completions` with `ocg/gpt-5.6-luna(high)` returns 200 with `finish_reason: stop`; non-Luna models (e.g. `kimi-k2.6`) still route to `/chat/completions`.

— post by BOT


AFG473319:
Hi @ridhoperdana — we maintain AFRouter, a downstream fork of 9Router. We tested this PR and it fixes the `ocg/gpt-5.6-luna(high)` routing bug for us — nice find on the two-part root cause (executor URL + the `findModel` suffix lookup).

We cherry-picked it onto our fork ([AFG473319/AFRouter@0b80424](https://github.com/AFG473319/AFRouter/commit/0b804246694d1e836063eede4a4278c28cb19e48)) and ran our full test suite against it — no regressions, so it is live for our users today. Hope that helps it land upstream.

One thing we had to add to get the tests green: the new test in `tests/unit/opencode-go-models.test.js` calls `getModelTargetFormat`, which the file does not import. One-line fix:

```js
import { PROVIDER_MODELS, getModelSupportedFormats, getModelTargetFormat } from "../../open-sse/config/providerModels.js";
```

Optional nit, nothing to change in this PR: `grok-4.6` is also registered responses-only on `ocg` but is not in `isResponsesModel()`, so a chat-format request for it without a runtime transport hits `/chat/completions` with a Responses-shaped body — same class of bug. Might be worth a follow-up.

ridhoperdana:
> Hi @ridhoperdana — we maintain AFRouter, a downstream fork of 9Router. We tested this PR and it fixes the `ocg/gpt-5.6-luna(high)` routing bug for us — nice find on the two-part root cause (executor URL + the `findModel` suffix lookup).
> 
> We cherry-picked it onto our fork ([AFG473319/AFRouter@0b80424](https://github.com/AFG473319/AFRouter/commit/0b804246694d1e836063eede4a4278c28cb19e48)) and ran our full test suite against it — no regressions, so it is live for our users today. Hope that helps it land upstream.
> 
> One thing we had to add to get the tests green: the new test in `tests/unit/opencode-go-models.test.js` calls `getModelTargetFormat`, which the file does not import. One-line fix:
> 
> ```js
> import { PROVIDER_MODELS, getModelSupportedFormats, getModelTargetFormat } from "../../open-sse/config/providerModels.js";
> ```
> 
> Optional nit, nothing to change in this PR: `grok-4.6` is also registered responses-only on `ocg` but is not in `isResponsesModel()`, so a chat-format request for it without a runtime transport hits `/chat/completions` with a Responses-shaped body — same class of bug. Might be worth a follow-up.

Thank you @AFG473319 for the check! i updated the code to also handle other model and fix the test.

decolua:
Thanks @ridhoperdana for the contribution! Reviewed and merged into master. 🙏

PR #4086, opened 2026-09-16
fix(commandcode): preserve images and reasoning_effort on /alpha/generate

KhuatHieu:
## Problem

Command Code through 9Router dropped vision and ignored client effort:

- OpenAI `image_url` / Claude image blocks became `"[image omitted]"`.
- HTTP image URLs never inlined, so they never reached `/alpha/generate`.
- `reasoning_effort` landed on the envelope wrapper; native CLI sends it in `params`. DeepSeek family mapping remapped `low` → `high`.
- Catalog treated `deepseek/deepseek-v4.1-flash` as text-only (`vision: false`), so the vision adapter stole requests to `oc/mimo-v2.5-free`.
- Quota Tracker had no Command Code usage surface.

## Root cause

1. `openai-to-commandcode.js` stubbed image parts instead of the native `{type:"image", image:"data:…;base64,…", mimeType}` block.
2. `prefetchRemoteImages` did not include `FORMATS.COMMANDCODE` in `TARGETS_NEED_BASE64`.
3. `applyThinking` used family `thinkingFormat: "deepseek"` after translation and wrote top-level `reasoning_effort` / `thinking`.
4. Capability lookup hit `*deepseek-v4*` before any Command Code provider default.
5. Registry had no `features.usage` / `usageApikey`.

## Fix

- Translator maps OpenAI/Claude image parts to native generate blocks (data-URI).
- Prefetch converts `http(s)` images with the existing SSRF-safe fetcher before translation.
- `targetFormat === "commandcode"` writes `params.reasoning_effort` and passthrough `low|medium|high|xhigh|max`.
- Provider-scoped capabilities: vision except the CLI text-only denylist; `thinkingFormat: "commandcode"`.
- Quota Tracker: whoami + billing credits/subscriptions (credits vs plan cap, 5h, weekly). Labels use `AI_PROVIDERS[].name` (`Command Code`).

## Tests

- `tests/unit/openai-to-commandcode.test.js`
- `tests/unit/prefetch-images.test.js`
- `tests/translator/thinking-unified.test.js`
- `tests/unit/capabilities.test.js`
- `tests/unit/commandcode-usage.test.js`
- `tests/unit/usage-dispatch.test.js`
- Known-bug `it.fails` for Command Code images flipped to a passing assert.

Live: `cmc/deepseek/deepseek-v4.1-flash` with `image_url` stays on `commandcode` (not `opencode`/`mimo-v2.5-free`); Quota Tracker shows credits + windows.


decolua:
Thanks @KhuatHieu for the contribution! Reviewed and merged into master. 🙏

The image fix is the important half here — the translator was stubbing every image block to `[image omitted]` unconditionally, so it would have stayed broken even with the capability fixed. Prefetch + provider-scoped caps + the params.reasoning_effort placement all landed too. The Quota Tracker surface is a nice addition.

Verified on master: full suite shows no regression against a clean upstream/master baseline, the 7 touched test files pass (104 passed, 4 expected-fail), and both the provider and alias baselines are byte-for-byte equal.

PR #4099, opened 2026-09-16
fix(opencode): route Union Alpha through Messages API

### LINKED ISSUE #4098: OpenCode Union Alpha is routed to the wrong Zen endpoint
## Bug description

OpenCode Zen now exposes **Union Alpha Free** with raw model ID `union-alpha`. Unlike most free Zen models, Union Alpha uses Anthropic Messages:

- Endpoint: `https://opencode.ai/zen/v1/messages`
- Request/response format: Anthropic Messages (`claude` in 9Router)
- Required header: `anthropic-version: 2023-06-01`

9Router's built-in `opencode` executor currently sends every non-Muse model to `/zen/v1/chat/completions`. Therefore adding `union-alpha` as a custom OpenCode model does not work.

## Steps to reproduce

1. Open the built-in **OpenCode Free** provider.
2. Add custom model ID `union-alpha`.
3. Run the model test.

## Actual behavior

9Router resolves the raw model ID correctly, but routes the request to:

```text
https://opencode.ai/zen/v1/chat/completions
```

OpenCode Zen returns HTTP 500:

```json
{"type":"error","error":{"type":"error","message":"Internal server error"}}
```

Using `opencode/union-alpha` as the raw ID produces HTTP 401 because that is an OpenCode client-qualified identifier, not the upstream model ID:

```text
Model opencode/union-alpha is not supported
```

## Expected behavior

For raw model ID `union-alpha`, 9Router should:

1. Declare target format as `claude` so OpenAI-compatible client input is translated to Anthropic Messages.
2. Route upstream requests to `/zen/v1/messages`.
3. Include `anthropic-version: 2023-06-01` only for Messages requests.
4. Leave Muse Responses routing and other Chat Completions models unchanged.

## Root cause

`open-sse/executors/opencode.js` currently distinguishes only Muse Responses models from all other models. There is no Messages-model route. The OpenCode provider registry also lacks a static `union-alpha` entry with `targetFormat: "claude"`.

## Proposed fix

- Register `union-alpha` under the built-in OpenCode provider with `targetFormat: "claude"`.
- Add a small Messages model set in the OpenCode executor.
- Route matching models to `/zen/v1/messages`.
- Add the shared Anthropic API-version header on that endpoint.
- Add regression coverage for target format, endpoint, headers, translated body, and unchanged non-Union headers.

## Verification

Validated against OpenCode Zen documentation and reproduced through 9Router's built-in model test. A local implementation passes the focused OpenCode, capability, and reasoning tests (24 tests total).

emi-ran:
## Summary

- register OpenCode Zen's `union-alpha` model as Anthropic Messages format
- route Union Alpha to `/zen/v1/messages`
- send the shared `anthropic-version` header only on Messages requests
- preserve existing Muse Responses and other Chat Completions routing
- add regression coverage for format, endpoint, headers, and translated request body

Fixes #4098

## Root cause

The built-in OpenCode executor only distinguished Muse Responses models from all other models. Union Alpha therefore fell through to `/zen/v1/chat/completions`, although Zen serves it from `/zen/v1/messages`.

## Test plan

- `cd tests && npx vitest run unit/opencode-muse-spark-thinking.test.js unit/reasoningContentInjector.test.js unit/capabilities.test.js` — 24 passed
- `node tests/__baseline__/verify-providers.mjs` — passed
- `node tests/__baseline__/verify-alias.mjs` — passed
- `node tests/__baseline__/verify-oauth-urls.mjs` — passed
- targeted ESLint — 0 errors; one pre-existing anonymous-default-export warning in the provider registry file
- `git diff --check upstream/master...HEAD` — passed


decolua:
Thanks @emi-ran for the contribution! Reviewed and merged into master. 🙏

PR #4112, opened 2026-09-17
fix(antigravity): preserve inlineData parts for image models (image edit/img2img)

Ndqitisme:
## Problem

`AntigravityExecutor.transformRequest` rebuilds `contents` for image models with a **text-only** filter:

```js
const textParts = (c.parts || []).filter(p => p.text !== undefined).map(p => ({ text: p.text }));
```

This silently drops every `inlineData` / `inline_data` part. So when a client calls `POST /v1/images/generations` with an input image (`image` or `images[]` field — which `imageProviders/antigravity.js` correctly converts into an `inlineData` part), the image never reaches the upstream generateContent call. The model receives only the prompt and returns a brand-new unrelated image instead of editing the provided one.

## Repro

```bash
curl -X POST http://localhost:20128/v1/images/generations   -H "Content-Type: application/json"   -d "{"model":"ag/gemini-3.1-flash-image","prompt":"add a yellow border","image":"data:image/png;base64,<SRC>"}"
```

Output does not preserve the source image at all (verified with a unique visual marker — the marker disappears, proving the input was dropped rather than used for editing).

## Fix

- `open-sse/executors/antigravity.js`: image-model branch keeps `inlineData`/`inline_data` parts (normalizing to `inlineData` camelCase) alongside text parts.
- `open-sse/handlers/imageProviders/antigravity.js`: forward **all** `images[]` entries instead of only the first (multi-image editing is supported by the model).
- `open-sse/providers/registry/antigravity.js`: advertise `edit` + `multiImage` capabilities for `gemini-3.1-flash-image` so the dashboard Edit UI and capability checks reflect reality.
- `tests/translator/bugs-antigravity.test.js`: regression tests covering both `inlineData` and `inline_data` shapes.

Verified end-to-end on a patched local build: the same request now returns an edited image that preserves a unique marker from the source image.


decolua:
Thanks @Ndqitisme for the contribution! Reviewed and merged into master. 🙏

PR #4114, opened 2026-09-17
fix(kiro): preserve underscores in tool names and restore sanitized names in responses

### LINKED ISSUE #4113: Kiro translator collapses consecutive underscores in tool names and never maps them back (MCP tools: "No such tool available")
### Problem
In `open-sse/translator/concerns/kiroConversation.js`, `uniqueName()` normalizes tool names with `.replace(/_+/g, "_")`. Consecutive underscores are collapsed, so MCP-namespaced tools such as `mcp__gitea__search_repos` are sent to Kiro as `mcp_gitea_search_repos`.

The Kiro translators keep a forward `nameMap` (client name -> Kiro name) for history reconciliation, but nothing maps the name back on the response path (`_toolNameMap` is only populated by `openai-to-claude`). The model therefore calls the collapsed name, it is forwarded to the client unchanged, and the client rejects it.

### Reproduction
1. Claude Code (or any MCP client) -> 9router -> Kiro, with at least one MCP server configured (tool names follow `mcp__<server>__<tool>`).
2. Ask the model to call an MCP tool.
3. The client reports: `Error: No such tool available: mcp_gitea_search_repos`.

The call only succeeds when the model happens to spell the double-underscore form by itself, which makes the failure look intermittent. Built-in tools (`Bash`, `Read`, ...) are unaffected because they contain no `__`.

Kiro accepts tool names with consecutive underscores (verified: with the collapse removed, `mcp__*` tools are declared and called normally).

### Expected
1. Consecutive underscores in otherwise valid tool names are preserved.
2. When a name still has to change (illegal characters, length limit, dedup suffix), the original client name is restored on the response path, streaming and non-streaming, for both OpenAI and Claude target formats.

Version: 0.5.75 (also present on current `master`).

qisthidev:
### Summary
Fixes MCP tool calls through the Kiro provider failing with `No such tool available: mcp_<server>_<tool>`.

- `uniqueName()` no longer collapses consecutive underscores, so `mcp__server__tool` reaches Kiro unchanged (Kiro accepts it).
- `openaiToKiroRequest` / `claudeToKiroRequest` attach a reverse map (`kiroName -> clientName`) as `_toolNameMap` whenever sanitizing, truncating or de-duplicating changed a name. `chatCore` already strips that field before the upstream call and passes it to the stream state.
- `kiroToOpenAIResponse` / `kiroToClaudeResponse` restore the client's original tool name from `state.toolNameMap` on tool-use events.
- New tests: `tests/translator/kiro-tool-name-roundtrip.test.js` (underscore preservation, illegal characters, dedup collisions, response restoration).

### Testing
`npx vitest run translator/kiro-tool-name-roundtrip.test.js translator/bugs-kiro.test.js unit/kiro-conversation-canonicalization.test.js` -> all pass.
`translator/claude-kiro-direct.test.js` has 9 failures on `master` without this change (thinking/effort cases); the same 9 fail with it, none added.

Also verified end to end with Claude Code + several MCP servers against Kiro: `ToolSearch` -> `mcp__gitea__search_repos` succeeds.

Closes #4113


decolua:
Thanks @qisthidev for the contribution! Reviewed and merged into master. 🙏

PR #4116, opened 2026-09-17
fix(opencode): resolve 403 FreeTierError with canonical session format and valid User-Agent

ErfanBagheri404:
OpenCode upstream now validates free-tier requests (\Authorization: Bearer public\) server-side. Two checks gate every call:

1. **User-Agent** must be \opencode/<version>\ with version >= 1.17.0. Bare \opencode\ (or any foreign client UA) gets **403 FreeTierError** (or 426 if < 1.17.0).
2. **\x-opencode-session\** must match OpenCode's canonical descending identifier format \ses_ + 12 hex timestamp digits + 14 Base62 chars\ (\/^ses_[0-9a-f]{12}[0-9A-Za-z]{14}$/\). Missing headers, raw UUIDs (\ses_<32hex>\) or foreign session formats (e.g. \claude:*\) trigger **403 FreeTierError**.

I verified this against the real client by reverse-engineering \opencode-ai@1.18.31\: it sends \User-Agent: opencode/\\, \x-opencode-session: <ses_...>\ (canonical, one long-lived session per conversation), \x-opencode-request: <user message id>\ (\msg_...\ canonical, stable per turn), \x-opencode-client\ (defaults to \cli\), \x-opencode-project\ (project id, \global\ when unset), and falls back to \piKey: public\ when logged out — same auth 9router uses.

**This PR:**
- Defaults \OPENCODE_UA\ to \opencode/1.18.31\; validates downstream UAs, preserving valid versions (>= 1.17.0) while upgrading bare/outdated strings.
- Implements canonical session/request ID generation plus deterministic session translation for foreign downstream sessions.
- Introduces request-local credentials via \prepareRequestCredentials\, eliminating shared instance state leakage on the \OpenCodeExecutor\ singleton.
- Adds \	ests/unit/opencode-session.test.js\ (16 tests: generation, translation, statelessness, UA validation).

**Verification:**
- \
px vitest run unit/opencode-session.test.js\ — 16/16 pass.
- Live check against a local 9router gateway: \oc/muse-spark-1.3-contributor-free\ and \oc/mimo-v2.5-free\ return HTTP 200 (previously 403) with correct \x-opencode-session\ format on the wire.

Closes #4101
Closes #4103

PR #4117, opened 2026-09-17
fix(opencode): reuse one stable upstream session per identity to stop premature 429s

### LINKED ISSUE #3270: opencode deepseek v4 flash 429

ErfanBagheri404:
Stacked on #4116 — merge that first. This adds the follow-up commit on top (branch is based on #4116's branch).

**Problem (#3270):** routing OpenCode Free (\oc/...\) models through 9router returns \Rate limit exceeded / 429 FreeUsageLimitError\ after only a few requests, while the same model in the OpenCode TUI on the same machine/IP keeps working. Upstream quota is clearly not exhausted — 9router is burning it.

**Root cause:** with no explicit session, every request minted a **fresh** \x-opencode-session\, and upstream free-tier quota is accounted per session. The reset-after backoff visibly grows per request (2s → 8s → 15s+). The real CLI (\opencode-ai@1.18.31\, verified from its bundle) reuses **one long-lived canonical session per conversation** and sends the current user message id (stable per turn) as \x-opencode-request\.

**This PR:**
- One stable canonical session per downstream identity (\connectionId\, else auth-header hash, else shared default), evicted after \MEMORY_CONFIG.sessionTtlMs\ like the other session stores (same pattern as \sessionManager.deriveSessionId\).
- Deterministic \x-opencode-request\ derived from session + last user message (stable across retries); valid downstream ids preserved.
- Explicit session hints (header / \providerSessionId\ / body fields) keep the previous deterministic-translation path; only the empty fallback now reuses instead of minting.
- 6 more unit tests (22 total in \	ests/unit/opencode-session.test.js\).

**Verification:**
- \
px vitest run unit/opencode-session.test.js\ — 22/22 pass.
- Same behavior validated live against a local gateway: \oc/muse-spark-1.3-contributor-free\ and \oc/mimo-v2.5-free\ return HTTP 200 with canonical session headers on the wire.

Closes #3270


decolua:
Thanks @ErfanBagheri404 for the contribution! Reviewed and merged into master. 🙏

PR #1402, opened 2026-05-24
Route Codex auto-review to Codex provider

Delcado19:
## Summary
- register codex-auto-review as a Codex review-quota model
- route bare codex-auto-review requests to Codex OAuth instead of OpenAI API-key credentials
- add focused routing regression coverage for #1398

Fixes #1398

## Tests
- npx vitest run --config ./tests/vitest.config.js tests/unit/model-routing.test.js
- node --check open-sse/config/providerModels.js
- node --check open-sse/services/model.js
- git diff --check
- npm run build

PR #4135, opened 2026-09-18
fix(codex): route bare codex-auto-review to the Codex provider

### LINKED ISSUE #1398: Got error when using Auto-review codex
When I using Auto-review mode on codex, something I got this error "Automatic approval review failed: unexpected status 404 Not Found: No active credentials for provider: openai, url: http://127.0.0.1:20128/v1/responses", upon inspect the log I foundout that when log show `[ROUTING] codex-auto-review → openai/codex-auto-review` it alway failed

amirsalaar:
## Summary

Fixes #1398. Supersedes #1402.

Codex CLI's auto-review sends the bare model id `codex-auto-review`. It matches no alias and no `MODEL_PREFIX_PROVIDERS` entry, so `inferProviderFromModelName` falls through to the `"openai"` default and the request fails with `No active credentials for provider: openai` at `/v1/responses`, which is exactly what the reporter's log line `[ROUTING] codex-auto-review → openai/codex-auto-review` shows.

## Why this supersedes #1402

#1402 was written against v0.4.71 and neither of its hunks applies to current master, because `PROVIDER_MODELS` is now assembled from `open-sse/providers/registry/*` rather than being a literal in `open-sse/config/providerModels.js`. Two details there are worth flagging for reviewers, since both are easy to reintroduce:

- It inserted the model as the **first** `cx` entry. `getDefaultModel` returns `models[0].id`, so that would have quietly made `codex-auto-review` the default Codex model.
- It added the routing rule as a hardcoded early `return` inside `inferProviderFromModelName`, rather than in the `MODEL_PREFIX_PROVIDERS` table that replaced those prefix checks.

## Changes

- **`open-sse/providers/registry/codex.js`** — register `codex-auto-review` as a review-quota model. It carries an explicit `upstreamModelId` because, unlike the neighbouring `-review` entries, it is not derived from a base model: `getModelUpstreamId` strips `CODEX_REVIEW_SUFFIX` for unregistered `cx` ids, which would send `codex-auto` upstream. Placed at the end of the llm block so `getDefaultModel("cx")` is unaffected.
- **`open-sse/services/model.js`** — add the rule to the `MODEL_PREFIX_PROVIDERS` table, per the config-driven convention in `open-sse/AGENTS.md`.
- **`tests/unit/codex-auto-review-routing.test.js`** — new file. #1402 put its tests in `tests/unit/model-routing.test.js`, which already exists on master covering unrelated provider-prefix routing. Alongside the routing and quota-family assertions, this pins the upstream id against `-review` stripping and pins that the model does not become the `cx` default.

## Tests

- `npx vitest run unit/codex-auto-review-routing.test.js` — 4 passed.
- Full suite before/after: **2198 → 2202** passing, failing unchanged at **114** (all pre-existing on master, plus the documented `known-fails.txt` set).
- `node tests/__baseline__/verify-providers.mjs` / `verify-alias.mjs` / `verify-oauth-urls.mjs` — all byte-identical, which matters since this touches the provider registry.
- `npm run build` — clean. `npx eslint` on the changed files — clean, aside from the pre-existing `import/no-anonymous-default-export` warning every registry file shares.

Credit to @Delcado19 for the original diagnosis and approach in #1402.


decolua:
Thanks @amirsalaar for the contribution! Reviewed and merged into master. 🙏

PR #4137, opened 2026-09-18
fix(antigravity): scope cached thought signatures to the model family

### LINKED ISSUE #4136: Antigravity: switching Claude <-> Gemini mid-conversation fails with 400 "Corrupted thought signature" (signature cache is not model-aware)
### Summary
With the `antigravity` provider, switching models in the middle of a conversation breaks it permanently. The usual trigger is Claude quota running out, after which the client continues on a Gemini model. Running `/compact` in Claude Code on another model breaks it the same way. Every later request fails on **all** accounts with:

```
400 INVALID_ARGUMENT: Corrupted thought signature.
```

Retrying or rotating accounts cannot recover the conversation.

### Root cause
`open-sse/services/thoughtSignatureStore.js` caches thought signatures by **tool call id only** and does not record which model produced them. The history is then re-sent to a different model family. At that point `openai-to-gemini.js` (Gemini and Claude-on-Antigravity paths) and `executors/antigravity.js` look up each tool call id and attach the cached signature. As a result, Claude-produced signatures (tool ids like `toolu_vrtx_*`) end up on Gemini `functionCall` parts.

Each backend only accepts its own signatures.

### Evidence
The calls below go straight to `v1internal:generateContent` with `gemini-3.8-flash-tiered`. The history is one `functionCall` + `functionResponse`, and only the `thoughtSignature` on the functionCall varies:

| thoughtSignature on the functionCall | Result |
|---|---|
| Signature cached by 9router from a `claude-opus-4-6-thinking` response | **400 Corrupted thought signature.** (3/3 tried) |
| Signature cached by 9router from a Gemini response | 200 |
| `DEFAULT_THINKING_AG_SIGNATURE` | 200 |
| `skip_thought_signature_validator` | 200 |
| none | 400 missing thought_signature |

Once the Claude signatures expire from the in-memory cache (1h TTL) or 9router restarts, the same conversation works again on Gemini. That confirms the cached cross-model signatures are the cause.

### Expected
A signature is replayed only to the same model family that produced it. For any other family, the lookup should be treated as a cache miss, so the existing default-signature fallback applies.

Fix proposed in the linked PR.

louisphamdev:
Fixes #4136

### Summary
Antigravity serves Gemini and Claude models behind one API, and the thought-signature cache is keyed only by tool call id. When a conversation switches model family mid-session, Claude signatures are replayed on Gemini `functionCall` parts. Google then rejects every account with `400 Corrupted thought signature.`

### Changes
- `open-sse/services/thoughtSignatureStore.js`
  - Add `signatureFamily(model)`: `claude`, `gemini`, or the lowercased model id.
  - `storeGeminiThoughtSignature(..., model)` records the family, both in memory and in the persisted entry.
  - `getGeminiThoughtSignatureSync` / `getGeminiThoughtSignature` take an optional target `model`. Entries from another family are treated as a miss.
  - Untagged (older) entries and lookups without a model behave exactly as before.
- Lookups now pass the target model, at all three call sites:
  - `translator/request/openai-to-gemini.js`: Gemini path and `wrapInCloudCodeEnvelopeForClaude`
  - `executors/antigravity.js`
- `translator/response/gemini-to-openai.js`
  - Passes `state.model` when storing.
  - Keeps the request model when a response has no `modelVersion`, instead of hard-coding `"gemini"`, so the family is recorded correctly.

On a family mismatch, the existing fallback applies unchanged: the default signature goes on the first call, and siblings stay unsigned.

### Tests
- New: `tests/unit/antigravity-thought-signature-family.test.js`, 7 tests covering:
  - family classification
  - cross-family isolation in both directions
  - legacy entries
  - recording from the response stream
  - Claude→Gemini and Gemini→Claude request translation
  - same-family reuse
- The new file fails 5/7 on `master` and passes 7/7 with this change.
- Full suite (`npx vitest run`, excluding `real/`): **no new failures** compared with `master` (master 125 failing, this branch 120; the difference is the new tests).

### Verification in production
The same logic is running as a local patch on a 0.5.75 deployment. After the change, three flows return 200:
- a Gemini tool-call round trip
- a history containing a `toolu_vrtx_*` id whose cached signature came from Claude
- streaming tool calls

Before the change, the Claude-produced signatures failed as described in #4136.


decolua:
Thanks @louisphamdev for the contribution! Reviewed and merged into master. 🙏

PR #4139, opened 2026-09-18
fix(antigravity): strip Claude Code billing header from system prompts

### LINKED ISSUE #4138: Antigravity: fake 429 RESOURCE_EXHAUSTED when Claude Code's x-anthropic-billing-header reaches the system prompt (OpenAI-format input)
### Summary
With the `antigravity` provider, requests from Claude Code that arrive via **OpenAI format** (`/v1/chat/completions`) fail on every account with:

```
429 RESOURCE_EXHAUSTED: Resource has been exhausted (e.g. check quota).
```

Quota is not the problem. `fetchAvailableModels` reports 98–99% `remainingFraction` on all accounts at the same moment. Common setups that hit this are proxies/switchers that convert Claude Code's `/v1/messages` to OpenAI chat.

Each failing request also takes ~14s per account because of the transport retry on 429, which makes the failure look like real rate limiting.

### Root cause
Claude Code prepends this line to its system prompt:

```
x-anthropic-billing-header: cc_version=2.1.275.f15; cc_entrypoint=cli;
```

`translator/request/claude-to-openai.js` strips it for Claude-format input. OpenAI-format input is never passed through that translator, so the line ends up in `systemInstruction`. Antigravity answers any system text containing it with a fake 429. This is the same class of problem `ANTIGRAVITY_PROMPT_REWRITES` already handles for other competing-client branding.

### Evidence
The calls below go straight to `v1internal:generateContent` with `gemini-3.8-flash-tiered`, changing only the system instruction:

| systemInstruction text | Result |
|---|---|
| `You are a helpful assistant.` | 200 |
| `You are Claude Code, Anthropic's official CLI for Claude.` | 200 |
| `x-anthropic-billing-header` (no colon) | 200 |
| `x-anthropic-billing-header: foo` | **429** |
| `x-anthropic-billing-header: cc_version=1.0.0; cc_entrypoint=cli;` | **429** |
| full Claude Code line (any account) | **429** |

Replaying a captured failing request through 9router gave:
- as captured: 429 after ~42s (3 accounts × retries)
- with only the header line removed: 200 in ~2s

Fix proposed in the linked PR.

louisphamdev:
Fixes #4138

### Summary
Claude Code prepends `x-anthropic-billing-header: ...` to its system prompt. For Claude-format input the line is already stripped. OpenAI-format clients pass it through to Antigravity, which answers with a fake `429 RESOURCE_EXHAUSTED` on every account even when quota is available.

### Changes
- `open-sse/config/appConstants.js`: add `/^x-anthropic-billing-header:[^\n]*(?:\r?\n)*/gim` to `ANTIGRAVITY_PROMPT_REWRITES`, next to the existing competing-client branding rules.
  - It uses the same pattern `claude-to-openai.js` already uses, with `g`/`m` flags because the rewrite loop uses `replaceAll` and the line may not be first.
  - It only applies to Antigravity system instructions. Other providers, including the Claude provider that relies on the header, are unaffected.

### Tests
- New: `tests/unit/antigravity-billing-header-rewrite.test.js`. It runs OpenAI → Antigravity translation plus `AntigravityExecutor.transformRequest` and checks three cases:
  - header as the first line
  - header on a later line
  - prompt without the header, left untouched
- The new file fails 2/3 on `master` and passes 3/3 with this change.
- Full suite (`npx vitest run`, excluding `real/`): **no new failures** compared with `master`.

### Verification in production
The same stripping runs in front of a 0.5.75 deployment. Replaying a previously failing Claude Code request (OpenAI format, 122 tools) now returns 200 in about 1–2s instead of 429 after ~42s.


decolua:
Thanks @louisphamdev for the contribution! Reviewed and merged into master. 🙏

PR #4142, opened 2026-09-18
fix(opencode): rename free-tier tool quartet so Claude Code CLI stops getting 403

yxxrn:
## Problem

The free-tier OpenCode model (`oc/muse-spark-1.3-contributor-free`) **always fails** when used from Claude Code CLI.

Claude Code sends its built-in tools capitalised (`Bash`, `Glob`, `Grep`, `Read`). The OpenCode Zen free-tier gate fingerprints its official client through the file-search tool names in **lowercase**, so it never matches and the request is rejected.

Measured directly against the upstream, one variable per row:

| Request body | Result |
|---|---|
| capitalised quartet only | `403 FreeTierError` |
| capitalised quartet **+** lowercase | `500 server_error` |
| lowercase quartet only | ✅ `200` |
| 43 Claude Code tools, quartet **renamed** | ✅ `200` |

The second row is the important one: the intuitive approach — *"if the quartet is missing, append it"* — produces **two tools with the same name** (`Bash` and `bash`), which the upstream rejects with 500. The capitalised variant has to be **renamed**, not supplemented. That is also why merely injecting the quartet is not enough.

## Fix

**`open-sse/utils/opencodeFingerprint.js`** (new)
- `applyFingerprintTools` renames quartet case-variants to the canonical lowercase form, drops duplicate names, and appends only the quartet members that are genuinely absent (a request with no tools is rejected too, so the quartet is still injected).
- An explicit `tool_choice` pointing at a renamed tool is retargeted.
- `restoreToolNames` puts the caller's own spelling back on the response so the client still recognises its tool calls. Shapes handled: Claude `content_block_start` (`tool_use`), Chat Completions `choices[].{delta,message}.tool_calls[]`, and Responses `output[]` (`function_call`).
- The rename map travels through a `WeakMap` keyed by the body object — `transformRequest()` receives a spread copy of `credentials`, so storing the map there would never be visible from `chatCore`.

**`open-sse/executors/opencode.js`**
- Runs `applyFingerprintTools` in `transformRequest`.

**`open-sse/handlers/chatCore.js`**
- Picks the rename map up after `executor.execute()` and merges it into `toolNameMap`, which the response handlers already thread through.

**`open-sse/translator/index.js`** + **`chatCore/nonStreamingHandler.js`** + **`chatCore/sseToJsonHandler.js`**
- Restore tool names on every response path: streaming, non-streaming, and the SSE→JSON route.

## Verification

Tested live on 9router 0.5.75, model `oc/muse-spark-1.3-contributor-free`:

| Case | Before | After |
|---|---|---|
| 43 capitalised tools, stream | ❌ 403 | ✅ 200 |
| 43 capitalised tools, non-stream | ❌ 403 | ✅ 200 |
| Multi-turn + 43 capitalised tools | ❌ 403 | ✅ 200 |
| `tool_choice` named `Bash` | ❌ 403 | ✅ 200 |
| Mixed case `BaSh/gLoB/GREP/rEaD` | ❌ 403 | ✅ 200 |
| Pure duplicate `Bash` + `bash` | ❌ 403 | ✅ 200 |
| No tools | ❌ 403 | ✅ 200 |
| Lowercase quartet (regression) | ✅ 200 | ✅ 200 |
| OpenAI `/chat/completions` (regression) | ❌ 403 | ✅ 200 |

The response still carries `Bash` — no lowercase name leaks back to the client.

## Tests

17 new cases in `tests/unit/opencode-fingerprint.test.js`, covering: case-variant duplicates, pure duplicates, injection when the quartet is absent, both tool shapes (flat and nested `function`), `tool_choice` retargeting, restoration in all three response shapes, malformed input (`null`/`undefined`/string/empty object must not throw), and the no-op path with an empty map.

`npx vitest run unit/opencode-fingerprint.test.js` → 17/17 pass. `npx eslint` on all changed files → clean. The existing opencode/translator suites still pass.

## Notes

The patch is purely additive and does not change any other code path. If the upstream gate changes again, only `OPENCODE_FINGERPRINT_TOOLS` and `fingerprintToolKey()` in the single new file need adjusting.


decolua:
Thanks @yxxrn for the contribution! Reviewed and merged into master with the stream aggregation and tool cloaking fixes. 🙏

PR #4150, opened 2026-09-18
fix(ollama): show free-plan monthly usage and reset time in quota tracker

x-adam:
## Problem

Ollama Cloud free accounts now report a single `limits.monthly` window from `GET /api/usage` instead of `session`/`weekly`. `getOllamaUsage` only knew the two paid-plan windows, so every free account showed **"Ollama Cloud connected. No usage limits reported."** in the quota tracker even though the dashboard at ollama.com shows e.g. "2.1% used, resets in 2 weeks".

Sample free-plan payload (real account, trimmed):

```json
{
  "activity": { "cost": "0.00000", "period": { "type": "last_4_weeks", "...": "..." }, "models": [] },
  "limits": {
    "monthly": {
      "usage": 0.021,
      "models": [ { "name": "gpt-oss:120b", "request_count": 6 } ]
    }
  }
}
```

`POST /api/me` → `"Plan": "free"`.

## Fix

- Map every known window (`session` → "Session (5h)", `weekly` → "Weekly (7d)", `monthly` → "Monthly") to a quota row via a small lookup table instead of two hard-coded checks. Paid plans are unchanged; the "No usage limits reported" message is kept for payloads with no known window.
- Tests: added the free-plan payload case (`Monthly` → used 2 / remaining 98, plan "Free") and an empty-`limits` case.
- Dashboard `parseQuotaData` comment updated; no logic change there.

## Tests

```
cd tests && npx vitest run unit/ollama-usage.test.js unit/usage-dispatch.test.js
Test Files  2 passed (2)
Tests       9 passed (9)
```


decolua:
Thanks @x-adam for the contribution! Reviewed and merged into master. 🙏

PR #4155, opened 2026-09-18
fix(opencode): cloak Responses requests that already have tools

AaronWong1999:
## Summary

Fix OpenCode free-tier `403 FreeTierError` for Responses-based models when the downstream agent already provides tools.

This affects cases such as:

```text
oc/muse-spark-1.3-contributor-free(xhigh)

Fix

Run cloakOpencodeTools(body, true) for all OpenCode Responses requests, not only requests with an empty tools array.

Existing client tools are preserved, and missing OpenCode fingerprint tools are added without duplication.

Test

Added a regression test covering:

muse-spark-1.3-contributor-free(xhigh)
+ existing client tools
+ reasoning_effort=xhigh

Verified with:

25/25 tests passed
node --check passed


decolua:
Thanks @AaronWong1999 for the contribution! Reviewed and merged into master. 🙏

PR #4157, opened 2026-09-18
feat(bedrock): add AWS Bedrock provider with AWS SSO credential support

amirsalaar:
Adds Amazon Bedrock as a provider, with real AWS SSO support rather than only pasted keys.
image

image

## Why

Bedrock is currently unsupported. Anyone on AWS today has to paste long-lived IAM keys, and most AWS shops have moved to IAM Identity Center where the only credentials available are short-lived and rotate. This adds both paths.

## Credential modes

```
static  — apiKey = AWS secret access key
          providerSpecificData { accessKeyId, sessionToken? }

profile — providerSpecificData { profile, region }   (no apiKey)
          then: aws sso login --profile <profile>
```

Profile mode resolves through `@aws-sdk/credential-providers` `fromIni`, which is what makes SSO work end to end: it follows `sso_session`, reads the token from `~/.aws/sso/cache`, calls `GetRoleCredentials`, and handles `source_profile` role chaining. Re-resolving on expiry gives refresh for free, so an expired SSO session recovers without touching the connection.

## Design notes for review

- **`transport.format` is `"claude"`.** Bedrock's Anthropic models accept the Anthropic Messages body verbatim (minus `model`, which lives in the URL, plus `anthropic_version`). So the existing `*-to-claude` translators are reused and **no new translator was added**.
- **SigV4 is hand-rolled** in `utils/awsSigv4.js` on `node:crypto`, adding no runtime dependency for signing. Every request here is a single POST with a JSON body and no query string, so the canonicalisation cases that make SigV4 fragile do not arise; it throws on a query string rather than signing one incorrectly. It is cross-verified **byte-for-byte against the `aws4` reference implementation**, not against documentation — re-derive with `node tasks/test-commands/verify-sigv4-against-aws4.mjs`.
- **Path segments are escaped twice**, which a `:0` model-version suffix requires: `…v1%3A0` on the wire, `…v1%253A0` in the canonical request. Getting this wrong is the usual cause of an opaque `SignatureDoesNotMatch` on Bedrock.
- **`executors/kiro.js` held the only AWS EventStream decoder in the tree** and Bedrock uses byte-identical framing, so `crc32` and `parseEventFrame` were extracted to `utils/awsEventStream.js` instead of being copied. `kiro.js` loses 102 lines and is otherwise untouched — its frame loop, error taxonomy and recovery paths are unchanged. **Kiro's suite reports 133 passing / 3 pre-existing failures before and after, identically**, which is the evidence that extraction was safe.
- **`passthroughModels: true`**, because which models an account can call varies by region and by enabled inference profiles.
- **Model ids come from a live `ListInferenceProfiles` call.** Worth knowing: every Anthropic model on Bedrock is `INFERENCE_PROFILE`-only, so the bare `anthropic.*` ids cannot be invoked on demand and the `us.` / `global.` profile ids are required.
- **The region is validated before use** because it is interpolated into the request hostname. Unvalidated, a value like `evil.com/x` resolves the host to `bedrock-runtime.evil.com` and ships the signed body and session token to an arbitrary origin. This is the same hardening class as the existing Vertex/Video "reject ids that escape the URL path" fix. The fetch also sets `redirect: "error"`.
- **`@aws-sdk/credential-providers` is in `optionalDependencies`** (npm still installs those by default) and imported lazily, so static-key users never load it and a failed optional install degrades to an actionable error. It is added to `serverExternalPackages` because it reads `~/.aws` from disk at runtime. **The `import()` specifier must stay a literal** — with a variable, Next's output tracing cannot see it, the package is omitted from `.next/standalone`, and profile mode fails with `MODULE_NOT_FOUND` only in the packaged CLI. There is a comment saying so at the call site.

## Verification

- **52 new unit tests** across `tests/unit/aws-sigv4.test.js`, `aws-credentials.test.js`, `bedrock-provider.test.js`.
- Full suite **2343 → 2398 passing, failing unchanged at 119** (the documented pre-existing set).
- `tests/__baseline__/verify-providers.mjs` and `verify-alias.mjs` regenerated for the new provider (81 → 82); diffs are purely additive Bedrock entries. `verify-oauth-urls.mjs` byte-identical. `verify-alias.mjs` gained `br`/`bedrock`/`aws-bedrock` probe tokens so the new alias is actually guarded.
- `npm run build` clean; eslint clean apart from the `import/no-anonymous-default-export` warning every registry file has.
- **Live AWS**, via a real IAM Identity Center profile: `ListFoundationModels` 200; streaming invoke 200 returning all six Claude SSE events with correct output in both `us-east-1` and `us-west-2`; non-streaming invoke 200 returning well-formed Anthropic Messages JSON. Reproducible with `node tasks/test-commands/verify-bedrock-live.mjs <profile>`.

12 of the 52 tests are regressions for defects an adversarial review pass found after the first implementation already passed 40 tests: a credential cache whose TTL never ran, a cache key that let two connections share a secret, concurrent callers inheriting each other's region, and a double `controller.close()` that left the upstream reader locked and never cancelled on every throttle. All are fixed and pinned.

## Known limitations

- The output stream has no backpressure, and buffer concatenation is quadratic under heavy fragmentation. Both match the existing pattern in the other executors here including `kiro.js`, so I left them alone rather than changing streaming behaviour repo-wide in this PR. Happy to follow up if you want it addressed engine-wide.
- Non-Anthropic Bedrock models (Llama, Titan, Mistral) and the Converse API are not handled; they need different body shapes.
- No bespoke dashboard UI; configuration goes through the existing `providerSpecificData` field, as `aws-polly` does.

Separately: `aws-polly` is currently a stub — a registry entry with `authHeader: "aws-sigv4"` but no signer and no handler for its `aws-polly` format. Now that a verified SigV4 signer exists it could be finished cheaply. Out of scope here; happy to open a follow-up.


amirsalaar:
---

## Update: second review round

A second adversarial pass verified all seven fixes above as correct and then found **two defects the fixes themselves introduced**. Both were reproduced with a probe before being fixed, and both are now pinned by regression tests.

- **`computeExpiresAt` returned `now`** for a credential with no declared expiry, so the cache entry failed its own `expiresAt > currentTime` check immediately and every request re-read and re-parsed `~/.aws` from disk on the hot path. `fromIni` returns no expiration for a plain `aws_access_key_id` profile in `~/.aws/credentials`, which is a common setup. Probed at 5 sequential requests → 5 disk resolutions; now 1, with a 60s floor that still refuses to pin a revoked profile for the process lifetime.
- **The new truncation check fired on client disconnect**, logging a false `ended before message_stop` and enqueueing onto an already-cancelled controller, which throws past the teardown. A client hanging up mid-answer legitimately never reaches `message_stop`. Now tracked as `downstreamCancelled` and skipped.

Worth calling out: **my own regression test had encoded the first of these as intended behaviour**, asserting two resolutions for two requests. It has been corrected to assert the floor and the lapse.

Three smaller items also taken from that pass: `resolveRegion` now trims, so a region pasted with a trailing space no longer hard-fails; the invalid-region error names its source (`providerSpecificData.region` vs `the AWS_REGION environment variable`) so an env problem does not read as a dashboard problem; and the provably-dead `searchParams.toString()` in the canonical request is now a literal `""`.

Test count 52 → 54, full suite 2398 → 2400 passing, failures still unchanged at 119. Live AWS re-verified green after every change.

### One architectural gap, flagged rather than fixed

In-band Bedrock exceptions — `ThrottlingException` above all — arrive inside an HTTP **200**, because `execute()` returns the `Response` before the body is read. Traced, the gate is a single early return:

- `src/sse/handlers/chat.js:310` — `if (result.success) return result.response;`

A 200 makes `result.success` true, so everything below that line is unreachable: `chat.js:327`'s `markAccountUnavailable(...)`, then `src/sse/services/auth.js:262`'s `checkFallbackError(status, errorText, backoffLevel)`, then the `ERROR_RULES` match in `open-sse/config/errorConfig.js`. Failover is decided from the **HTTP status**, so a throttled Bedrock account is never marked rate-limited and multi-account fallback never fires for it.

Worth splitting in two: a **mid-stream** exception genuinely cannot be surfaced as a non-200, since the status is already on the wire. A **first-frame** exception could be, by peeking at the first frame before returning the `Response` — which is the common `ThrottlingException` shape. That is the tractable half if you want it.

I did not fix this here because `executors/kiro.js:1134` does the same thing — `status: response.status`, with the in-band error delivered as SSE inside it. Fixing it for Bedrock alone would make Bedrock inconsistent with the one existing provider that shares this framing, and fixing it properly means buffering the first frame before returning the `Response`, which is an engine-wide change to streaming behaviour. Happy to do that as a separate PR covering both providers if you think it is worth it.

amirsalaar:
Converting to draft while I finish a review round — please hold off merging.

An adversarial review pass has surfaced a further set of findings (C2–C13) that I have not yet seen in full; the two disclosed so far are both HIGH. I would rather this sat in draft than be merged on a merge-ready label I put on it prematurely.

Two already triaged:

- **My `refreshCredentials` override was dead code.** `handlers/chatCore.js:420` gates the refresh-and-retry path on `newCredentials?.accessToken || newCredentials?.copilotToken`, and SigV4 has no bearer token to put there, so the `{ expiresAt }` it returned never took effect — while its doc comment claimed it was how an expired SSO session recovered. Removed in favour of the base `null`, with a comment recording why so it does not get re-added. Refresh does still work a layer down: `execute()` calls `resolveAwsCredentials` on every request and that re-resolves past expiry. The real cost is that a 401/403 is not transparently retried inside the same request. Flagging the gate shape as feedback: it assumes bearer-token auth and has no shape for signature-based providers.

- **No connect timeout on this executor** (`FETCH_CONNECT_TIMEOUT_MS`), and no configured 429/502 retry, because overriding `execute()` skips the `connectCtrl`/`tryRetry` machinery in `executors/base.js:135-173`. I checked before treating it as mine: **18 of the 19 executors that override `execute()` have no connect timeout**, `qoder.js` being the sole exception, and that includes `vertex.js`, which this provider was modelled on. So it is a pre-existing engine-wide gap rather than something this PR introduces, and worth its own change across all of them rather than a one-provider divergence here.

That second point also blocks the first-frame failover peek discussed above: peeking converts an open-stream stall into a pre-header stall, so it must land together with a first-frame timeout or not at all.

I will report the rest of the findings, fix what is real, and mark this ready again once the list is closed.

amirsalaar:
## Round 3 — findings list now closed, and back out of draft

The adversarial pass confirmed its inventory is closed at 13 findings. All triaged; verified each against the code rather than taking the report on trust, which mattered — two claims did not survive.

### Fixed

| Finding | Defect | Why it mattered |
|---|---|---|
| C6 | `anthropic_version` sat **before** the body spread, so a claude-format client sending its own `"2023-06-01"` overrode the Bedrock-required value | every such request earned a `ValidationException` |
| C4 | profile resolution was unbounded, and every caller awaits the same in-flight promise | one hung `GetRoleCredentials` blocked **all** traffic on that profile indefinitely |
| C7 | the profile name reached the AWS SDK unvalidated | the SDK follows `source_profile` chains and runs a `credential_process` subprocess for whatever it names; now allowlisted like the region |
| C9 | a Nova/Llama/Titan id passed the passthrough catalogue, then died mid-stream on a chunk with no `type` | the upstream call was already billed; `buildUrl` now rejects upfront and names the limitation |
| C10 | `DataView` over the frame buffer had no `byteLength`, so it ran to the end of undici's pooled `ArrayBuffer` | benign under the current `>= 12` guard, but any added read would parse neighbouring pooled bytes |
| C3 | `refreshCredentials` was **dead code** — `chatCore.js:420` gates on `accessToken`/`copilotToken` and SigV4 has neither — while its docstring claimed automatic SSO recovery | removed in favour of the base `null`; credentials are still re-resolved every `execute()` |
| C8 | moot once C3's override was removed (it was the source of the epoch-ms `expiresAt`) | — |
| C12a | the extraction left an orphaned `Parse AWS EventStream frame` comment in `kiro.js` | — |

### The one that actually blocked the feature

**Profile/SSO mode could not be configured at all.** The registry notice says to leave the API key empty; `src/app/api/providers/route.js:119` and `validate/route.js:92` both returned 400 for exactly that. Anyone following this provider's own setup instructions hit a wall.

Rather than hardcode a provider name in a shared route, the registry now declares `apiKeyOptionalWith: "profile"` — naming the `providerSpecificData` field that stands in for a key — and both routes honour it generically. Reusable for any future provider whose credential lives outside the key field.

### Not fixed, with reasons

- **C2, no connect timeout or configured retry.** Overriding `execute()` skips `base.js:135-173`. Checked before attributing it: **18 of the 19 executors that override `execute()` have no connect timeout**, `qoder.js` being the only exception, `vertex.js` included. Pre-existing engine-wide, worth one change across all of them.
- **C1, in-band exceptions inside HTTP 200.** As discussed above; `kiro.js` shares the shape.
- **C11, signed headers logged unmasked.** Pre-existing sink with masking deliberately disabled; `x-amz-security-token` is a new credential class landing in it, so worth its own change.
- **Backpressure and quadratic concat**, same pattern as the other executors.

### Two claims that did not survive verification

- **The model list was called into question** on the grounds that mixed suffix shapes implied stale provenance. Re-ran `ListInferenceProfiles` live: **all 11 ids exist and are ACTIVE**. AWS genuinely serves both `us.anthropic.claude-opus-5` and `us.anthropic.claude-sonnet-4-5-20250929-v1:0`. No change.
- **`registry/index.js` was called a hand-edited generated file.** Both `scripts/migrate-registry.mjs` and `scripts/injectDisplayToRegistry.mjs` explicitly exclude it (`f !== "index.js"`), so **no generator for it is committed**. Appending was the only option, and the non-monotonic `p124` matches the file's own convention — `p123` is already taken at line 69.

### State

60 unit tests (21 regressions, one per defect). Full suite **2406 passing, failures unchanged at 119**. Kiro **133 passing / 3 pre-existing — identical to before the extraction**. All three baselines byte-identical. Build clean, SDK still traced. Live AWS re-verified green on streaming and non-streaming after every change.

Marking ready for review again.

amirsalaar:
## Round 4 — one more real defect, found only by running it through the gateway

Everything so far was verified against the executor directly. Wiring the provider into a running gateway and sending real client traffic surfaced a defect none of the 60 unit tests or any live-AWS executor test could see.

**`forceStream: true` in the registry broke non-streaming.** With it set, a non-streaming client is routed to `handleForcedSSEToJson` (`handlers/chatCore.js:482`) instead of the normal non-streaming path, and that aggregation returned:

```json
{"choices":[{"message":{"role":"assistant","content":""},"finish_reason":"stop"}],
 "usage":{"input_tokens":28,"output_tokens":6}}
```

Empty `content` despite 6 output tokens, with Claude-shaped `usage` leaking into an OpenAI envelope. Bedrock has a real non-streaming endpoint (`/invoke`) that returns a complete Anthropic Messages response — verified directly — so `forceStream` was simply wrong for this provider. Removed.

After the fix, through the running gateway on all three client shapes:

| Client | Result |
|---|---|
| OpenAI `stream:false` | `content: "NONSTREAM OK"`, `usage: {prompt_tokens, completion_tokens, total_tokens}` |
| OpenAI `stream:true` | `delta.content: "STREAM OK"`, `finish_reason: "stop"` |
| Claude `/v1/messages` | correct Claude SSE, `message_start` onward |

The request log confirms the routing changed from `· STREAM ·` to `· JSON ·` for a non-streaming client.

**Also verified end to end: the C5 fix works through the real API.** `POST /api/providers` with `providerSpecificData: {profile, region}` and **no `apiKey`** returns **201**. That is the documented SSO setup path, which returned 400 before.

### Regression position, stated carefully

119 failing / 2419 passing. Two intermediate runs showed 106 failing, which was an artifact of my own inconsistent deletion of `tests/translator/__snapshots__/golden-url-header.test.js.snap` between runs — not an improvement, and I am not claiming one.

`unit/force-stream-config.test.js` fails 2, and I checked whether removing `forceStream` caused it: the file never mentions bedrock (it tests `provider: "openai"`), is byte-identical to `upstream/master`, and fails on a pre-existing vitest mock gap (`No "formatHeadroomSizeLog" export is defined on the mock`). Not mine.

Worth flagging for the repo: `tests/__baseline__/verify-no-regression.mjs` cannot be used on a normal checkout. It derives test keys with `f.name.split("/app/")[1]`, a container path, so every key is `undefined`, nothing matches `known-fails.txt`, and it reports all 119 failures as regressions. Same class of hardcoded-path problem CLAUDE.md already documents for `tests/package.json`.

Provider baseline regenerated; the only diff is the removed `forceStream`. Alias and OAuth baselines byte-identical.

amirsalaar:
## Dashboard support and provider logo

Two follow-ups after testing the provider through the actual UI rather than curl.

**The setup notice was unfollowable.** `AddApiKeyModal.js`'s `buildProviderSpecificData()` is a hardcoded per-provider switch (ollama-local, azure, cloudflare-ai, anything declaring `regions`), so Bedrock had **no input field at all** and the form would have posted `providerSpecificData: undefined`. Added an **AWS Bedrock Credentials** section with Profile, Region, Access Key ID and Session Token.

Worth calling out for reviewers, because it is the interesting part: the "is a credential present" rule was encoded in **two** places — the Save button's `disabled` expression and an independent early `return` in `handleSubmit`. I fixed only the first, so Save looked clickable and silently did nothing. Both now read one `apiKeySatisfied()` helper, keyed off the same registry-declared `apiKeyOptionalWith` field as the API-route exemption, so they cannot drift apart again.

`aws-polly` has the same dead end in its notice, if anyone wants to pick that up alongside finishing the stub.

**Provider logo.** `public/providers/bedrock.png` — the AWS wordmark, 128×128 RGBA, matching the `/providers/<id>.png` convention the other 143 logos use. Rendered from devicon's MIT-licensed SVG. It carries a light background rather than transparency because the card tints its container with `provider.color + "15"`, which on the dark theme would leave the navy wordmark unreadable; `anthropic.png` solves it the same way. `icon`/`textIcon` stay as fallbacks, consistent with every other PNG-having provider.

The registry notice now names the actual form fields instead of describing raw JSON.

### Verified

- 60 unit tests pass; provider, alias and OAuth baselines byte-identical.
- New modal fields present in the shipped client chunks; `GET /providers/bedrock.png` returns **200 image/png**, byte-identical to the committed asset.
- Gateway re-verified after restart: `stream=false` and `stream=true` both return correct content.

Total footprint outside the new Bedrock files stays small: **83 insertions across 5 files**, of which the three shared files account for 14 lines.

amirsalaar:
## Adds `bedrock-xai` — xAI Grok on Bedrock

Second Bedrock entry, `bedrock-xai` (alias `brx`), covering `us.xai.grok-4.6` and `global.xai.grok-4.6`.

### Probed before designing

I did not assume the wire format. Against live AWS, all three candidate schemas returned 200, and the answer was decisive:

| Probe | Result |
|---|---|
| `/invoke` with an Anthropic Messages body | 200, but **`anthropic_version` ignored** — returned `choices[0].message` |
| `/invoke` with an OpenAI body | 200, identical Chat Completions response |
| `/converse` | 200, Converse schema |

So **Grok on Bedrock speaks OpenAI Chat Completions natively**, and its streaming frames carry `chat.completion.chunk` payloads. That means no response translation at all — the existing OpenAI path applies end to end.

### Design

The entry omits `format`, so it defaults to `"openai"` (`providers/schema.js`), and **one `BedrockExecutor` serves both entries** off the registry transport format — the pattern `vertex` / `vertex-partner` already establishes.

Everything format-dependent keys off that single field:

- `anthropic_version` is injected **only** on the claude wire (Grok ignores it, and sending it is noise)
- SSE is emitted as named Claude events, or as bare OpenAI `data:` chunks closed by `[DONE]` — an `event:` line on the OpenAI wire is junk the parsers must skip
- completion is detected from `message_stop`, or from a non-null `finish_reason` on a choice
- each entry **rejects the other family's ids upfront**, because passing one through would fail mid-stream after the call was billed

Credentials, SigV4, region validation, EventStream decoding and the credential cache are reused unchanged. Net new code is one registry file plus branching in the executor.

### Verified

- **67 unit tests** (up from 60), including both directions of the family guard, the absence of `anthropic_version` on the OpenAI wire, bare-`data:` emission with `[DONE]`, `finish_reason` truncation detection, and rejection of a chunk with no `choices`.
- Full suite **2416 passing, failures unchanged at 119**. Provider baseline 82 → 83, alias probe gained `brx`/`bedrock-xai`/`bedrock-grok`; both regenerated and byte-identical after.
- **Live through the running gateway**, assembled from the SSE rather than eyeballed:
  - `bedrock-xai/us.xai.grok-4.6` non-streaming → `"GATEWAY GROK"`
  - streaming → `"GATEWAY GROK STREAM"`, `finish_reason: stop`, `[DONE]` present, no stray `event:` lines, no error payloads
  - `brx/` alias → `"ALIAS OK"`
  - `bedrock/` Anthropic entry → `"CLAUDE OK"`, unaffected

### Operational note

Grok 4.6 is a reasoning model that spends output budget before it answers. At `max_tokens: 16` and even `200` it returns `finish_reason: "length"` with **empty content** — 476 reasoning tokens on a trivial prompt. Anyone testing with a small budget will think it is broken, so the registry notice says to allow a few thousand tokens.

`public/providers/bedrock-xai.png` uses the xAI mark so the two Bedrock entries are distinguishable at a glance, while the card keeps AWS orange to signal it is still Bedrock.

Converse remains unimplemented, and is still the better route if the other Bedrock families (Llama, Qwen, DeepSeek, Nova, Mistral — 109 models in the account I tested against) are ever wanted; that is a separate change with its own translator pair.

amirsalaar:
## Fix: `bedrock-xai` was unconfigurable from the dashboard

The AWS credentials form was gated on `provider === "bedrock"`, so the new `bedrock-xai` entry rendered **no Profile or Region inputs at all**, `buildProviderSpecificData()` returned `undefined`, and Save stayed blocked. SSO mode was impossible from the UI, even though the API accepted it.

**This is the third time this class of bug shipped in this PR** — the API-key exemption, then the Save guard, now the form — every time because a capability was written as a provider id. So the fix is structural rather than another conditional:

- both entries declare `credentialForm: "aws"` in the registry
- the modal gates on that flag, not on an id
- a test asserts **every** provider whose id starts with `bedrock` declares both `credentialForm` and `apiKeyOptionalWith`, so the next AWS entry cannot regress it silently

A future Bedrock entry now gets the credential form by declaring what it needs, instead of by someone remembering to edit a conditional in the dashboard.

### Verified after the change

- 68 unit tests pass; all three baselines byte-identical (83 providers, 123 alias tokens).
- Both connections survive a restart, and through the running gateway: `bedrock-xai/us.xai.grok-4.6` → `"FIXED"`, `bedrock/…claude-sonnet-4-5` → `"FIXED"`.
- Both logos serve 200: `bedrock.png` (3633b), `bedrock-xai.png` (4232b).

### Separately, on Grok's token budget

I checked whether the empty-content behaviour needed code. It does not: **omitting `max_tokens` works correctly** (`finish_reason: stop`, 311 reasoning tokens, content present). The failure only occurs with an explicitly small budget, where Grok returns `finish_reason: "length"` and `content: null` after spending it all reasoning — which is standard OpenAI semantics, correctly reported. Rewriting a caller's `max_tokens` would be worse than documenting it, so the registry notice carries the warning and no code intervenes.

PR #4158, opened 2026-09-18
fix(huggingface): complete the Inference Providers router migration

kimono381:
## Problem

HuggingFace was broken in three independent ways, not one.

### 1. Dead host

`imageConfig.baseUrl` pointed at `https://api-inference.huggingface.co` — the retired Inference API. The domain no longer resolves (DNS ENOTFOUND), so every image request failed inside `fetch()` before reaching HuggingFace. This was the only reference to the old host in the repository.

### 2. Missing `sttConfig`

Speech-to-text did not work at all. `open-sse/handlers/sttCore.js:172` returned before any network call:

```
Provider 'huggingface' does not support STT
```

`serviceKinds` still advertised `"stt"` and Whisper models were listed, so the dashboard offered a feature that did not exist.

### 3. Dead model

`openai/whisper-small` has an empty `inferenceProviderMapping` — no live provider serves it. It was listed as working.

## What replaced what

HuggingFace moved to the **Inference Providers router** (`https://router.huggingface.co`). It is addressed by provider-resolved id, not Hub id:

```
<baseUrl>/<provider>/<providerModelId>
```

So `black-forest-labs/FLUX.1-schnell` becomes `https://router.huggingface.co/fal-ai/fal-ai/flux/schnell`. The doubled `fal-ai/fal-ai/` is correct, not a typo — the first segment is the router's provider name, the second is fal.ai's own namespace prefix. Verified against the Hub API:

```
GET https://huggingface.co/api/models/black-forest-labs/FLUX.1-schnell?expand[]=inferenceProviderMapping
→ fal-ai: { providerId: "fal-ai/flux/schnell", status: "live" }
```

Images and speech go through **different** router providers: audio is served by `hf-inference`, which keeps the Hub id verbatim (`/hf-inference/models/<hubId>`). Hence `sttConfig.baseUrl` carries the `/hf-inference/models` suffix while `imageConfig.baseUrl` does not. Confirmed against `sttCore.js:127`, which builds `baseUrl.replace(/\/+$/, "") + "/" + model`.

## Image-to-image models could never work

Nine registered models are `pipeline_tag: image-to-image`, not text-to-image. The old `buildBody` always sent the text-to-image shape:

```js
buildBody: (_model, body) => ({ inputs: body.prompt }),
```

For image-to-image the [task contract](https://huggingface.co/docs/inference-providers/tasks/image-to-image) puts the source image in `inputs` and the prompt under `parameters.prompt`. These models would have failed even with a live host — they received text where an image was expected.

`modelMap` values are now either a bare path or `{path, task}`, and the adapter branches on the task. It accepts the source image as a data URL, bare base64, remote URL or array, and strips the data-URL prefix the router does not want. A missing source image raises a descriptive error; `handleImageGenerationCore.js:100` already wraps `buildBody` in try/catch, so callers get a `400` with a readable message rather than a confusing upstream failure.

## Smaller defects found along the way

- **Prototype pollution in lookup.** `MODEL_MAP[model]` returned inherited values, so a request naming a model `toString` or `__proto__` built a URL out of `function Object() { [native code] }`. Now resolved with `Object.hasOwn`.
- **Dead parameter.** Whisper declared `params: ["language"]`, but `transcribeHuggingFace` posts raw audio and never reads `formData`, and the router's ASR payload has no `language` field. The dashboard showed a selector whose value was silently discarded.
- **Edit models unreachable from the UI.** `GenericExampleCard` renders the source-image field only when a model declares `capabilities: ["edit"]` (`GenericExampleCard.js:47`). Added to the nine edit models, using the same format `cloudflare-ai` already uses upstream, with a sample image pre-filled so Run works immediately.

## Models

The table grows from 2 to 23 image models — FLUX.2 and FLUX.2-klein families, Qwen-Image-Edit-2511, Ideogram 4, Z-Image-Turbo, Krea-2-Turbo, HiDream-I1-Fast, Playground v2.5 — and replaces the dead `whisper-small` with `whisper-large-v3-turbo`.

Only providers the router actually forwards to are mapped. `replicate`, `wavespeed`, `deepinfra` and `ideogram` appear in `inferenceProviderMapping` but reject router traffic with `400 Model not supported by provider <name>` or `404`, so they are excluded and the reason is documented in the registry comment.

**Live probe:** all 23 built URLs return `402 credits-depleted` (path accepted, billing reached), while a control path returns `400` and the legacy host returns `000`.

## Tests

| File | Tests | Coverage |
|---|---|---|
| `tests/unit/huggingface-router-migration.test.js` | 39 | registry, `sttConfig`, URL building, payload shapes, prototype guard, `modelMap` completeness |
| `tests/unit/huggingface-image-end-to-end.test.js` | 9 | full path through the real `handleImageGenerationCore` against a mocked router |

## Verification

- **0 regressions.** Full suite compared one-to-one against a pristine upstream checkout: 2583 tests, no pass-to-fail transitions. 48 new tests pass.
- All three baseline gates green (`verify-providers`, `verify-alias`, `verify-oauth-urls`).
- `eslint` clean on every changed file.
- `registry/index.js` untouched — no provider was added or removed.

## One honest limitation

Live image generation could not be verified. The available access token has no PRO plan and no remaining credits, and the router returns `402` **before** validating the payload — confirmed experimentally: empty body, garbage body and a correct body all produce an identical response. So a live probe can prove a path exists but never that the body is right.

Payload correctness is therefore established two other ways: against the published task contract, and against the official `@huggingface/inference` client v4.13.30, plus the end-to-end suite against a mocked router. Confirming real generation needs a token with balance.

## Note on scope

The first commit in this branch was pushed earlier and is superseded by the second, which completes the migration. Reviewers may prefer to read the two together as one logical change.


decolua:
Thanks @kimono381 for the contribution! Reviewed and merged into master. 🙏 Two small tweaks were applied during review: removed a source-attribution comment line, and added a model-id guard (rejects `..`, `//`, `?`, `#`) on custom-endpoint connections, mirroring the existing check in sttCore — plus a test for it. All 49 tests and the three baseline gates pass.

Issue #4152, opened 2026-09-18
Opencode free


luongtri540-blip:
Still got "OpenCode's free tier can only be used from within OpenCode" when use through Claude Code

mohammad653:
on hermes too
i was did the path by my self and it was worked for hermes but claude code didnt fixed
now both of them didnt work

Issue #4127, opened 2026-09-17
OpenCode Free Tier returns 403 when used through 9Router


TEGAR-SRC:
OpenCode free-tier models are currently returning HTTP 403 when accessed through 9Router.

The error returned is:

```text
HTTP 403: [403]: {
  "type": "error",
  "error": {
    "type": "FreeTierError",
    "message": "Error from provider (Console): OpenCode's free tier can only be used from within OpenCode"
  }
}
```

It looks like OpenCode has changed or restricted its free-tier access so that free models can only be used directly from the OpenCode client.

Because 9Router acts as a proxy/router, requests to OpenCode free-tier models are now rejected.

## Environment

* 9Router version: `v0.5.75`
* Deployment: Docker
* OS: Ubuntu
* API endpoint: `/v1`
* Provider: OpenCode
* Authentication: OpenCode account / free tier

## Affected models

Examples:

```text
mimo-v2.5-free
muse-spark-1.2-contributor-free
```

Potentially all OpenCode free-tier models are affected.

## Steps to reproduce

1. Deploy 9Router `v0.5.75`
2. Login or configure OpenCode provider
3. Select an OpenCode free-tier model
4. Send a request through the 9Router OpenAI-compatible API
5. Request fails with HTTP 403

Example request flow:

```text
Client / Hermes
      ↓
9Router
      ↓
OpenCode
      ↓
HTTP 403 FreeTierError
```

## Actual behavior

The request fails with:

```text
OpenCode's free tier can only be used from within OpenCode
```

## Expected behavior

If OpenCode free-tier access is still supported by 9Router, requests should work normally.

If OpenCode no longer allows third-party/proxy access to free-tier models, 9Router should ideally:

* Detect this restriction
* Show a clearer error message in the UI
* Mark affected OpenCode free-tier models as unavailable
* Avoid repeatedly routing requests to an unusable free-tier endpoint
* Optionally fallback to another configured provider/model

## Additional notes

This appears to be a provider-side restriction rather than a Docker, networking, or 9Router API configuration issue.

Paid OpenCode models or other providers may still work normally.

It would be useful to know whether:

1. OpenCode free-tier support can still be maintained in 9Router
2. Authentication or API behavior has changed
3. Free-tier models should be removed/disabled from the provider list
4. A fallback mechanism can be added when this specific `FreeTierError` occurs

Bonbocat:
Dupe #4101 #4103

MuhWldns:
PR #4128

[...]

Issue #4124, opened 2026-09-17
OpenCode free doesnt work anymore


mnickmagham:
OpenCode free now doesnt work and says it needs to be requested from the opencode harness iteself, the result of requests: HTTP 403: [403]: {"type":"error","error":{"type":"FreeTierError","message":"Error from provider (Console): OpenCode's free tier can only be used from within OpenCode"}}

Someone on reddit had posted something worth of checking: https://www.reddit.com/r/hermesagent/comments/1w9x0on/fix_opencode_freetier_models_working_in_hermes/

NoReplyUI5:
i am getting the same error
HTTP 403: [403]: {"type":"error","error":{"type":"FreeTierError","message":"Error from provider (Console): OpenCode's free tier can only be used from within OpenCode"}}

1nlyr:
same

TEGAR-SRC:
same

TEGAR-SRC:
HTTP 403: [403]: {"type":"error","error":{"type":"FreeTierError","message":"Error from provider (Console): OpenCode's free tier can only be used from within OpenCode"}}

mnickmagham:
someone fix this i want to burn tokens

blessedplummer:
It appears that the current fixes are incomplete, and does not fully address the enforcement mechanisms. Here's what my agent has to say on the matter after it got it working:



I can confirm this and I've bisected it further — #4128 and #4131 cover part of it (UA + session shape, both real: bare `opencode` UA and random session IDs do 403), but there are **two more gates** upstream enforces that neither PR addresses yet. Tested against 0.5.75, `oc/muse-spark-1.3-contributor-free`, same session + genuine `opencode/1.18.x` UA throughout:

**1. Tool-signature gate.** Identical headers/session, different bodies:

| Body | Result |
|---|---|
| minimal `{messages:[{user:"hi"}]}` | 403 FreeTierError |
| \+ system prompt | 403 |
| \+ 1 tool | 403 |
| full body minus `tools` (10 KB) | 403 |
| minimal messages **+ OpenCode's full 10-tool set** (`bash, edit, glob, grep, read, skill, task, todowrite, webfetch, write`) | **200** |

So upstream also fingerprints the agentic tool signature. A client sending no/few tools is rejected no matter how valid the session is.

[...]

Issue #4103, opened 2026-09-17
opencode zen报错


qq458249269:
应该是又更新了请求头了

URL: https://opencode.ai/zen/v1/chat/completions [403]: {"type":"error","error":{"type":"FreeTierError","message":"Error from provider (Console): OpenCode's free tier can only be used from within OpenCode"}}

Azzarnuji:
i have same issue

tkhs101:
i have same issue

arwinwijaya:
i have same issue

zinatomas:
i have same issue

anojndr:
确实是上游更新了请求头校验规则：

上游 OpenCode 在免费模型端点（`Authorization: Bearer public`）部署了两个严格的服务端格式校验：

1. **User-Agent 请求头**：必须匹配 `opencode/<version>` 且版本号 >= `1.17.0`。如果是裸字符 `opencode` 或第三方客户端 User-Agent，会直接返回 `403 FreeTierError`（版本号 < `1.17.0` 会返回 `426 Upgrade Required`）。9router 原先默认发送的是没有版本号的 `opencode`。
2. **x-opencode-session 请求头**：必须符合 OpenCode 官方的 descending session ID 格式（`ses_` + 12位十六进制时间戳 + 14位 Base62 字符，正则 `/^ses_[0-9a-f]{12}[0-9A-Za-z]{14}$/`）。如果请求头缺失或使用普通的 UUID 格式，上游也会判定为非 OpenCode 客户端并拒绝（403）。

修复 PR 已提交：#4105
- 将 `User-Agent` 默认设置为 `opencode/1.18.31`，并保留下游符合要求的版本号。
- 实现了 OpenCode 官方规范的 session ID 生成和确定性映射转换，使下游客户端会话（如 Claude Code 等）能无缝映射为合规格式并保留缓存亲和性。
- 移除了共享实例状态，使用请求局部凭据保证并发请求安全。

zinatomas:
有发布的版本号吗？速度搞一下

wuhanwuyanzu123:
OpenCode最牛的模型是哪个？

anojndr:
Update — my earlier comment here (PR #4105) was an incomplete fix: it covered only 2 of the 4 gates upstream enforces, which is why the 403 came back within hours.

Live bisection with genuine `opencode/1.18.31` + canonical `ses_` held constant found the two missing gates Zen enforces with `Bearer public`:
[...]

PR #4165, opened 2026-09-19
fix(opencode): stop 403 FreeTierError on Responses requests with tools

suryacagur:
## Problem

`POST oc/muse-spark-1.3-contributor-free` via `/v1` fails whenever the
request carries tools, while the dashboard provider test stays green:

[403]: {"type":"error","error":{"type":"FreeTierError","message":"Error from provider (Console): OpenCode's free tier can only be used from within OpenCode"}}

Tool-less calls succeed; calls with N tools (e.g. 46) always 403.

## Root cause

Verified live against `https://opencode.ai/zen/v1/responses`: upstream
requires **both** `bash` + `read` function tools **and**
`tool_choice: "auto"` on every Responses request (403 without, 200 with).
`OpenCodeExecutor` only injected the decoy tools when the payload had
zero tools, so every real agentic call went out naked. The dashboard
probe only does `GET /zen/v1/models`, so it can never catch this.

## Changes

- `open-sse/executors/opencode.js` - cloak decoys unconditionally on the
  Responses path (was: empty-tools only)
- `open-sse/providers/registry/opencode.js` - extend
  `forceAutoToolChoiceModels` to `muse-spark-1.2-contributor-free`
  (same auto-only rule, verified live with 400 otherwise)
- `tests/unit/opencode-free-tool-choice.test.js` - updated expectations
  plus a 46-external-tools → 48 regression test
- `src/app/api/providers/[id]/test/testUtils.js` - comment documenting
  the dashboard probe blind spot

## Verification

- Live upstream matrix: no-decoy 46-tool body → 403 (exact reported
  error); same body + decoys + auto → 200 SSE, for both 1.2 and 1.3
- `node --check` passes on all touched files
- Unit tests: `cd tests && npx vitest run
  unit/opencode-free-tool-choice.test.js unit/opencode-session.test.js`

> Note: the gateway must be restarted/rebuilt after merge - a stale
> bundle keeps serving the old gate.

## Issue

Fixed OpenCode Free Tier requests returning HTTP 403 `FreeTierError` when routed through 9Router. This resolves compatibility issues affecting OpenCode Free Tier models and requests from clients such as Claude Code.

### Resolved issues:

- [#4101 - OpenCode Free Tier models returning HTTP 403 `FreeTierError` via 9Router](https://github.com/decolua/9router/issues/4101)
- [#4103 - OpenCode Zen returning `FreeTierError` after request changes](https://github.com/decolua/9router/issues/4103)
- [#4124 - OpenCode Free Tier stopped working through 9Router](https://github.com/decolua/9router/issues/4124)
- [#4127 - OpenCode Free Tier returning HTTP 403 when used through 9Router](https://github.com/decolua/9router/issues/4127)
- [#4152 - OpenCode Free Tier still failing when used through Claude Code](https://github.com/decolua/9router/issues/4152)


ManiTheFerfri:
@decolua please push this!!!

